PR #210, opened 2017-07-12
Fix: Media controls and usability fixes for mobile

jeremypress:
- Scrubber responds to touch events
- Controls are toggled on touch
- Controls are not always force show due to video360 css


src/lib/viewers/media/MediaControls.js
```diff
@@ -67,6 +68,8 @@ const FILMSTRIP_FRAME_HEIGHT = 90;
         this.setupSettings();
         this.setupScrubbers();
         this.attachEventHandlers();
+
+        this.hasTouch = Browser.hasTouch();
```
JustinHoldstock: Could you set `this.hasTouch` in BaseViewer?

JustinHoldstock: Also, can you double check and see if you can use this in a declaration above the constructor? I'm curious of when that code executes, then you can remove it from the constructor entirely

jeremypress: Forgot it was defined in base, I removed it elsewhere. It could be moved out of the constructor but we need to set it as a class property anyways..

PR #213, opened 2017-07-13
Feature/drawing annotations

MinhHNguyen:
Drawing Icon SVGs from: https://iconmonstr.com/license/
Rbush (performant RTree JS implementation): https://github.com/mourner/ for fast(relatively) hit detection on rectangles

Currently allows for drawing on documents across pages. Does not yet work with resizing or saving. Also need to finish updating and creating new tests.

Some thoughts from discussion with @jeremypress:
- Drawing annotations are behind a feature toggle (BoxAnnotations.js Annotators.TYPE array in conjunction with BaseViewer.isAnnotatable(type))
- The way that annotations are baked into viewer is not an ideal design. BaseViewer should not have to deal with annotation controller logic besides a generic setup function. In its current state, BaseViewer will become very unmanageable with more custom annotation types.
- We might want to consider having DocPointThread/Dialog and ImagePointThread/Dialogue should inherit from a PointThread/Dialogue class which shares common functions (seems like it is only createDialogue that can be shared in this case, but show is similar in both) This will make it simpler to add FilePointThreads classes such as VideoPointThread in the future.
- Reminder (tracked in trello): need to deal with language localization for drawing mode notification and fix tests

@pramodsum @JustinHoldstock @tonyjin


pramodsum:
I would maybe move your localization changes into a separate PR so we can focus on reviewing the code changes here

src/lib/annotations/drawing/DrawingPath.js
```diff
@@ -0,0 +1,101 @@
...
+    /** @property {number} - The minimum Y position of all coordinates */
+    minY;
+
+    /**
+     * [constructor]
+     * @return {DrawingPath} DrawingPath instance
+     */
+    constructor() {
+        this.path = [];
+        this.maxX = -Infinity;
```
JustinHoldstock: You can set all of these defaults in the declarations above :) Then you can nuke the constructor!

src/lib/annotations/drawing/DrawingPath.js
```diff
@@ -0,0 +1,101 @@
...
+
+    /**
+     * Add position to coordinates and update the bounding box
+     *
+     * @param {number} xPos xPosition to be part of the drawing
+     * @param {number} yPos yPosition to be part of the drawing
+     * @return {void}
+     */
+    addCoordinate(xPos, yPos) {
+        if (xPos < this.minX) {
```
JustinHoldstock: AFAIK none of these ifs will ever take, as nothing can be greater or less than Infinity/-Infinity

JustinHoldstock: Also, newlines after your if blocks

src/lib/annotations/drawing/DrawingPath.js
```diff
@@ -0,0 +1,101 @@
...
+
+    /**
+     * Draw the recorded coordinates onto a CanvasContext
+     *
+     * @param {CanvasContext} context Context to draw the recorded path on
+     * @return {void}
+     */
+    drawPath(context) {
+        const ctx = context;
+        if (ctx) {
```
JustinHoldstock: To remove some arrowness, you can, at the top of the block, instead do
```
if(!ctx) {
  return;
}
```

src/lib/annotations/drawing/DrawingThread.js
```diff
@@ -0,0 +1,131 @@
...
+    //--------------------------------------------------------------------------
+    // Protected
+    //--------------------------------------------------------------------------
+
+    /**
+     * @protected
+     * @param {number} scale The document scale
+     * @return {void}
+     */
+    updateContextStyles(scale) {
```
JustinHoldstock: Description docs

src/lib/annotations/drawing/DrawingThread.js
```diff
@@ -0,0 +1,131 @@
...
+        if (this.context) {
+            this.context.lineCap = 'round';
+            this.context.lineJoin = 'round';
+            this.context.strokeStyle = 'black';
+            this.context.lineWidth = BASE_LINE_WIDTH * scale;
+        }
+    }
+
+    /**
+     * @protected
```
JustinHoldstock: Description docs

src/lib/annotations/drawing/DrawingThread.js
```diff
@@ -0,0 +1,131 @@
...
+     * @return {void}
+     */
+    render(timestamp) {
+        if (this.drawingFlag === DRAW_POINTER_DOWN) {
+            const elapsed = timestamp - (this.lastRenderTimestamp || 0);
+            if (elapsed >= DRAW_RENDER_THRESHOLD && this.context) {
+                this.lastRenderTimestamp = timestamp;
+                const context = this.context;
+                if (this.savedState) {
+                    context.putImageData(this.savedState, 0, 0);
```
JustinHoldstock: `get/putImageData` are rather slow operations. Is there a reason for this? Why can't the saved state be a record of previous point data, and re-render that as lines to the canvas, if savedState is required?

src/lib/annotations/doc/DocDrawingThread.js
```diff
@@ -0,0 +1,88 @@
...
+        const canvas = context.canvas;
+
+        this.savedState = context.getImageData(0, 0, canvas.width, canvas.height);
+
+        if (!this.pendingPath) {
+            this.pendingPath = new DrawingPath();
+        }
+        this.context = context;
+        this.updateContextStyles(scale);
+        window.requestAnimationFrame(this.render.bind(this));
```
JustinHoldstock: Whenever you `requestAnimationFrame()`, remember you need a `cancelAnimationFrame()` for cleanup :)

src/lib/annotations/drawing/DrawingThread.js
```diff
@@ -0,0 +1,131 @@
...
+
+    /**
+     * @protected
+     * @param {number} timestamp The time when the function was called;
+     * @return {void}
+     */
+    render(timestamp) {
+        if (this.drawingFlag === DRAW_POINTER_DOWN) {
+            const elapsed = timestamp - (this.lastRenderTimestamp || 0);
+            if (elapsed >= DRAW_RENDER_THRESHOLD && this.context) {
```
JustinHoldstock: Also, pair this delta time check with input events, to have an even MORE performant render loop. There's no sense in re-rendering the whole scene if you're just sitting with the pointer in a "down" state

src/lib/annotations/doc/DocDrawingThread.js
```diff
@@ -0,0 +1,88 @@
...
+        const scale = getScale(this.annotatedElement);
+        const context = DocAnnotatorUtil.getContext(
+            this.pageEl,
+            CLASS_ANNOTATION_LAYER_DRAW,
+            PAGE_PADDING_TOP,
+            PAGE_PADDING_BOTTOM
+        );
+        const canvas = context.canvas;
+
+        this.savedState = context.getImageData(0, 0, canvas.width, canvas.height);
```
MinhHNguyen: Each render clears the canvas, puts the savedState onto the canvas, then draws the path of all of the current point. It seems that drawImage(canvas) is faster than get/putImageData ( as a result, it would be good to have a memory canvas) but I am unsure how it performs in contrast to redrawing all of the paths on each render.

src/lib/annotations/drawing/DrawingPath.js
MinhHNguyen: Yeah made a mistake here, 
For the minimums, anything fitting in memory should be less than infinity 
For the maximums, anything fitting in memory should be greater than negative infinity

src/lib/annotations/doc/DocDrawingThread.js
JustinHoldstock: Also, cache the result of `this. render.bind(this)`, and pass that in, so you aren't creating a new bound function object every frame. (60fps * function.bind(this) = :[ )

src/lib/annotations/doc/DocDrawingThread.js
JustinHoldstock: AFAIK re-rendering each line on the whole canvas should be more performant than an put/get of the whole canvas. It passes a series of points that will leverage hardware acceleration to render VS get/put which uses JS to get/replace the pixel data of the canvas (width * height * rgba)

MinhHNguyen: Seems like drawing a canvas over might be faster. Likely depends on both canvas size and the number of coordinates being drawn. I would argue that it is faster to copy over a memory canvas.

https://jsperf.com/drawimagequadraticcurveto

JustinHoldstock: I'm getting some wicked inconsistent behaviour with this test. One moment, drawImage is 48% slower, the next 70% faster. Each time I run it, I take down my browser for about 30 seconds. Lemme investigate

PR #215, opened 2017-07-13
Chore: Cleaning up annotation strings/classes

pramodsum:


src/lib/annotations/AnnotationDialog.js
```diff
@@ -388,41 +403,43 @@ const CLASS_ANIMATE_DIALOG = 'bp-animate-show-dialog';
...
                 this.postReply();
                 break;
             // Clicking trash icon to initiate deletion
-            case 'delete-btn':
+            case DATA_TYPE_DELETE:
                 this.showDeleteConfirmation(annotationID);
                 break;
             // Clicking 'Cancel' button to cancel deletion
-            case 'cancel-delete-btn':
+            case CLASS_CANCEL_DELETE:
```
JustinHoldstock: I'm a tiny bit hesitant for us to use the same variable, for a class, as a data-type property. Can you make constants for the data-types, and use those instead?

src/lib/annotations/AnnotationDialog.js
```diff
@@ -475,24 +492,24 @@ const CLASS_ANIMATE_DIALOG = 'bp-animate-show-dialog';
...
                     ${__('annotation_delete_confirmation_message')}
                 </div>
-                <div class="button-container">
-                    <button class="bp-btn cancel-delete-btn" data-type="cancel-delete-btn">
+                <div class="${constants.CLASS_BUTTON_CONTAINER}">
+                    <button class="bp-btn ${CLASS_CANCEL_DELETE}" data-type="${CLASS_CANCEL_DELETE}">
                         ${__('annotation_cancel')}
                     </button>
-                    <button class="bp-btn bp-btn-primary confirm-delete-btn" data-type="confirm-delete-btn">
+                    <button class="bp-btn bp-btn-primary ${DATA_TYPE_CONFIRM_DELETE}" data-type="DATA_TYPE_CONFIRM_DELETE">
```
JustinHoldstock: See above note about `data-type` vs `class name` variables

src/lib/annotations/annotationConstants.js
```diff
@@ -1,31 +1,42 @@
...
-export const CLASS_REPLY_CONTAINER = 'reply-container';
-export const CLASS_REPLY_TEXTAREA = 'reply-textarea';
+export const CLASS_MOBILE_ANNOTATION_DIALOG = 'bp-mobile-annotation-dialog';
+export const CLASS_MOBILE_DIALOG_HEADER = 'bp-annotation-mobile-header';
+export const CLASS_DIALOG_CLOSE = 'bp-annotation-dialog-close';
+export const CLASS_TEXTAREA = 'bp-textarea';
+export const CLASS_HIGHLIGHT_BTNS = 'bp-annotation-highlight-btns';
+export const CLASS_ADD_HIGHLIGHT_BTN = 'bp-add-highlight-btn';
+export const CLASS_ADD_HIGHLIGHT_COMMENT_BTN = 'bp-highlight-comment-btn';
+export const CLASS_ANNOTATION_LAYER = 'bp-annotation-layer';
```
JustinHoldstock: Do you wanna create selector constants for each class name?

pramodsum: Sure. I only added selectors for classes that made use of them so far

src/lib/annotations/AnnotationDialog.js
pramodsum: Yea I wasn't entirely sure about this. I agree that it just makes sense to just do data-type constants instead even if they're technically the same string.

PR #220, opened 2017-07-17
Update: Render Vera-protected HTML files

tonyjin:
Render Vera-protected files instead of displaying the raw HTML.


src/lib/viewers/text/TextLoader.js
```diff
@@ -43,6 +44,18 @@ class TextLoader extends AssetLoader {
         super();
         this.viewers = VIEWERS;
     }
+
+    /**
+     * @inheritdoc
+     */
+    determineViewer(file, disabledViewers = []) {
+        const viewer = super.determineViewer(file, disabledViewers);
+        if (viewer && isVeraProtectedFile(file)) {
```
jeremypress: Maybe add a comment explaining this?

PR #222, opened 2017-07-17
Fix settings pullup for model3D files on mobile

jeremypress:
- Resizes/repositions the pullup based on device size
- Fixes inputs in the settings panel


src/lib/viewers/box3d/Box3DControls.scss
```diff
@@ -0,0 +1,12 @@
+@media (max-width: 375px) {
+    .bp-box3d .bp-pullup {
+        right: 0;
+    }
+}
+
+@media (max-height: 550px) {
```
JustinHoldstock: If you could make a sass variable for each width, it'll make this easier to read :)

```
// Then use like
@media($compact-width){}
@media{$less-compact-width}{}
```

PR #225, opened 2017-07-18
Fix: Limit iOS font fix to 10.3.1

tonyjin:
Apple fixed the font issue in 10.3.2, so we only limit the disableFontFace fix to 10.3.1


jeremypress:
Looked good on 10.3.3

PR #231, opened 2017-07-19
Fix: Allow legacy annotations to render in PDF.js

jeremypress:
PDF.js relies on the presence of annotation icons in the DOM to render popups. Once our migration tool removes the hidden flag from the native annotations it creates, PDF.js will be able to render migrated annotations fully.


jeremypress:
@JustinHoldstock yep!

jeremypress:
We could do this, or we could add a few SVGs and support Native PDF annotations as well as PDF.js does!

PR #233, opened 2017-07-19
Fix: Don't swallow 'touchstart' event to prevent iOS inertia scrolling

jeremypress:
This was blocking annotation creation on mobile.


src/lib/viewers/doc/PresentationViewer.js
```diff
@@ -228,10 +230,10 @@ class PresentationViewer extends DocBaseViewer {
             return;
         }
 
-        event.preventDefault();
-
         if (event.type === 'touchstart') {
             this.scrollStart = event.changedTouches[0].clientY;
+        } else if (event.type === 'touchmove') {
+            event.preventDefault();
```
jeremypress: @pramodsum you don't use touchmove for anything correct? And @MinhHNguyen you prevent default behavior anyways?

pramodsum: nope! we don't

PR #242, opened 2017-07-25
Fix: Don't trigger highlight dialogs while mouse is over another dialog

pramodsum:
- This also prevents highlight dialogs from triggering while the mouse is over a point dialog

### DIFF
--- a/src/lib/annotations/doc/DocAnnotator.js
+++ b/src/lib/annotations/doc/DocAnnotator.js
@@ -594,6 +594,12 @@ class DocAnnotator extends Annotator {
                 return;
             }
 
+            // Determine if mouse is over any highlight dialog currently
+            // and ignore hover events of any highlights below
+            if (docAnnotatorUtil.isDialogDatatype(event.target)) {
+                return;
+            }
+
             // Determine if the mouse is hovering over any highlight threads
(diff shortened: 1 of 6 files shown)


src/lib/annotations/doc/docAnnotatorUtil.js
```diff
@@ -359,3 +386,8 @@ export function getContext(pageEl, annotationLayerClass, paddingTop, paddingBott
 export function getPageEl(annotatedEl, pageNum) {
     return annotatedEl.querySelector(`[data-page-number="${pageNum}"]`);
 }
+
+export function isDialogDatatype(eventTarget) {
```
JustinHoldstock: Docs

src/lib/annotations/doc/DocAnnotator.js
```diff
@@ -598,6 +598,12 @@ class DocAnnotator extends Annotator {
                 return;
             }
 
+            // Determine if mouse is over any highlight dialog currently
+            // and ignore hover events of any highlights below
+            if (docAnnotatorUtil.isDialogDatatype(event.target)) {
```
JustinHoldstock: Move out of the `for`

PR #243, opened 2017-07-25
New: Adding 'annotationfocus' event to threads

pramodsum:
- "focus" behavior is specific to the current annotations icon
- Slightly enlarges the icon when 'annotationfocus' event is emitted for a thread number matching the current thread
- Ensuring image point annotations have thread numbers associated with the threads


JustinHoldstock:
@pramodsum In this case "focus" is confusing, as this can refer to actually giving focus to the element and focus can be used for accessibility purposes. Can we consider calling this something along the lines of "call to action", "prompt", "prod", or "jiggle"?

PR #246, opened 2017-07-25
New: Add pseudo fullscreen to mobile Safari

jeremypress:
- Mobile Safari is the only browser we support that doesn't support the Fullscreen API. This change sets the content to fill the viewport by `setting position: fixed`
- Added touch-action manipulation to prevent double-tap zoom on the content

### DIFF
--- a/src/lib/viewers/BaseViewer.js
+++ b/src/lib/viewers/BaseViewer.js
@@ -330,25 +330,9 @@ class BaseViewer extends EventEmitter {
      */
     addCommonListeners() {
         // Attach common full screen event listeners
-        /* istanbul ignore next */
-        fullscreen.addListener('enter', () => {
-            this.containerEl.classList.add(CLASS_FULLSCREEN);
-            if (!fullscreen.isSupported()) {
-                this.containerEl.classList.add(CLASS_FULLSCREEN_UNSUPPORTED);
-            }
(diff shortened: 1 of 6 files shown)


src/lib/viewers/BaseViewer.js
```diff
@@ -361,6 +354,20 @@ class BaseViewer extends EventEmitter {
     }
 
     /**
+     * Applies appropriate styles and resizes the document depending on fullscreen state
+     * @protected
```
tonyjin: Nit - space after description before @. We also stopped marking things as protected

src/lib/viewers/doc/DocBaseViewer.js
```diff
@@ -361,7 +361,7 @@ class DocBaseViewer extends BaseViewer {
         const previousPageButtonEl = this.containerEl.querySelector('.bp-previous-page');
         const nextPageButtonEl = this.containerEl.querySelector('.bp-next-page');
 
-        // Safari disables keyboard input in fullscreen
+        // Safari disables keyboard input in fullscreen before Safari 10.1
```
tonyjin: Can we detect this and re-enable page selector in Safari 10.1?

jeremypress: I'll do that in a separate PR since it's not trivial to select 10.1 on iOS and Safari at the same time.

PR #247, opened 2017-07-26
Fix: thread save incorrectly rejects when no dialogue exists

MinhHNguyen:
Promise would reject upon successful creation of annotation on the backend due to attempted dereferences on null/undefined objects. 

Might have future issues with 
`this.thread = this.thread || savedAnnotation.thread;`
if create(annotationData) returns a null annotation object.
This is not likely, I just haven't read through all of AnnotationService.

For 
`this.annotations[tempIdx] = savedAnnotation;`,
let me know if this.annotations[-1] is actually intentional. I saw that ImageAnnotator uses threads.page[-1] as a valid index.


pramodsum:
The line `this.thread = this.thread || savedAnnotation.thread;` was added so that either the thread will always have an associated threadNumber. If the thread has annotations already in it, then it will set the threadNumber to the one associated with all those annotations. If the thread was just created, then it'll set it as the threadNumber for the newly created annotation aka savedAnnotation

pramodsum:
ImageAnnotator uses -1 to indicate that it is a single paged image. I just added a PR to change page -1 to page 1 for images. Having -1 pages was just how it was implemented beforehand to indicate that we didn't care for pages in images, but that definitely changed once we added the multi-page image annotations.

JustinHoldstock:
After @pramodsum and @MinhHNguyen's discussion, is this still a valid PR?

MinhHNguyen:
@JustinHoldstock 
Yes it is still valid, if the code is left as is, newly created annotations without dialogues will be created on the server but throw an error on the client side. Let me know if you see any unforseen errors that can occur with the changes.

jeremypress:
LGTM but someone with more annotations expertise should verify

MinhHNguyen:
tests incoming

src/lib/annotations/AnnotationThread.js
```diff
@@ -151,20 +151,23 @@ class AnnotationThread extends EventEmitter {
...
                 this.threadNumber = this.threadNumber || savedAnnotation.threadNumber;
-                this.dialog.element.dataset.threadNumber = this.threadNumber;
-
-                // Otherwise, replace temporary annotation with annotation saved to server
-                this.annotations[tempIdx] = savedAnnotation;
 
                 if (this.dialog) {
+                    // Add thread number to associated dialog and thread
+                    if (this.dialog.element && this.dialog.element.dataset) {
+                        this.dialog.element.dataset.threadNumber = this.thread;
```
pramodsum: This will be setting the dialog thread number to null
`this.dialog.element.dataset.threadNumber = this.threadNumber;`

src/lib/annotations/AnnotationThread.js
```diff
@@ -151,20 +151,23 @@ class AnnotationThread extends EventEmitter {
...
             .create(annotationData)
             .then((savedAnnotation) => {
-                // If no temporary annotation is found, save to thread normally
                 const tempIdx = this.annotations.indexOf(tempAnnotation);
                 if (tempIdx === -1) {
+                    // If no temporary annotation is found, save to thread normally
                     this.saveAnnotationToThread(savedAnnotation);
+                } else {
+                    // Otherwise, replace temporary annotation with annotation saved to server
+                    this.annotations[tempIdx] = savedAnnotation;
```
pramodsum: Is it possible that `this.annotations[-1]` would have `tempAnnotation` in it? Would we need to replace it with the correct one no matter what?

src/lib/annotations/AnnotationThread.js
MinhHNguyen: fixed

MinhHNguyen: @pramodsum tests in and variable name updated

src/lib/annotations/AnnotationThread.js
```diff
@@ -397,6 +378,38 @@ class AnnotationThread extends EventEmitter {
...
     /**
+     * Update a temporary annotation with the annotation saved on the backend and set the threadNumber if it does
+     * not exist. Propogate the threadnumber to dialog if applicable.
+     *
+     * @private
+     * @param {Annotation} serverAnnotation - The annotation returned by the backend. Use this as the source of truth
+     * @param {Annotation} localAnnotation - The temporary annotation holding place of the server validated annotation
+     * @return {void}
+     */
+    updateLocalAnnotationFromServer(serverAnnotation, localAnnotation) {
```
pramodsum: Thanks for moving this out and writing tests! 👍

src/lib/annotations/AnnotationThread.js
```diff
@@ -146,29 +146,10 @@ class AnnotationThread extends EventEmitter {
...
-
-                // Otherwise, replace temporary annotation with annotation saved to server
-                this.annotations[tempIdx] = savedAnnotation;
-
-                if (this.dialog) {
-                    this.dialog.addAnnotation(savedAnnotation);
-                    this.dialog.removeAnnotation(tempAnnotationID);
-                }
-            })
+            .then((savedAnnotation) => this.updateLocalAnnotationFromServer(savedAnnotation, tempAnnotation))
```
pramodsum: I'd like to keep the naming consistent. Can we make this method either updateTempAnnotation(...) or change tempAnnotation, tempIdx, and other temp variables to localAnnotation, etc?

src/lib/annotations/AnnotationThread.js
```diff
@@ -397,6 +378,38 @@ class AnnotationThread extends EventEmitter {
     //--------------------------------------------------------------------------
 
     /**
+     * Update a temporary annotation with the annotation saved on the backend and set the threadNumber if it does
+     * not exist. Propogate the threadnumber to dialog if applicable.
+     *
+     * @private
+     * @param {Annotation} serverAnnotation - The annotation returned by the backend. Use this as the source of truth
+     * @param {Annotation} localAnnotation - The temporary annotation holding place of the server validated annotation
```
pramodsum: The temporary annotation stored locally holding place of the server validated annotation

src/lib/annotations/AnnotationThread.js
```diff
@@ -397,6 +378,38 @@ class AnnotationThread extends EventEmitter {
...
+     * @return {void}
+     */
+    updateLocalAnnotationFromServer(serverAnnotation, localAnnotation) {
+        const tempIdx = this.annotations.indexOf(localAnnotation);
```
pramodsum: choose temp or local

src/lib/annotations/AnnotationThread.js
```diff
@@ -397,6 +378,38 @@ class AnnotationThread extends EventEmitter {
...
+
+        this.threadNumber = this.threadNumber || serverAnnotation.threadNumber;
+
+        if (this.dialog) {
+            // Add thread number to associated dialog and thread
+            if (this.dialog.element && this.dialog.element.dataset) {
+                this.dialog.element.dataset.threadNumber = this.threadNumber;
+            }
+
+            this.dialog.addAnnotation(serverAnnotation);
```
pramodsum: // Removes temporary/local annotation and replaces it with the annotation saved to server

src/lib/annotations/__tests__/AnnotationThread-test.js
```diff
@@ -174,6 +174,54 @@ describe('lib/annotations/AnnotationThread', () => {
...
+        });
+
+        it('should overwrite a local annotation to the thread if it does exist as an associated annotation', () => {
+            const serverAnnotation = 'im a real annotation';
+            const replacementAnnotation = 'im a replacement annotation';
+            const isReplacementAnnotation = (annotation) => annotation === replacementAnnotation;
+            stubs.saveAnnotationToThread = sandbox.stub(thread, 'saveAnnotationToThread');
+            thread.annotations.push(serverAnnotation)
+            thread.updateLocalAnnotationFromServer(replacementAnnotation, serverAnnotation);
+            thread.on('annotationcreateerror', () => {
```
pramodsum: Not sure the check on `'annotationcreateerror'` needs to be in the unit tests for `updateLocalAnnotationFromServer()`. Might make sense to move the portion of code called in the catch block into it's own method and move this check there.

src/lib/annotations/AnnotationThread.js
MinhHNguyen: changing all localAnnotation variable names to tempAnnotation

src/lib/annotations/__tests__/AnnotationThread-test.js
MinhHNguyen: Good point, I've removed this portion from the tests.

pramodsum:
Just make sure you test this thoroughly with all the different annotation types just to be sure 👍

MinhHNguyen:
I'll go through a quick sanity check

PR #248, opened 2017-07-26
Fix: Add clickHandlers back for all enabled annotation modes

pramodsum:
- click handlers were moved into docBase which meant that the image point annotation button no longer had a handler associated with it

### DIFF
--- a/src/lib/viewers/BaseViewer.js
+++ b/src/lib/viewers/BaseViewer.js
@@ -767,9 +767,16 @@ class BaseViewer extends EventEmitter {
      * @param {string} mode - Target annotation mode
      * @return {Function|null} Click handler
      */
-    /* eslint-disable no-unused-vars */
-    getAnnotationModeClickHandler(mode) {}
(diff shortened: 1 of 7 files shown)


src/lib/viewers/BaseViewer.js
```diff
@@ -767,9 +767,16 @@ class BaseViewer extends EventEmitter {
      * @param {string} mode - Target annotation mode
      * @return {Function|null} Click handler
      */
-    /* eslint-disable no-unused-vars */
-    getAnnotationModeClickHandler(mode) {}
-    /* eslint-enable no-unused-vars */
+    getAnnotationModeClickHandler(mode) {
+        if (!mode || !this.isAnnotatable(mode)) {
```
JustinHoldstock: Can `isAnnotatable()` just return false if null is passed in?

pramodsum: It defaults to the value of previewOptions.showAnnotations if no viewer specific options are specified

PR #249, opened 2017-07-26
Chore: Default single page images to page 1 for annotations

pramodsum:


src/lib/annotations/image/ImageAnnotator.js
```diff
@@ -89,6 +89,13 @@ class ImageAnnotator extends Annotator {
      */
     createAnnotationThread(annotations, location, type) {
         let thread;
+
+        // Corrects any image annotation page number to 1 instead of -1
+        const fixedLocation = location;
```
pramodsum: @JustinHoldstock this should cover anyone who created an image annotation before this commit was merged in

src/lib/annotations/annotatorUtil.js
```diff
@@ -41,7 +41,7 @@ export function findClosestElWithClass(element, className) {
 */
 export function getPageInfo(element) {
     const pageEl = findClosestElWithClass(element, 'page') || null;
-    let page = -1;
+    let page = 1;
```
JustinHoldstock: Make sure that this works. I know we do some explicit checks for `page === -1` so that we can check where events occur on the page.

pramodsum: yup it does!

PR #254, opened 2017-07-27
[WIP] New: Preview file from offline

disist:
Hy guys! 

This pull request is like the proposal to preview files from offline

We have a hybrid mobile application based on Angular2+ and we are using your previewer to preview files.

Our goal: learn previewer to show files from offline storage when we have no internet.

As we understand we have the sequence of 3 requests to preview files (fetch `file` info from the server, get `content` based the first call, and some `event` call). 
So we created `fetchFileInfo` public method for getting `file` info (file info will be placed in our secure storage). Then we try to preview file from offline storage, we pass `offline` option link and file (which was fetched before).

Is approach ok with yours?

PS: Unit tests will be added:) after finalizing with offline approach


src/lib/Preview.js
```diff
@@ -467,6 +461,33 @@ class Preview extends EventEmitter {
         });
     }
 
+    /**
+     * Fetch file info
```
tonyjin: How about `Fetch and return file info, which can then be cached or stored externally`? We also like adding a line break after the description before the first @ param

src/lib/Preview.js
```diff
@@ -612,6 +639,9 @@ class Preview extends EventEmitter {
         // Save the reference to any additional custom options for viewers
         this.options.viewers = options.viewers || {};
 
+        // Save the reference to offline
+        this.options.offline = options.offline || {};
```
tonyjin: How about `offlineConfig` to be clearer that `offline` isn't a boolean?

src/lib/Preview.js
```diff
@@ -782,6 +848,23 @@ class Preview extends EventEmitter {
     }
 
     /**
+     * Enrich file entries with offline links.
+     *
+     * @private
+     * @return {void}
+     */
+    enrichOfflineLinks() {
```
tonyjin: We probably need to rethink this implementation since it assumes we want to replace all of the representation's URLs with one offline link. This may work for a simple document use case but breaks with multi-asset viewers like video and 3D.

Handling offline for HD video is particularly difficult since the main representation for DASH is actually a manifest which the open source library (Shaka Player) knows how to read. Shaka will then make requests to the URLs described in the manifest.

I would also move this logic to https://github.com/box/box-content-preview/blob/af861cfc33c6c4648a2cd31598c22a9dcf06819e/src/lib/file.js since we already have some logic there to manipulate representations in a file. 

Speaking of... looking at the logic there for `addOriginalRepresentation` gives me the idea that we could potentially solve this by creating `addOfflineRepresentation`, which takes in a file + an offline URL and adds an `OFFLINE` representation to the file, which is the URL you configure in `previewOptions.offlineConfig`. Then, in https://github.com/box/box-content-preview/blob/af861cfc33c6c4648a2cd31598c22a9dcf06819e/src/lib/viewers/doc/DocLoader.js, you could add two new configurations for Presentation and Document that accept this new `OFFLINE` representation.

Not sure yet - any other thoughts on how we could scope this change to only documents or single-rep viewers @JustinHoldstock @jeremypress @pramodsum @MinhHNguyen ?

priyajeet: Another approach, that will require some refactoring, is to change the networking layer. From the top of my head, it could be:

To show a preview one does `show()` --> `getTokens()` --> `getFile()` --> `loadRep()` --> preview. Other than the networking done inside `pdf.js` or `shaka-player.js` or done by the browser (think ``), which is out of scope, everything goes through `util.js`.

Network calls can be refactored out of `util.js` and into a class `Xhr.js`. An override of `Xhr.js`, lets call it `Offline.js` can implement `get()`, `post()` etc...which fetches data out of some local offline cache instead of getting it over the wire. As long as the data returned is in the same format as our V2 API, `preview.js` should consume without knowing where it came from.

`preview.js` on `show()` could initialize its networking layer based on some passed in option or a fallback from a network error.
```js
if (options.offline) {
    this.xhr = new Offline();
} else {
    this.xhr = new Xhr();
}```

and then inside `preview.js` and the viewers, one will use `this.xhr.get()` as opposed to the current `util.get()`

`getFile()` will have to revert back to `Xhr` if the file loaded ends up being a non-document (non-offline) file and call `this.load()` again to repeat the above process and go the online route or error out.

Obviously, this won't solve the prior mentioned issue where the networking is done inside `pdf.js` or `shaka-player.js` to load the actual representation data (unless the offline links being provided are actually URLs that open without a internet connection, like from localhost).

Comment (disist) [inline src/lib/Preview.js]: Thank you guys for the fast and deliberate response. 

@tonyjin I also agree that this solution requires rethinking and you approach sound good. 
After better understanding about `addOriginalRepresentation` method, it made me think that we can enrich file info with offline file links on our side (property `authenticated_download_url` also, which base for creating original representation). So we can download all content files on our side after that put there to the storage and enrich `file` with their offline links. 

@priyajeet Seems like your solution requires some refactoring which can affect core part of box previewer so it will be overtime for me, but I like some solutions.  

So, guys, I've updated pull request with the latest rethinking, what you think?

**Comment by disist** [comment]
@tonyjin Thanks for the fast response. 

> Also, how does your offline storage work? I would double check that pdf.js (which our document viewer uses) properly handles the offline links.
Yes, we checked our solution with offline links. The `pdfjs` loads from offline link correctly (besides, he ignores all query parameters). 

So, please not merge this pull request, this is like the proposal, only for discussion.

**Comment by priyajeet** [comment]
@disist Hey, thanks for the comments and the patch updates. I am still a bit confused, can you explain what exactly you are doing here, as in how the whole online/offline workflow will happen? I see the public `fetchFileInfo` method. But who is calling that? Is that being called externally from your application? Where are the auth tokens coming from in the offline case? How is the actual content (image or pdf or video binaries) being stored offline? How is `fetchFileInfo` providing the offline file to preview? I also see your stuff modifying one file (the current file), but preview also works over a collection of files and does pre-fetching, where and how will those files will be fetched/pre-fetched?

Box at the moment doesn't do offline files outside of native apps (and native apps have their own preview). On the web it was never prioritized mostly because of security and compliance, and this still holds true. Unless I am misunderstanding something, at the moment your patch seems very specific to your use case. If offline storage needs to be implemented the correct way, it has to be done by separating out the network layer so that core preview code is agnostic of where the data comes from. So that it works for all cases, not just for this specific one.

Alternatively, if you want to attempt offline with least modifications to preview, you should manage the `file` object in your own app. In order to store the `file` object within your storage you should either make your own V2 API request to get it before even using preview OR you can store the `file` object after it has been fetched from our server by listening to preview's `load` event after the 1st online preview.

Now that you have the `file` object, and somehow your app detects being offline, you should enhance this object with your offline URLs. And then pass this object to preview's show method indicating offline. Currently the `show()` method takes in a fileId, it will need to be modified to take the whole `file` object. Roughly, something along these lines:

```js
    show(file, undefined, { offline: true });
```
```js
    show(file, token, options = {}) {
        const isValidFile = checkFileValid(file);
        const { offline } = options;

        if (offline && isValidFile) {
            token = 'bogus';
        }

        // Save a reference to the options to be used later
        if (typeof token === 'string' || typeof token === 'function') {
        } else if (token) {
            this.previewOptions = Object.assign({}, token || {});
        } else {
            throw new Error('Missing access token!');
        }

        // load the preview
        this.load(file);
    }

    load(file) {
        ...
        // Use cached file data if available, otherwise create empty file object
        this.file = typeof file === 'object' ? file : this.cache.get(file) || { id: file };
        ...
    }

    loadFromCache() {
        ...
        // Also refresh from server to update cache
        if (!this.options.offline) {
            this.loadFromServer();
        }
    }

```

Other than that code, some offline guards will have to be placed in other areas that do a network call.

tonyjin:
Hey @disist I just had a chat with @priyajeet and we may have a way forward for you!

Can you please share with us your requirements or use case and the call pattern you'd be expecting to make? Also in general, we're trying to find accept features that will be useful for all Box customers (web application and platform), anything more specific to one use case we'd recommend a fork of this repo. In this case and given our current understanding of what you want to do, this is a suggestion that we can pull in:

- Modify `preview.show()` to accept either a fileId (current behavior) or well-formed file object. Given a well-formed file object, the token is then unnecessary.
- Add an option for preview to `skipServerUpdate` or something similar. Right now, we try loading a file object from our cache and then make a call to `loadFromServer()` to update that cache in case it is out of date. With your use case, you'd pass in `skipServerUpdate`, which then skips this check since you're passing in a file object with offline URLs

We can then look at either adding a util method for `getFileInfo()`, similar to what you have for `fetchFileInfo()`, but with a function signature closer to `prefetch()` since we also potentially need shared name and shared name password in order to fetch file info. Once you have this file object, you can save it and enrich it with offline URLs on your end.

I think doing the above would enable your use case while also preventing the need to add any reference to `offline` in our code base.

disist:
Hello guys, thanks for the answers.

Our main goal to preview documents (no video, just pdf, doc, ect) via `box-content-preview` when we have no internet. 

@priyajeet 

> can you explain what exactly you are doing here, as in how the whole online/offline workflow will happen?

So, once when we need to have an ability to offline preview, we use `fetchFileInfo` method to get `file` object and store it in our secure storage, and after (when we have no internet and want to preview file) we use `show` method and pass `file` object (which was stored on our side and enriched by offline links to files, which were downloaded for each representation), so with this approach we can preview file when we have no internet. 

> I see the public fetchFileInfo method. But who is calling that?

Yes, we calling this method for getting `file` info object and store it in our secure storage to pass this object into `show` method

> Where are the auth tokens coming from in the offline case?

So, it's very interesting question, we store it in our secure storage, and pass it for following default load flow with usage `getTokens(this.file.id, this.previewOptions.token)`, but I would love to skip this step and not use token anymore when we pass `file` object directly. 

> How is the actual content (image or pdf or video binaries) being stored offline?

On our side, we will download all files from template URL (in `representation` and download URL)  and store it locally in mobile phone file system, and we will have offline link to this file

>  I also see your stuff modifying one file (the current file), but preview also works over a collection of files and does pre-fetching, where and how will those files will be fetched/pre-fetched?

So, seems like I not familiar with all behaviors, but I tested this offline approach and everything is working. Could you please explain, in which cases the files will be fetched/pre-fetched?

@tonyjin 
> Can you please share with us your requirements or use case and the call pattern you'd be expecting to make?

So, our main use case: 
Given: 
 - User open our application
 - User goes to the file list
 - User marks file "as available for offline" (without previewing) 
When: 
 - When we have no internet
Then:
 - User can preview this file

Much likely that in future, business will want make files available offline by default (when you preview file online, you will be able to preview file offline)

> We can then look at either adding a until method for getFileInfo(), similar to what you have for fetchFileInfo(), but with a function signature closer to prefetch() since we also potentially need shared name and shared name password in order to fetch file info. Once you have this file object, you can save it and enrich it with offline URLs on your end.

Yep, that's I do. First of all, I looked into `prefetch` method, but as I understand this method works a bit another, it works with `viewerInstance.prefetch`. 

So, guys, lets finilize steps, what we need to do:
 - Rename `fetchFileInfo` to `getFileInfo`
 - Remove all references `offline` and will pass `file` option directly into the options with a well-formed file object
 - Add option `skipServerUpdate`, it will be prevent a call to `loadFromServer()` 
 - Do that the token will be unnecessary when we pass `file` object 

Right? 


Sorry for the delay it was a vacation :)

priyajeet:
Those steps seem accurate.

Also, when someone marks file "as available for offline", is it marked on just 1 file OR can be it marked on an entire folder too? If I were implementing that, I would you just do a V2 API request of your own to cache the file (without involving preview at all). Because you can just cache the entire folder instead. Might be more performant (1 request vs many) depending upon the use case. For example if you do this:

```js
fetch('https://api.box.com/2.0/folders/FOLDERID?fields=permissions,shared_link,sha1,file_version,name,size,extension,representations,watermark_info,authenticated_download_url', {
    headers: {
        Authorization: 'Bearer TOKEN'
    }
})
```

That will get you all the items in that folder with those fields (see `item_collection` in the response), and you can cache them all (or one by one depending upon what the user selected). Now that you have all the files in memory/cached, you can just pass the `file` object in preview (after making the changes mentioned above) and things should hopefully just work.

Only downside of this approach is you need to know what ?fields= to fetch, its hardcoded above. So if we change those fields (additive change) in the future, you will have to update your code to include the new fields. But then, I assume you will not just start using a new version of preview without updating your codebase, so this should not be a problem.

disist:
Great, just one clarification:

@priyajeet 
> is it marked on just 1 file OR can be it marked on an entire folder too?

Yep, we will cache entire folder too, great API, thanks.  
So, More and more come to mind that we could move `getFileInfo()` logic on our side and get file info or folder info and remove this public method on your side. What do you think?

priyajeet:
Yeah, that will be better.

disist:
So, guys! @priyajeet @tonyjin 
Are you ok with this implementation?

PS: The tests and build I will fix

PR #256, opened 2017-07-28
Fix: Load annotator with the correct initial scale

MinhHNguyen:
Annotator is currently always being loaded with scale 1. When the true scale is set, the annotations are not re-rendered. As a result, annotations on page load will always be the same size.

Proposed changes:
-  annotator.setScale() returns this (annotatorInstance) so that calls can be used in a builder format. ie annotator.setScale().init(), unfortunately prettier formats these calls all on one line. Might be useful to anyone who has forked the repository.
-  Update the viewer scale just before loading the annotations
-  Assign annotatedElement in Annotator constructor rather than init() so that annotatorInstance.setScale(trueScale) can be called before annotator.init(). Init() will render the annotations and require a rescale. As init() appears to be called right after constructing Annotator instances, I do not see an issue with this. In addition, child classes will be able to use this.annotatedElement.

### DIFF
--- a/src/lib/annotations/Annotator.js
+++ b/src/lib/annotations/Annotator.js
@@ -61,11 +61,6 @@ class Annotator extends EventEmitter {
         this.isMobile = data.isMobile;
         this.previewUI = data.previewUI;
         this.annotationModeHandlers = [];
-        this.annotatedElement = this.getAnnotatedEl(this.container);
-
-        if (!this.annotatedElement) {
-            throw new Error('No annotatable element found.');
-        }
     }
 
     /**
@@ -91,9 +86,11 @@ class Annotator extends EventEmitter {
     /**
      * Initializes annotator.
      *
+     * @param {number} [initialScale] - The initial scale factor to render the annotations
      * @return {void}
      */
-    init() {
+    init(initialScale = 1) {
+        this.annotatedElement = this.getAnnotatedEl(this.container);
         this.notification = new Notification(this.annotatedElement);
 
         const { apiHost, fileId, token } = this.options;
@@ -110,7 +107,7 @@ class Annotator extends EventEmitter {
             this.setupMobileDialog();
         }
 
-        const scale = annotatorUtil.getScale(this.annotatedElement);
+        const scale = initialScale;
         this.setScale(scale);
         this.setupAnnotations();
         this.showAnnotations();
--- a/src/lib/annotations/Annotator.js
+++ b/src/lib/annotations/Annotator.js
@@ -61,6 +61,7 @@ class Annotator extends EventEmitter {
         this.isMobile = data.isMobile;
         this.previewUI = data.previewUI;
         this.annotationModeHandlers = [];
+        this.annotatedElement = this.getAnnotatedEl(this.container);
     }
 
     /**
@@ -89,7 +90,6 @@ class Annotator extends EventEmitter {
      * @return {void}
      */
     init() {
-        this.annotatedElement = this.getAnnotatedEl(this.container);
         this.notification = new Notification(this.annotatedElement);
 
         const { apiHost, fileId, token } = this.options;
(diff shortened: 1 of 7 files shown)


src/lib/annotations/Annotator.js
```diff
@@ -89,7 +90,6 @@ class Annotator extends EventEmitter {
      * @return {void}
      */
     init() {
-        this.annotatedElement = this.getAnnotatedEl(this.container);
```
pramodsum: Maybe throw an error here if a developer forgets to override getAnnotatotedEl() in their viewer specific annotator

src/lib/annotations/Annotator.js
```diff
@@ -236,10 +237,11 @@ class Annotator extends EventEmitter {
      * Sets the zoom scale.
      *
      * @param {number} scale - current zoom scale
-     * @return {void}
+     * @return {Annotator} Annotator instance returned for chaining
      */
     setScale(scale) {
         this.annotatedElement.setAttribute('data-scale', scale);
+        return this;
```
JustinHoldstock: I understand the benefits of chaining, but since we already have access to the annotator instance, what is the benefit here over just calling `instance.whatever()` after calling `instance.setScale()`?

My thought process is that it goes against any coding paradigms we currently follow, in Preview

src/lib/viewers/BaseViewer.js
```diff
@@ -344,7 +344,10 @@ class BaseViewer extends EventEmitter {
         // Add a resize handler for the window
         document.defaultView.addEventListener('resize', this.debouncedResizeHandler);
 
-        this.addListener('load', () => {
+        this.addListener('load', (event) => {
+            // this.scale set to 1 if event.scale does not exist
+            ({ scale: this.scale = 1 } = event);
```
JustinHoldstock: I know that's not how it works now, but wouldn't it make more sense if set scale set `this.scale` as well as set the attribute on the annotated element?

If so, then move to 

```
const { scale } = event.scale;
//...
this.annotator.setScale(scale);
```
and `setScale()` can be

```
setScale(scale = 1) { 
  this.scale = scale;
  // ... set attribute and stuff
}
```

src/lib/annotations/Annotator.js
MinhHNguyen: Anytime you can chain you will have access to the instance. Its just for readability so you can chain calls in a clear readable fashion.

```
NumberArray.setFirstFiveElements(5)
                      .setMultiplier(3)
                      .transform()
                      .collect()
```
As opposed to
```
NumberArray.setFirstFiveElements(5)
NumberArray.setMultiplier(3)
NumberArray.transform()
NumberArray.collect()
```
Very minute semantic idea, although I understand it is probably not a good idea to throw it into the mix now. Essentially your functions will build properties of your instance and a few getter methods will actually display the transformed properties.

src/lib/viewers/BaseViewer.js
MinhHNguyen: That would indeed make more sense and I wish it would work. Issue here is as follows:

BaseViewer gets scale from 'scale' event from BaseViewer child class
Annotator gets scale from data-scale which it also sets. Which creates a cold start problem. Annotator.init() creates the annotatedElement and sets data-scale to be 1 at the time of initialization. init() then renders the annotations. Afterwards, data-scale is correctly set through BaseViewer's scale listener doing Annotator.setScale(). At this point, annotations have been shown and they are not redrawn.

MinhHNguyen: I originally misread what you meant. 

Problem here is that setScale is in Annotator and the eventListener is in BaseViewer.

MinhHNguyen: We can make this more clear by doing 
```
if(event && event.scale) {
        this.scale = event.scale;
}
```
since this.scale is set to 1 by the property default.

src/lib/annotations/Annotator.js
tonyjin: When in Rome :) Good discussion around patterns here but since we don't utilize that sort of chaining in Preview, please add it in a feature change.

MinhHNguyen:
Decided it was not good to rely on an abstract method in the constructor. Instead the initial scale will be passed in as an option for the init() method.

PR #257, opened 2017-07-28
New: add blank annotation icon to allow native PDF annotations to be shown

jeremypress:
In the short term this will fix the mouse over detection for any Native PDF annotations that rely on their own built in icon.

In the long term, we can add our own icons for the various types of native annotations that PDF.js supports. We can then support native PDF annotations, which we are currently hidden via CSS.


tonyjin:
Could you implement this by setting the PDF.js image path using the existing version and just adding the empty SVG to the 0.130.0 directory? I ask since every update to pdf.js cache-busts, which isn't a huge deal but if this can be done without doing so that's better!

jeremypress:
@tonyjin, yep, I can do that.

PR #260, opened 2017-07-28
Chore: Duplicating any util.js methods used into annotatorUtil.js

pramodsum:
- Removing any reference to files/methods in preview.js


MinhHNguyen:
Looks good, I would just suggest copying the tests from util-test into annotatorUtil-test

PR #261, opened 2017-07-29
Fix: Emit scale does not always pass events

MinhHNguyen:
Thanks to @pramodsum for the catch.
Fancy destructure syntax was no good here.

### DIFF
--- a/src/lib/viewers/__tests__/BaseViewer-test.js
+++ b/src/lib/viewers/__tests__/BaseViewer-test.js
@@ -284,8 +284,9 @@ describe('lib/viewers/BaseViewer', () => {
             sandbox.stub(document.defaultView, 'addEventListener');
             sandbox.stub(base, 'loadAnnotator');
             base.annotationsPromise = {
-                then: () => {
+                then: (arg) => {
                     expect(base.scale).to.equal(1.5);
+                    expect(arg).to.equal(base.loadAnnotator);
                     done();
                 }
             };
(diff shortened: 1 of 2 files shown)


src/lib/viewers/__tests__/BaseViewer-test.js
```diff
@@ -278,6 +278,36 @@ describe('lib/viewers/BaseViewer', () => {
...
+            sandbox.stub(fullscreen, 'addListener');
+            sandbox.stub(document.defaultView, 'addEventListener');
+            sandbox.stub(base, 'loadAnnotator');
+            base.annotationsPromise = {
+                then: () => {
+                    done();
+                }
+            };
+
+            base.addCommonListeners();
```
pramodsum: Not sure this is actually checking any expectations. I'd just do something like
```
this.annotationPromise = {};
base.emit('load', ....);
expect(base.loadAnnotor).to.be.called;
```
etc. for w/e is called and check that scale is set if it exists in event.scale

MinhHNguyen: The expectation is that the annotations promise is called. done() is only resolved when the annotations promise is called. Unfortunately we cannot expect base.loadAnnotator to be called since events are async.

I've added a check to ensure that base.scale is set and that base.loadAnnotator is passed into the promise.then function.

pramodsum: Oops you're right. Totally missed that! This works 👍

PR #262, opened 2017-07-29
Chore: Initializing notifications in viewer rather than annotator

pramodsum:
- This will allow us to display other types of notifications should we have the need

### DIFF
--- a/src/lib/PreviewUI.js
+++ b/src/lib/PreviewUI.js
@@ -292,7 +292,7 @@ class PreviewUI {
     /**
      * Shows a notification message.
      *
-     * @private
+     * @public
      * @param {string} message - Notification message
      * @param {string} [buttonText] - Optional text to show in button
      * @return {void}
@@ -304,7 +304,7 @@ class PreviewUI {
     /**
      * Hides the notification message. Does nothing if the notification is already hidden.
      *
-     * @private
+     * @public
      * @return {void}
      */
     hideNotification() {
--- a/src/lib/PreviewUI.js
+++ b/src/lib/PreviewUI.js
@@ -1,5 +1,6 @@
 import ProgressBar from './ProgressBar';
 import shellTemplate from './shell.html';
+import Notification from './Notification';
 import { insertTemplate } from './util';
 import {
     CLASS_HIDDEN,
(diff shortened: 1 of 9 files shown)


pramodsum:
@JustinHoldstock So I thought about this and I think that the notification should be kept inside PreviewUI.js rather than Preview.js

This does mean that we will instantiate a notification on a per viewer basis still but it seems to make the most sense for it to live there with other UI components such as the progress bar, etc

JustinHoldstock:
@pramodsum Are we creating a new instance of the Loading bar and spinner for every viewer?

src/lib/PreviewUI.js
```diff
@@ -285,6 +289,28 @@ class PreviewUI {
         this.progressBar.finish();
     }
 
+    /**
+     * Shows a notification message.
+     *
+     * @private
```
JustinHoldstock: Should be public, as it's being called in Preview.js (and needs to be called by anything else using it)

Ditto to `hideNotification()`

src/lib/annotations/Annotator.js
```diff
@@ -725,9 +718,29 @@ class Annotator extends EventEmitter {
...
+     * @protected
+     * @emits viewerevent
+     * @param {string} event - Event name
+     * @param {Object} data - Event data
+     * @return {void}
+     */
+    emit(event, data) {
+        const { annotator, fileId } = this.options;
+        super.emit(event, data);
+        super.emit('annotationevent', {
```
JustinHoldstock: What's the reason we're emitting double events?

pramodsum:
Actually I look another look. Seems like we're instantiating one instance of PreviewUI in Preview.js so that should apply for all viewers.

src/lib/annotations/image/ImageAnnotator.js
```diff
@@ -105,7 +105,7 @@ class ImageAnnotator extends Annotator {
             type
         };
 
-        if (!annotatorUtil.validateThreadParams(threadParams)) {
+        if (annotatorUtil.validateThreadParams(threadParams)) {
```
JustinHoldstock: Does this mean that if `validateThreadParams()` returns `true`, that it isn't valid?

pramodsum: Ah! I was forcing this to trigger the emission of annotation errors. Will fix that up

PR #265, opened 2017-07-31
Fix: Don't trigger highlight if a new highlight is being created

pramodsum:
- Prevents other highlight dialogs from being triggered while a shared dialog for creating highlight annotations is already open


MinhHNguyen:
Any way we can leverage a pending thread state here instead of adding a new variable?

pramodsum:
isCreatingHighlight already exists actually. We set/unset it during mouseDown/mouseUp to prevent any annotations from triggering while you're dragging the mouse and highlighting text

JustinHoldstock:
And it's used BEFORE any state exists for an annotation(because the annotations don't exist yet)

PR #267, opened 2017-08-01
Feature: drawing annotation scaling

MinhHNguyen:
Changes
- Use lodash's round function to truncate to a fixed number of decimal places
- requestAnimationFrame is kicked off by handlestart and ended when the drawingflag is no longer 'draw'
- We now use two canvases: a canvas for the committed drawings and a canvas for a drawing in progress (drawing-canvas-mem). In the future, I plan to cut out the boundary of the pending drawing and paste it into the commit canvas, thereby potentially saving on the number of bytes copied.
- Added in DocDrawingThread show(), which is similar to render except render only displays pending drawings whereas show displays committed and pending drawings. In the future, I will investigate how to combine shared logic in both functions. show() is called by the Annotator when it is emitted by a Viewer.
- A drawing path now stores the path as an array of document coordinates. Before drawing, the browser coordinates (browserPath) needs to be computed from the document coordinates. addCoordinate now takes a document coordinate has the option to be given a browser coordinate that will be inserted into the browserPath.


src/lib/annotations/doc/DocDrawingThread.js
```diff
@@ -1,78 +1,88 @@
...
+     * @return {DocDrawingThread} Drawing annotation thread instance
+     */
+    constructor(data) {
+        super(data);
+
+        this.handleStart = this.handleStart.bind(this);
+        this.handleMove = this.handleMove.bind(this);
+        this.handleStop = this.handleStop.bind(this);
+        this.checkAndHandleScaleUpdate = this.checkAndHandleScaleUpdate.bind(this);
+        this.reconstructBrowserCoordFromLocation = this.reconstructBrowserCoordFromLocation.bind(this);
```
MinhHNguyen: :(

src/lib/annotations/Annotator.js
```diff
@@ -321,12 +323,25 @@ class Annotator extends EventEmitter {
...
 
-            const thread = this.createAnnotationThread([], {}, TYPES.draw);
             this.unbindDOMListeners();
-            this.bindDrawModeListeners(thread, postButtonEl);
+
+            const thread = this.createAnnotationThread([], {}, TYPES.draw);
+            if (thread) {
+                this.bindCustomListenersOnThread(thread);
+                this.bindDrawModeListeners(thread, postButtonEl);
+                // TODO: @minhnguyen deal with page changes in a better way
```
jeremypress: Move this to a trello card.

src/lib/annotations/doc/DocDrawingThread.js
```diff
@@ -1,78 +1,88 @@
...
-import { getScale } from '../annotatorUtil';
 import DrawingPath from '../drawing/DrawingPath';
 import DrawingThread from '../drawing/DrawingThread';
 import * as docAnnotatorUtil from './docAnnotatorUtil';
+import * as annotatorUtil from '../annotatorUtil';
 
-class DocDrawingThread extends DrawingThread {
-    /** @property {number} - Drawing state */
-    lastPage;
+const CLASS_ANNOTATION_LAYER_DRAW_MEMORY = `${CLASS_ANNOTATION_LAYER_DRAW}-mem`;
```
jeremypress: Doesn't hurt to add a new constant

src/lib/annotations/doc/DocDrawingThread.js
jeremypress: @tonyjin @JustinHoldstock @pramodsum Is this what we decided on going forward?

jeremypress: @JustinHoldstock is this what we decided on going forward?

src/lib/annotations/doc/DocDrawingThread.js
```diff
@@ -88,6 +98,132 @@ class DocDrawingThread extends DrawingThread {
...
+                CLASS_ANNOTATION_LAYER_DRAW,
+                PAGE_PADDING_TOP,
+                PAGE_PADDING_BOTTOM
+            );
+            this.setContextStyles(config, context);
+        }
+
+        // Draw the paths to the annotation layer canvas
+        /* eslint-disable require-jsdoc */
+        const scaleAndDraw = (drawing) => {
```
jeremypress: Break this outside?

src/lib/annotations/drawing/DrawingPath.js
```diff
@@ -18,21 +23,39 @@ class DrawingPath {
...
      * @return {void}
      */
-    addCoordinate(xPos, yPos) {
-        if (!xPos || !yPos) {
+    addCoordinate(documentLocation, browserLocation) {
+        if (!documentLocation || !documentLocation.x || !documentLocation.y) {
             return;
         }
 
         // OPTIMIZE (@minhnguyen): We convert a number to a string using toFixed and then back a number.
```
jeremypress: Is this optimized now?

MinhHNguyen: Unfortunately not, the native function to truncate is number.toFixed(precision) which provides a string.

src/lib/annotations/Annotator.js
pramodsum: Can we only bind these listeners if draw annotations are enabled?

src/lib/annotations/Annotator.js
```diff
@@ -321,12 +323,25 @@ class Annotator extends EventEmitter {
...
+                thread.addListener('drawthreadcommited', () => {
```
pramodsum: Same thing w/ this

src/lib/annotations/Annotator.scss
```diff
@@ -325,6 +325,7 @@ $avatar-color-9: #f22c44;
 // CSS for highlights
 //------------------------------------------------------------------------------
 .bp-annotation-layer-draw,
+.bp-annotation-layer-draw-mem,
```
pramodsum: rename this to something a little clearer

src/lib/annotations/annotatorUtil.js
```diff
@@ -388,6 +388,23 @@ export function eventToLocationHandler(locationFunction, callback) {
...
 
+/**
+ * Create a JSON object containing x/y coordinates and optionally dimensional information
+ *
+ * @param {number} x - The x position of the location object
+ * @param {number} y - The y position of the location object
+ * @param {Object} [dimensions] - The dimensional information of the location object
+ * @return {Object} - A location object with x/y position information as well as provided dimensional information
+ */
+export function createLocation(x, y, dimensions) {
```
pramodsum: I'm not sure we need a separate method for this? Is there a strong reason for breaking this out into a method?

src/lib/annotations/doc/DocAnnotator.js
```diff
@@ -308,11 +308,14 @@ class DocAnnotator extends Annotator {
             thread = new DocDrawingThread(threadParams);
         } else if (type === TYPES.point) {
             thread = new DocPointThread(threadParams);
-        } else {
```
pramodsum: Maybe instead of an error (or getting rid of that error altogether) we can do `this.emit('annotationerror', __('annotations_create_error'));` and return, which will trigger a notification in Preview informing the user that their annotation failed to create?

src/lib/annotations/doc/DocAnnotator.js
```diff
@@ -308,11 +308,14 @@ class DocAnnotator extends Annotator {
...
-            throw new Error(`Unhandled document annotation type: ${type}`);
         }
 
-        this.addThreadToMap(thread);
+        if (!thread && this.notification) {
```
pramodsum: see note above

src/lib/annotations/doc/DocDrawingThread.js
```diff
@@ -1,78 +1,88 @@
...
-            CLASS_ANNOTATION_LAYER_DRAW,
-            PAGE_PADDING_TOP,
-            PAGE_PADDING_BOTTOM
-        );
+    handleStart(location) {
+        const pageChanged = this.hasPageChanged(location);
+        if (pageChanged) {
+            this.handleStop();
+            this.saveAnnotation();
+            this.emit('drawthreadcommited');
```
pramodsum: Should we be creating/committing the draw annotation on page change or just prevent any drawing on other pages, keeping w/ the idea that users must intentionally commit their annotations

src/lib/annotations/doc/DocDrawingThread.js
```diff
@@ -1,78 +1,88 @@
...
+        // Assign a location and dimension to the annotation thread
+        if (!this.location || (this.location && !this.location.page)) {
+            this.location = {
+                page: location.page,
+                dimensions: location.dimensions
+            };
+            this.checkAndHandleScaleUpdate();
+        }
 
+        this.drawingFlag = STATES_DRAW.draw;
```
pramodsum: Rename STATES_DRAW to DRAW_STATES?

src/lib/annotations/doc/DocDrawingThread.js
```diff
@@ -88,6 +98,132 @@ class DocDrawingThread extends DrawingThread {
...
+        }
+
+        // Get DrawingPath objects to be shown
+        const drawings = this.getDrawings();
+        if (this.pendingPath && !this.pendingPath.isEmpty()) {
+            drawings.push(this.pendingPath);
+        }
+
+        this.checkAndHandleScaleUpdate();
+        // Get the annotation layer context to draw with
```
pramodsum: new line before comment

src/lib/annotations/doc/DocDrawingThread.js
```diff
@@ -88,6 +98,132 @@ class DocDrawingThread extends DrawingThread {
...
+        let context;
+        if (this.state === STATES.pending) {
+            context = this.drawingContext;
+        } else {
+            this.pageEl = docAnnotatorUtil.getPageEl(this.annotatedElement, this.location.page);
```
pramodsum: When would you be trying to save a draw annotation that isn't in the pending state?

src/lib/annotations/doc/DocDrawingThread.js
```diff
@@ -88,6 +98,132 @@ class DocDrawingThread extends DrawingThread {
...
+            this.setContextStyles(config, context);
+        }
+
+        // Draw the paths to the annotation layer canvas
+        /* eslint-disable require-jsdoc */
+        const scaleAndDraw = (drawing) => {
+            drawing.generateBrowserPath(this.reconstructBrowserCoordFromLocation);
+            drawing.drawPath(context);
+        };
+        /* eslint-enable require-jsdoc */
```
pramodsum: new line before this eslint comment

src/lib/annotations/drawing/DrawingPath.js
```diff
@@ -1,10 +1,15 @@
+import * as annotatorUtil from '../annotatorUtil';
```
pramodsum: import only what you need from annotatorUtil

src/lib/annotations/drawing/DrawingThread.js
```diff
@@ -96,55 +112,65 @@ class DrawingThread extends AnnotationThread {
...
     /**
-     * Set the drawing styles
+     * Set the drawing styles for a provided context. Sets the context of the memory context if
+     * no other context is provided.
      *
      * @protected
      * @param {Object} config - The configuration Object
      * @param {number} config.scale - The document scale
      * @param {string} config.color - The brush color
+     * @param {CanvasContext} [otherContext] - Optional context to be set
```
pramodsum: The naming of otherContext is a bit confusing. Is it a new canvas context from the new drawing or the memory context? I'm assuming this.drawingContext is the memory context so might make sense to rename otherContext to something a little descriptive

src/lib/annotations/doc/DocAnnotator.js
pramodsum: See the error in docAnnotator.getLocationFromEvent()

src/lib/annotations/annotatorUtil.js
MinhHNguyen: This is sort of like a namedtuple, purely for convenience. I don't think it is detrimental to the codebase but if there is a strong reason against including it then this method can be removed.

src/lib/annotations/doc/DocDrawingThread.js
MinhHNguyen: For now, I will be using the model of soft commiting any drawings done on the page assigned to the drawing thread and creating a new thread assigned to the new page.

src/lib/annotations/doc/DocDrawingThread.js
MinhHNguyen: show() is called when rescaling is done as well as when a drawing thread is initially created. This control block determines whether to draw the drawing to the in-progress context or to the commited drawing context.

src/lib/annotations/doc/DocDrawingThread.js
MinhHNguyen: improvements coming with the undo and redo pr. For now I will be moving this directly to the drawings.forEach call.

src/lib/annotations/drawing/DrawingThread.js
MinhHNguyen: The otherContext was meant to be any other provided context to be drawn to. In the case of this PR, the other context was intended to be the commited drawings context.

In any case, this will be done away with if SVGs are used in place of canvas drawings. For now, otherContext will be renamed to context (ambiguity intended as it should be able to style any context provided)

src/lib/annotations/annotationConstants.js
```diff
@@ -62,7 +63,7 @@ export const SELECTOR_DIALOG_CLOSE = `.${CLASS_DIALOG_CLOSE}`;
 export const SELECTOR_HIGHLIGHT_BTNS = `.${CLASS_HIGHLIGHT_BTNS}`;
 export const SELECTOR_ADD_HIGHLIGHT_BTN = `.${CLASS_ADD_HIGHLIGHT_BTN}`;
 
-export const STATES_DRAW = {
+export const DRAW_STATES = {
     draw: 'draw',
```
pramodsum: Thoughts on making these progressive tense? e.g. `DRAW_STATES.drawing`, `DRAW_STATES.erasing`

MinhHNguyen: Sure, and idle would stay as 'idle' or would it become 'idling'?

MinhHNguyen:
Merged the refactor from master.

src/lib/annotations/doc/DocDrawingThread.js
JustinHoldstock: I'll have @tonyjin and @pramodsum weigh in on this. I prefer this, but it does seem cumbersome

src/lib/annotations/Annotator.js
```diff
@@ -205,6 +204,17 @@ class Annotator extends EventEmitter {
     }
 
     /**
+     * Returns true if the annotator has an annotation type enabled
+     *
+     * @param {string} type - Annotation type to check
+     * @return {boolean} Whether or not the annotation type is enabled
+     */
+    isTypeEnabled(type) {
```
JustinHoldstock: Bump @MinhHNguyen

MinhHNguyen: Yeah I initially added it as a check before isModeAnnotatable was there. Gonna remove this

src/lib/annotations/Annotator.js
```diff
@@ -663,6 +673,7 @@ class Annotator extends EventEmitter {
             });
         } else if (mode === TYPES.draw) {
             const drawingThread = this.createAnnotationThread([], {}, TYPES.draw);
```
MinhHNguyen: I don't think that keeps in line with the pending -> commit pattern. But I am creating a new DrawingController class to manage threads separately. And developers will be able to plug in their own controllers.

pramodsum: ^ yes that works for me.

src/lib/annotations/doc/DocDrawingThread.js
```diff
@@ -81,13 +90,139 @@ class DocDrawingThread extends DrawingThread {
...
+
+    /**
+     * Requires a DocDrawingThread to have been started with DocDrawingThread.start(). Reconstructs a browserCoordinate
+     * relative to the dimensions of the DocDrawingThread page element.
+     *
+     * @private
+     * @param {Location} documentLocation - The location coordinate relative to the document
+     * @return {Location} The location coordinate relative to the browser
+     */
+    reconstructBrowserCoordFromLocation(documentLocation) {
```
MinhHNguyen: I think this is not generic. This is a really specific case where we need to create a location from the thread information and use docAnnotatorUtil.getBrowserCoordinatesFromLocation(). Its sort of a shallow wrapper for creating a location, passing a location into the docannotatorutil function and then creating a new location based on the return of docAnnotatorUtil.getBrowserCoord.

MinhHNguyen:
@pramodsum removed the isTypeEnabled method!

PR #268, opened 2017-08-01
Chore: Remove code supporting token in options object

tonyjin:
Remove code that supports access token in the options object - this has been deprecated for some time now.


tonyjin:
Please do not merge yet! Need to confirm that this change doesn't break anyone regardless of how long it's been deprecated.

jeremypress:
Is this good to go?

tonyjin:
@jeremypress not yet, haven't had a chance to confirm that this is okay.

PR #269, opened 2017-08-01
Chore: Hide viewer controls when creating annotations

pramodsum:
- This prevents the viewer controls from being displayed in the following situations: 
1) user is creating an annotation and the shared highlight dialog is shown
2) user is adding the first annotation comment to a point annotation
3) user is adding the first annotation comment to a highlight-comment annotation

### DIFF
--- a/src/lib/annotations/image/ImageAnnotator.js
+++ b/src/lib/annotations/image/ImageAnnotator.js
@@ -85,6 +85,8 @@ class ImageAnnotator extends Annotator {
      * @return {AnnotationThread} Created annotation thread
      */
     createAnnotationThread(annotations, location, type) {
+        this.emit('annotationmodeexit');
+
         let thread;
 
         // Corrects any image annotation page number to 1 instead of -1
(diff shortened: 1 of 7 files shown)


src/lib/annotations/image/ImageAnnotator.js
```diff
@@ -85,6 +85,8 @@ class ImageAnnotator extends Annotator {
      * @return {AnnotationThread} Created annotation thread
      */
     createAnnotationThread(annotations, location, type) {
+        this.emit('annotationmodeexit');
```
JustinHoldstock: Can you emit what kind of mode was exited? I'm wondering if developers may want to hide/show different UI depending on what annotation mode is left/entered

jeremypress:
I thought we only hide the controls when explicitely in an annotation mode (point, draw, etc.)?

pramodsum:
@jeremypress the mode is technically exited on first click so this ensures that no other annotations can be triggered when you're still typing the comment for the annotation you just created

src/lib/annotations/image/ImageAnnotator.js
pramodsum: I already pass along w/e mode was entered, I can make sure to pass along what mode was exited as well

PR #270, opened 2017-08-02
Fix: Ensures that annotation listeners are removed/added on hide/show

pramodsum:
- Cleans up listeners in CommentBox.js on hide/show 
- This ensures that there will not be any interaction when creating highlight annotations on mobile

### DIFF
--- a/src/lib/annotations/__tests__/CommentBox-test.js
+++ b/src/lib/annotations/__tests__/CommentBox-test.js
@@ -108,6 +108,14 @@ describe('lib/annotations/CommentBox', () => {
         beforeEach(() => {
             // Kickstart creation of UI
             commentBox.show();
+
+            commentBox.cancelEl = {
+                removeEventListener: sandbox.stub()
+            };
+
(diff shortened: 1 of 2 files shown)


src/lib/annotations/__tests__/CommentBox-test.js
```diff
@@ -121,10 +129,22 @@ describe('lib/annotations/CommentBox', () => {
         it('should add the hidden class to the comment box element', () => {
             commentBox.hide();
```
MinhHNguyen: might be a good idea to expect commentbox.containerel.contains(CLASS_HIDDEN) to be false before calling commentbox.hide() as an extra check

pramodsum: We call commentBox.show() in the beforeEach() block of the unit test

pramodsum:
@JustinHoldstock can you confirm that this is ok with your mobile changes?

pramodsum:
Shouldn't be seeing this issue anymore

PR #272, opened 2017-08-02
Chore: Cleaning up annotations methods in BaseViewer

pramodsum:
Do NOT merge before https://github.com/box/box-content-preview/pull/271

### DIFF
--- a/src/lib/viewers/BaseViewer.js
+++ b/src/lib/viewers/BaseViewer.js
@@ -647,10 +647,21 @@ class BaseViewer extends EventEmitter {
             // Users can currently only view annotations on mobile
             this.canAnnotate = checkPermission(file, PERMISSION_ANNOTATE);
             if (this.canAnnotate) {
+                const annotationButtons = {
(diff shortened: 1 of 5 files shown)


src/lib/viewers/BaseViewer.js
```diff
@@ -724,48 +739,36 @@ class BaseViewer extends EventEmitter {
...
-     * @param {Function} handler - Point annotation button handler
+     * @param {string} currentMode - Annotation mode
      * @return {void}
      */
-    showPointAnnotateButton(handler) {
-        if (!this.isAnnotatable('point')) {
-            return;
-        }
+    showModeAnnotationButton(currentMode) {
+        const modes = {
```
JustinHoldstock: Move this out to a higher level constant. Doesn't need to be created every time we call this method

pramodsum: moved this one level higher. I'll add the unit tests if this is good to go

PR #277, opened 2017-08-04
Chore: Removing references to constants.js and Browser.js in Annotations code

pramodsum:


src/lib/annotations/Annotator.js
```diff
@@ -21,6 +17,8 @@ import {
     SELECTOR_ANNOTATION_BUTTON_DRAW_CANCEL,
     SELECTOR_ANNOTATION_BUTTON_DRAW_ENTER,
     SELECTOR_ANNOTATION_BUTTON_DRAW_POST,
+    SELECTOR_BOX_PREVIEW_BTN_ANNOTATE_POINT,
```
pramodsum: Ideally I don't want preview specific button css to exist in annotations. I will look into ways to make that configurable and passed in through annotationOptions or something

src/lib/viewers/BaseViewer.js
```diff
@@ -690,7 +690,8 @@ class BaseViewer extends EventEmitter {
             fileVersionId,
             isMobile: this.isMobile,
             locale: location.locale,
-            previewUI: this.previewUI
+            previewUI: this.previewUI,
+            modeButtons: ANNOTATION_BUTTONS
```
pramodsum: As per our discussion @JustinHoldstock

pramodsum: This will allow us to pass in any selector which will indicate the annotation button css selector in Preview aka ultimately platform and webapp can use different annotation button icons.

PR #286, opened 2017-08-09
Update: use pixelDeviceRatio to choose canvas pixel size

MinhHNguyen:
Suggestion from @JustinHoldstock for clearer drawings. Highlight annotations also use the getContext method from annotatorUtil. Quick test shows no issues.


src/lib/annotations/doc/docAnnotatorUtil.js
```diff
@@ -358,21 +358,32 @@ export function getContext(pageEl, annotationLayerClass, paddingTop, paddingBott
...
         const pageDimensions = pageEl.getBoundingClientRect();
         const pagePaddingTop = paddingTop || 0;
         const pagePaddingBottom = paddingBottom || 0;
-        annotationLayerEl.width = pageDimensions.width;
-        annotationLayerEl.height = pageDimensions.height - pagePaddingTop - pagePaddingBottom;
+        const pxRatio = window.devicePixelRatio || 1;
+        const width = pageDimensions.width;
+        const height = pageDimensions.height - pagePaddingTop - pagePaddingBottom;
+
+        annotationLayerEl.style.width = `${width}px`;
```
jeremypress: do we not have to scale the CSS?

MinhHNguyen: The canvas height and width should be the ratio * display height and width. Meaning if the ratio is 2 we have 2 canvas pixels but the element appears to be only 1.

src/lib/shell.html
```diff
@@ -13,8 +13,8 @@
                 </svg>
             </button>
             <button class="bp-btn-plain bp-btn-annotate-draw bp-is-hidden">
-                <svg class="bp-btn-annotate-draw-enter" width="22" height="21" viewBox="0 0 24 24" focusable="false">
-                  <path d="M7.127 22.562l-7.127 1.438 1.438-7.128 5.689 5.69zm1.414-1.414l11.228-11.225-5.69-5.692-11.227 11.227 5.689 5.69zm9.768-21.148l-2.816 2.817 5.691 5.691 2.816-2.819-5.691-5.689z"/>
+                <svg class="bp-btn-annotate-draw-enter" width="22" height="21" viewBox="0 0 14.88 14.88" focusable="false">
+                  <path d="M11.65,6.64,5.11,13.17.06,14.94,1.83,9.89,8.37,3.35Zm1.41-1.41L9.78,1.94,11.37.35a1,1,0,0,1,1.41,0l1.87,1.87a1,1,0,0,1,0,1.41Z" transform="translate(-0.06 -0.06)" fill-rule="evenodd"/>
```
jeremypress: ;)

PR #287, opened 2017-08-09
Feature: Drawing Annotations Undo Redo

MinhHNguyen:
Contains changes from #267 that need to be merged in first.

- Added an event observer function in annotator.bindModeListeners that will be moved to a new controller class in upcoming changes. The controller will use an RTree to contain each DrawingThread and its axis aligned bounding box. It might also help clean up some logic in annotator.
- Added undo and redo functionality for drawings that have not yet been committed


MinhHNguyen:
Unit tests are in

src/lib/annotations/Annotator.js
```diff
@@ -677,23 +690,53 @@ class Annotator extends EventEmitter {
...
-            handlers.push({
-                type: 'mouseup',
-                func: annotatorUtil.eventToLocationHandler(locationFunction, drawingThread.handleStop),
-                eventObj: this.annotatedElement
+            const undoButtonEl = this.previewUI.getAnnotateButton(SELECTOR_ANNOTATION_BUTTON_DRAW_UNDO);
+            const redoButtonEl = this.previewUI.getAnnotateButton(SELECTOR_ANNOTATION_BUTTON_DRAW_REDO);
+
+            // NOTE (@minhnguyen): Move this logic to a new controller class
+            const that = this;
+            drawingThread.addListener('annotationevent', (data = {}) => {
```
pramodsum: move this into a separate method + tests

src/lib/annotations/__tests__/Annotator-test.js
```diff
@@ -560,7 +536,7 @@ describe('lib/annotations/Annotator', () => {
                     'click',
                     sinon.match.func
                 );
-                expect(annotator.annotationModeHandlers.length).equals(4);
+                expect(annotator.annotationModeHandlers.length).equals(6);
```
pramodsum: Add tests for when only some of the buttons exist i.e postButtonEl, undoButton...etc

src/lib/annotations/doc/DocAnnotator.js
```diff
@@ -490,9 +490,11 @@ class DocAnnotator extends Annotator {
         // We need to redraw highlights on the page if a thread was deleted
         // since deleting 'cuts' out the highlight, which may have been
         // overlapping with another
-        thread.addListener('threaddeleted', () => {
-            this.showHighlightsOnPage(thread.location.page);
-        });
+        if (annotatorUtil.isHighlightAnnotation(thread.type)) {
```
pramodsum:
```
if (!annotatorUtil.isHighlightAnnotation(thread.type)) {
    return;
}

thread.addListener(.....
```

src/lib/annotations/doc/DocDrawingThread.js
```diff
@@ -116,16 +114,23 @@ class DocDrawingThread extends DrawingThread {
      * @return {void}
      */
     saveAnnotation(type, text) {
-        super.saveAnnotation(type, text);
+        this.emit('threadcleanup');
```
pramodsum: What's the reason for deleting the thread on save?

src/lib/annotations/drawing/DrawingContainer.js
```diff
@@ -0,0 +1,101 @@
...
+    }
+
+    /**
+     * Retrieve a JSON blob containing the number of undo and redo in each stack.
+     *
+     * @return {AvailableItemData} The number of undo and redo items available.
+     */
+    getNumberOfItems() {
+        return {
+            undo: this.undoStack.length,
```
pramodsum: maybe rename to undoCount and redoCount for clarity?

src/lib/annotations/doc/DocDrawingThread.js
MinhHNguyen: Clean-up removes the listeners, 'threaddeleted' deletes the thread from the thread map.

src/lib/annotations/Annotator.js
MinhHNguyen: This will be done in a separate PR. I don't think it makes sense to separate and test it at the moment when it will be moved in a coming class change.

src/lib/annotations/doc/DocDrawingThread.js
MinhHNguyen: Realized that bindings might be made to bind to 'threaddeleted' so I'll add a custom 'annotationevent' of type 'drawingcommit' to removeAllListeners('annotationevent'). We should be able to clean this up in the future once some logic is pulled out of Annotator.

PR #288, opened 2017-08-09
Chore: Cleaning up and consolidating annotation methods

pramodsum:
- Combines annotation mode specific handler methods in Annotator.js
- Fixes point annotation button so that it turns blue when in annotation mode
- Moves mode handler creation/destruction into Annotator.js
- Cleans up annotations methods in BaseViewer.js (removing as much as possible)


src/lib/annotations/Annotator.js
```diff
@@ -687,23 +695,15 @@ class Annotator extends EventEmitter {
...
      * @protected
-     * @return {boolean} True if drawing mode is on, otherwise returns false.
+     * @param {string} mode - Current annotation mode
+     * @return {boolean} Whether or not in the specified annotation mode
      */
-    isInDrawMode() {
-        return this.annotatedElement.classList.contains(CLASS_ANNOTATION_DRAW_MODE);
+    isInAnnotationMode(mode) {
+        return this.annotationMode === mode;
+        // return this.annotatedElement.classList.contains(this[`CLASS_ANNOTATION_${mode.toUpperCase()}_MODE`]);
```
MinhHNguyen: Remove this?

src/lib/annotations/Annotator.js
```diff
@@ -241,93 +240,113 @@ class Annotator extends EventEmitter {
...
+        this.emit('annotationmodeenter', mode);
+        this.annotatedElement.classList.add(CLASS_ANNOTATION_MODE);
+        if (buttonEl) {
+            buttonEl.classList.add(CLASS_ACTIVE);
 
-            if (buttonEl) {
+            const drawEnterEl = buttonEl.querySelector(SELECTOR_ANNOTATION_BUTTON_DRAW_ENTER);
+            const drawCancelEl = buttonEl.querySelector(SELECTOR_ANNOTATION_BUTTON_DRAW_CANCEL);
+            const postButtonEl = this.previewUI.getAnnotateButton(SELECTOR_ANNOTATION_BUTTON_DRAW_POST);
+            if (postButtonEl && drawEnterEl && drawCancelEl) {
```
MinhHNguyen: We can use annotatorUtil.show/hideElement on each buttonElement in this case. On `(postButtonEl && drawEnterEl && drawCancelEl)` we assume that all buttons will be included. This allows for custom implementations to include an enter and cancel but no post. Or post/enter/draw but no undo/redo and so forth.

pramodsum: You're right... I didn't really read what the method did while refactoring it haha

MinhHNguyen: Also, I believe the draw button showing should only be done on draw mode.

src/lib/annotations/__tests__/Annotator-test.js
```diff
@@ -6,7 +6,7 @@ import AnnotationService from '../AnnotationService';
 import {
     STATES,
     TYPES,
-    CLASS_ANNOTATION_POINT_MODE,
+    ,
```
MinhHNguyen: Extra ,

src/lib/annotations/Annotator.js
```diff
@@ -241,93 +240,112 @@ class Annotator extends EventEmitter {
...
+        this.emit('annotationmodeexit');
+        this.annotatedElement.classList.remove(CLASS_ANNOTATION_MODE);
+        if (buttonEl) {
+            buttonEl.classList.remove(CLASS_ACTIVE);
 
-        const drawButtonSelector = this.modeButtons[TYPES.draw].selector;
-        const buttonEl = event.target || this.previewUI.getAnnotateButton(drawButtonSelector);
-        const postButtonEl = this.previewUI.getAnnotateButton(SELECTOR_ANNOTATION_BUTTON_DRAW_POST);
+            const drawEnterEl = buttonEl.querySelector(SELECTOR_ANNOTATION_BUTTON_DRAW_ENTER);
+            annotatorUtil.hideElement(drawEnterEl);
```
MinhHNguyen: It is actually the case that the enterButton and drawButton are just SVGs under the same button. On mode change, we simply toggle the graphics.

On disable, we want to show the drawEnterEl (this button lets us enter draw mode), hide the cancelEl (exit draw mode without saving) and hide the postButtonEl (post is the hard commit button).

src/lib/annotations/Annotator.js
```diff
@@ -241,93 +240,112 @@ class Annotator extends EventEmitter {
...
+     * @return {void}
+     */
+    exitAnnotationModes(mode) {
+        Object.keys(this.modeButtons).forEach((type) => {
+            if (mode === type) {
+                return;
+            }
+
+            const buttonSelector = this.modeButtons[type].selector;
+            const buttonEl = event.target || this.previewUI.getAnnotateButton(buttonSelector);
```
MinhHNguyen: I don't think event is in the scope of this function. Might be a good idea to have modeButtons hold a tuple of modeName -> (selector, element). Otherwise previewUI.getAnnotateButton should be good enough.

src/lib/annotations/Annotator.js
```diff
@@ -541,25 +559,67 @@ class Annotator extends EventEmitter {
...
+                        drawingThread.saveAnnotation(TYPES.draw);
+                        this.toggleAnnotationHandler(TYPES.draw);
+                    },
+                    eventObj: postButtonEl
+                });
             }
-        ];
+        }
 
         handlers.forEach((handler) => {
```
MinhHNguyen: for each handler in handlers, we need to push the object to this.annotationModeHandler. That way it will be unbound when unbindModeHandlers is called. Handlers is a local array.

src/lib/annotations/Annotator.js
```diff
@@ -541,25 +559,67 @@ class Annotator extends EventEmitter {
...
-                type: 'mousedown',
-                func: pointFunc,
-                eventObj: this.annotatedElement
-            },
-            {
-                type: 'touchstart',
-                func: pointFunc,
-                eventObj: this.annotatedElement
+    bindModeListeners(mode) {
+        let handlers = [];
```
MinhHNguyen: for each handler in handlers, we need to push the object to this.annotationModeHandler. (line 624) That way it will be unbound when unbindModeHandlers is called. Handlers is a local array.

src/lib/annotations/Annotator.js
pramodsum: yea i wasn't really sure why it was being checked here. it seemed like it was always null. I'll just make use of getAnnotateButton

src/lib/annotations/Annotator.js
```diff
@@ -241,93 +240,116 @@ class Annotator extends EventEmitter {
...
+     * @param {string} mode - Current annotation mode
+     * @param {HTMLElement} buttonEl - Annotation button element
      * @return {void}
      */
-    toggleDrawAnnotationHandler(event = {}) {
-        this.destroyPendingThreads();
-        if (this.isInPointMode()) {
-            this.togglePointAnnotationHandler();
+    disableAnnotationMode(mode, buttonEl) {
+        this.emit('annotationmodeexit');
```
JustinHoldstock: Could we move `annotationmodeexit`, "enter" and any other message strings to a constant we can reference around the code base? Or are they ever only used in one place, each?

src/lib/annotations/Annotator.js
```diff
@@ -541,25 +563,66 @@ class Annotator extends EventEmitter {
...
-            },
-            {
-                type: 'touchstart',
-                func: pointFunc,
-                eventObj: this.annotatedElement
+    bindModeListeners(mode) {
+        const handlers = [];
+        if (mode === TYPES.point) {
+            const pointFunc = this.pointClickHandler.bind(this.annotatedElement);
+            handlers.extend([
```
MinhHNguyen: My mistake on this syntax, extend is a python function.
Looks like we need to use a spread operator: `handlers.push(...[h1,h2,h3])`

src/lib/annotations/Annotator.js
pramodsum: done

src/lib/annotations/Annotator.js
pramodsum: yea i realized that once my dev vm was up and running for me to test this! :)

pramodsum:
@MinhHNguyen @JustinHoldstock I ended up moving cleaning up more annotations methods. Was able to remove more from BaseViewer.js Let me know what you think!

pramodsum:
Also moved any handler creation/deletion into Annotator.js itself rather than in BaseViewer.js so that we're not making extra checks to see if annotations is even enabled in order to clean up the viewer

pramodsum:
I still have one more edge case to fix up with these (relating to highlight handlers)

src/lib/annotations/Annotator.js
```diff
@@ -101,13 +107,76 @@ class Annotator extends EventEmitter {
...
+        }
+
+        return true;
+    }
+
+    /**
+     * Shows the annotate button for the specified mode
+     *
+     * @param {string} currentMode - Annotation mode
+     * @param {Object[]} modeButtons - Annotation modes which require buttons
```
MinhHNguyen: This isn't a param anymore is it?

pramodsum: nope i removed it in a later commit

src/lib/annotations/Annotator.js
```diff
@@ -75,6 +77,13 @@ class Annotator extends EventEmitter {
             });
         }
 
+        // Destroy all annotate buttons
+        Object.keys(this.modeButtons).forEach((type) => {
```
MinhHNguyen: Are there cases in which modeButtons is null or undefined?
Perhaps this should be wrapped just to make sure its safe (just like line 72 above)
`if (this.modeButtons) {}`

src/lib/annotations/Annotator.js
```diff
@@ -103,13 +111,79 @@ class Annotator extends EventEmitter {
             this.setupMobileDialog();
         }
 
+        // Show the annotate button for all enabled types for the
+        // current viewer
+        Object.keys(this.modeButtons).forEach((type) => {
```
MinhHNguyen: Same comment as line 81

src/lib/annotations/Annotator.js
MinhHNguyen: Just looked at the code in BaseViewer. Looks like this is fine since baseviewer passes in a constant object that always exists.

src/lib/annotations/doc/DocAnnotator.js
```diff
@@ -445,11 +445,12 @@ class DocAnnotator extends Annotator {
...
         this.annotatedElement.removeEventListener('mouseup', this.highlightMouseupHandler);
 
-        if (this.annotationService.canAnnotate) {
+        if (this.canAnnotate) {
             this.annotatedElement.removeEventListener('dblclick', this.highlightMouseupHandler);
             this.annotatedElement.removeEventListener('mousedown', this.highlightMousedownHandler);
             this.annotatedElement.removeEventListener('contextmenu', this.highlightMousedownHandler);
-            this.annotatedElement.removeEventListener('mousemove', this.getHighlightMouseMoveHandler());
+            this.annotatedElement.removeEventListener('mousemove', this.getHighlightMouseMoveHandler);
+            this.highlightMousedownHandler = null;
```
MinhHNguyen: Is this supposed to be this.highlightMousemoveHandler? I noticed this.getHightlightMouseMoveHandler() sets this.highlightMousemoveHandler.

MinhHNguyen:
Asides from the possible change in DocAnnotator, looks good to me.

src/lib/annotations/doc/DocAnnotator.js
pramodsum: Yea i need to clarify this w/ @JustinHoldstock

pramodsum: And also this needed to be `this.highlightMousemoveHandler = null` not `this.highlightMouseDownHandler = null`

pramodsum: I talked to Justin and it should be ...this.highlightMousemoveHandler); the addition of () was a mistake

PR #292, opened 2017-08-10
Fix: Forcing HD/SD should disable adaptation

bhh1988:
Shaka-player changed its behavior when selecting a variant. We now have
to explicitly disable adaptation. See notes below (which also apply to
upgrade from 2.0 to 2.1):

https://github.com/google/shaka-player/blob/master/docs/tutorials/upgrade-v2-0.md

NOTE: From my testing, their example doesn't actually work, and I need
to disable adaptation BEFORE selecting the variant. I filed https://github.com/google/shaka-player/issues/962 on shaka team for this

### DIFF
--- a/src/lib/viewers/media/DashViewer.js
+++ b/src/lib/viewers/media/DashViewer.js
@@ -216,10 +216,10 @@ class DashViewer extends VideoBaseViewer {
      * Enables or disables automatic adaptation
      *
      * @private
-     * @param {boolean} [adapt] - Enable or disable adaptation
+     * @param {boolean} adapt - Enable or disable adaptation
      * @return {void}
      */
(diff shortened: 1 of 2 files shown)


src/lib/viewers/media/DashViewer.js
```diff
@@ -219,7 +219,7 @@ class DashViewer extends VideoBaseViewer {
      * @param {boolean} [adapt] - Enable or disable adaptation
      * @return {void}
      */
-    enableAdaptation(adapt = true) {
+    enableAdaptation(adapt) {
```
JustinHoldstock: Is `adapt` a required param, now? If so, update doc above to remove square braces

bhh1988: Done

PR #293, opened 2017-08-10
Chore: Only remove annotation listeners if enabled on viewer

pramodsum:


src/lib/viewers/BaseViewer.js
```diff
@@ -168,7 +168,7 @@ class BaseViewer extends EventEmitter {
         }
 
         const { container } = this.options;
-        if (container) {
+        if (container && this.annotatorConf) {
```
tonyjin: Move this check to `isAnnotatable()` - that helper function shouldn't error out if there is no annotatorConf

pramodsum: moved

PR #294, opened 2017-08-10
Fix: annotation handler rebinding

MinhHNguyen:
I previously (unnecessarily) bound pointClickHandler to the wrong object. Also moving DrawingThread handler bindings to the constructor.

### DIFF
--- a/src/lib/annotations/BoxAnnotations.js
+++ b/src/lib/annotations/BoxAnnotations.js
@@ -7,7 +7,7 @@ const ANNOTATORS = [
         NAME: 'Document',
         CONSTRUCTOR: DocAnnotator,
         VIEWER: ['Document', 'Presentation'],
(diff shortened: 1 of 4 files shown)


src/lib/annotations/BoxAnnotations.js
```diff
@@ -7,7 +7,7 @@ const ANNOTATORS = [
         NAME: 'Document',
         CONSTRUCTOR: DocAnnotator,
         VIEWER: ['Document', 'Presentation'],
-        TYPE: [TYPES.point, TYPES.highlight]
+        TYPE: [TYPES.point, TYPES.highlight, TYPES.draw]
```
pramodsum: Remove this

MinhHNguyen: Good catch

PR #296, opened 2017-08-11
Update: Increase file size limit for range requests

tonyjin:
Increase document file size limit for disabling range requests for users with a non-US locale from 5MB to 25MB. This will hopefully improve performance of previews for larger documents outside the US since we are still limited by latency from round trips.


tonyjin:
@jeremypress this just removes the exception for Excel files. Excel files will now go through the same logic - if the user is not on an English locale and the original file is <25MB, we disable range requests. Before, WinExcel representations were significantly larger than the original file size. With WinExcel now turned to minimum output mode, the file sizes are now more in line.

Excel files through Libre will still be affected with the file size difference. We can pull data afterwards to see if this change significantly affects overall Excel performance.

PR #297, opened 2017-08-11
Fix: scale being unbound when annotating

MinhHNguyen:
Bug: open a file with annotations,
click point mode and change the browser window size and the annotations will not redraw.

Cause: unbindDOMModeListeners is being called on entering annotation mode. This method removes the scale listener.


pramodsum:
Good catch

PR #298, opened 2017-08-11
Upgrade: Upgrade to shaka-player 2.1.6

bhh1988:
The past few releases contain some new commits around alternate audio
tracks that are needed for supporting multiple audio tracks


src/lib/constants.js
```diff
@@ -80,7 +80,7 @@ export const X_REP_HINT_VIDEO_MP4 = '[mp4]';
 // These should be updated to match the Preview version in package.json
 // whenever a file in that third party directory is updated
 export const DOC_STATIC_ASSETS_VERSION = '0.130.0';
-export const MEDIA_STATIC_ASSETS_VERSION = '0.127.0';
+export const MEDIA_STATIC_ASSETS_VERSION = '1.5.0';
```
tonyjin: Latest release is actually 1.5.1, which wouldn't be reflected in package.json due to our new patch process. 1.5.0 is probably okay in this situation but I wonder if we should update the comments to mention latest version in https://github.com/box/box-content-preview/releases

bhh1988:
Actually, I realized this version of shaka still has the bug I reported here: https://github.com/google/shaka-player/issues/948

They fixed it on master, but not for this version (and I get the feeling they're not going to backport it). So I might want to hold off on this and wait until shaka v2.2 comes out.

PR #299, opened 2017-08-11
Multi audio

bhh1988:
New: Support multiple audio tracks

    This commit adds support for selection of different audio tracks when
    available in the manifest. It selects based on a combination of the
    language and "role" of the audio track. This combination should be
    unique in order to unambiguously select audio tracks - roles are made
    unique by the backend, as audio0, audio1, etc.

    One thing to be careful of in all this is how this interacts with
    selecting HD/SD video. Previous implementation would cache the HD/SD
    variant to switch to. Now, there are multiple HD variants and multiple
    SD variants, because the HD stream could have different audio. Thus this
    required some refactoring of how we pick the HD/SD streams, so that we
    maintain the same audio stream while changing video quality. In the
    process, I've also fixed a bug in how we size the video for audio-only
    representations (before, the size would end up being so small that the
    play button icon would leak into the controls).

### DIFF
--- a/src/lib/viewers/media/DashViewer.js
+++ b/src/lib/viewers/media/DashViewer.js
@@ -12,6 +12,8 @@ const CSS_CLASS_HD = 'bp-media-controls-is-hd';
 const SEGMENT_SIZE = 5;
 const MAX_BUFFER = SEGMENT_SIZE * 12; // 60 sec
 const MANIFEST = 'manifest.mpd';
+const DEFAULT_VIDEO_WIDTH_PX = 854;
+const DEFAULT_VIDEO_HEIGHT_PX = 480;
 
 @autobind
 class DashViewer extends VideoBaseViewer {
@@ -27,9 +29,10 @@ class DashViewer extends VideoBaseViewer {
         this.switchHistory = [];
 
         // tracks
-        this.hdRepresentation = {};
-        this.sdRepresentation = {};
+        this.hdVideoId = -1;
+        this.sdVideoId = -1;
         this.textTracks = []; // Must be sorted by representation id
+        this.audioTracks = [];
 
         // dash specific class
         this.wrapperEl.classList.add(CSS_CLASS_DASH);
@@ -191,25 +194,21 @@ class DashViewer extends VideoBaseViewer {
     }
 
     /**
-     * Handler for hd video
+     * Given a videoId (e.g. hd video id), enables the track with that video ID
+     * while maintaining the SAME AUDIO as the active track.
      *
      * @private
+     * @param {number} videoId - The id of the video used in the variant (provided by Shaka)
      * @return {void}
      */
-    enableHD() {
-        this.showLoadingIcon(this.hdRepresentation.id);
-        this.player.selectVariantTrack(this.hdRepresentation, true);
-    }
-
-    /**
-     * Handler for sd video
-     *
-     * @private
-     * @return {void}
-     */
-    enableSD() {
-        this.showLoadingIcon(this.sdRepresentation.id);
-        this.player.selectVariantTrack(this.sdRepresentation, true);
+    enableVideoId(videoId) {
+        const tracks = this.player.getVariantTracks();
+        const activeTrack = this.getActiveTrack();
+        const newTrack = tracks.find((track) => track.videoId === videoId && track.audioId === activeTrack.audioId);
+        if (newTrack && newTrack.id !== activeTrack.id) {
+            this.showLoadingIcon(newTrack.id);
+            this.player.selectVariantTrack(newTrack, true);
+        }
     }
 
     /**
@@ -245,6 +244,21 @@ class DashViewer extends VideoBaseViewer {
     }
 
     /**
+     * Handler for audio track
+     *
+     * @private
+     * @emits audiochange
+     * @return {void}
+     */
+    handleAudioTrack() {
+        const audioIdx = parseInt(this.cache.get('media-audiotracks'), 10);
+        if (this.audioTracks[audioIdx] !== undefined) {
+            const track = this.audioTracks[audioIdx];
+            this.player.selectAudioLanguage(track.language, track.role);
+        }
+    }
+
+    /**
      * Handler for hd/sd/auto video
      *
      * @private
@@ -257,11 +271,11 @@ class DashViewer extends VideoBaseViewer {
         switch (quality) {
             case 'hd':
                 this.enableAdaptation(false);
-                this.enableHD();
+                this.enableVideoId(this.hdVideoId);
                 break;
             case 'sd':
                 this.enableAdaptation(false);
-                this.enableSD();
+                this.enableVideoId(this.sdVideoId);
                 break;
             case 'auto':
             default:
@@ -283,7 +297,7 @@ class DashViewer extends VideoBaseViewer {
      */
     adaptationHandler() {
         const activeTrack = this.getActiveTrack();
-        if (activeTrack.id === this.hdRepresentation.id) {
+        if (activeTrack.videoId === this.hdVideoId) {
             this.wrapperEl.classList.add(CSS_CLASS_HD);
         } else {
             this.wrapperEl.classList.remove(CSS_CLASS_HD);
@@ -329,6 +343,7 @@ class DashViewer extends VideoBaseViewer {
         super.addEventListenersForMediaControls();
         this.mediaControls.addListener('qualitychange', this.handleQuality);
         this.mediaControls.addListener('subtitlechange', this.handleSubtitle);
+        this.mediaControls.addListener('audiochange', this.handleAudioTrack);
     }
 
     /**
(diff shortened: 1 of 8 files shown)


src/lib/viewers/media/DashViewer.js
```diff
@@ -191,25 +192,24 @@ class DashViewer extends VideoBaseViewer {
...
+    enableVideoId(videoId) {
+        const tracks = this.player.getVariantTracks();
+        const activeTrack = this.getActiveTrack();
+        const newTrack = tracks.find((track) => track.videoId === videoId && track.audioId === activeTrack.audioId);
+        if (newTrack && newTrack.id !== activeTrack.id) {
+            this.showLoadingIcon(newTrack.id);
+            this.player.selectVariantTrack(newTrack, true);
+        } else {
+            // This should never happen...
+            console.log(`Could not find a track with videoId=${videoId}!`);
```
bhh1988: Not sure if I'm allowed to do this? I feel like we should log something though...

src/lib/viewers/media/DashViewer.js
```diff
@@ -27,9 +27,10 @@ class DashViewer extends VideoBaseViewer {
         this.switchHistory = [];
 
         // tracks
-        this.hdRepresentation = {};
-        this.sdRepresentation = {};
+        this.hdVideoId = -1;
+        this.sdVideoId = -1;
```
bhh1988: Not so sure if this is ok to do - using -1 as sentinel value.

src/lib/viewers/media/DashViewer.js
jeremypress: If you consider  this to be an error feel free to use `console.error`. I would probably remove that comment though

src/lib/viewers/media/DashViewer.js
jeremypress: We would normally do `null` or just initialize it as `undefined` but if -1 makes more sense since these are numeric IDs I'm ok with that. What are other peoples thoughts on this?

src/lib/viewers/media/DashViewer.js
bhh1988: Done

src/lib/viewers/media/DashViewer.js
```diff
@@ -402,14 +448,25 @@ class DashViewer extends VideoBaseViewer {
 
         // Iterate over all available video representations and find the one that
         // seems the biggest so that the video player is set to the max size
-        const hdRepresentation = tracks.reduce((a, b) => (a.width > b.width ? a : b));
-        const sdRepresentation = tracks.reduce((a, b) => (a.width < b.width ? a : b));
+        const hdRep = tracks.reduce((a, b) => (a.width > b.width ? a : b));
+        const sdRep = tracks.reduce((a, b) => (a.width < b.width ? a : b));
+        if (sdRep.videoId === null) {
+            // There is only audio, no video
+            this.videoWidth = 854;
```
jeremypress: Magic Numbers

src/lib/viewers/media/MediaControls.js
```diff
@@ -877,6 +887,17 @@ class MediaControls extends EventEmitter {
...
     }
+
+    /**
+     * Takes a list of audio tracks and populates the settings menu
+     *
+     * @param {Array} audioLanguages - An ordered list of languages of the audio tracks
+     * @return {void}
+     */
+    initAlternateAudio(audioLanguages) {
+        this.settings.addListener('audiotracks', this.handleAudioTrack);
```
jeremypress: Is this emitted by shakaplayer?

src/lib/viewers/media/DashViewer.js
bhh1988: Defined as constants at top

src/lib/viewers/media/MediaControls.js
bhh1988: No. Settings menu emits this whenever you choose an option. Shaka player doesn't seem to emit events when you manually choose a track. Why?

build/karma.conf.js
```diff
@@ -2,7 +2,7 @@
 const webpackConfig = require('./webpack.karma.config');
 
 const DOC_STATIC_ASSETS_VERSION = '0.130.0';
-const MEDIA_STATIC_ASSETS_VERSION = '0.127.0';
+const MEDIA_STATIC_ASSETS_VERSION = '1.5.1';
```
tonyjin: When you do manage to merge this in after Shaka 2.2, please make sure to update this to the latest version at that point. Right now, it's 1.6.2

src/lib/constants.js
```diff
@@ -80,7 +80,7 @@ export const X_REP_HINT_VIDEO_MP4 = '[mp4]';
 // These should be updated to match the Preview version in package.json
 // whenever a file in that third party directory is updated
 export const DOC_STATIC_ASSETS_VERSION = '0.130.0';
-export const MEDIA_STATIC_ASSETS_VERSION = '0.127.0';
+export const MEDIA_STATIC_ASSETS_VERSION = '1.5.1';
```
tonyjin: Same as above.

src/lib/viewers/media/DashViewer.js
```diff
@@ -12,6 +12,8 @@ const CSS_CLASS_HD = 'bp-media-controls-is-hd';
 const SEGMENT_SIZE = 5;
 const MAX_BUFFER = SEGMENT_SIZE * 12; // 60 sec
 const MANIFEST = 'manifest.mpd';
+const DEFAULT_VIDEO_WIDTH = 854;
```
tonyjin: I like appending the unit, e.g. `DEFAULT_VIDEO_WIDTH_PX`

src/lib/viewers/media/DashViewer.js
tonyjin: -1's probably fine.

src/lib/viewers/media/DashViewer.js
```diff
@@ -191,25 +194,21 @@ class DashViewer extends VideoBaseViewer {
     }
 
     /**
-     * Handler for hd video
+     * Given a videoId (e.g. hd video id), enables the track with that video ID
+     * while maintaining the SAME AUDIO as the active track.
      *
      * @private
+     * @param {number} videoId - the id of the video used in the variant (provided by Shaka)
```
tonyjin: Nit: Capitalize `The` in `The id of the video`

src/lib/viewers/media/DashViewer.js
```diff
@@ -245,6 +244,21 @@ class DashViewer extends VideoBaseViewer {
...
 
     /**
+     * Handler for audio track
+     *
+     * @private
+     * @emits audiochange
+     * @return {void}
+     */
+    handleAudioTrack() {
+        const audioIdx = parseInt(this.cache.get('media-audiotracks'), 10);
```
tonyjin: Is there a situation where the cache could be empty for `media-audiotracks`?

src/lib/viewers/media/DashViewer.js
```diff
@@ -345,6 +360,35 @@ class DashViewer extends VideoBaseViewer {
...
+     * Loads alternate audio streams
+     *
+     * @return {void}
+     */
+    loadAlternateAudio() {
+        const variants = this.player.getVariantTracks().sort((track1, track2) => track1.audioId - track2.audioId);
+        const audioIds = [];
+        const uniqueAudioVariants = [];
+        let i = 0;
+        for (i = 0; i < variants.length; i++) {
```
tonyjin: Please space out your code more. I'd suggest a line break before `let i = 0;`, `this.audioTracks = `, and `if (this.audioTracks.length > 1) {`

src/lib/viewers/media/DashViewer.js
```diff
@@ -402,14 +447,25 @@ class DashViewer extends VideoBaseViewer {
...
+        } else if (hdRep.videoId === sdRep.videoId) {
+            // Only one video quality
+            this.videoWidth = sdRep.width;
+            this.videoHeight = sdRep.height;
+            this.sdVideoId = sdRep.videoId;
+        } else {
+            this.videoWidth = hdRep.width;
+            this.videoHeight = hdRep.height;
+            this.hdVideoId = hdRep.videoId;
+            this.sdVideoId = sdRep.videoId;
```
tonyjin: Could be simplified:

```
if (sdRep.videoId === null) {
    // There is only audio, no video
    this.videoWidth = DEFAULT_VIDEO_WIDTH;
    this.videoHeight = DEFAULT_VIDEO_HEIGHT;
} else {
    this.videoWidth = hdRep.width;
    this.videoHeight = hdRep.height;
    this.hdVideoId = hdRep.videoId !== sdRep.videoId ? hdRep.videoId : undefined;
    this.sdVideoId = sdRep.videoId;
}
```

build/karma.conf.js
bhh1988: Will do

src/lib/viewers/media/DashViewer.js
bhh1988: Done

src/lib/viewers/media/DashViewer.js
bhh1988: Done

src/lib/viewers/media/DashViewer.js
bhh1988: There should be no such situation, because handleAudioTrack() is only ever called after selection in Settings.js, which sets the cache as the only means of communicating to this file what audio track should be switched to.

src/lib/viewers/media/DashViewer.js
bhh1988: Done

src/lib/viewers/media/DashViewer.js
bhh1988: Thanks. Also shaka 2.2.0 has an isAudioOnly() method, which I can call now instead of checking sdRep.videoId === null

PR #300, opened 2017-08-11
Add en-x-pseudo language to support pseudolocalization in WebApp

jeeyi:
Pseudolocalization is a feature to test internationalization. It catches common internationalization errors such as hard-coded string, concatenated string, horizontal/vertical expansions, encoding, etc. It can be enabled using cookieswitch pseudolocalization in WebApp.


build/webpack.config.js
```diff
@@ -59,7 +60,10 @@ const languages = isRelease
         'zh-CN',
         'zh-TW'
     ]
-    : ['en-US']; // Only 1 language needed for dev
+    : [
+        'en-US',
+        'en-x-pseudo'
+    ]; // Only 2 languages needed for dev
```
jeeyi: Sounds good.  I've removed en-x-pseudo from dev.

PR #302, opened 2017-08-12
New: Enable support for gslide documents

tonyjin:


tonyjin:
Do not merge until backend support is added

gentooist:
@tonyjin please add support for `.gslides` (plural) as well.

PR #309, opened 2017-08-15
Fix: only hide the filmstrip if it exists

jeremypress:


src/lib/viewers/media/MediaControls.js
```diff
@@ -611,7 +611,9 @@ class MediaControls extends EventEmitter {
 
         if (this.wrapperEl && this.wrapperEl.parentNode) {
             this.wrapperEl.parentNode.classList.remove(SHOW_CONTROLS_CLASS);
-            this.filmstripHideHandler();
+            if (this.filmstripEl) {
```
MinhHNguyen: Any thoughts on putting this check inside of the handler? That way filmstripHideHandler can be a safe function itself.

jeremypress: good call, done.

PR #310, opened 2017-08-15
Upgrade: Upgrade to shaka-player 2.1.7

bhh1988:
The past few releases contain some new commits around alternate audio
tracks that are needed for supporting multiple audio tracks


bhh1988:
Found a safari bug this time. I think I should wait until v2.2...

PR #311, opened 2017-08-15
Chore: Updating method decorators in base files

jeremypress:


pramodsum:
I also like having methods near things that call them. Makes them easier to find when you're digging through the code

tonyjin:
I generally organize by access and then within access order by starting out with call order and then putting methods near things that call them (granted they're in the right access group).

PR #316, opened 2017-08-17
Chore: Store annotation threads by threadID in the thread map

pramodsum:
Includes some docAnnotator cleanup as well


src/lib/annotations/image/ImageAnnotator.js
```diff
@@ -127,6 +127,7 @@ class ImageAnnotator extends Annotator {
         // Set existing thread ID if created with annotations
         if (annotations.length > 0) {
             threadParams.threadID = annotations[0].threadID;
+            threadParams.threadNumber = annotations[0].threadNumber;
```
pramodsum: Apparently we never associated a thread number with image annotations...

src/lib/annotations/doc/DocAnnotator.js
```diff
@@ -499,25 +500,6 @@ class DocAnnotator extends Annotator {
...
 
-    /**
-     * Checks whether mouse is inside any dialog on the current page
-     *
-     * @protected
-     * @param {Event} event - Mouse event
-     * @param {number} page - Current page number
-     * @return {boolean} Whether or not mouse is inside a dialog on the page
-     */
-    isInDialogOnPage(event, page) {
```
pramodsum: We used to use this method in the monstrous highlightMouseMove() method but not that we're simply checking the datatype being hovered over, we no longer need this!

src/lib/annotations/doc/DocAnnotator.js
```diff
@@ -834,59 +806,39 @@ class DocAnnotator extends Annotator {
...
         }
     }
-    /**
-     * Returns all threads with a state in the specified states.
-     *
-     * @private
-     * @param {...string} states - States of highlight threads to find
-     * @return {AnnotationThread[]} threads with the specified states
-     * */
-    getThreadsWithStates(...states) {
```
pramodsum: This was also being used only once where we would end up iterating through ALL the file version's annotations once to generate the array in this method and then iterate through that array again wherever getThreadsWithStates() was being used.

src/lib/annotations/doc/DocAnnotator.js
```diff
@@ -834,59 +806,39 @@ class DocAnnotator extends Annotator {
         let consumed = false;
         let activeThread = null;
 
-        // Destroy any pending highlights on click outside the highlight
-        const pendingThreads = this.getThreadsWithStates(PENDING_STATES);
```
pramodsum: Cleaned it up so we only iterate through the threads on the current page once instead of using getThreadsWithStates()

pramodsum:
@JustinHoldstock can you take a look at this w/ the updated changes when you get a chance?

src/lib/annotations/Annotator.js
```diff
@@ -272,8 +275,8 @@ class Annotator extends EventEmitter {
      * @return {void}
      */
     renderAnnotations() {
-        Object.keys(this.threads).forEach((page) => {
-            this.threads[page].forEach((thread) => {
+        Object.values(this.threads).forEach((pageThreads) => {
```
JustinHoldstock: Use `this.renderAnnotationsOnPage()`

src/lib/annotations/Annotator.js
```diff
@@ -287,11 +290,14 @@ class Annotator extends EventEmitter {
...
     renderAnnotationsOnPage(pageNum) {
-        if (this.threads && this.threads[pageNum]) {
-            this.threads[pageNum].forEach((thread) => {
-                thread.show();
-            });
+        if (!this.threads) {
+            return;
         }
+
+        const pageThreads = this.getThreadsOnPage(pageNum);
```
JustinHoldstock: Could this be setup so that if it is given a certain value, it renders ALL annotations?

src/lib/annotations/Annotator.js
```diff
@@ -244,8 +244,8 @@ class Annotator extends EventEmitter {
      * @return {void}
      */
     hideAnnotations() {
-        Object.keys(this.threads).forEach((page) => {
-            this.threads[page].forEach((thread) => {
+        Object.values(this.threads).forEach((pageThreads) => {
```
JustinHoldstock: Can you follow the same idea for what I mention below, about `showAnnotationsOnPage()`

src/lib/annotations/Annotator.js
```diff
@@ -825,6 +836,19 @@ class Annotator extends EventEmitter {
...
+     *
+     * @private
+     * @param {number} page - Current page number
+     * @return {Map|[]} Threads on page
+     */
+    getThreadsOnPage(page) {
+        if (!(page in this.threads)) {
+            this.threads[page] = {};
+        }
+        return this.threads[page];
```
JustinHoldstock: Newline after the `if`!

src/lib/annotations/doc/DocAnnotator.js
JustinHoldstock: 👍

src/lib/annotations/doc/DocAnnotator.js
```diff
@@ -880,59 +850,36 @@ class DocAnnotator extends Annotator {
...
-            // mousedown activated some highlight, and then informs the other
-            // keydown handlers to not activate
-            const threadActive = thread.onClick(event, consumed);
-            if (threadActive) {
-                activeThread = thread;
-            }
+        const pageThreads = this.getThreadsOnPage(page);
+
+        Object.values(pageThreads).forEach((thread) => {
+            if (PENDING_STATES.indexOf(thread.state) >= 0) {
```
JustinHoldstock: Don't we have a method along the lines of `threadIsPending()`?

src/lib/viewers/doc/DocBaseViewer.js
```diff
@@ -410,8 +410,6 @@ class DocBaseViewer extends BaseViewer {
      * @return {boolean} consumed or not
      */
     onKeydown(key) {
-        // test
```
JustinHoldstock: :P

src/lib/annotations/Annotator.js
pramodsum: I'm not sure what you mean? We are currently rendering all annotations on that page

src/lib/annotations/doc/DocAnnotator.js
pramodsum: Yup! Good catch!

src/lib/annotations/Annotator.js
JustinHoldstock: My bad, nvm

PR #324, opened 2017-08-18
Fix: Inconsistent natural image dimensions in IE

MinhHNguyen:
@pramodsum 
The data-scale attribute is set to infinity when an Image element has a naturalWidth and naturalHeight of 0. This circumvents the problem even though IE11 is supposed to support naturalWidth and naturalHeight. It might be that naturalWidth and naturalHeight are eventually set but the setting is not done on load. In any case, this seems to fix the issue (tested on IE11).

### DIFF
--- a/src/lib/viewers/image/ImageBaseViewer.js
+++ b/src/lib/viewers/image/ImageBaseViewer.js
@@ -43,7 +43,7 @@ class ImageBaseViewer extends BaseViewer {
             return;
         }
 
-        ImageBaseViewer.setOriginalImageSize(this.imageEl);
+        this.setOriginalImageSize(this.imageEl);
         this.loadUI();
         this.zoom();
 
@@ -172,10 +172,11 @@ class ImageBaseViewer extends BaseViewer {
      * Sets the original image width and height on the img element. Can be removed when
      * naturalHeight and naturalWidth attributes work correctly in IE 11.
      *
+     * @private
      * @param {Image} imageEl - The image to set the original size attributes on
      * @return {Promise} A promise that is resolved if the original image dimensions were set.
      */
-    static setOriginalImageSize(imageEl) {
+    setOriginalImageSize(imageEl) {
         const image = imageEl;
         const promise = new Promise((resolve, reject) => {
             // Do not bother loading a new image when the natural size attributes exist
--- a/src/lib/viewers/image/ImageBaseViewer.js
+++ b/src/lib/viewers/image/ImageBaseViewer.js
@@ -43,6 +43,7 @@ class ImageBaseViewer extends BaseViewer {
             return;
         }
 
+        ImageBaseViewer.setOriginalImageSize(this.imageEl);
         this.loadUI();
         this.zoom();
 
@@ -165,6 +166,33 @@ class ImageBaseViewer extends BaseViewer {
     loadUI() {
         this.controls = new Controls(this.containerEl);
         this.bindControlListeners();
+    }
+
+    /**
+     * Sets the original image width and height on the img element. Can be removed when
+     * naturalHeight and naturalWidth attributes work correctly in IE 11.
+     *
+     * @param {Image} imageEl - The image to set the original size attributes on
+     * @return {void}
+     */
+    static setOriginalImageSize(imageEl) {
+        const image = imageEl;
+
+        // Do not bother loading a new image when the natural size attributes exist
+        if (imageEl.naturalWidth > 0 && imageEl.naturalHeight > 0) {
+            image.originalWidth = imageEl.naturalWidth;
+            image.originalHeight = imageEl.naturalHeight;
+        } else {
+            const originalImg = new Image();
+            image.originalWidth = 1;
+            image.originalHeight = 1;
+
+            originalImg.onload = () => {
+                image.originalWidth = originalImg.width || 1;
+                image.originalHeight = originalImg.height || 1;
+            };
+            originalImg.src = imageEl.src;
+        }
     }
 
     //--------------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


src/lib/viewers/image/ImageBaseViewer.js
```diff
@@ -43,6 +43,7 @@ class ImageBaseViewer extends BaseViewer {
             return;
         }
 
+        ImageBaseViewer.setOriginalImageSize(this.imageEl);
```
MinhHNguyen: setOriginalImageSize is implemented as a static method right now. Is there any preference for avoiding static methods? I know that only one Viewer will actually be instantiated so there will not be much of a difference in our use case but I am advocating for the use of static methods to avoid the creation of monolithic utility files.

tonyjin: We don't have a policy one way or another, but I'd lean towards making this a normal private function in the base viewer until otherwise needed since that's how we've organized other similar helper functions. The moment this is required in a different viewer outside of the ImageBaseViewer inheritance chain, we can easily update this to a static method.

MinhHNguyen: Updated to be non-static

src/lib/viewers/image/ImageBaseViewer.js
```diff
@@ -167,6 +168,42 @@ class ImageBaseViewer extends BaseViewer {
         this.bindControlListeners();
     }
 
+    /**
+     * Sets the original image width and height on the img element. Can be removed when
+     * naturalHeight and naturalWidth attributes work correctly in IE 11.
+     *
+     * @private
+     * @param {Image} imageEl - The image to set the original size attributes on
```
tonyjin: {HTMLElement}

src/lib/viewers/image/ImageBaseViewer.js
```diff
@@ -167,6 +168,42 @@ class ImageBaseViewer extends BaseViewer {
...
+     * @return {Promise} A promise that is resolved if the original image dimensions were set.
+     */
+    setOriginalImageSize(imageEl) {
+        const image = imageEl;
```
tonyjin: I think @JustinHoldstock mentioned at some point he prefers just disabling the eslint rule that warns against this when you mean to explicitly modify the passed in param.

Would look like:
```
setOriginalImageSize(imageEl) {
    // implementation
}
/* eslint-enable no-param-reassign */
```

src/lib/viewers/image/ImageBaseViewer.js
```diff
@@ -167,6 +168,42 @@ class ImageBaseViewer extends BaseViewer {
...
+        const promise = new Promise((resolve, reject) => {
+            // Do not bother loading a new image when the natural size attributes exist
+            if (imageEl.naturalWidth > 0 && imageEl.naturalHeight > 0) {
```
tonyjin: Is this the preferred way of checking whether natural size attributes are available? What happens for 0x0 or 0 by Y / X by 0 images?

src/lib/viewers/image/ImageBaseViewer.js
```diff
@@ -167,6 +168,42 @@ class ImageBaseViewer extends BaseViewer {
...
+
+                originalImg.error = () => {
+                    reject();
+                };
+                originalImg.onload = () => {
+                    image.originalWidth = originalImg.width || 1;
+                    image.originalHeight = originalImg.height || 1;
+                    resolve();
+                };
+                originalImg.src = imageEl.src;
```
tonyjin: Do you need to do this? `finishLoading()` is triggered after the image element loads (see ImageViewer.js) and this should be before `zoom()` is called so the image should already be in its natural width/height.

This implementation makes sense for the generic use case, but is probably less efficient for the exact use case you're solving for here.

src/lib/viewers/image/ImageBaseViewer.js
```diff
@@ -167,6 +168,42 @@ class ImageBaseViewer extends BaseViewer {
...
+        const promise = new Promise((resolve, reject) => {
+            // Do not bother loading a new image when the natural size attributes exist
+            if (imageEl.naturalWidth > 0 && imageEl.naturalHeight > 0) {
+                image.originalWidth = imageEl.naturalWidth;
+                image.originalHeight = imageEl.naturalHeight;
+                resolve();
+            } else {
+                const originalImg = new Image();
+                image.originalWidth = 1;
+                image.originalHeight = 1;
```
tonyjin: Instead of adding properties directly to an HTML element, it's better to add a data attribute, see: https://developer.mozilla.org/en-US/docs/Learn/HTML/Howto/Use_data_attributes

```
// Set
imageEl.dataset.originalWidth = 1;
imageEl.dataset.originalHeight = 1;

// Get
const originalWidth = imageEl.dataset.originalWidth;
const originalHeight = imageEl.dataset.originalHeight;
```

src/lib/viewers/image/ImageBaseViewer.js
MinhHNguyen: I had previously tested out just checking image.width and image.height in the onload function. It appears to have already been styled at that point, hence the need for this. I can go back through and see if there is a point in which the original loaded image can set its data attributes.

src/lib/viewers/image/ImageBaseViewer.js
MinhHNguyen: for 0x0, the original width and height will be set to 1x1. It is not ideal but the original width/height are used to measure scale. Therefore if they are set to 0 we will have division by 0 that results in scale = Infinity.

PR #326, opened 2017-08-21
New: Preview file from offline passed file

disist:
Hello guys, 

It continues of the pull request https://github.com/box/box-content-preview/pull/254 
I have created a new (clean) pull request with our latest agreements 

Guys, @tonyjin, @priyajeet Are you ok with this implementation?

To remind our goal: learn previewer to show files from offline storage when we have no internet.


src/lib/Preview.js
```diff
@@ -486,6 +486,11 @@ class Preview extends EventEmitter {
         // Init performance logging
         this.logger = new Logger(this.location.locale);
 
+        // Save reference to the file when a well-formed file object is present in options
+        if (this.previewOptions.file) {
+            this.updateFileCache(this.previewOptions.file);
+        }
```
tonyjin: Remove this. I'll try and get a different PR in this week for allowing the file object to be directly sent into preview.show() as the first param.

The signature will be like:

```
    /**
     * Primary function for showing a preview of a file.
     *
     * @param {string|Object} fileIdOrFile - Box File ID or well-formed Box File object
     * @param {string|Function} token - Access token string or generator function
     * @param {Object} [options] - Optional preview options
     * @return {void}
     */
```

src/lib/Preview.js
```diff
@@ -502,6 +507,11 @@ class Preview extends EventEmitter {
             this.retryCount = 0;
         }
 
+        if (this.previewOptions.file) {
+            this.loadPreviewWithTokens({});
+            return;
+        }
```
tonyjin: Remove this, see above.

src/lib/Preview.js
```diff
@@ -613,6 +623,12 @@ class Preview extends EventEmitter {
         // Save the reference to any additional custom options for viewers
         this.options.viewers = options.viewers || {};
 
+        // Save the reference to file object
+        this.options.file = options.file || null;
```
tonyjin: Remove this, see above.

src/lib/Preview.js
```diff
@@ -879,7 +897,9 @@ class Preview extends EventEmitter {
             });
 
             // If there wasn't an error, use Events API to log a preview
-            this.logPreviewEvent(this.file.id, this.options);
+            if (!this.options.skipServerUpdate) {
+                this.logPreviewEvent(this.file.id, this.options);
+            }
```
tonyjin: I have conflicting thoughts about this. On one hand I see that your offline use case wouldn't require this logging call but on the other hand there could be generic use case where we don't want to refresh from the server but the developer is not using an offline use case.

I don't think we need a separate option for this, and I'd err on the side of having this logging call fail in your offline use case (ie remove the if check)

priyajeet: Yeah lets remove this `if` check. And just let the request fail it was offline.

priyajeet: Might still have to do something about line 924

src/lib/Preview.js
priyajeet: Will need a guard for the line below

tonyjin:
FYI, the initial PR for updating `show()` is in code review.

src/lib/Preview.js
disist: Okay, removed, agree

src/lib/Preview.js
disist: So, when we pass well-formed file object, we need to call `this.loadPreviewWithTokens` but we don't pass the token and in lines below, we call `this.loadPreviewWithTokens` after executing `getTokens`, so as a proposal with according to your pull request https://github.com/box/box-content-preview/pull/328/, can I use guard here: 
```
if (typeof fileIdOrFile === 'object') {
     this.loadPreviewWithTokens({});
     return;
}
```
?

src/lib/Preview.js
disist: Okay, removed

src/lib/Preview.js
disist: Guys, agree, let the request fail it was offline. 
@priyajeet Thanks for good catch about `this.prefetchNextFiles()`. I've added the guard.

disist:
@tonyjin Thanks for the fast response. It's not a problem to wait, it's a life :)
Besides, thanks for the PR with new signature, I agree with you. 

After your comments I have just one question regarding one comment on 513 line in Previewer, could you please look here.  

Described flow above is correct.

src/lib/Preview.js
```diff
@@ -502,6 +502,11 @@ class Preview extends EventEmitter {
             this.retryCount = 0;
         }
 
+        if (typeof fileId === 'object') {
+            this.loadPreviewWithTokens({});
+            return;
+        }
```
tonyjin: Good point. How about this formatting?

```
if (typeof fileId === 'object') {
    // Load without token if using well-formed file object
    this.loadPreviewWithTokens({});
} else {
    // Fetch tokens before proceeding
    getTokens(this.file.id, this.previewOptions.token)
        .then(this.loadPreviewWithTokens)
        .catch(this.triggerFetchError);
}
```

tonyjin:
@disist good catch. Faking the 'original' representation is a construct on Preview's end and isn't something we should expect the API to return. To account for other use cases, I think you should add a `cacheFile()` in `loadPreviewWithTokens()` like:

```
// If file object is well-formed, load with it. Otherwise, fetch file information from the server.
if (checkFileValid(this.file)) {
    // Save file in cache. This also adds the 'ORIGINAL' representation.
    cacheFile(this.file);
    this.loadFromCache();
} else {
    this.loadFromServer();
}
```

src/lib/Preview.js
disist: No problem, thanks

disist:
@tonyjin Thank you, I have added `cacheFile()` in `loadPreviewWithTokens()`. 
For the first time it's worked it's worked approach for us. 

And also one question, should we add check on the well-formed file object when I pass file object and not pass token, I mean in `show(wellFormedFileObject, null, options)`?

tonyjin:
Good thought! We added it to https://github.com/box/box-content-preview/blob/master/src/lib/Preview.js#L503 as part of the `show()` change. At some point in the future we may be calling `load()` directly with a file object internally so we thought the validation worked better there.

disist:
Hello, Guys! If everything is ok, let's merge this PR.

tonyjin:
@disist done - thanks for your contribution!

PR #327, opened 2017-08-21
Chore: Rsync.plugin should update files rather than delete and replace

pramodsum:
- This is so that annotations.js/annotations.css and
  preview.js/preview.css can be synced from separate 
  repositories simultaneously

### DIFF
--- a/build/RsyncPlugin.js
+++ b/build/RsyncPlugin.js
@@ -10,7 +10,7 @@ RsyncPlugin.prototype.apply = function rsync(compiler) {
     compiler.plugin('done', () => {
         console.log('');
         console.log(`🔄 🔄 🔄  Rsync starting for ${this.source} 🔄 🔄 🔄`);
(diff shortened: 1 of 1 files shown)


build/RsyncPlugin.js
```diff
@@ -10,7 +10,7 @@ RsyncPlugin.prototype.apply = function rsync(compiler) {
     compiler.plugin('done', () => {
         console.log('');
         console.log(`🔄 🔄 🔄  Rsync starting for ${this.source} 🔄 🔄 🔄`);
-        execSync(`rsync -avz --delete --exclude=".*" "${this.source}" "${this.destination}"`, { stdio: [0, 1, 2] });
```
tonyjin: Note that this can potentially result in a lot of cruft if you change file names around or do some larger configuration change.

pramodsum: Hmm good point. Let me look into alternative setups

PR #328, opened 2017-08-21
New: Support well-formed file object in show()

tonyjin:
- Partially reverts changes from #24 and will enable #326
- Adds 'id' to FILE_FIELDS since we use that to validate the file object passed in

### DIFF
--- a/src/lib/Preview.js
+++ b/src/lib/Preview.js
@@ -148,12 +148,12 @@ class Preview extends EventEmitter {
     /**
      * Primary function for showing a preview of a file.
      *
-     * @param {string} fileId - Box File ID
+     * @param {string|Object} fileIdOrFile - Box File ID or well-formed Box File object
      * @param {string|Function} token - Access token string or generator function
      * @param {Object} [options] - Optional preview options
      * @return {void}
      */
-    show(fileId, token, options = {}) {
+    show(fileIdOrFile, token, options = {}) {
         // Save a reference to the options to be used later
         if (typeof token === 'string' || typeof token === 'function') {
             this.previewOptions = Object.assign({}, options, { token });
@@ -161,8 +161,8 @@ class Preview extends EventEmitter {
             throw new Error('Missing access token!');
         }
 
-        // load the preview
-        this.load(fileId);
+        // Load the preview
+        this.load(fileIdOrFile);
     }
 
     /**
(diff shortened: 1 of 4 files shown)


src/lib/file.js
```diff
@@ -1,9 +1,10 @@
 import { ORIGINAL_REP_NAME } from './constants';
 
-// List of Box Content API fields that the Preview SDK requires for every file. Updating this list is most likely
+// List of Box Content API fields that the Preview library requires for every file. Updating this list is most likely
 // a breaking change and should be done with care. Clients that leverage functionality dependent on this format
 // (e.g. Box.Preview.updateFileCache()) will need to be updated if this list is modified.
 const FILE_FIELDS = [
+    'id',
```
tonyjin: Normally, `id` is returned with any get files API call. Now, since we also use this list for simple validation of a passed in file object, I'm explicitly adding this so developers cannot pass in a file object without an ID.

src/lib/Preview.js
```diff
@@ -148,21 +148,28 @@ class Preview extends EventEmitter {
...
         if (typeof token === 'string' || typeof token === 'function') {
             this.previewOptions = Object.assign({}, options, { token });
         } else {
             throw new Error('Missing access token!');
         }
 
-        // load the preview
-        this.load(fileId);
+        // Check if passed in file is a well-formed Box File object
+        if (typeof fileIdOrFile === 'object' && !checkFileValid(fileIdOrFile)) {
```
priyajeet: Should move the === object check inside `checkFileValid()` since it already has a similar null check

src/lib/Preview.js
```diff
@@ -492,8 +499,14 @@ class Preview extends EventEmitter {
         // Save reference to the currently shown file, if any
         const currentFileId = this.file ? this.file.id : undefined;
 
-        // Use cached file data if available, otherwise create empty file object
-        this.file = this.cache.get(fileId) || { id: fileId };
+        // Check if file ID or well-formed file object was passed in
+        if (typeof fileIdOrFile === 'string') {
+            // Use cached file data if available, otherwise create empty file object
+            this.file = this.cache.get(fileIdOrFile) || { id: fileIdOrFile };
+        } else {
```
priyajeet: Make this `else if checkValidFile()` else throw the error. So that `load()` function is guarded.

src/lib/Preview.js
priyajeet: Also, move the checkFileValid check below

src/lib/Preview.js
tonyjin: Good suggestions, thanks. Updated.

PR #332, opened 2017-08-23
Fix: Safari font rendering issue

tonyjin:
Should fix https://github.com/mozilla/pdf.js/issues/8700 and https://cloud.box.com/s/xd92oda655j808n0zvklga0qt2tltn6v on Safari


tonyjin:
@jeremypress @JustinHoldstock let's make sure I don't break Walmart with this change :)

PR #337, opened 2017-08-25
Feature: drawing annotation deletion

MinhHNguyen:
Unfortunately I didn't have the foresight to split selection and deletion into separate pull requests, but ~900 lines are tests.

- I'm proposing the ability to pass a controller into the annotation object (see BaseViewer.js for an example). This will make more sense with custom annotations in the sense that only a controller needs to be accounted for within the official library.
- The controller will be able to take over some of the responsibilities currently done in annotator. Annotator should be able to be made more generic as any annotation type specific logic should be delegated to a controller. Eventually the controller should not need to know about the annotator. Instead should emit events and the annotator should observe these events.
- Deletion is currently done upon selecting a drawing twice. Next up is to make UI for selection and deletion. An additional improvement is to use two canvases for hit detection so that it can be done in O(1) time, allowing selection to be done on hover without a huge loss of performance.


src/lib/viewers/BaseViewer.js
```diff
@@ -3,6 +3,8 @@ import EventEmitter from 'events';
 import debounce from 'lodash.debounce';
 import fullscreen from '../Fullscreen';
 import RepStatus from '../RepStatus';
+// TODO (@minhnguyen): Import controller from annotations npm module once the library is separate
+import DrawingController from '../annotations/drawing/DrawingController';
```
pramodsum: I really want to avoid calling annotation specific code in BaseViewer. Is it possible to move this into Annotator or BoxAnnotations?

src/lib/annotations/drawing/DrawingController.js
```diff
@@ -0,0 +1,303 @@
...
+} from '../annotationConstants';
+
+class DrawingController extends AnnotationController {
+    /* eslint-disable new-cap */
+    /** @property {Array} - The array of annotation threads */
+    threads = new rbush();
+    /* eslint-enable new-cap */
+
+    /** @property {DrawingThread} - The selected DrawingThread */
+    selected;
```
jeremypress: maybe selectedThread? I initially thought selected was a boolean.

src/lib/annotations/drawing/DrawingController.js
```diff
@@ -0,0 +1,303 @@
...
+    /**
+     * Handle an annotation event.
+     *
+     * @inheritdoc
+     * @protected
+     * @param {AnnotationThread} thread - The thread that emitted the event
+     * @param {Object} data - Extra data related to the annotation event
+     * @return {void}
+     */
+    handleAnnotationEvent(thread, data = {}) {
```
jeremypress: +1 on the comments here, nice.

src/lib/annotations/drawing/DrawingController.js
```diff
@@ -0,0 +1,303 @@
...
+            return;
+        }
+
+        const location = this.annotator.getLocationFromEvent(event, TYPES.point);
+        if (!location) {
+            return;
+        }
+
+        const eventBoundary = {
+            minX: +location.x - DRAW_BORDER_OFFSET,
```
jeremypress: What do these +'s do?

src/lib/annotations/drawing/DrawingController.js
```diff
@@ -0,0 +1,303 @@
...
+            this.selected.drawingContext.clearRect(0, 0, canvas.width, canvas.height);
+        }
+
+        // Selected a region with no drawing threads, remove the reference to the previously selected thread
+        if (intersectingThreads.length === 0) {
+            this.selected = undefined;
+            return;
+        }
+
+        // Randomly select a thread in case there are multiple
```
jeremypress: o_0

src/lib/annotations/drawing/DrawingController.js
```diff
@@ -0,0 +1,303 @@
...
+        const index = Math.floor(Math.random() * intersectingThreads.length);
+        const selected = intersectingThreads[index];
+        this.select(selected);
+    }
+
+    /**
+     * Select the indicated drawing thread. Deletes a drawing thread upon the second consecutive selection
```
jeremypress: This will change once we introduce UI, correct?

src/lib/annotations/drawing/DrawingController.js
```diff
@@ -0,0 +1,303 @@
...
+     * Toggle the undo and redo buttons based on the number of actions available
+     *
+     * @private
+     * @param {number} undoCount - The number of objects that can be undone
+     * @param {number} redoCount - The number of objects that can be redone
+     * @return {void}
+     */
+    updateUndoRedoButtonEls(undoCount, redoCount) {
+        if (this.undoButtonEl) {
+            if (undoCount === 1) {
```
jeremypress: Can these numbers be > 1?

src/lib/viewers/BaseViewer.js
MinhHNguyen: Good point, I'll move it into annotator as discussed.

src/lib/annotations/AnnotationController.js
```diff
@@ -0,0 +1,146 @@
...
+     * @protected
+     * @param {AnnotationThread} thread - Thread to unbind events from
+     * @return {void}
+     */
+    unbindCustomListenersOnThread(thread) {
+        if (!thread) {
+            return;
+        }
+
+        thread.removeAllListeners('threaddeleted');
```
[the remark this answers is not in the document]
MinhHNguyen: Looks like you're right about this, but there is a comment in annotator regarding calling unbindCustomListenersOnThread on 'threadcleanup':
>// Thread should be cleaned up, unbind listeners - we don't do 
>// in threaddeleted listener since thread may still need to respond
>// to error messages

src/lib/annotations/Annotator.js
```diff
@@ -536,6 +545,17 @@ class Annotator extends EventEmitter {
...
+
+                const modeData = this.modeButtons[firstAnnotation.type];
+                if (
+                    modeData &&
+                    modeData.controller &&
+                    modeData.controller.constructor.name === DrawingController.name
+                ) {
+                    modeData.controller.bindCustomListenersOnThread(thread);
+                    modeData.controller.registerThread(thread);
+                    this.addThreadToMap(thread);
```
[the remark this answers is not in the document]
MinhHNguyen: This particular binding is done on saved annotations fetched from the server. This binding allows the annotator to react to the deletion of the AnnotationThread.

src/lib/annotations/Annotator.js
```diff
@@ -836,12 +771,26 @@ class Annotator extends EventEmitter {
...
      * @protected
+     * @param {string} mode - Annotation mode to be unbound
      * @return {void}
      */
-    unbindModeListeners() {
+    unbindModeListeners(mode) {
         while (this.annotationModeHandlers.length > 0) {
             const handler = this.annotationModeHandlers.pop();
-            handler.eventObj.removeEventListener(handler.type, handler.func);
+            const eventNames = handler.type.split(' ');
```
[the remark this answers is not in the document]
MinhHNguyen: I was initially checking out a way to bind the same listener to multiple events ie. 'mousedown', 'touchstart' by simply adding it to the string in bindModeListeners. 

This is done in AnnotationController (see drawingcontroller line 119 for an example) and doesn't need to be in annotator. I'll remove this.

src/lib/annotations/drawing/DrawingController.js
MinhHNguyen: Sometimes location.x/y are strings (I suspect this is when they are loaded from the server) and + ensures they are processed as numbers. We can also use parseFloat() here if that is more clear.

src/lib/annotations/drawing/DrawingController.js
MinhHNguyen: This will change with better hit detection 👍  (using a hitmap canvas)

src/lib/annotations/drawing/DrawingController.js
MinhHNguyen: Yes, with the UI, a selection should show the GUI. A deletion will be done by clicking delete on the GUI option.

src/lib/annotations/drawing/DrawingController.js
MinhHNguyen: Yes, in which case we don't need to do anything because they should already be enabled. In theory, the undo and redo counts will always start at 0. When they hit 1 they will be enabled and stay enabled for values > 1. If they drop to 0 they will return to the starting state (of being disabled).

src/lib/annotations/doc/DocAnnotator.js
```diff
@@ -436,7 +437,9 @@ class DocAnnotator extends Annotator {
 
         if (this.hasTouch && this.isMobile) {
             document.addEventListener('selectionchange', this.onSelectionChange);
+            document.addEventListener('touchstart', this.drawingSelectionHandler);
```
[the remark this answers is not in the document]
MinhHNguyen: I'm not quite sure. I will test it out but it might be the case that mobile browsers are functioning a bit differently.

src/lib/annotations/drawing/DrawingController.js
MinhHNguyen: It looks like +number is actually the fastest. https://jsperf.com/number-vs-plus-vs-toint-vs-tofloat/16

src/lib/annotations/AnnotationController.js
```diff
@@ -0,0 +1,146 @@
+import EventEmitter from 'events';
+
+class AnnotationController extends EventEmitter {
+    /** @property {Array} - The array of annotation threads */
+    threads = [];
```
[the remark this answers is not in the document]
MinhHNguyen: The goal is to have each controller manage their respective type of threads. Right now we are at an in-between state where highlight and point annotation doesn't use a controller so we still need to store threads in the annotator and controller. In addition, the threadmap is in use in other functions that I did not want to modify.

The good news is that these data structures simply store a reference to the thread so the memory overhead isn't that bad. The downside is that the thread reference needs to be removed from both (this is accomplished by listening to 'threaddeleted').

MinhHNguyen:
- Changed controller names
- Controllers instantiated inside of BoxAnnotations and loaded through annotatorConf
-  Handler array uses an array of event names rather than one string being parsed

src/lib/annotations/doc/DocAnnotator.js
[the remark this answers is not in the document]
MinhHNguyen: Just tested it out, looks like highlight selection needs to be bound to document but the drawing selection can be bound to annotated element.

src/lib/annotations/drawing/DrawingModeController.js
```diff
@@ -0,0 +1,303 @@
...
+                eventObj: this.annotator.annotatedElement
+            },
+            {
+                type: ['mouseup', 'touchcancel', 'touchend'],
+                func: annotatorUtil.eventToLocationHandler(locationFunction, this.currentThread.handleStop),
+                eventObj: this.annotator.annotatedElement
+            }
+        );
+
+        if (this.postButtonEl) {
```
pramodsum: move these into their own helper method and just call it for each button. May even make sense to move it to the base class!
```
addButtonHandlers(buttonEl, handler) {
  if (buttonEl) {
    const buttonHandler = {
      type: 'click,
      func: handler, 
      eventObj: buttonEl
    };
    handlers.push(buttonHandler);
}
```

src/lib/annotations/drawing/DrawingModeController.js
```diff
@@ -0,0 +1,303 @@
...
+
+        const eventBoundary = {
+            minX: +location.x - DRAW_BORDER_OFFSET,
+            minY: +location.y - DRAW_BORDER_OFFSET,
+            maxX: +location.x + DRAW_BORDER_OFFSET,
+            maxY: +location.y + DRAW_BORDER_OFFSET
+        };
+
+        // Get the threads that correspond to the point that was clicked on
+        const intersectingThreads = this.threads
```
pramodsum: is drawController.threads an array?

src/lib/annotations/drawing/DrawingThread.js
```diff
@@ -93,21 +120,41 @@ class DrawingThread extends AnnotationThread {
...
+     *
+     * @public
+     * @return {void}
+     */
+    deleteThread() {
+        this.annotations.forEach(this.deleteAnnotationWithID);
+
+        // Calculate the bounding rectangle
+        const [x, y, width, height] = this.getRectangularBoundary();
+        // Clear the drawn thread and destroy it
```
pramodsum: new line before comment

src/lib/annotations/Annotator.js
```diff
@@ -167,6 +170,11 @@ class Annotator extends EventEmitter {
 
             const handler = this.getAnnotationModeClickHandler(currentMode);
             annotateButtonEl.addEventListener('click', handler);
+
+            const { CONTROLLERS } = this.options.annotator;
```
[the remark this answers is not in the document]
MinhHNguyen: This.options.annotator can't be null because of the isModeAnnotatable check above

src/lib/annotations/Annotator.js
```diff
@@ -836,13 +771,23 @@ class Annotator extends EventEmitter {
      * Unbinds event listeners for annotation modes.
      *
      * @protected
+     * @param {string} mode - Annotation mode to be unbound
      * @return {void}
      */
-    unbindModeListeners() {
+    unbindModeListeners(mode) {
         while (this.annotationModeHandlers.length > 0) {
```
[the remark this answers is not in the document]
MinhHNguyen: This part delegates the unbinding to the controller rather than the annotator. Meaning once a point annotation controller is in (and preferably highlight as well) this call just needs to be a call to controller[mode].unbindModeListeners()

src/lib/annotations/BoxAnnotations.js
```diff
@@ -17,6 +18,10 @@ const ANNOTATORS = [
     }
 ];
 
+const ANNOTATOR_TYPE_CONTROLLERS = {
+    [TYPES.draw]: DrawingModeController
```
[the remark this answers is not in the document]
MinhHNguyen: We can do this but the time complexity of access will be linear (linear search to match type) rather than using the type as the key (constant time). This shouldn't be too bad as long as the number of controllers stays small. Let me know if this small time tradeoff is preferable for style.

src/lib/annotations/annotatorUtil.js
```diff
@@ -420,12 +420,10 @@ export function eventToLocationHandler(locationFunction, callback) {
             return;
         }
 
-        evt.stopPropagation();
         evt.preventDefault();
+        evt.stopPropagation();
         const location = locationFunction(evt);
-        if (location) {
```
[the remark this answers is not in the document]
MinhHNguyen: I found it preferable to let the callback handle an invalid location rather than avoiding a call on the callback. This is because the callback can handle/display an error rather than hiding the error overall.

src/lib/annotations/doc/DocAnnotator.js
```diff
@@ -315,7 +316,7 @@ class DocAnnotator extends Annotator {
 
         if (!thread && this.notification) {
             this.emit('annotationerror', __('annotations_create_error'));
-        } else if (thread) {
+        } else if (thread && (type !== TYPES.draw || location.page)) {
```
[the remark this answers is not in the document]
MinhHNguyen: It's a little bit messy but a DrawingThread shouldn't be added to the threadmap until it has a page. On DrawingThread creation, the page is undefined so adding it to the threadmap would add it to threadmap[page1] rather than the correct page.

Then on lookup we would be looking for the thread in threadmap[correctPage] where correct page may not be page1.

src/lib/annotations/doc/DocDrawingThread.js
```diff
@@ -32,42 +32,50 @@ class DocDrawingThread extends DrawingThread {
     /**
      * Handle a pointer movement
      *
+     * @public
      * @param {Object} location - The location information of the pointer
      * @return {void}
      */
     handleMove(location) {
-        if (this.drawingFlag !== DRAW_STATES.drawing) {
+        if (this.drawingFlag !== DRAW_STATES.drawing || !location) {
```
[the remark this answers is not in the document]
MinhHNguyen: It would not have a valid location upon moving the mouse on the side of the document.

src/lib/annotations/drawing/DrawingModeController.js
MinhHNguyen: It is an array by default (AnnotationModeController) and an RTree in DrawingModeController

src/lib/annotations/doc/DocAnnotator.js
```diff
@@ -774,6 +779,25 @@ class DocAnnotator extends Annotator {
...
 
     /**
+     * Drawing selection handler. Delegates to the drawing controller
+     *
+     * @private
+     * @param {Event} event - DOM event
+     * @return {void}
+     */
+    drawingSelectionHandler(event) {
+        const { annotator } = this.options;
```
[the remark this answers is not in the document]
MinhHNguyen: Yes that is a valid assumption since the controllers are only instantiated for enabled types inside of BoxAnnotations

PR #338, opened 2017-08-25
Fix: pass image setOriginalImageSize failure through

MinhHNguyen:
In the case that ImageViewer's imageEl is a div of images (.tiff) or more generically, some other tag - doing a get on a non-existent source will fail. Before, the promise would reject but it should always resolve in order to stay consistent with the behavior of previous versions.


pramodsum:
So this would prevent the image from failing to load if we aren't able to set the original image size?

MinhHNguyen:
@pramodsum exactly, which is what was happening before this change went in

src/lib/viewers/image/ImageBaseViewer.js
```diff
@@ -205,15 +205,15 @@ class ImageBaseViewer extends BaseViewer {
...
                             // Assume 300x150 that chrome does by default
                             imageEl.setAttribute('originalWidth', 300);
                             imageEl.setAttribute('originalHeight', 150);
-                            resolve(e);
+                        } finally {
+                            resolve();
                         }
                     })
-                    .catch(reject);
+                    .catch(resolve);
```
jeremypress: What does resolving in this catch do? Also is there a reason this is all async? Shouldn't everything dimension-wise be available after the image is done loading?

MinhHNguyen: The get call is async so we need to know when it resolves in order to proceed. Otherwise zoom will get called in finishloading() and the scale will be set to infinity. Sometimes the get call is not called (original dimensions are set) and the promise is resolved right away. On any errors, we should resolve so that the image still loads (although with an incorrect data-scale attribute).

jeremypress:
Testing looked good on my end.

PR #340, opened 2017-08-25
New: Add method to get top right corner of a highlight annotation

pramodsum:
This is required to flip the annotations dialog above highlight annotations


pramodsum:
It should ever get called with an undefined value or empty array. It will be used similar to how getLowerRightCornerOfLastQuadPoint is used to position the dialog on a highlight. The method it's being called in should deal w/ any kind of quad point validation if needed. 

As far as I know, we used the quad points to calculate where to position the highlights because that's how they're being stored in annotation thread and that's what we end up having access to in the dialog.

JustinHoldstock:
@MinhHNguyen @pramodsum Also, AFAIK we store the points as per the PDF annotations spec.

PR #343, opened 2017-08-26
Chore: log actual error message

jeremypress:
We now attempt to log the error's 'message' if it exists instead of a display message. 

Also, Moving the error parsing to the error viewer itself. We're not using the stack right now because its really long and won't be super useful with minified/bundled code.


src/lib/viewers/error/__tests__/PreviewErrorViewer-test.js
```diff
@@ -64,12 +64,13 @@ describe('lib/viewers/error/PreviewErrorViewer', () => {
                 const extension = testCase[0];
                 const expectedIcon = testCase[1];
 
-                const message = 'reason';
+                const err = new Error()
+                err.displayMessage = 'reason'
```
MinhHNguyen: possible ';' for style on 67 & 68

jeremypress: good catch, I've been switching back and for with Python a lot...

src/lib/viewers/doc/DocBaseViewer.js
```diff
@@ -492,11 +492,12 @@ class DocBaseViewer extends BaseViewer {
                 console.error(err);
                 /* eslint-enable no-console */
 
-                // Display a generic error message but log the real one
+                // Ensure a generic display message but log the real one
```
tonyjin: ?

src/lib/viewers/error/PreviewErrorViewer.js
```diff
@@ -77,8 +76,11 @@ class PreviewErrorViewer extends BaseViewer {
             }
         }
 
+        // Figure out what error message to log and what error message to display
+        const displayMessage = err instanceof Error && err.displayMessage ? err.displayMessage : __('error_default');
```
tonyjin: I don't think this logic is completely right. We want messages in this order of preference:

1. err.displayMessage
2. err.message
3. __('error_default')

Maybe:
```
// Up top where const message = reason || __('error_default');
err = err instanceof Error ? err : new Error(__('error_default'));
/* eslint-enable no-param-reassign */

// And here
const displayMessage = err.displayMessage || err.message;
```

This helps guard below (`error: err.message`) if err was not properly passed in as an instance of Error as well.

jeremypress: I don't think we ever want 2 to be the display message. That could be some stacktrace like `a is undefined`.

For display:
1. err.displayMessage
2. __('error_default')

for logging:
1. err.message
2. err.displayMessage
3.  __('error_defualt')

src/lib/viewers/error/PreviewErrorViewer.js
```diff
@@ -77,8 +76,15 @@ class PreviewErrorViewer extends BaseViewer {
             }
         }
 
+        /* eslint-disable no-param-reassign */
+        err = err instanceof Error ? err : new Error(__('error_default'));
+        /* eslint-enable no-param-reassign */
+
+        // If there is no display message fallback to the message from above
+        const displayMessage = err.displayMessage || err.message;
```
jeremypress: @tonyjin just to make sure we're on the same page here.

Display message will be either:
1. display message from passed in error
2. message of locally created error (default)

Log message will be:
1. message from passed in error (hopefully useful)
2. message of locally created error (default)
3. display message from passed in error (case with display message but no message)

src/lib/viewers/error/PreviewErrorViewer.js
```diff
@@ -77,17 +76,27 @@ class PreviewErrorViewer extends BaseViewer {
...
+        this.messageEl.textContent = displayMessage;
 
         // Add optional download button
         if (checkPermission(file, PERMISSION_DOWNLOAD) && showDownload && Browser.canDownload()) {
             this.addDownloadButton();
         }
 
         this.loaded = true;
+
+        // The error will either be the message from the original error, the displayMessage from the orignal error,
```
jeremypress: @tonyjin added the comment

PR #345, opened 2017-08-29
Fix: Re-scale annotation canvases on zoom

pramodsum:
- Fixes the bug where the annotations canvas fails to re-render on zoom which results in highlights and drawings being incorrectly positioned
- Includes a CSS fix for highlight create dialog on mobile

### DIFF
--- a/src/lib/annotations/doc/DocAnnotator.js
+++ b/src/lib/annotations/doc/DocAnnotator.js
@@ -36,7 +36,7 @@ const CLASS_DEFAULT_CURSOR = 'bp-use-default-cursor';
 // Required by rangy highlighter
 const ID_ANNOTATED_ELEMENT = 'bp-rangy-annotated-element';
 
-const annotationLayers = [CLASS_ANNOTATION_LAYER_HIGHLIGHT, CLASS_ANNOTATION_LAYER_DRAW];
+const ANNOTATION_LAYER_CLASSES = [CLASS_ANNOTATION_LAYER_HIGHLIGHT, CLASS_ANNOTATION_LAYER_DRAW];
 
 /**
  * For filtering out and only showing the first thread in a list of threads.
@@ -388,7 +388,7 @@ class DocAnnotator extends Annotator {
      */
     renderAnnotationsOnPage(pageNum) {
         // Scale existing canvases on re-render
-        this.scaleAnnotationCanvases(pageNum, annotationLayers);
+        this.scaleAnnotationCanvases(pageNum);
 
         super.renderAnnotationsOnPage(pageNum);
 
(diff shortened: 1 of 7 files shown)


src/lib/annotations/doc/DocAnnotator.js
```diff
@@ -384,6 +384,14 @@ class DocAnnotator extends Annotator {
      * @return {void}
      */
     renderAnnotationsOnPage(pageNum) {
+        // Force any annotation canvases on page to clear & re-render
+        const pageEl = docAnnotatorUtil.getPageEl(this.annotatedElement, pageNum);
+        const annotationLayerEl = pageEl.querySelector(`.${CLASS_ANNOTATION_LAYER_HIGHLIGHT}`);
```
JustinHoldstock: This isn't actually clearing the canvas, it's just removing the canvas from the DOM. You can either use canvas's `clearRect()` or set the `width` and or `height` (I've never been a fan of this method as an explicit clear call makes most sense to the user/reader. But it works!)

src/lib/annotations/doc/docAnnotatorUtil.js
```diff
@@ -359,47 +359,34 @@ export function isValidSelection(selection) {
...
-        annotationLayerEl.height = pxRatio * height;
-        context = annotationLayerEl.getContext('2d');
-
-        if (pxRatio !== 1) {
-            annotationLayerEl.style.width = `${width}px`;
-            annotationLayerEl.style.height = `${height}px`;
-            context.scale(pxRatio, pxRatio);
-        }
+        annotationLayerEl.width = pageDimensions.width;
+        annotationLayerEl.height = pageDimensions.height - pagePaddingTop - pagePaddingBottom;
```
JustinHoldstock: Move this to below the `if` so that the canvas is resized on every call of `getContext()`, like we discussed yesterday. If you want bonus points, I'd love it if you moved it out entirely from this function and have one that's job is to resize the canvas.

ie)
```
// clear and resize
resizeContext() {
  canvas.width = blah;
  canvas.height = blargh;
}
```

src/lib/annotations/doc/DocAnnotator.js
pramodsum: Changed this logic to re-scale rather than remove and add again

pramodsum:
tests coming soon

src/lib/annotations/doc/DocAnnotator.js
```diff
@@ -394,6 +400,17 @@ class DocAnnotator extends Annotator {
         });
     }
 
+    scaleAnnotationCanvases(pageNum, layerClasses) {
```
MinhHNguyen: missing jsdoc here

PR #348, opened 2017-08-29
Temp hide show ui for plain and comment highlights

JustinHoldstock:
This change hides UI elements in the CRUD process of Plain and Comment highlight annotations.

**THIS IS A TEMPORARY CHANGE UNTIL WE CAN PROPERLY REFACTOR COMMENT AND PLAIN HIGHLIGHTS**

As always, tests to come :)


src/lib/annotations/BoxAnnotations.js
```diff
@@ -7,7 +7,7 @@ const ANNOTATORS = [
         NAME: 'Document',
         CONSTRUCTOR: DocAnnotator,
         VIEWER: ['Document', 'Presentation'],
-        TYPE: [TYPES.point, TYPES.highlight]
+        TYPE: [TYPES.point, TYPES.highlight, TYPES.highlight_comment]
```
pramodsum: 🎉

src/lib/annotations/doc/CreateHighlightDialog.js
```diff
@@ -315,6 +312,39 @@ class CreateHighlightDialog extends EventEmitter {
      * @return {HTMLElement} The element containing Highlight creation UI
      */
     createElement() {
+        const caretTemplate = this.isMobile
```
pramodsum: should we move these templates out into constants somewhere?

src/lib/annotations/doc/CreateHighlightDialog.js
```diff
@@ -323,42 +353,52 @@ class CreateHighlightDialog extends EventEmitter {
...
-        this.commentCreateEl.addEventListener('click', this.onCommentClick);
-        this.commentBox.addListener(CommentBox.CommentEvents.post, this.onCommentPost);
-        this.commentBox.addListener(CommentBox.CommentEvents.cancel, this.onCommentCancel);
+        // Events for highlight button
+        if (this.allowHighlight) {
+            this.highlightCreateEl = containerEl.querySelector(constants.SELECTOR_ADD_HIGHLIGHT_BTN);
+            this.highlightCreateEl.addEventListener('click', this.onHighlightClick);
+
+            if (this.hasTouch) {
+                this.highlightCreateEl.addEventListener('touchstart', this.stopPropagation);
```
pramodsum: maybe move the adding of these 2 event handlers into its own method? so you can just call it anywhere on w/e element you pass in?

src/lib/annotations/doc/DocAnnotator.js
```diff
@@ -99,19 +105,35 @@ class DocAnnotator extends Annotator {
...
+            allowHighlight: this.plainHighlightEnabled
         });
-        this.createHighlightDialog.addListener(CreateEvents.plain, this.createPlainHighlight);
-        this.createHighlightDialog.addListener(CreateEvents.comment, this.highlightCurrentSelection);
-        this.createHighlightDialog.addListener(CreateEvents.commentPost, this.createHighlightThread);
+
+        if (this.commentHighlightEnabled) {
+            this.highlightCurrentSelection = this.highlightCurrentSelection.bind(this);
+            this.createHighlightDialog.addListener(CreateEvents.comment, this.highlightCurrentSelection);
+            this.createHighlightThread = this.createHighlightThread.bind(this);
```
pramodsum: new line before this

src/lib/annotations/doc/DocAnnotator.js
```diff
@@ -390,7 +423,17 @@ class DocAnnotator extends Annotator {
         // Scale existing canvases on re-render
         this.scaleAnnotationCanvases(pageNum);
 
-        super.renderAnnotationsOnPage(pageNum);
+        if (this.threads && this.threads[pageNum]) {
```
pramodsum: Actually if you just leave the call to `super.renderAnnotationsOnPage(pageNum)`
We can modify annotator.renderAnnotationsOnPage() to this since w/e annotation types aren't supported will get filtered out w/ the call to `this.isModeAnnotatable(thread.type)`
```
        if (this.threads && this.threads[pageNum]) {
            this.threads[pageNum].forEach((thread) => {
                if (!this.isModeAnnotatable(thread.type)) {
                    return;
                }
                
                thread.show();
            });
        }
    }
```

Plus then it'll be applicable to any other annotation types we don't want to render :)

pramodsum: Check if there's anywhere else we can just make calls to `isModeAnnotatable()` to Annotator or AnnotationThread methods to clean up some of hackier logic

src/lib/annotations/doc/CreateHighlightDialog.js
JustinHoldstock: That would've made more sense than what I did :D DONE!

src/lib/annotations/doc/CreateHighlightDialog.js
JustinHoldstock: I'm not sure this'd be valid for the `touchend` -> `highlightClick` event binding, as it only ever happens once. Would it be valid abstracting the binding for `stopPropagation()`?

Maybe instead of 
```
element.addEventListener('event', this.stopPropagation);
```
We'd do
```
this.bindStopPropagation(element, event);
```

Though, after seeing it, I'm not sure what the benefit of that is

src/lib/annotations/doc/DocAnnotator.js
JustinHoldstock: done

src/lib/annotations/doc/DocAnnotator.js
JustinHoldstock: I'll override the super call to do this, but I still have to not call it, as we're calling our butchered version of `show()` that takes in the flags for plain and comment highlights

JustinHoldstock: I'll make a ticket for the cleanup, as I don't want this task to turn into a refactor/cleanup PR. Then we can do it, next

src/lib/annotations/doc/CreateHighlightDialog.js
pramodsum: hmm true. I guess it wouldn't really clean up too much.

src/lib/annotations/doc/DocAnnotator.js
```diff
@@ -751,7 +800,7 @@ class DocAnnotator extends Annotator {
         // hovered over at the same time, only the top-most highlight
         // dialog will be displayed and the others will be hidden
         // without delay
-        delayThreads.forEach(showFirstDialogFilter);
+        delayThreads.forEach(showFirstDialogFilter.bind(this));
```
MinhHNguyen: is this able to be bound elsewhere?

src/lib/annotations/doc/CreateHighlightDialog.js
```diff
@@ -315,50 +328,75 @@ class CreateHighlightDialog extends EventEmitter {
...
+            }
+        }
+
+        // Events for comment button
+        if (this.allowComment) {
+            // Create comment box
+            this.commentBox = new CommentBox(containerEl);
+            this.commentCreateEl = containerEl.querySelector(`.${constants.CLASS_ADD_HIGHLIGHT_COMMENT_BTN}`);
+            this.commentCreateEl.addEventListener('click', this.onCommentClick);
+            // Event listeners
```
pramodsum: new line above comment

src/lib/annotations/doc/CreateHighlightDialog.js
```diff
@@ -315,50 +328,75 @@ class CreateHighlightDialog extends EventEmitter {
...
+            this.commentBox.addListener(CommentBox.CommentEvents.post, this.onCommentPost);
+            this.commentBox.addListener(CommentBox.CommentEvents.cancel, this.onCommentCancel);
+            // Hide comment box, by default
```
pramodsum: new line before comment

src/lib/annotations/doc/DocAnnotator.js
```diff
@@ -122,12 +149,19 @@ class DocAnnotator extends Annotator {
      */
     destroy() {
         super.destroy();
+        if (this.createHighlightDialog) {
```
pramodsum:
```
if (!this.createHighlightDialog) {
  return;
}
```

src/lib/annotations/doc/DocAnnotator.js
```diff
@@ -485,7 +542,11 @@ class DocAnnotator extends Annotator {
             this.highlightThrottleHandle = null;
         }
 
-        if (this.hasTouch && this.isMobile) {
+        if (!this.canAnnotate) {
```
pramodsum: This is now stored as `this.permissions.canAnnotate`. Update this around the code! See https://github.com/box/box-content-preview/blob/master/src/lib/annotations/Annotator.js#L70

src/lib/annotations/doc/DocHighlightDialog.js
```diff
@@ -100,6 +100,33 @@ class DocHighlightDialog extends AnnotationDialog {
...
+    show(showPlain = true, showComment = true) {
+        const plainButtonEl = this.highlightDialogEl.querySelector(`button.${constants.CLASS_ADD_HIGHLIGHT_BTN}`);
+        const commentButtonEl = this.highlightDialogEl.querySelector(
+            `button.${constants.CLASS_ADD_HIGHLIGHT_COMMENT_BTN}`
+        );
+
+        if (showPlain) {
+            annotatorUtil.showElement(plainButtonEl);
+        } else {
+            annotatorUtil.hideElement(plainButtonEl);
```
pramodsum: You wouldn't ever show the plain highlight dialog AND the comments dialog correct? Can we combine these 2 if/else statements?

src/lib/annotations/doc/CreateHighlightDialog.js
JustinHoldstock: done

src/lib/annotations/doc/CreateHighlightDialog.js
JustinHoldstock: done

src/lib/annotations/doc/DocHighlightDialog.js
JustinHoldstock: Yes we would, as that's how it currently is. This code is for only showing/hiding the buttons

src/lib/annotations/doc/DocAnnotator.js
```diff
@@ -485,6 +547,10 @@ class DocAnnotator extends Annotator {
             this.highlightThrottleHandle = null;
         }
 
+        if (!this.permissions.canAnnotate) {
```
pramodsum: i don't think this is needed?

src/lib/annotations/doc/DocHighlightDialog.js
pramodsum: ah right not showing/hiding specific types of dialogs. got it

src/lib/annotations/doc/DocAnnotator.js
```diff
@@ -391,7 +432,19 @@ class DocAnnotator extends Annotator {
         // Scale existing canvases on re-render
         this.scaleAnnotationCanvases(pageNum);
 
-        super.renderAnnotationsOnPage(pageNum);
+        if (!this.threads) {
+            return;
+        }
+
+        // TODO (@jholdstock|@spramod) remove this if statement, and make super call, upon refactor.
+        const pageThreads = this.getThreadsOnPage(pageNum);
```
pramodsum: at this point, isn't this the same as calling `super.renderAnnotationsOnPage()`? Since that also checks `isModeAnnotatable(type)` and that's essentially when `this.plainHighlightEnabled` and `this.commentHighlightEnabled` is?

JustinHoldstock: Note that the override is passing in the two flags, which are **not** done in the parent class
```
 thread.show(this.plainHighlightEnabled, this.commentHighlightEnabled);
```

src/lib/annotations/doc/DocAnnotator.js
JustinHoldstock: I'd say it's unnecessary to remove it, as we don't want to be removing event listeners that don't exist. I'm not sure what our standard is for that. @jeremypress @tonyjin @MinhHNguyen Thoughts?

tonyjin: It's safe to remove event listeners that don't exist. I'd personally prefer fewer lines of code over adding a guard before removing event listeners but I don't think this is something we need to standardize on.

PR #355, opened 2017-08-30
New: Emit file version id with viewer events

pramodsum:
- Also emits the current page number from DocBaseViewer.js

### DIFF
--- a/src/lib/viewers/doc/DocBaseViewer.js
+++ b/src/lib/viewers/doc/DocBaseViewer.js
@@ -437,6 +437,30 @@ class DocBaseViewer extends BaseViewer {
     //--------------------------------------------------------------------------
 
     /**
+     * Emits a generic viewer event
+     *
+     * @override
+     * @protected
+     * @emits viewerevent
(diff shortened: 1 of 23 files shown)


src/lib/viewers/doc/DocBaseViewer.js
```diff
@@ -437,6 +437,30 @@ class DocBaseViewer extends BaseViewer {
...
+        const { file, viewer } = this.options;
+
+        super.emit(event, data);
+        super.emit('viewerevent', {
+            event,
+            data,
+            viewerName: viewer ? viewer.NAME : '',
+            fileId: file,
+            fileVersionId: file.file_version.id,
+            currentPage: this.getCachedPage()
```
jeremypress: Why are we overriding this here? Can't you get the page number from the relevant events that will fire from the 'viewerevents' coming from docbase?

pramodsum:
Unsure why this change was required. I'll re-submit a modified version of this if it is still needed.

PR #356, opened 2017-08-31
Feature: Drawing boundary updates on each action

MinhHNguyen:
Draw the bounded drawing on stroke, undo, redo, and page rescale. Also contains scaffolding for dialog UI.


src/lib/annotations/doc/DocDrawingDialog.js
```diff
@@ -0,0 +1,11 @@
+import AnnotationDialog from '../AnnotationDialog';
+
+class DocDrawingDialog extends AnnotationDialog {
+    addAnnotation() {}
```
JustinHoldstock: Docs for these functions? And maybe for the class

MinhHNguyen: Will add them

PR #358, opened 2017-09-01
New: Separate create and view permissions for annotations

pramodsum:
Enables annotations to be shown in a view-only setting with the correct scopes

Annotations permissions will be as follows:
- can_annotate - whether a user can add any new annotations
- can_view_annotations_all - whether a user can view all annotations
- can_view_annotations_self - whether a user can view their own annotations (The API should return only the current user's annotations for this scope)

### DIFF
--- a/src/lib/annotations/Annotator.js
+++ b/src/lib/annotations/Annotator.js
@@ -52,7 +52,6 @@ class Annotator extends EventEmitter {
     constructor(data) {
         super();
 
-        this.canAnnotate = data.canAnnotate;
         this.container = data.container;
         this.options = data.options;
         this.fileVersionId = data.fileVersionId;
@@ -65,6 +64,14 @@ class Annotator extends EventEmitter {
 
         const { CONTROLLERS } = this.options.annotator || {};
         this.modeControllers = CONTROLLERS || {};
+
+        // Get annotation permissions
+        const permissions = data.options.permissions || {};
+        this.permissions = {
+            canAnnotate: permissions.can_annotate || false,
+            canViewAllAnnotations: permissions.can_view_annotations_all || false,
+            canViewOwnAnnotations: permissions.can_view_annotations_self || false
+        };
     }
 
     /**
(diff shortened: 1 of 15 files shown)


src/lib/annotations/Annotator.js
```diff
@@ -162,7 +168,7 @@ class Annotator extends EventEmitter {
      */
     showModeAnnotateButton(currentMode) {
         const mode = this.modeButtons[currentMode];
-        if (!mode || !this.isModeAnnotatable(currentMode)) {
+        if (!mode || !this.permissions.canAnnotate || !this.isModeAnnotatable(currentMode)) {
```
pramodsum: Ensures button to create point/draw annotations does not display

src/lib/annotations/Annotator.js
```diff
@@ -534,12 +540,20 @@ class Annotator extends EventEmitter {
     fetchAnnotations() {
         this.threads = {};
 
+        // Do not load any pre-existing annotations if the user does not have
+        // the correct permissions
+        if (!this.permissions.canViewAllAnnotations || !this.permissions.canViewOwnAnnotations) {
```
pramodsum: Does not fetch any annotations if the user cannot view any/their own annotations

src/lib/annotations/doc/DocHighlightDialog.js
```diff
@@ -293,9 +293,12 @@ class DocHighlightDialog extends AnnotationDialog {
             ]);
             annotatorUtil.showElement(highlightLabelEl);
 
-            // Hide delete button on plain highlights if user doesn't have
-            // permissions
-            if (annotations[0].permissions && !annotations[0].permissions.can_delete) {
+            if (!this.canAnnotate) {
```
pramodsum: Hides buttons in the plain highlight dialog if the user does not have `can_annotate` permissions

pramodsum:
Tests coming soon!

PR #359, opened 2017-09-01
Update: Allow passed in collections to be of file objects

priyajeet:
Allows tokens to be null or undefined. Useful when the passed in file or collection has well formed file objects. This is for offline use case or the use case where the parent project (like ContentExplorer UI Element) has already made the call.

Also changes the token getter to pass in typed ids for files and handle cases when typed ids are returned from the token service to make it consistent with other UI Elements. Since all the UI elements can either be requesting `folder` or `file` ids, the token service needs to know what type of id we are sending it, as such either `file_` or `folder_` is prefixed to all the ids. Likewise the token service can/will return typed ids after fetching tokens. So it needs to be mapped back to what preview wants.

### DIFF
--- a/src/lib/Preview.js
+++ b/src/lib/Preview.js
@@ -155,12 +155,19 @@ class Preview extends EventEmitter {
      * @return {void}
      */
     show(fileIdOrFile, token, options = {}) {
-        // Save a reference to the options to be used later
-        if (typeof token === 'string' || typeof token === 'function') {
+        // Save a reference to the options to be re-used later.
+        // Token should either be a function or a string.
+        // Token can also be null or undefined for offline use case.
+        // But it cannot be a random object.
+        if (token === null || typeof token !== 'object') {
             this.previewOptions = Object.assign({}, options, { token });
         } else {
-            throw new Error('Missing access token!');
+            throw new Error('Bad access token!');
         }
+
+        // Update the optional file navigation collection and caches
+        // if proper valid file objects were passed in.
+        this.updateCollection(options.collection);
 
         // Load the preview
         this.load(fileIdOrFile);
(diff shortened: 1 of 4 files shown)


priyajeet:
Still fixing tests

src/lib/Preview.js
```diff
@@ -155,13 +155,20 @@ class Preview extends EventEmitter {
...
-        if (typeof token === 'string' || typeof token === 'function') {
+        // Save a reference to the options to be re-used later.
+        // Token should either be a function or a string.
+        // Token can also be null or undefined for offline use case.
+        // But it cannot be a random object.
+        if (token === null || typeof token !== 'object') {
             this.previewOptions = Object.assign({}, options, { token });
         } else {
-            throw new Error('Missing access token!');
+            throw new Error('Bad access token!');
```
JustinHoldstock: Should this be a translated string?

src/lib/Preview.js
```diff
@@ -187,16 +194,41 @@ class Preview extends EventEmitter {
...
+        fileOrIds.forEach((fileOrId) => {
+            if (fileOrId && typeof fileOrId === 'string') {
+                // String id found in the collection
+                fileIds.push(fileOrId);
+            } else if (fileOrId && typeof fileOrId === 'object' && typeof fileOrId.id === 'string') {
+                // Possible well-formed file object found in the collection
+                fileIds.push(fileOrId.id);
+                files.push(fileOrId);
+            } else {
+                throw new Error('Bad collection provided!');
```
JustinHoldstock: Translated string?

src/lib/tokens.js
```diff
@@ -1,19 +1,46 @@
 // Create an error to throw if needed
-const error = new Error('Missing Auth Token!');
+const error = new Error('Bad Auth Token!');
```
JustinHoldstock: Another translated string

JustinHoldstock: Also, I can't say I've ever seen anyone reuse an Error Object. Should there be any gotchas with that?

PR #362, opened 2017-09-02
Feature: drawingSelectionUI

MinhHNguyen:
- Select saved drawings
- Pop up dialog and position dialog in drawing mode
- Dialog implementation is in a separate PR


src/lib/annotations/doc/DocAnnotator.js
```diff
@@ -454,6 +454,10 @@ class DocAnnotator extends Annotator {
     bindDOMListeners() {
         super.bindDOMListeners();
 
+        if (!this.permissions.canAnnotate) {
```
pramodsum: double check that this doesn't block triggering highlight annotations

src/lib/annotations/doc/DocAnnotator.js
```diff
@@ -485,6 +489,12 @@ class DocAnnotator extends Annotator {
             this.highlightThrottleHandle = null;
         }
 
+        if (!this.permissions.canAnnotate) {
```
pramodsum: same as above

src/lib/annotations/doc/DocDrawingDialog.js
```diff
@@ -1,6 +1,8 @@
 import AnnotationDialog from '../AnnotationDialog';
 
 class DocDrawingDialog extends AnnotationDialog {
+    visible = false;
```
jeremypress: JSDoc in here

src/lib/annotations/doc/DocDrawingThread.js
```diff
@@ -134,23 +139,18 @@ class DocDrawingThread extends DrawingThread {
...
     saveAnnotation(type, text) {
-        this.emit('annotationevent', {
-            type: 'drawcommit'
-        });
         this.reset();
 
         // Only make save request to server if there exist paths to save
-        const { undoCount } = this.pathContainer.getNumberOfItems();
-        if (undoCount === 0) {
+        if (this.pathContainer.isUndoEmpty()) {
```
jeremypress: If you're reusing undo logic to determine if we should save I might rename to just isEmpty since you're already accessing the path container.

src/lib/annotations/drawing/DrawingModeController.js
```diff
@@ -96,6 +97,34 @@ class DrawingModeController extends AnnotationModeController {
...
+        annotatorUtil.disableElement(this.undoButtonEl);
+        annotatorUtil.disableElement(this.redoButtonEl);
+    }
+
+    /**
+     * Deselect a saved and selected thread
+     *
+     * @return {void}
+     */
+    cleanSelector() {
```
jeremypress: "clean" doesn't really mean much to me, maybe destroy or remove?

src/lib/annotations/annotatorUtil.js
```diff
@@ -448,7 +448,8 @@ export function validateThreadParams(thread) {
 export function eventToLocationHandler(locationFunction, callback) {
     return (event) => {
         const evt = event || window.event;
-        if (!evt) {
+        // Do nothing when the target isn't the text layer in case the text layer receives event precedence over buttons
+        if (!evt || (event.target && event.target.className !== 'textLayer')) {
```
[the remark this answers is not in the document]
MinhHNguyen: This is indeed only for documents. I have only tested this on documents with textlayers. Clearly this would not work for images.

src/i18n/en-US.properties
```diff
@@ -135,6 +135,12 @@ annotation_highlight_toggle=Highlight text
 annotation_highlight_comment=Add comment to highlighted text
 # Text for which user made the highlight annotation
 annotation_who_highlighted={1} highlighted
+# Text for which user made the drawn annotation
+annotation_who_drew={1} drew
```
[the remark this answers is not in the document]
pramodsum: Personally I prefer keeping this consistent w/ highlights so "{1} highlighted" or "{1} drew"

src/lib/annotations/doc/DocAnnotator.js
MinhHNguyen: Removing these. For some reason I had thought that the upstream codebase had `if (!this.canAnnotate)`

src/lib/annotations/doc/DocDrawingDialog.js
MinhHNguyen: JSDoc is in #364, I separated them to make it easier to read. DocDrawingDialog is simply a stub in this PR.

src/lib/annotations/annotatorUtil.js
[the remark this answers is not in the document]
MinhHNguyen: Adding a note about this for future development

src/lib/annotations/drawing/DrawingModeController.js
MinhHNguyen: changed to removeSelection

PR #363, opened 2017-09-02
Fix: annotation mode exits on wrong button and fetch annotations returns non-promise

MinhHNguyen:
- Annotation mode exits using the wrong button element. This causes the wrong one to become in-active
- When calling fetchannotations() without permissions to view all or own annotations, a non-promise unexpectedly returned.

### DIFF
--- a/src/lib/annotations/Annotator.js
+++ b/src/lib/annotations/Annotator.js
@@ -369,7 +369,7 @@ class Annotator extends EventEmitter {
         const buttonEl = event.target || this.getAnnotateButton(buttonSelector);
 
         // Exit any other annotation mode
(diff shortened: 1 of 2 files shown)


src/lib/annotations/Annotator.js
```diff
@@ -543,7 +545,7 @@ class Annotator extends EventEmitter {
         // Do not load any pre-existing annotations if the user does not have
         // the correct permissions
         if (!this.permissions.canViewAllAnnotations || !this.permissions.canViewOwnAnnotations) {
-            return this.threads;
```
pramodsum: oh shoot. Great catch! Thanks!

PR #366, opened 2017-09-05
New: Scroll file to annotation position on 'scrolltoannotation'

pramodsum:
- This expects that 'scrolltoannotation' is emitted to the viewer with
  an object of the format { threadID: '123abc', page: ### }


src/lib/annotations/Annotator.js
```diff
@@ -832,6 +831,28 @@ class Annotator extends EventEmitter {
...
+     *
+     * @private
+     * @param {Object} data - annotation data for thread that should scroll into view
+     * @param {string} data.threadID - annotation thread id
+     * @param {string} data.page - annotation page number
+     * @return {void}
+     */
+    scrollToAnnotation(data) {
+        const { threadID, page } = data;
+        if (!threadID || !page) {
```
pramodsum: Should we be displaying or emitting some kind of error if incorrect params are passed into this method? Say if the threadID corresponds to a thread that is not on the specified page?

JustinHoldstock: I'd almost say that we should have some sort of invalidation occur if a thread doesn't exist on a page, which in turn will potentially delete the annotation. Thoughts? Is it possible for an annotation to be saved onto a page that it doesn't belong to?

src/lib/annotations/doc/DocHighlightThread.js
```diff
@@ -118,6 +118,15 @@ class DocHighlightThread extends AnnotationThread {
         }
     }
 
+    scrollIntoView() {
```
JustinHoldstock: Docs

src/lib/annotations/Annotator.js
pramodsum: I guess the only time I could see that occurring is if annotations were created from the API. I don't think the thread would be deleted in this case, there would just be no action, or indication that anything even happened

pramodsum: Since annotation page numbers are stored in the details blob in the DB, there's not really any validation that the API would do if "invalid" annotations were created. When rendering that annotation on our end, it would trigger a notification at the top saying "some annotations failed to load" so maybe triggering similar behavior if someone tries to scroll to something that doesn't exist on the file?

@varunmaker would it make sense to not display any kind of error here or show some kind of custom annotations error here?

pramodsum: Removing the page number from the emitted event. This should prevent any kind of invalidation checks that would needed.

src/lib/annotations/AnnotationThread.js
```diff
@@ -226,6 +226,45 @@ class AnnotationThread extends EventEmitter {
...
+
+    /**
+     * Scroll to the annotation's page
+     *
+     * @private
+     * @return {void}
+     */
+    scrollToPage() {
+        const pageEl = this.annotatedElement.querySelector(`[data-page-number="${this.location.page}"]`);
+        pageEl.scrollIntoView();
```
JustinHoldstock: maybe a guard here, in case the `.location` doesn't contain a page?

src/lib/annotations/Annotator.js
```diff
@@ -835,6 +834,26 @@ class Annotator extends EventEmitter {
...
+     * @private
+     * @param {Object} threadID - annotation threadID for thread that should scroll into view
+     * @return {void}
+     */
+    scrollToAnnotation(threadID) {
+        if (!threadID) {
+            return;
+        }
+
+        Object.values(this.threads).forEach((pageThreads) => {
```
JustinHoldstock: I don't remember, do we have a polyfill for `Object.values()`? IE does not support it

pramodsum: shoot. swapping it for object.keys

pramodsum: Fixed that and others here https://github.com/box/box-content-preview/pull/371

PR #377, opened 2017-09-08
New: Emit annotation thread events to the viewer and beyond

pramodsum:
Emits annotator and annotation thread specific events to the Viewer and Preview

Annotator Events: 
- annotationmodeenter: Enter an annotation mode
- annotationmodeexit: Exit an annotation mode
- annotationsfetched: Annotations have been fetched from the API
- annotationerror: Annotation error occurred

Annotation Thread Events:
- annotationpending: Thread was created but not saved
- annotationthreadsaved: Annotation thread saved on the server
- annotationthreaddeleted: Annotation thread deleted on the server
- annotationsaved: Annotation saved on an existing thread on the server
- annotationdeleted: Annotation deleted from an existing thread on the server
- annotationdeleteerror: Error while deleting an annotation on either an existing/new thread
- annotationcanceled: Annotation cancelled from posting on either an existing/new thread
- annotationcreateerror: Error while creating an annotation on either an existing/new thread

### DIFF
--- a/src/lib/annotations/AnnotationThread.js
+++ b/src/lib/annotations/AnnotationThread.js
@@ -134,7 +134,7 @@ class AnnotationThread extends EventEmitter {
      *
      * @param {string} type - Type of annotation
      * @param {string} text - Text of annotation to save
-     * @return {void}
+     * @return {Promise} - Annotation create promise
      */
     saveAnnotation(type, text) {
         const annotationData = this.createAnnotationData(type, text);
@@ -155,16 +155,10 @@ class AnnotationThread extends EventEmitter {
         // Changing state from pending
         this.state = STATES.hover;
         // Save annotation on server
-        this.annotationService
+        return this.annotationService
             .create(annotationData)
             .then((savedAnnotation) => this.updateTemporaryAnnotation(tempAnnotation, savedAnnotation))
-            .catch(() => {
-                // Remove temporary annotation
-                this.deleteAnnotation(tempAnnotationID, /* useServer */ false);
-
-                // Broadcast error
-                this.emit(THREAD_EVENT.createError);
-            });
+            .catch(() => this.handleThreadSaveError(tempAnnotationID));
     }
 
     /**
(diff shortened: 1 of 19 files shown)


pramodsum:
@varunmaker the events listed above will be emitted from Annotations

src/lib/annotations/AnnotationThread.js
```diff
@@ -149,30 +155,24 @@ class AnnotationThread extends EventEmitter {
...
      * @param {boolean} [useServer] - Whether or not to delete on server, default true
-     * @return {void}
+     * @return {Promise} - Annotation delete promise
      */
     deleteAnnotation(annotationID, useServer = true) {
         // Ignore if no corresponding annotation exists in thread or user doesn't have permissions
         const annotation = this.annotations.find((annot) => annot.annotationID === annotationID);
         if (!annotation || (annotation.permissions && !annotation.permissions.can_delete)) {
-            return;
+            return Promise.resolve();
```
JustinHoldstock: Could this be a Promise.reject() with a message that the annotation either "does not exist" or "cannot be deleted"?

pramodsum: Oh. Yea i think that's totally fair. I'm not sure it make sense to reject the promise later on in this method if `useServer = false` but here i think it makes sense. You don't have permission to do what you're trying to do so therefore should get an error

src/lib/annotations/Annotator.js
```diff
@@ -669,7 +668,7 @@ class Annotator extends EventEmitter {
         if (!service || !(service instanceof AnnotationService)) {
             return;
         }
-        service.removeAllListeners('annotatorerror');
+        service.removeAllListeners(ANNOTATOR_EVENT.error);
```
jeremypress: `removeListeners` + the function name

src/lib/annotations/Annotator.js
```diff
@@ -705,10 +694,7 @@ class Annotator extends EventEmitter {
      * @return {void}
      */
     unbindCustomListenersOnThread(thread) {
-        thread.removeAllListeners('threaddeleted');
-        thread.removeAllListeners('threadcleanup');
-        thread.removeAllListeners('annotationsaved');
-        thread.removeAllListeners('annotationevent');
+        thread.removeAllListeners('threadevent');
```
jeremypress: I think if you remove the args this should work. Make sure they're actually getting removed though

PR #378, opened 2017-09-09
New: More specific error messages from conversion

jeremypress:
Tests coming soon.

### DIFF
--- a/src/i18n/en-US.properties
+++ b/src/i18n/en-US.properties
@@ -48,8 +48,6 @@ error_rate_limit=We're sorry, the preview didn't load because your request was r
 error_reupload=We're sorry, the preview didn't load. Please re-upload the file or contact Box support.
 # Preview error message shown when the user's browser doesn't support previews of this file type
 error_browser_unsupported=We're sorry, your browser doesn't support preview for {1}.
-# Preview error message shown when the file has invalid formatting (most likely a 3D file)
-error_file_type_unsupported=We're sorry, this file format is not supported.
 # Preview error message shown when an iWork file is previewed
 error_iwork=We're sorry, iWork files are not currently supported.
 # Preview error message shown when document loading fails (most likely due to password or watermark)
@@ -59,7 +57,7 @@ error_password_protected=We're sorry the preview didn't load. This document is p
 # Preview error message shown when conversion was unable to process the file at the given time.
 error_try_again_later=We're sorry the preview didn't load. Please try again later.
 # Preview error message shown when conversion failed due to file contents
-error_unsupported_format=We're sorry the preivew didn't load. This file may be corrupted.
+error_bad_file=We're sorry the preivew didn't load. This file may be corrupted.
 
 # Media Preview
 # Label for changing speed in media player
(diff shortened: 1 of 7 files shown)


src/lib/viewers/BaseViewer.js
```diff
@@ -229,13 +229,13 @@ class BaseViewer extends EventEmitter {
     };
 
     /**
-     * Emits an error when an asset (static or representation) fails to load.
+     * triggers an error when an asset (static or representation) fails to load.
```
pramodsum: Capitalize "triggers"

src/lib/RepStatus.js
```diff
@@ -4,13 +4,18 @@ import { STATUS_SUCCESS, STATUS_VIEWABLE } from './constants';
...
 class RepStatus extends EventEmitter {
     /**
      * Gets the status out of represenation
      *
      * @public
      * @param {Object} representation - representation object
-     * @return {string} rep status instance
+     * @return {string} rep status state
      */
     static getStatus(representation) {
```
pramodsum: Not sure this makes sense to be called getStatus() anymore since it's technically returning the state of the RepStatus rather than the status itself. Might just be a nitpick but thoughts?

src/lib/viewers/BaseViewer.js
```diff
@@ -244,11 +244,11 @@ class BaseViewer extends EventEmitter {
      *
      * @protected
      * @emits error
-     * @param {Error} [err] - Optional error with message
+     * @param {Error|string} [err] - Optional error or stringwith message
```
pramodsum: or string with

src/lib/RepStatus.js
jeremypress: Agreed that it is a little odd but I think that the `state` is actually the `status` since the `status` as we knew it became an object a while back.

src/i18n/en-US.properties
```diff
@@ -54,6 +54,12 @@ error_file_type_unsupported=We're sorry, this file format is not supported.
 error_iwork=We're sorry, iWork files are not currently supported.
 # Preview error message shown when document loading fails (most likely due to password or watermark)
 error_document=We're sorry, the preview didn't load. This document may be protected.
+# Preview error message shown when the document is password protected
+error_password_protected=We're sorry the preview didn't load. This document is protected.
+# Preview error message shown when conversion was unable to process the file at the given time.
+error_try_again_later=We're sorry the preview didn't load. Please try again later.
+# Preview error message shown when conversion failed due to file contents
+error_unsupported_format=We're sorry the preivew didn't load. This file may be corrupted.
```
JustinHoldstock: Is this supposed to be `error_unsupported_format`? Or should it be changed to `error_corrupted_file`?

src/lib/RepStatus.js
```diff
@@ -4,13 +4,18 @@ import { STATUS_SUCCESS, STATUS_VIEWABLE } from './constants';
 
 const STATUS_UPDATE_INTERVAL_MS = 2000;
 
+const ERROR_PASSWORD_PROTECTED = 'error_password_protected';
+const ERROR_TRY_AGAIN_LATER = 'error_try_again_later';
+const ERROR_UNSUPPORTED_FORMAT = 'error_unsupported_format';
```
JustinHoldstock: See above comment

src/i18n/en-US.properties
jeremypress: good point. Harris is going to look into all of the cases that cause this error to fire.

jeremypress: This applies to a variety of situations including both unsupported files and corrupted ones. Technically a corrupted file is unsupported so that's slightly more of an umbrella term?

JustinHoldstock: Also, spelling: 'preview' not 'preivew'

PR #379, opened 2017-09-09
Fix: Respect view permissions when viewer.initAnnotations() is called

pramodsum:
Annotations should be rendered in view-only if at LEAST one of the correct annotation permissions are true. This fully completes the separation of create + view permissions from #358.

### DIFF
--- a/src/lib/viewers/BaseViewer.js
+++ b/src/lib/viewers/BaseViewer.js
@@ -12,12 +12,8 @@ import {
     prefetchAssets,
     createAssetUrlCreator
 } from '../util';
-import { checkPermission } from '../file';
 import Browser from '../Browser';
 import {
-    PERMISSION_ANNOTATE,
-    PERMISSION_CAN_VIEW_ANNOTATIONS_ALL,
-    PERMISSION_CAN_VIEW_ANNOTATIONS_SELF,
     CLASS_FULLSCREEN,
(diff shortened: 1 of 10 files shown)


src/lib/viewers/BaseViewer.js
```diff
@@ -666,8 +668,10 @@ class BaseViewer extends EventEmitter {
 
         const { file } = this.options;
         this.canAnnotate = checkPermission(file, PERMISSION_ANNOTATE);
+        this.canViewAllAnnotations = checkPermission(file, PERMISSION_CAN_VIEW_ANNOTATIONS_ALL);
+        this.canViewOwnAnnotations = checkPermission(file, PERMISSION_CAN_VIEW_ANNOTATIONS_SELF);
 
-        if (this.canAnnotate) {
+        if (this.canAnnotate || this.canViewAllAnnotations || this.canViewOwnAnnotations) {
```
JustinHoldstock: Could we abstract this into the annotator util we have? Then you could do something like:
```
if(annotatorUtil.enabledForFile(file)) {
  this.initAnnotations();
}
```

pramodsum: Moved the permissions check into BoxAnnotations.determineAnnotator() so that we don't introduce more annotations specific code into BaseViewer.js

PR #380, opened 2017-09-09
New: New design for point annotation icons

pramodsum:
- The current method of color assignment does not guarantee unique
  avatar colors for each user. That will be implemented in a separate
  pull request

screen shot 2017-09-08 at 8 43 36 pm


JustinHoldstock:
Curious, how are you creating the initials? Does it work if there's no first name/last name username?

src/lib/annotations/AnnotationThread.js
```diff
@@ -551,6 +568,27 @@ class AnnotationThread extends EventEmitter {
...
+     * @return {void}
+     */
+    fillPointAnnotationIcon(annotation) {
+        // Update point annotation icon with user initials
+        const userName = annotation.user.name;
+        const userInitials = userName
+            .split(' ')
+            .map((n) => n[0])
+            .join('');
+        this.element.innerHTML = userInitials;
```
JustinHoldstock: Answered my own question. Does this work with missing username? Does this cap at two initials?

pramodsum: It doesn't cap at 2 initials. I'm not entirely sure how we'd deal w/ people with a lot of names in their full name. Maybe cap it at 2 to include the most common case of first, middle and last? I'll confirm this  

All users are required to have a name so we wouldn't run into a null case here but we do have the workaround that we've been seeing where a user can have a space for the name? We decided to allow a blank circle in those cases.

pramodsum: Going to only generate initials from the user's first and last name. See the tests here
https://github.com/box/box-content-preview/pull/380/commits/5255bb2c03e996f915f8dce911b7dd0db1882fba

pramodsum:
Shoot you're right. Accounting for that and adding a test case!

src/lib/annotations/AnnotationThread.js
```diff
@@ -612,6 +624,22 @@ class AnnotationThread extends EventEmitter {
...
+     * Inserts user initials and updates point annotation icon with a
+     * color pseudo-unique to the user
+     *
+     * @private
+     * @param {Object} annotation - first annotation in the thread
+     * @return {void}
+     */
+    fillPointAnnotationIcon(annotation) {
+        // Update point annotation icon with user initials
+        this.element.innerHTML = annotatorUtil.getUserInitials(annotation.user.name);
```
JustinHoldstock: Can we not use `innerHTML` and instead create a span, and add the initials to it?
Or use `textContent` ?

src/lib/annotations/Annotator.scss
```diff
@@ -492,4 +456,43 @@ $avatar-color-9: #f22c44;
     pointer-events: none;
 }
 
+.bp-point-annotation-marker,
+.bp-annotation-profile {
+    &.avatar-color-0 {
```
JustinHoldstock: Can you somehow use a SASS Mixin, here?

src/lib/annotations/annotatorUtil.js
```diff
@@ -288,8 +320,7 @@ export function getAvatarHtml(avatarUrl, userId, userName) {
             .substring(0, 3);
     }
 
-    const index = parseInt(userId, 10) || 0;
-    return `<div class="bp-annotation-profile avatar-color-${index % AVATAR_COLOR_COUNT}">${initials}</div>`.trim();
+    return `<div class="bp-annotation-profile ${getUserColor(userId)}">${initials}</div>`.trim();
```
JustinHoldstock: I'd rather we return an HTML node instead of text (avoid using `innerHTML`)

pramodsum: I'm going to add this to a different ticket to do after because it requires a refactoring of how we create the entire annotation element

PR #384, opened 2017-09-12
Fix: fix page change flickering when zoomed in

jeremypress:
@pramodsum I reintroduced the concept of previous scroll position that you had before to fix the case where you are zoomed in enough that you would scroll back and for after a page change.


src/lib/util.js
```diff
@@ -730,20 +730,24 @@ export function removeActivationListener(element, handler) {
...
     const currentPageMiddleY = currentPageEl.offsetTop + currentPageEl.clientHeight / 2;
 
-    if (currentScrollTop > currentPageMiddleY) {
-        return currentPageNum + 1;
-    } else if (currentScrollBottom < currentPageMiddleY) {
-        return currentPageNum - 1;
+    // Scrolling down
+    if (currentScrollTop > previousScrollTop) {
+        pageNum = currentScrollTop > currentPageMiddleY ? currentPageNum + 1 : currentPageNum;
+        // Scrolling up
```
pramodsum: move comment inside `else{}`

src/lib/util.js
```diff
@@ -730,20 +730,24 @@ export function removeActivationListener(element, handler) {
...
+    let pageNum = currentPageNum;
     const currentScrollTop = wrapperEl.scrollTop;
     const currentScrollBottom = wrapperEl.scrollTop + wrapperEl.offsetHeight;
     const currentPageMiddleY = currentPageEl.offsetTop + currentPageEl.clientHeight / 2;
 
-    if (currentScrollTop > currentPageMiddleY) {
-        return currentPageNum + 1;
-    } else if (currentScrollBottom < currentPageMiddleY) {
-        return currentPageNum - 1;
+    // Scrolling down
```
pramodsum: move comment inside `if() {}`

jeremypress: done

PR #322, opened 2017-08-17
Selectively disable annotations

JustinHoldstock:
Annotations can now be selectively disabled through preview options. 
Viewer options override global `showAnnotations` value, for that viewer.
Tests to follow.
```javascript
// Turn on/off annotations for all viewers
preview.show(..., {
    showAnnotations: Boolean
});
```
Combined with the following:
```javascript
// [LEGACY] BOOLEAN - turn on/off annotations for a viewer
preview.show(..., {
    viewers: {
        VIEWER_NAME: {
            annotations: Boolean
        }
    }
});

// [NEW] OBJECT - turn on/off annotations for a viewer.
preview.show(..., {
    viewers: {
        VIEWER_NAME: {
            annotations: {
                enabled: Boolean, // Enables/disables if set. Respects "showAnnotations" if empty
                disabledTypes: String[] // List of annotation types to disable
            }
        }
    }
});
```

And an example. Enable all annotations, turn off for Image Viewers, and
disable `point` annotations on Document viewer:

```javascript
preview.show(fileId, token, {
    showAnnotations: true,
    viewers: {
        Image: {
            annotations: {
                enabled: false
            }
        },
        Document: {
            annotations: {
                disabledTypes: ['point']
            }
        }
    }
});
```

PR #387, opened 2017-09-12
Update annotations readme

JustinHoldstock:
Updating annotation related READMEs with new properties added by https://github.com/box/box-content-preview/pull/322


JustinHoldstock:
@pramodsum @tonyjin @jeremypress Do you think I should link to https://github.com/box/box-content-preview/pull/322 in the README, so users can see this in play?

tonyjin:
Feels a bit odd to have documentation in a PR, we should perhaps have an annotations documentation file and link to that, like what we do for the viewers.

pramodsum:
Yea I have it in my next sprint i think to start writing up documentation for the entire annotations codebase. It makes sense to add it to there.

PR #388, opened 2017-09-12
Chore: Adding console errors for annotation errors

pramodsum:
@JustinHoldstock are there any other areas you think we could use console errors?


src/lib/viewers/BaseViewer.js
```diff
@@ -346,7 +346,9 @@ class BaseViewer extends EventEmitter {
             }
 
             if (this.annotationsPromise) {
-                this.annotationsPromise.then(this.loadAnnotator);
+                this.annotationsPromise.then(this.loadAnnotator).catch((error) => {
```
JustinHoldstock: newline the `catch()`

pramodsum: Our linter puts it back in one line 😛

PR #391, opened 2017-09-13
Chore: Optimize annotations fetch by separating permissions

pramodsum:
- Storing thread.annotations as an object by annotationId rather than an
  array of annotations


pramodsum:
We are making optimizations in the API requests which should negate the need for this change

PR #394, opened 2017-09-15
Fix: Only call setOriginalImageSize() on single page images

pramodsum:
Will add tests soon


src/lib/viewers/image/ImageViewer.js
```diff
@@ -397,12 +412,59 @@ class ImageViewer extends ImageBaseViewer {
...
+    /**
+     * Sets the original image width and height on the img element. Can be removed when
+     * naturalHeight and naturalWidth attributes work correctly in IE 11.
+     *
+     * @private
+     * @param {HTMLElement} imageEl - The image to set the original size attributes on
+     * @return {Promise} A promise that is resolved if the original image dimensions were set.
+     */
+    setOriginalImageSize(imageEl) {
+        const promise = new Promise((resolve) => {
```
priyajeet: Don't need to wrap stuff.

```js
if (imageEl.naturalWidth && imageEl.naturalHeight) {
    imageEl.setAttribute('originalWidth', imageEl.naturalWidth);
    imageEl.setAttribute('originalHeight', imageEl.naturalHeight);
    return Promise.resolve();
}

return get().then(...).catch(...)
```

PR #396, opened 2017-09-15
New: Allow documents to be viewed in presentation mode

jeremypress:
Approach here is to add another entry for the presentationViewer which includes all document types. This comes after the document viewer in the hierarchy, so it's only triggered when the document viewer is disabled via the options.


src/lib/viewers/doc/DocLoader.js
```diff
@@ -5,9 +5,81 @@ import PresentationViewer from './PresentationViewer';
 import RepStatus from '../../RepStatus';
 import { ORIGINAL_REP_NAME, STATUS_SUCCESS } from '../../constants';
 
-// Order of the viewers matters. Prefer original before others. Go from specific to general.
```
JustinHoldstock: Can we make a note to move out document formats to a different file, and make a format constants file? Maybe `formats.js`, or something like that? Thoughts?

PR #403, opened 2017-09-19
New: Draw annotations UI

jeremypress:
Tests incoming

### DIFF
--- a/src/lib/PreviewUI.js
+++ b/src/lib/PreviewUI.js
@@ -5,6 +5,7 @@ import { insertTemplate } from './util';
 import {
     CLASS_HIDDEN,
     CLASS_INVISIBLE,
+    CLASS_BOX_PREVIEW_BASE_HEADER,
     CLASS_BOX_PREVIEW_HAS_HEADER,
     CLASS_BOX_PREVIEW_HEADER,
     CLASS_BOX_PREVIEW_THEME_DARK,
@@ -311,6 +312,24 @@ class PreviewUI {
         this.notification.hide();
     }
 
+    /**
+     * Replaces the currently active header with a specified header
+     *
+     * @public
+     * @param {string} replacementHeader - Class name of new header
+     * @return {void}
+     */
+    replaceHeader(replacementHeader) {
+        // First hide all possible headers
+        this.container.querySelectorAll(`.${CLASS_BOX_PREVIEW_HEADER}`).forEach((element) => {
+            element.classList.add(CLASS_HIDDEN);
+        });
+
+        // Show the specified header
+        const headerToShow = this.container.querySelector(replacementHeader);
+        headerToShow.classList.remove(CLASS_HIDDEN);
+    }
+
     //--------------------------------------------------------------------------
     // Private
     //--------------------------------------------------------------------------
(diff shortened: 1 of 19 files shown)


src/lib/PreviewUI.js
```diff
@@ -311,6 +312,24 @@ class PreviewUI {
...
+    /**
+     * Replaces the currently active header with a specified header
+     *
+     * @public
+     * @param {string} replacementHeader - Class name of new header
+     * @return {void}
+     */
+    replaceHeader(replacementHeader) {
+        // First hide all possible headers
+        this.container.querySelectorAll(`.${CLASS_HIDDEN}`).forEach((element) => {
```
JustinHoldstock: Do you mean to add `CLASS_HIDDEN` to all elements that have `CLASS_HIDDEN`?
Shouldn't this just be selecting the header element and hiding it, if it exists?

src/lib/annotations/Annotator.scss
```diff
@@ -15,6 +15,72 @@ $avatar-color-7: #159f45;
...
+        background-color: $white;
+    }
+}
+
+.bp-annotate-draw-post-cancel-container {
+    position: absolute;
+    right: 10px;
+}
+
+.bp-btn-annotate-draw-undo,
```
JustinHoldstock: What happens if you don't have :hover, :active, and :focus?
I'm just curious as to why those are included

src/lib/annotations/annotatorUtil.js
```diff
@@ -282,10 +309,7 @@ export function getAvatarHtml(avatarUrl, userId, userName) {
     let initials = '';
     if (userId !== '0') {
         // http://stackoverflow.com/questions/8133630/spliting-the-first-character-of-the-words
-        initials = userName
-            .replace(/\W*(\w)\w*/g, '$1')
-            .toUpperCase()
-            .substring(0, 3);
+        initials = userName.replace(/\W*(\w)\w*/g, '$1').toUpperCase().substring(0, 3);
```
JustinHoldstock: Just for record, we're planning on moving away from initials, and moving towards colours and symbols

src/lib/annotations/Annotator.scss
jeremypress: We need to override the normal button styles since they assume things like background/border for all of these states. The undo redo buttons are an svg in a clear button. I considered not using `'bp-btn' but we gain convenience with the disabled state, cursor, and using the correct hit box.

src/lib/annotations/annotatorUtil.js
jeremypress: That was prettier, I didn't touch this line.

src/lib/annotations/Annotator.scss
jeremypress: @tonyjin have you dealt with anything like this before?

tonyjin: Have you tried using 'bp-btn-plain`?

I'd do something like `class="bp-btn bp-btn-plain bp-btn-annotate-draw-undo"` - not sure if you need both `bp-btn` and `bp-btn-plain`

PR #405, opened 2017-09-21
Update: Support additional query params in requests

tonyjin:
- Replace manual URL manipulation with jsuri library
- Add support for new `queryParams` Preview option. When used, each key/value pair will be appended to all API requests as query params

### DIFF
--- a/src/lib/util.js
+++ b/src/lib/util.js
@@ -1,9 +1,10 @@
+import Uri from 'jsuri';
 import 'whatwg-fetch';
 
 const HEADER_CLIENT_NAME = 'X-Box-Client-Name';
 const HEADER_CLIENT_VERSION = 'X-Box-Client-Version';
-const PARAM_CLIENT_NAME = 'box_client_name';
-const PARAM_CLIENT_VERSION = 'box_client_version';
+const CLIENT_NAME_KEY = 'box_client_name';
+const CLIENT_VERSION_KEY = 'box_client_version';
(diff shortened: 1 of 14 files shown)


src/lib/util.js
```diff
@@ -393,11 +401,6 @@ export function appendAuthParams(url, token = '', sharedLink = '', password = ''
  * @return {string} Content url
  */
 export function createContentUrl(template, asset) {
-    // @NOTE(tjin): Remove the next 3 lines after reps API is stabilized after 4/6/17
```
jeremypress: 👍

src/lib/util.js
```diff
@@ -351,37 +375,21 @@ export function appendAuthParams(url, token = '', sharedLink = '', password = ''
...
-    if (CLIENT_NAME) {
-        params = `${params}&${PARAM_CLIENT_NAME}=${encodeURI(CLIENT_NAME)}`;
-    }
-
-    if (CLIENT_VERSION) {
-        params = `${params}&${PARAM_CLIENT_VERSION}=${encodeURI(CLIENT_VERSION)}`;
-    }
-
-    return `${url}${delim}${params}`;
+    return appendQueryParams(uri.toString(), ANALYTICS_QUERY_PARAMS);
```
jeremypress: A little confusing that addQueryParam takes a URI while appendQueryParam takes a string. Not much we could do since add is coming from the library. Maybe create a small wrapper in util to make things consistent?

PR #409, opened 2017-09-26
Fix: Moved guards so drawing works when highlights disabled

JustinHoldstock:


JustinHoldstock:
@jeremypress  Done! Had one heck of a time trying to figure out how to chain mock calls!

PR #414, opened 2017-09-29
Fix: Ensure we are displaying a valid error string

jeremypress:


src/lib/viewers/error/PreviewErrorViewer.js
```diff
@@ -77,7 +77,7 @@ class PreviewErrorViewer extends BaseViewer {
         }
 
         /* eslint-disable no-param-reassign */
-        err = err instanceof Error ? err : new Error(__('error_default'));
+        err = err instanceof Error && typeof err.displayMessage === 'string' ? err : new Error(__('error_default'));
```
pramodsum: Would it make sense to do some sort of check a little below w/ this line instead? 
`const displayMessage = err.displayMessage || err.message;`

PR #416, opened 2017-09-29
Chore: Temporarily allow draw annotations to be enabled via config

jeremypress:


src/lib/annotations/BoxAnnotations.js
```diff
@@ -111,6 +112,12 @@ class BoxAnnotations {
             });
         }
 
+        // Remove draw annotations if they aren't explicitly enabled.
+        const drawIdx = modifiedAnnotator.TYPE.indexOf(TYPES.draw);
+        if (!viewerConfig.drawEnabled && drawIdx > -1) {
+            delete modifiedAnnotator.TYPE[drawIdx];
```
pramodsum: Can do this in one line w/ this 
```
if (!viewerConfig.drawEnabled) {
  modifiedAnnotator.TYPE = modifiedAnnotator.TYPE.filter((type) => type !== TYPES.draw);
}
```

PR #418, opened 2017-10-02
Fix: Prevent canvas flickering on tap for mobile safari

jeremypress:


src/lib/viewers/doc/_docBase.scss
```diff
@@ -9,6 +9,7 @@
     overflow: hidden;
     position: absolute;
     right: 0;
+    -webkit-tap-highlight-color: rgba(0, 0, 0, 0);
```
jeremypress: Changing this to only impact canvases that are children of bp-doc

PR #427, opened 2017-10-06
Disable right click when user does not have download permissions

jeremypress:
In Box3d, we already disable the context menu event so that custom functionality can occur. 

@pramodsum, will this block any annotations functionality? I see contextmenu is bound to the highlightmousedown handler


pramodsum:
Should be fine. If the user doesn't have download permissions, they can't create highlights anyways so this shouldn't affect anything

PR #433, opened 2017-10-12
New: add autoplay to media viewers

jeremypress:


src/lib/viewers/media/MediaBaseViewer.js
```diff
@@ -212,6 +208,40 @@ class MediaBaseViewer extends BaseViewer {
...
     /**
+     * Handler for autoplay
+     *
+     * @private
+     * @emits autoplay
+     * @return {void}
+     */
+    handleAutoplay() {
+        const isAutoplayEnabled =
+            this.cache.has(MEDIA_AUTOPLAY_CACHE_KEY) && this.cache.get(MEDIA_AUTOPLAY_CACHE_KEY) === 'Enabled';
```
tonyjin: I don't think you have to do the first check. `get()` in Cache.js already does a check

src/lib/viewers/media/MediaBaseViewer.js
```diff
@@ -212,6 +208,40 @@ class MediaBaseViewer extends BaseViewer {
...
+    /**
+     * Determines if media should autoplay based on cached settings value.
+     *
+     * @private
+     * @emits volume
+     * @return {void}
+     */
+    checkAutoplay() {
+        if (
+            (!this.cache.has(MEDIA_AUTOPLAY_CACHE_KEY) || this.cache.get(MEDIA_AUTOPLAY_CACHE_KEY) === 'Disabled') &&
```
tonyjin: This code implies if nothing is set in the cache, then don't autoplay. Is that what you want? (I think so?)

src/lib/viewers/media/MediaBaseViewer.js
```diff
@@ -212,6 +208,40 @@ class MediaBaseViewer extends BaseViewer {
...
+            !Browser.isIOS()
+        ) {
+            return;
+        }
+
+        // iOS doesn't fire loadeddata event until some data loads
+        // Adding autoplay prevents that, but iOS Safari won't autoplay media
+        // https://webkit.org/blog/6784/new-video-policies-for-ios/
```
tonyjin: I'd move this comment above the return check

src/lib/viewers/media/MediaBaseViewer.js
```diff
@@ -212,6 +208,40 @@ class MediaBaseViewer extends BaseViewer {
...
+        this.mediaEl.autoplay = true;
```
tonyjin: Could this be triggered when the media element isn't loaded yet?

src/lib/viewers/media/MediaBaseViewer.js
jeremypress: Yep, I'll add a clarifying comment.

src/lib/viewers/media/MediaBaseViewer.js
jeremypress: It should be set before the media element is loaded. From MDN: `A Boolean attribute; if specified, the video automatically begins to play back as soon as it can do so without stopping to finish loading the data.`

src/lib/viewers/media/MediaBaseViewer.js
```diff
@@ -321,6 +368,7 @@ class MediaBaseViewer extends BaseViewer {
         this.hideLoadingIcon();
         this.handleRate();
         this.handleVolume();
+        this.emit('play');
```
jeremypress: Moved this here so that all autoplay related plays also emit.

src/lib/viewers/media/MediaBaseViewer.js
```diff
@@ -107,6 +108,7 @@ class MediaBaseViewer extends BaseViewer {
             .getPromise()
             .then(() => {
                 this.mediaEl.src = this.mediaUrl;
+                this.checkAutoplay();
```
jeremypress: Added this after the src is set because doing so can disrupt play. Noted here: https://webkit.org/blog/7734/auto-play-policy-changes-for-macos/

src/lib/viewers/media/Settings.js
```diff
@@ -160,6 +182,11 @@ class Settings extends EventEmitter {
         addActivationListener(this.settingsEl, this.menuEventHandler);
         this.containerEl.classList.add(CLASS_SETTINGS_SUBTITLES_UNAVAILABLE);
         this.containerEl.classList.add(CLASS_SETTINGS_AUDIOTRACKS_UNAVAILABLE);
+
+        if (Browser.isIOS()) {
```
jeremypress: Autoplay is not supported on iOS.

Issue #431, opened 2017-10-11
Make FindBar Optional


georgediaz88:
Hello,
	
First off, thanks for your hard work on this JavaScript library. I wanted to know if making the find bar optional is something you'd consider? The reason for such a change is we're using this library to display a PDF that spans the height of the page rather than in a container element. In trying to achieve this, we removed the preview.css stylesheet from our code and overwrote a few styles. The last change involves disabling the "findBar" functionality since we need the scrollbar to happen along the page window and not the container element. I also noticed this library disables the browers native "search bar" which we'd be fine using if we could turn off the "findBar" functionality which disables it. So to do this, I've forked your library and made turning off the "findBar" optional -- default true. It would be used like:

`preview.show(..., { useFindBar: false })`

Here's a [link](https://github.com/georgediaz88/box-content-preview/commit/534fbdddb3f18c1a9c58a8f96b32f67617386f4f) to those code changes which I can turn into a pull request if this seems like a change you'd accept. Otherwise, let me know if there's an alternate route I can take to achieve this.

Thanks!
George

jeremypress:
Hey @georgediaz88,
Thanks for the suggestion, we would definitely appreciate a PR for this! I think useFindBar would be better suited as a Document/Presentation viewer option, as it doesn't apply to other viewers. The config would look like this:

`preview.show({
    viewers: {
        Document: { // Or presentation
            useFindBar: false
        }
    }
});`

georgediaz88:
@jeremypress, okay here's the pull request with the suggested changes: https://github.com/box/box-content-preview/pull/438.

Let me know if there's anything else I need to do on my end. Thanks again!

georgediaz88:
@jeremypress, just leaving an update here that I updated the [PR](https://github.com/box/box-content-preview/pull/438) with your suggested changes.

pramodsum:
Hey @georgediaz88, we approved your PR #438! Feel free to merge it in when you're ready 👍

pramodsum:
@georgediaz88 merged your PR in! Thanks! It will be included in our next release v1.17.0

georgediaz88:
Fantastic. Thanks!

PR #438, opened 2017-10-17
New: Make findBar optional

georgediaz88:
For Issue: https://github.com/box/box-content-preview/issues/431

New option `disableFindBar: true` replaces the viewer's findBar with the browser's native findBar (currently disabled). Very useful for when you are showing the PDF on the entire page vs a container.

Example:
`preview.show(..., { viewers: { Document: { disableFindBar: true } } })`

Here's a screenshot showing the _browser's_ findBar at work:
screenshot 2017-10-16 21 31 37

Minor: I did notice a small text shadow around the highlighted words when searching which might come from the pdf_viewer.css? I'm not entirely sure. Let me know if you know offhand how I can disable this.

Again, thanks for being open to this PR and your great work on this library.


README.md
```diff
@@ -122,6 +122,17 @@ preview.show(fileId, accessToken, {
...
+Viewer specific options:
+```javascript
+preview.show(fileId, accessToken, {
+    ...,
+    viewers: { Document: { useFindBar: false } }
+});
+```
+| Option | Default | Description |
+| --- | --- | --- |
+| useFindBar | true | Setting to false will enable the browser's findBar in place of the viewers. Applies to `Document`, `Presentation` viewers. |
```
jeremypress: Please move this to the document specific readme https://github.com/box/box-content-preview/blob/master/src/lib/viewers/doc/README.md.

src/lib/viewers/doc/DocBaseViewer.js
```diff
@@ -271,7 +271,10 @@ class DocBaseViewer extends BaseViewer {
         // the file. Users without download permissions shouldn't be able to
         // interact with the text layer
         const canDownload = checkPermission(this.options.file, PERMISSION_DOWNLOAD);
-        this.findBar = new DocFindBar(this.findBarEl, this.findController, canDownload);
+
+        if (this.getViewerOption('useFindBar') !== false) {
```
jeremypress: Could be simplified to `if (!this.getViewerOption('useFindBar') {`

jeremypress: I think this might be more clear as `disableFindBar`, since the default behavior is to use the find bar. Something like `if (!this.getViewerOption('disableFindBar') {` cc @tonyjin

georgediaz88:
@jeremypress, okay made the suggested changes. Let me know if there's anything else needed. Thanks!

src/lib/viewers/doc/DocBaseViewer.js
```diff
@@ -271,7 +271,9 @@ class DocBaseViewer extends BaseViewer {
         // the file. Users without download permissions shouldn't be able to
         // interact with the text layer
         const canDownload = checkPermission(this.options.file, PERMISSION_DOWNLOAD);
-        this.findBar = new DocFindBar(this.findBarEl, this.findController, canDownload);
+        if (!this.getViewerOption('disableFindBar')) {
```
pramodsum: I think it make sense to structure this like so we don't have a double negative
```
if (this.getViewerOption('disableFindBar') {
    return;
}
this.findBar = new DocFindBar(this.findBarEl, this.findController, canDownload);
```

georgediaz88: @pramodsum, sure I'll change it. It reads better too.

PR #439, opened 2017-10-17
New: Added enabledTypes property. Backwards compat for disabledTypes

JustinHoldstock:
Adding `enabledTypes` to the annotation configuration, as well as a default enabled annotation types(`DEFAULT_ENABLED`), if none provided, to `ANNOTATORS`. This allows us to specify all of the *allowed* kinds of annotators, and the ones that are enabled by default (if no config provided).

IE)
```
// for document annotator, results in only 'point' annotations, as timestamp are not allowed
enabledTypes: ['point', 'timestamp']
```

### DIFF
--- a/src/lib/annotations/BoxAnnotations.js
+++ b/src/lib/annotations/BoxAnnotations.js
@@ -9,7 +9,7 @@ import { canLoadAnnotations } from './annotatorUtil';
  * CONSTRUCTOR: Constructor for the annotator.
  * VIEWER: The kinds of viewers that can be annotated by this.
  * TYPE: The types of annotations that can be used by this annotator.
- * DEFAULT_ENABLED: The default annotation types enabled if none provided.
+ * DEFAULT_TYPES: The default annotation types enabled if none provided.
  */
 const ANNOTATORS = [
     {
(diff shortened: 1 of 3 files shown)


src/lib/annotations/BoxAnnotations.js
```diff
@@ -4,18 +4,27 @@ import DrawingModeController from './drawing/DrawingModeController';
...
+        TYPE: [TYPES.point, TYPES.highlight, TYPES.highlight_comment, TYPES.draw],
+        DEFAULT_ENABLED: [TYPES.point, TYPES.highlight, TYPES.highlight_comment]
     },
     {
         NAME: 'Image',
         CONSTRUCTOR: ImageAnnotator,
         VIEWER: ['Image', 'MultiImage'],
-        TYPE: [TYPES.point]
+        TYPE: [TYPES.point],
+        DEFAULT_ENABLED: [TYPES.point]
```
jeremypress: I agree with using `DEFAULT_TYPES`

src/lib/annotations/BoxAnnotations.js
```diff
@@ -104,12 +113,19 @@ class BoxAnnotations {
 
         modifiedAnnotator = Object.assign({}, annotator);
 
-        // Filter out disabled annotation types
-        if (Array.isArray(viewerConfig.disabledTypes)) {
-            modifiedAnnotator.TYPE = modifiedAnnotator.TYPE.filter((type) => {
-                return !viewerConfig.disabledTypes.includes(type);
-            });
-        }
+        const enabledTypes = viewerConfig.enabledTypes || [...modifiedAnnotator.DEFAULT_ENABLED];
```
jeremypress: Why do we need the spread operator and then an array? Isn't `modifiedAnnotator.DEFAULT_ENABLED` already an array?

JustinHoldstock: I want to make a copy of the array, not modify the original

src/lib/annotations/BoxAnnotations.js
JustinHoldstock: done

src/lib/annotations/README.md
```diff
@@ -139,15 +139,16 @@ preview.show(..., {
         VIEWER_NAME: {
             annotations: {
                 enabled: Boolean, // Enables/disables if set. Respects "showAnnotations" if empty
-                disabledTypes: String[] // List of annotation types to disable
+                enabledTypes: String[] | null // List of annotation types to enable for this viewer. If empty, will respect default types for that annotator.
+                disabledTypes: String[] | null // DEPRECATED. List of annotation types to disable.
```
pramodsum: let's not add this line so people won't use disabledTypes from here on out

PR #440, opened 2017-10-17
Fix: Fix draw selection and related bugs

jeremypress:
* Fixes selection on Firefox/Safari
* Fixes interaction between text selection and drawings
* Replaces text cursor with crosshair

### DIFF
--- a/src/lib/annotations/Annotator.js
+++ b/src/lib/annotations/Annotator.js
@@ -266,6 +266,7 @@ class Annotator extends EventEmitter {
         this.destroyPendingThreads();
 
         if (this.createHighlightDialog.isVisible) {
+            document.getSelection().removeAllRanges();
             this.createHighlightDialog.hide();
         }
 
--- a/src/lib/annotations/Annotator.js
+++ b/src/lib/annotations/Annotator.js
@@ -265,6 +265,10 @@ class Annotator extends EventEmitter {
 
         this.destroyPendingThreads();
 
+        if (this.createHighlightDialog.isVisible) {
+            this.createHighlightDialog.hide();
+        }
+
         // No specific mode available for annotation type
         if (!(mode in this.modeButtons)) {
             return;
--- a/src/lib/annotations/Annotator.scss
+++ b/src/lib/annotations/Annotator.scss
@@ -407,8 +407,6 @@ $tablet: 'max-width: 768px';
 //------------------------------------------------------------------------------
 // CSS for highlights
 //------------------------------------------------------------------------------
-.bp-annotation-layer-draw,
-.bp-annotation-layer-draw-in-progress,
 .bp-annotation-layer-highlight {
     cursor: text;
     left: 0;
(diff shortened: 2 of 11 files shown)


src/lib/annotations/doc/CreateHighlightDialog.js
```diff
@@ -173,6 +173,7 @@ class CreateHighlightDialog extends EventEmitter {
         }
 
         hideElement(this.containerEl);
+        document.getSelection().removeAllRanges();
```
JustinHoldstock: Can you please move this into the highlight annotator? I want to keep this a dumb piece of UI, nothing more than that.

jeremypress: Moved, had to move it to two places :(. One where we enter another mode, and one for the mouseupHandler for clicking the document.

src/lib/annotations/Annotator.scss
```diff
@@ -550,6 +548,22 @@ $tablet: 'max-width: 768px';
...
 
+.bp-annotation-draw .textLayer {
+    pointer-events: none;
+
+    > div {
+        pointer-events: auto;
+    }
+}
+
+.bp-annotation-layer-draw,
```
jeremypress: @pramodsum @JustinHoldstock There's a little duplication with some highlight CSS, but it makes more sense to me since there are separate sections for annotation types and we didn't want the cursor to be text while drawing.

src/lib/annotations/drawing/DrawingModeController.js
```diff
@@ -53,6 +54,8 @@ class DrawingModeController extends AnnotationModeController {
         this.postButtonEl = annotator.getAnnotateButton(SELECTOR_ANNOTATION_BUTTON_DRAW_POST);
         this.undoButtonEl = annotator.getAnnotateButton(SELECTOR_ANNOTATION_BUTTON_DRAW_UNDO);
         this.redoButtonEl = annotator.getAnnotateButton(SELECTOR_ANNOTATION_BUTTON_DRAW_REDO);
+
+        this.annotator.annotatedElement.classList.add(CLASS_ANNOTATION_DRAW);
```
jeremypress: I'm adding this class so that we only affect CSS when we have drawing enabled

src/lib/annotations/Annotator.scss
```diff
@@ -550,6 +548,22 @@ $tablet: 'max-width: 768px';
     }
 }
 
+.bp-annotation-draw .textLayer {
+    pointer-events: none;
+
+    > div {
+        pointer-events: auto;
```
jeremypress: @tonyjin The trick here is to remove pointer events on the overall blocking div but re-enable them on the text layer divs so that both behaviors an coexist.

src/lib/annotations/Annotator.js
```diff
@@ -263,6 +263,11 @@ class Annotator extends EventEmitter {
 
         this.destroyPendingThreads();
 
+        if (this.createHighlightDialog.isVisible) {
```
pramodsum: there's actually a few places where we just check `if(this. createHighlightDialog) {}` would you be able to swap those out for a check for whether or not the dialog is visible?

pramodsum: Just noticed that you caught one of them below! So perfect :)

PR #441, opened 2017-10-18
Chore: Passing localized strings into annotations

pramodsum:
Since BoxAnnotations will be included as an npm package, it will not be able to use the mojito generated versions of the code based on the locale. Until BoxAnnotations can be written in React with it's own localization strategy, a temporary solution will be to pass in the localized strings as an option from Preview into BoxAnnotations on initialization. 

- Tests + documentation coming soon


pramodsum:
I think the end goal would be to have something inside the BoxAnnotations repo so this is more a temporary workaround since we can't use Mojito like how we do in Preview

PR #461, opened 2017-10-30
Fix: update autoplay behavior

jeremypress:


src/lib/viewers/media/Settings.js
```diff
@@ -183,7 +183,7 @@ class Settings extends EventEmitter {
         this.containerEl.classList.add(CLASS_SETTINGS_SUBTITLES_UNAVAILABLE);
         this.containerEl.classList.add(CLASS_SETTINGS_AUDIOTRACKS_UNAVAILABLE);
 
-        if (Browser.isIOS()) {
+        if (Browser.isMobile()) {
```
jeremypress: Chrome doesn't allow play without a user action and setting the autoplay tag is not respected, so i've turned off autoplay for mobile in general. https://developers.google.com/web/updates/2016/07/autoplay

src/lib/Preview.js
```diff
@@ -905,6 +905,9 @@ class Preview extends EventEmitter {
             case 'notificationhide':
                 this.ui.hideNotification();
                 break;
+            case 'previewnextfile':
```
jeremypress: I'm thinking this should be `mediaend` related and not `previewnextfile` because in the future we may not want to preview the next file if it is not a media file.

tonyjin: Generally, the message you broadcast should be for what happened, not what you want to happen, and then the subscriber then determines what it wants to do.

I'm thinking something like... `mediaendautoplay`? `mediaend` to me could be fired as an event that has nothing to do with autoplay, so appending autoplay makes it a bit more specific

PR #465, opened 2017-11-01
Chore: Separating fetch/init of annotations

pramodsum:
Dependent on https://github.com/box/box-annotations/pull/16


pramodsum:
Need to verify that moving this request earlier won't block any preview requests

pramodsum:
Not needed anymore

PR #470, opened 2017-11-05
New: SSR Viewer Initial Implementation

jeremypress:
Missing a few things:
* Reps should swap out based on resizing
* Placeholder pages
* Tests


src/lib/viewers/doc/ImagePageLoader.js
```diff
@@ -0,0 +1,89 @@
...
+    }
+
+    /**
+     * Gets a representation based on preview size
+     *
+     * @param {number} previewHeight - Height of preview window
+     * @param {number} previewWidth - Width of preview window
+     * @return {Object} Instance of RepStatus
+     */
+    getRepresentation(previewHeight, previewWidth) {
```
jeremypress: Right now this only considers preview dimensions, but could eventually consider device type, network conditions, zoom level etc.

src/lib/viewers/doc/SSRRenderer.js
```diff
@@ -0,0 +1,169 @@
...
+        const image = page.firstChild;
+
+        const maxScale = MAX_AUTO_SCALE / (getDimensionsFromRep(this.currentRep) / MINIMUM_REP_DIMENSION);
+        const scaleValue = Math.min((this.docEl.clientWidth - ROW_HEIGHT_PADDING) / image.naturalWidth, maxScale);
+
+        image.width = image.naturalWidth * scaleValue;
+        image.height = image.naturalHeight * scaleValue;
+
+        this.cache.clear(pageNumber - 1);
+        measure();
```
jeremypress: This is a method provided by the CellMeasurer which allows the cell to re-check its size, which will then update the row width calculated by the List component.

src/lib/viewers/doc/SSRRenderer.js
```diff
@@ -0,0 +1,169 @@
...
+        this.list.recomputeRowHeights(pageNumber - 1);
+
+        image.classList.remove('bp-hidden');
+
+        if (pageNumber === 1) {
+            this.emit('load');
+            this.cache._defaultHeight = image.height + ROW_HEIGHT_PADDING;
```
jeremypress: This may not be respected, but I'm attempting to update the default cache height for each page size after we learn the first page height.

src/lib/viewers/doc/SSRRenderer.js
```diff
@@ -0,0 +1,169 @@
...
+     * @public
+     * @return {void}
+     */
+    resize() {
+        this.cache.clearAll();
+        const renderedPages = this.docEl.querySelectorAll('.bp-page');
+        renderedPages.forEach((page) => {
+            const image = page.firstChild;
+            const src = image.src;
+            image.src = null;
```
jeremypress: This hack allows the 'onload' event to fire each time a page is rendered, which will cause a re-measurement of the page.

src/lib/viewers/doc/ImagePageLoader.js
JustinHoldstock: To make this a little more future proof, could you have the parameters be a single object, and then extract `previewHeight` and `previewWidth` from it?

src/lib/viewers/doc/ImagePageLoader.js
```diff
@@ -0,0 +1,89 @@
...
+        let rep;
+        let size;
+        const maxDimension = Math.max(previewHeight, previewWidth);
+        Object.keys(this.availableReps).forEach((repType) => {
```
JustinHoldstock: Could pull out the enclosed logic into it's own function? `getLargestAllowedRep()` or something like that? Also, if you want the first one, and not to do this check through the whole array, you could use `Array.prototype.some()` instead of `forEach()`

PR #472, opened 2017-11-07
 Update: npm packages to latest versions

pramodsum:
Packages that still need to be updated

- sinon


src/lib/viewers/doc/DocBaseViewer.js
```diff
@@ -365,7 +365,7 @@ class DocBaseViewer extends BaseViewer {
      * @return {void}
      */
     setPage(pageNumber) {
-        if (pageNumber < 1 || pageNumber > this.pdfViewer.pagesCount) {
+        if (!pageNumber || pageNumber < 1 || pageNumber > this.pdfViewer.pagesCount) {
```
jeremypress: Good catch!

package.json
```diff
@@ -116,7 +111,7 @@
     "prettier": "prettier-eslint \"src/lib/**/*.js\" --print-width 120 --single-quote --tab-width 4 --write",
     "prod": "BABEL_ENV=production NODE_ENV=production node --max_old_space_size=4096 ./node_modules/webpack/bin/webpack.js --progress --colors --config build/webpack.config.js",
     "release": "yarn run clean && yarn run build-rb && yarn run lint && yarn run test && yarn run prod",
-    "test": "NODE_ENV=test ./node_modules/.bin/karma start build/karma.conf.js",
+    "test": "NODE_ENV=test node --max_old_space_size=4096 ./node_modules/.bin/karma start build/karma.conf.js",
```
jeremypress: Is there a way we can tell if we have a memory leak? Or is this just a reasonable amount of memory to run tests?

pramodsum: This is a temporary solution while we look into ways to optimize the unit tests. We shouldn't be hitting this limit (especially after annotations has been removed). I'm going to go through the tests and verify that they're all being cleaned up properly.

PR #474, opened 2017-11-07
Upgrade: pdf.js v2.0.104

tonyjin:


tonyjin:
This upgrade occasionally gives me a few bad offset errors in the console, but it doesn't look like they affect functionality.

tonyjin:
This upgrade enables streaming via fetch - if we determine streaming gives worse performance than what we have already, we can disable it via `PDFJS.disableStream`

PR #478, opened 2017-11-08
Fix: toggleannotationmode event to trigger the right method

pramodsum:


pramodsum:
Decided to fix this in webapp directly rather than adding backwards compatibility

PR #481, opened 2017-11-09
Update: babel-preset-es201* to babel-preset-env

pramodsum:
NOTE: This should be upgraded to use @babel/preset-env when it is no
longer a beta

https://github.com/babel/babel/tree/master/experimental/babel-preset-env


pramodsum:
Babel still seems to be experimenting a lot w/ these packages and how they're structuring them in their new monorepo. Will upgrade these later when `@babel/preset-env` is more stable and no longer on beta

PR #482, opened 2017-11-13
Fix: Correctly revoke objectURL for print blob in office viewer

jeremypress:
Original fix here: https://github.com/box/box-content-preview/pull/444
feeling the pain of repeated code


pramodsum:
Is there any way we can consolidate this revoke logic amongst viewers so we won't run into an issue like this again?

tonyjin:
@pramodsum: https://github.com/box/box-content-preview/pull/62

We tabled that due to complications, but may be worth revisiting.

jeremypress:
I don't think it would be worth it to just abstract the printURL without doing the rest, which ended up being pretty messy the first time I tried.

PR #487, opened 2017-11-15
Fix: Presentation preloading scales without max

tonyjin:
This now matches the behavior of the presentation viewer, which doesn't have a max zoom scale.

### DIFF
--- a/src/lib/viewers/doc/DocPreloader.js
+++ b/src/lib/viewers/doc/DocPreloader.js
@@ -7,17 +7,16 @@ import {
     CLASS_INVISIBLE,
     CLASS_IS_TRANSPARENT,
     CLASS_PREVIEW_LOADED,
-    CLASS_SPINNER
+    CLASS_SPINNER,
+    PDFJS_CSS_UNITS,
+    PDFJS_MAX_AUTO_SCALE,
+    PDFJS_WIDTH_PADDING_PX,
+    PDFJS_HEIGHT_PADDING_PX
 } from '../../constants';
 import { get, setDimensions } from '../../util';
 
 const EXIF_COMMENT_TAG_NAME = 'UserComment'; // Read EXIF data from 'UserComment' tag
 const EXIF_COMMENT_REGEX = /pdfWidth:([0-9.]+)pts,pdfHeight:([0-9.]+)pts,numPages:([0-9]+)/;
-
-const PDFJS_CSS_UNITS = 96.0 / 72.0; // Should match CSS_UNITS in pdf_viewer.js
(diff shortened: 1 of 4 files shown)


src/lib/viewers/doc/PresentationPreloader.js
```diff
@@ -24,6 +29,28 @@ class PresentationPreloader extends DocPreloader {
...
 
     /**
+     * Get dimensions for the preload image. For presentations, this is always scaled to fit inside the viewport since
+     * pdf.js auto-scales presentations to fit the window with no maximum zoom.
+     *
+     * @param {number} pdfWidth - Width of PDF in pixels
+     * @param {number} pdfHeight - Height of PDF in pixels
+     * @return {Object} Width and height in pixels
+     */
+    getScaledDimensions(pdfWidth, pdfHeight) {
```
JustinHoldstock: If this is identical to DocPreloader's, can you move it to `util.js`, with an additional parameter for `wrapperEl` OR `clientWidth` and `clientHeight`?

JustinHoldstock: I guess there's a slight difference. Is there anything you can abstract to remove duplicated code?

tonyjin: Hmm yeah, can probably make max scale an optional property and keep it in docpreloader. Good point, thanks.

src/lib/viewers/doc/DocPreloader.js
```diff
@@ -33,6 +32,9 @@ class DocPreloader extends EventEmitter {
     /** @property {HTMLElement} - Preload image element */
     imageEl;
 
+    /** @property {HTMLElement} - Maximum auto-zoom scale - set to 0 for no limit */
+    maxZoomScale;
```
JustinHoldstock: Instead of setting in the constructor, you could set it here!

src/lib/viewers/doc/PresentationPreloader.js
```diff
@@ -3,6 +3,9 @@ import { CLASS_INVISIBLE, CLASS_BOX_PREVIEW_PRELOAD_WRAPPER_PRESENTATION } from
 import { setDimensions } from '../../util';
 
 class PresentationPreloader extends DocPreloader {
+    /** @property {HTMLELement} - Maximum auto-zoom scale - set to 0 for no limit */
+    maxZoomScale;
```
JustinHoldstock: AFAIK setting to 0 here will override the default, if set in the parent class in the same manner. See above comment. This cleans up the constructor quite a bit.

PR #488, opened 2017-11-15
Fix: Reset retry count after successful preview

tonyjin:
Solves:

```
There is one issue in preview right now...if you call preview.show(...,...,options)...and then after call preview.show() again with different options, the 2nd time the options are not parsed and ignored
cuz of https://github.com/box/box-content-preview/blob/master/src/lib/Preview.js#L599
returns after loadFromServer()
probably the retryCount should be reset to 0 when stuff has loaded successfully
```

### DIFF
--- a/src/lib/Preview.js
+++ b/src/lib/Preview.js
@@ -860,6 +860,10 @@ class Preview extends EventEmitter {
         // Once the viewer instance has been created, emit it so that clients can attach their events.
         // Viewer object will still be sent along the load event also.
         this.emit('viewer', this.viewer);
+
+        // Reset retry count after successful load so we don't go into the retry short circuit when the same file
+        // previewed again
+        this.retryCount = 0;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


src/lib/Preview.js
```diff
@@ -860,6 +860,10 @@ class Preview extends EventEmitter {
         // Once the viewer instance has been created, emit it so that clients can attach their events.
         // Viewer object will still be sent along the load event also.
         this.emit('viewer', this.viewer);
+
+        // Reset retry count after successful load so we don't go into the retry short circuit when the same file
+        // previewed again
+        this.retryCount = 0;
```
jeremypress: Would this get reset for the error viewer as well? In that case I'm not sure if we want to reset the retry count.

tonyjin: The error viewer is a final state AFAIK - we no longer retry after we hit the error viewer.

jeremypress: Why is loading other viewers not a final state? because of prefetching next files?

tonyjin: They should all be a final state. This is what we're trying to solve with this PR:

```
There is one issue in preview right now...if you call preview.show(...,...,options)...and then after call preview.show() again with different options, the 2nd time the options are not parsed and ignored
cuz of https://github.com/box/box-content-preview/blob/master/src/lib/Preview.js#L599
returns after loadFromServer()
probably the retryCount should be reset to 0 when stuff has loaded successfully
```

PR #490, opened 2017-11-16
Chore: extract file loading icons to be used globally throughout preview

jeremypress:


src/lib/viewers/error/PreviewErrorViewer.js
```diff
@@ -59,17 +59,15 @@ class PreviewErrorViewer extends BaseViewer {
...
         // Generic errors will not have the file object
         if (file) {
             switch (file.extension) {
                 case 'zip':
                 case 'tgz':
-                    this.icon = ICON_FILE_ZIP;
-                    break;
                 case 'flv':
-                    this.icon = ICON_FILE_MEDIA;
+                    this.icon = getIconFromExtension(file.extension);
```
jeremypress: This is here to maintain current behavior. Eventually we can just rely on the setup on base viewer and then fall back if needed

jeremypress:
@JustinHoldstock I don't want to use extensions because that's used for determining which viewers we want to use. I'd like to keep it here for now and if our list expands greatly we can break it out. Thanks to @tonyjin on the map adding one liners

PR #494, opened 2017-11-18
Fix: Media player control menus show scrollbars during transition

tonyjin:
Add class to hide overflow during menu transitions and hide after transitions.

### DIFF
--- a/src/lib/viewers/media/Settings.js
+++ b/src/lib/viewers/media/Settings.js
@@ -183,6 +183,11 @@ class Settings extends EventEmitter {
         this.containerEl.classList.add(CLASS_SETTINGS_SUBTITLES_UNAVAILABLE);
         this.containerEl.classList.add(CLASS_SETTINGS_AUDIOTRACKS_UNAVAILABLE);
 
+        // Allow scrollbars after animations end
+        this.settingsEl.addEventListener('transitionend', () => {
+            this.settingsEl.classList.remove('bp-media-settings-in-transition');
+        });
+
         if (Browser.isMobile()) {
             this.containerEl.classList.add(CLASS_SETTINGS_AUTOPLAY_UNAVAILABLE);
         }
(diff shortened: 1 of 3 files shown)


src/lib/viewers/media/Settings.js
```diff
@@ -330,6 +337,9 @@ class Settings extends EventEmitter {
         const type = menuItem.getAttribute('data-type');
         const value = menuItem.getAttribute('data-value');
 
+        // Hide scrollbars while menu animation is happening - this is removed when transition ends
+        this.settingsEl.classList.add('bp-media-settings-in-transition');
```
tonyjin: I add this here instead of just for showing a sub menu since resetting a menu can theoretically trigger a vertical scrollbar (I'm thinking a submenu has two options + the top return to main menu option and goes back to the main menu with 4 options?)

Adding this class and then clicking on a menu option with a value is find since you animate back to the previous menu with that value set.

jeremypress: Could we still get a scrollbar on a reset? The flow I see with this change is:
1. click on an option and hide overflow
2. transition ends and overflow is permitted again

Is it possible that on a reset() we see the scroll bar since we are doing the same transition but in reverse (with a different code flow)? Could we add the class on transition start and remove it on transition end? Or will that not fire in time?

tonyjin: This is the entry point for any kind of menu action - entering in submenu, resetting, and clicking a menu option. Thus we are effectively adding on transition start and removing on transition end.

The actual DOM event for `transitionstart` is not widely supported yet: https://developer.mozilla.org/en-US/docs/Web/Events/transitionstart#Browser_compatibility

jeremypress: ah the file concatenation had me messed up, you're correct. Good to know about `transitionStart`

PR #495, opened 2017-11-20
Fix: Enable HD settings only if HD rep is available

jeremypress:


src/lib/viewers/media/Settings.js
```diff
@@ -445,11 +450,14 @@ class Settings extends EventEmitter {
...
      * @param {string} value - of menu option
+     * @param {boolean} setCache - should we set the cache with the given settings value
      * @return {void}
      */
-    chooseOption(type, value) {
-        // Save the value
-        this.cache.set(`media-${type}`, value, true);
+    chooseOption(type, value, setCache = true) {
+        // Save the value if specified
+        if (setCache) {
```
jeremypress: Added this so we don't set the cache if we are choosing sd because we have no other options. Now if switch to an HD video we can retain our previous quality setting value

src/lib/viewers/media/Settings.js
```diff
@@ -699,6 +707,19 @@ class Settings extends EventEmitter {
...
+     * Adds the quality options to the settings menu
+     *
+     * @return {void}
+     */
+    enableHD() {
+        this.containerEl.classList.remove(CLASS_SETTINGS_HD_UNAVAILABLE);
+
+        // Check our cached settings value now that we know the HD rep is available
+        const quality = this.cache.get('media-quality') || MEDIA_QUALITY_AUTO;
+        this.chooseOption(TYPE_QUALITY, quality);
```
bhh1988: I'm wondering if you'll need to call this.reset() here, and/or set the menu-dimensions right.

Do the quality-submenu's dimensions end up right in either case, whether HD is available or not?
Is the menu getting properly reset to the main menu after enableHD is getting called? I'm not entirely sure all the edge-cases this.reset() was trying to protect against, but maybe the edge case where the user is looking at the quality submenu and then enableHD() gets called, the menu should switch to the main menu...

jeremypress: 1. Dimensions end up correct in both cases. SetmenucontainerDimensions gets called when we show the sub menu, enableHD will have already been called at this point.
2. enableHD is only called once before any menu interaction has happened, the regular chooseOption flow will execute when clicking on the various quality options.
3. Same as 2, reset will still be called like when other entries are selected.

PR #509, opened 2017-11-22
Fix: BaseViewer debounced resize handler binding

tonyjin:
Context was bound to the fetch resize handler function instead of the actual handler


boxcla:

PR #513, opened 2017-11-22
Chore: Change webpack source maps to 'inline-cheap-source-map'

jeremypress:
Options can be found here: https://webpack.js.org/configuration/devtool/

### DIFF
--- a/build/webpack.config.js
+++ b/build/webpack.config.js
@@ -91,7 +91,7 @@ function updateConfig(conf, language, index) {
         }
 
         // Add inline source map
-        config.devtool = 'inline-source-map';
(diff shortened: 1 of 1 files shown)


boxcla:

jeremypress:
Doesn't affect our initial build time too much, but it reduces the rebuild time (when using watch) by almost 50%.

priyajeet:
Does it work? We used inline-source-map, because it was more accurate and less brittle.

jeremypress:
@priyajeet I'll do a full browser check

jeremypress:
Looks like IE and Edge need something inline, so switching to 'inline-cheap-source-map'

PR #514, opened 2017-11-22
Chore: only use source maps when running unit tests in debug mode

jeremypress:
Source maps don't work in phantom, so we should turn them off unless running our tests in debug mode in a browser. This should shave off between 30-45 seconds every place we run our tests; locally, jenkins, and travis.

### DIFF
--- a/build/webpack.karma.config.js
+++ b/build/webpack.karma.config.js
@@ -1,18 +1,27 @@
 require('babel-polyfill');
 
+const { argv } = process;
+const isDebug = argv.find((arg) => {
+    return arg === '--auto-watch' || arg === '--no-single-run';
+});
+
 const { IgnorePlugin } = require('webpack');
 const commonConfig = require('./webpack.common.config');
 
 const baseConfig = commonConfig('en-US');
 
 const config = Object.assign(baseConfig, {
-    devtool: 'inline-source-map',
     resolve: {
         alias: {
             sinon: 'sinon/pkg/sinon'
         }
     }
 });
+
+if (isDebug) {
+    config.devtool = 'inline-source-map';
+}
+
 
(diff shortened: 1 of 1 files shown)


boxcla:

build/webpack.karma.config.js
```diff
@@ -1,19 +1,28 @@
 require('babel-polyfill');
 
+const { argv } = process;
```
jeremypress: Looking for a better way to determine the original command to run  the tests

build/webpack.karma.config.js
```diff
@@ -1,19 +1,28 @@
...
-    devtool: 'inline-source-map',
     resolve: {
         alias: {
             sinon: 'sinon/pkg/sinon'
         }
     }
 });
 
+if (isDebug) {
+    config.devtool = 'inline-source-map';
```
jeremypress: Will change this to the other source map type once I check it on all of our browsers

build/webpack.karma.config.js
pramodsum: We do this in webpack.config.js, would it work here? 
```
const isRelease = process.env.NODE_ENV === 'production';
const isDev = process.env.NODE_ENV === 'dev';
```

jeremypress: That's a good catch to turn them off for prod, but we also want to turn them off if we're not running 'debug' in any environment. We never run debug during a production build so it should already be covered.

build/webpack.karma.config.js
jeremypress: changed

build/webpack.karma.config.js
pramodsum: ah good point.

PR #516, opened 2017-11-28
Chore: Lazy load annotations using new chunk

jeremypress:
To improve this we'll need to change box-annotations so that it exports BoxAnnotations globally.


boxcla:

jeremypress:
Wenbo and I tried using the `commonChunksPlugin` from webpack to do the import automatically but we ran into an issue. If we end up wanted to split more viewers out as separate chunks it would be worth attempting again.

src/lib/viewers/BaseViewer.js
```diff
@@ -370,10 +371,6 @@ class BaseViewer extends EventEmitter {
             if (event && event.scale) {
                 this.scale = event.scale;
             }
-
-            if (this.annotatorConf) {
```
pramodsum: I would keep the initialization of annotations separate because there's no guarantee that when the annotations assets are loaded that the viewer would have loaded the content as well.

src/lib/viewers/BaseViewer.js
```diff
@@ -673,13 +670,28 @@ class BaseViewer extends EventEmitter {
...
         }
 
-        try {
-            const boxAnnotations = new BoxAnnotations();
-            this.annotatorConf = boxAnnotations.determineAnnotator(this.options, this.viewerConfig);
-        } catch (err) {
-            /* eslint-disable no-console */
-            console.error('Annotation assets failed to load');
-            /* eslint-enable no-console */
+        this.loadAssets([BOX_ANNOTATIONS_JS])
```
pramodsum: `this.annotationsPromise = this.loadAssets([BOX_ANNOTATIONS_JS]);`

src/lib/viewers/BaseViewer.js
pramodsum: try something like this
```
if (!this.annotationPromise) {
  return;
}

this.annotationPromise
    .then(this.onAnnotationsLoaded)
    .catch(() => { ... });
```

jeremypress: done

build/webpack.config.js
```diff
@@ -58,6 +58,7 @@ const languages = isRelease
 function updateConfig(conf, language, index) {
     const config = Object.assign(conf, {
         entry: {
+            boxAnnotations: ['box-annotations/lib/BoxAnnotations'],
```
pramodsum: Thoughts on renaming this to annotations? So that the file loads as annotations.js

src/lib/viewers/BaseViewer.js
```diff
@@ -1,5 +1,4 @@
 import 'box-annotations/lib/Annotator.scss';
```
pramodsum: I think once we bundle annotations with webpack properly we should be able to get rid of this CSS import in BaseViewer.js

src/lib/viewers/BaseViewer.js
```diff
@@ -371,9 +372,11 @@ class BaseViewer extends EventEmitter {
                 this.scale = event.scale;
             }
 
-            if (this.annotatorConf) {
-                this.initAnnotations();
-            }
+            this.annotationsLoadPromise.then(this.onAnnotationsLoaded).catch(() => {
```
pramodsum: Make sure to add a check here if the annotationsLoadPromise doesn't exist!

build/webpack.config.js
jeremypress: Sure, we're going to have to test your change to see about annotations.js vs. annotations.min.js

src/lib/viewers/BaseViewer.js
jeremypress: Will change this once I can test with yours.

PR #520, opened 2017-11-29
Fix: Bind download function in Preview.js

jeremypress:
same requirement as print()


boxcla:

PR #522, opened 2017-11-29
Fix: Fix inheritance chains in DashViewer

jeremypress:
There's a problem with our babel config that causes inheritance to fail in the following situation.

Class A -> Class B -> Class C
1. Class A has method foo()
2. Class B does not override the method
3. Class C overrides foo and calls super()

This may be related to `modules: false` in our .babelrc for the es2015 preset


boxcla:

priyajeet:
You sure this is still an issue after removing autobind?

jeremypress:
Looks like I wasn't installing the latest babel packages correctly

PR #523, opened 2017-11-30
Fix: Fix inheritance and add bindings for controls options

jeremypress:


boxcla:

PR #525, opened 2017-12-01
Fix: Only remove the timeupdate listener if it has been bound before

jeremypress:


boxcla:

PR #527, opened 2017-12-01
Fix: Properly bind download button in PreviewErrorViewer

jeremypress:


boxcla:

PR #533, opened 2017-12-07
Upgrade: pdf.js v2.0.191

tonyjin:
Smaller file size from removed polyfills


boxcla:

PR #536, opened 2017-12-09
Update: Do not prefetch for mp4 and mp3 viewers

bhh1988:
Prefetching for these viewers can lead to more problems than it's worth,
because for a single view the same token is used later when you seek
with a range-request (making it more likely for token-expiration issues
to occur).


boxcla:

PR #537, opened 2017-12-11
Fix: Building preview before running tests

jeremypress:


boxcla:

PR #538, opened 2017-12-11
Fix: build-rb required before building ci

jeremypress:


boxcla:

PR #539, opened 2017-12-11
New: Allow BoxAnnotations to be passed in as a Preview option

pramodsum:
This change combined with https://github.com/box/box-annotations/pull/68 will allow a developer to instantiate annotations outside of Preview like this: 
```
const options = {
    Image: {
        enabledTypes: ['point']
    },

    // Document/Presentation Annotators are the same so the config is simplified here
    Document: {
        enabledTypes: ['highlight', 'draw']
    }
};


/* global BoxAnnotations */
const boxAnnotations = new BoxAnnotations(options);

preview.show(FILE_ID, ACCESS_TOKEN, {
    container: '.preview-container',
    showDownload: true,
    boxAnnotations
});
```

### DIFF
--- a/src/lib/viewers/BaseViewer.js
+++ b/src/lib/viewers/BaseViewer.js
@@ -695,9 +695,10 @@ class BaseViewer extends EventEmitter {
         }
 
         // Auto-resolves promise if BoxAnnotations is passed in as a Preview option
-        this.annotationsLoadPromise = this.options.boxAnnotations
-            ? Promise.resolve()
-            : this.loadAssets([ANNOTATIONS_JS], [ANNOTATIONS_CSS]);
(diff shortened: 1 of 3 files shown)


boxcla:

src/lib/viewers/BaseViewer.js
```diff
@@ -755,13 +764,21 @@ class BaseViewer extends EventEmitter {
      */
     areAnnotationsEnabled() {
         // Respect viewer-specific annotation option if it is set
-        // #TODO(@spramod|@jholdstock): remove this after we have annotation instancing
```
pramodsum: @JustinHoldstock we can finally get rid of this!

PR #542, opened 2017-12-11
Fix: Validate passed in boxAnnotations is instance of the proper type

pramodsum:


boxcla:

PR #546, opened 2017-12-12
Fix: Auto hide scrollbars for CSV on IE/Edge

jeremypress:
https://cloud.box.com/s/xv1zu89asqwnq9nm73zabi1l9djw64rv


boxcla:

PR #548, opened 2017-12-14
New: Enable download of watermarked representations

jeremypress:


boxcla:

src/lib/file.js
```diff
@@ -77,6 +77,18 @@ export function checkPermission(file, operation) {
...
 
+/**
+ * Checks permission
+ *
+ * @public
+ * @param {Object} file - Box file
+ * @param {string} operation - Action to check permission for
+ * @return {boolean} Whether or not action is permitted
+ */
+export function canDownloadWatermarkedRep(file) {
```
jeremypress: This is temporary, Albert will be provided a `can_download_watermarked` boolean to preview that accurately dictates if we can download the watermarked rep

src/lib/Preview.js
```diff
@@ -415,13 +428,25 @@ class Preview extends EventEmitter {
...
                 openUrlInsideIframe(data.download_url);
             });
+        } else if (canDownloadWatermarkedRep(this.file)) {
+            // If we can download the watermarked rep but not the original, we assume that the viewer is using the
+            // watermarked rep for preview, so we will download it.
+            // This allows the content URL to be downloaded by the browser
+            const downloadQueryParams = { response_content_disposition_type: 'attachment' };
+            const template = this.viewer.options.representation.content.url_template;
+            downloadUrl = appendQueryParams(
+                this.viewer.createContentUrlWithAuthParams(template, this.viewer.options.viewer.ASSET),
```
jeremypress: Not ideal to access the viewer from Preview like this, thinking about alternatives.

src/lib/file.js
tonyjin: 👍

src/lib/Preview.js
```diff
@@ -462,13 +467,25 @@ class Preview extends EventEmitter {
...
         if (checkPermission(this.file, PERMISSION_DOWNLOAD)) {
             // Append optional query params
-            const downloadUrl = appendQueryParams(getDownloadURL(this.file.id, apiHost), queryParams);
+            downloadUrl = appendQueryParams(getDownloadURL(this.file.id, apiHost), queryParams);
             get(downloadUrl, this.getRequestHeaders()).then((data) => {
                 openUrlInsideIframe(data.download_url);
             });
+        } else if (canDownloadWatermarkedRep(this.file)) {
+            // If we can download the watermarked rep but not the original, we assume that the viewer is using the
+            // watermarked rep for preview, so we will download it.
```
tonyjin: I like putting the comment for the block above so you can separate the comments for the block from the comment for the first line, like:

```
    get(downloadUrl, this.getRequestHeaders()).then((data) => {
        openUrlInsideIframe(data.download_url);
    });
    
    // If we can download the watermarked rep but not the original, we assume that the viewer is using the
    // watermarked rep for preview, so we will download it.
} else if (canDownloadWatermarkedRep(this.file)) {
    // This allows the content URL to be downloaded by the browser
    const downloadQueryParams = { response_content_disposition_type: 'attachment' };
```

src/lib/Preview.js
```diff
@@ -462,13 +467,25 @@ class Preview extends EventEmitter {
...
+            // This allows the content URL to be downloaded by the browser
+            const downloadQueryParams = { response_content_disposition_type: 'attachment' };
+            const template = this.viewer.options.representation.content.url_template;
```
tonyjin: Will `this.viewer.options.representation` ever be null? One potential example: clicking the download button for a ZIP file that's in a watermarked folder?

src/lib/Preview.js
tonyjin: I'm thinking we should instead expose something like `this.viewer.getRepUrl()` that abstracts away some of this complexity and adds some error/null checks.

Then you could have:
```
const watermarkedRepUrl = this.viewer.getRepUrl();
const downloadUrl = appendQueryParams(this.viewer.appendAuthParams(watermarkedRepUrl), {
    response_content_disposition_type: 'attachment' // Tells browser to download, not display
});
openUrlInsideIframe(downloadUrl);
```

A few stylistic things:
- No need to have downloadQueryParams on a separate line since there aren't that many query param pairs and you don't need to optionally add any based on if statements
- Put `const downloadUrl` inside each block to save the line `let downloadUrl` above since you're not using `downloadUrl` outside of each block

jeremypress:
This is blocked for the time being.

PR #549, opened 2017-12-15
Fix: Add bindings for more audio-track and subtitles controls

bhh1988:
Audio-track and subtitle selection is not working, since autobind
removal


boxcla:

PR #553, opened 2017-12-19

bhh1988:
This fixes a bug caused by a backwards incompatible change on the server
to look for "watermark_content" query param instead of watermark-cache.


boxcla:

PR #556, opened 2017-12-19

bhh1988:
This fixes a bug caused by a backwards incompatible change on
the server to look for "watermark_content" query param instead of
watermark-cache.


boxcla:

bhh1988:
@tonyjin no implications, other than that this commit fixes what's already broken, because the server's expecting watermark_content now. When they changed the "watermark-cache" query param into "watermark_content", they made sure already that they did it in a way that didn't break preview watermarking at all...but they forgot about video watermarking.

PR #557, opened 2017-12-19

jeremypress:
Committing for @bhh1988 since Travis is not running his jobs.


boxcla:

PR #561, opened 2017-12-22
Fix: Temporarily disable requireJS if specified to load 3rd party deps

### LINKED ISSUE #528: Preview fails to load when used with RequireJS
### Environment:
- Preview version: 1.20.2, 1.20.1 and 1.19.1
- Browser name and version: Chrome V62.0.3202.94 
- Operating system and version: macOS Seirra 10.12.6

### Steps to reproduce the problem:
Code used

```javascript 
preview.show('<file-id>', '<acess_token>', {
   container: '.preview-container',
   showDownload: true
});
```

Files were imported using cdn `https://cdn01.boxcdn.net/platform/preview/1.20.1/en-US/preview.js` and css also.

### What is the expected behavior? (Screenshots can be helpful here)
Preview for PDF file

### What went wrong? (Screenshots can be helpful here)
An icon appears with text message `PDFJS is not defined`. (Please note same code is working for an image file)

jeremypress:
Resolves #528

### DIFF
--- a/src/lib/Preview.js
+++ b/src/lib/Preview.js
@@ -757,6 +757,10 @@ class Preview extends EventEmitter {
         // Optional additional query params to append to requests
         this.options.queryParams = options.queryParams || {};
 
+        // Option to pause requireJS while Preview loads third party dependencies
+        // RequireJS will be re-enabled on the 'assetsloaded' event fired by Preview
+        this.options.pauseRequireJS = !!options.pauseRequireJS;
+
         // Prefix any user created loaders before our default ones
         this.loaders = (options.loaders || []).concat(loaderList);
 
--- a/src/lib/util.js
+++ b/src/lib/util.js
@@ -467,11 +467,19 @@ export function loadStylesheets(urls) {
  *
  * @public
  * @param {Array} urls - Asset urls
+ * @param {string} [disableRequireJS] - Should requireJS be temporarily disabled
  * @return {Promise} Promise to load scripts
  */
-export function loadScripts(urls) {
+export function loadScripts(urls, disableRequireJS = false) {
     const { head } = document;
     const promises = [];
+    const { define, require, requirejs } = window;
+
+    if (disableRequireJS) {
+        window.define = undefined;
+        window.require = undefined;
+        window.requirejs = undefined;
+    }
 
     urls.forEach((url) => {
         if (!head.querySelector(`script[src="${url}"]`)) {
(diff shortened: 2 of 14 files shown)


boxcla:

src/lib/Preview.js
```diff
@@ -702,6 +702,10 @@ class Preview extends EventEmitter {
         // Optional additional query params to append to requests
         this.options.queryParams = options.queryParams || {};
 
+        // Option to pause requireJS while Preview loads third party dependencies
+        // RequireJS will be re-enabled on the 'assetsloaded' event fired by Preview
+        this.options.pauseRequireJS = !!options.pauseRequireJS;
```
jeremypress: still not 100% on this name.

src/lib/util.js
```diff
@@ -486,7 +494,21 @@ export function loadScripts(urls) {
...
-    return Promise.all(promises);
+    return Promise.all(promises)
+        .then(() => {
+            if (disableRequireJS) {
+                window.define = define;
+                window.require = require;
+                window.requirejs = requirejs;
+            }
+        })
+        .catch(() => {
```
jeremypress: Added a catch in case we fail to load the script, we should still relinquish control over require.

src/lib/viewers/BaseViewer.js
```diff
@@ -588,9 +588,11 @@ class BaseViewer extends EventEmitter {
      * @protected
      * @param {Array} [js] - js assets
      * @param {Array} [css] - css assets
+     * @param {boolean} [isThirdParty] is the asset to load third party
      * @return {Promise} Promise to load scripts
      */
-    loadAssets(js, css) {
+    loadAssets(js, css, isThirdParty = true) {
+        const disableRequireJS = isThirdParty && !!this.options.pauseRequireJS;
```
jeremypress: I added the third party boolean because we don't want to emit our message when loading annotations or any other potential code split points. These do not rely on require and should not be considered in this process.

src/lib/Preview.js
JustinHoldstock: `disableRequireJS`, since you've already mentioned in the comment that it `will be re-enabled on the 'assetloaded' event`?

JustinHoldstock: Another argument for `disableRequireJS`, you use that term later on as parameter names

jeremypress: I feel like disable means that it won't come back

tonyjin: disableRequireJSWhileLoadingAssets :)

src/lib/viewers/BaseViewer.js
tonyjin: `isViewerAsset`? I'm thinking of a future where we have 3rd party assets for annotations or other splits?

src/lib/util.js
tonyjin: Too bad we can't use `finally` yet... :(

Chrome 63/Firefox 58 is probably too new: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/finally

src/lib/viewers/BaseViewer.js
jeremypress: true but there are a ton of places where we are loading viewer assets, so those would all have to set a boolean

jeremypress: and it would also created a flipped negative `!isViewerAsset`

PR #562, opened 2017-12-22
Fix: Never cache watermarked files, add reload() method

tonyjin:
- Never cache watermarked files
- Separate loading and re-loading (when cache is stale or file is newly watermarked) when loading from server
- Remove superfluous caching before loading from cache
- Add new public method to reload current preview and replace legacy methods of reloading preview with new method

### DIFF
--- a/src/lib/Preview.js
+++ b/src/lib/Preview.js
@@ -20,7 +20,16 @@ import {
     findScriptLocation,
     appendQueryParams
 } from './util';
-import { getURL, getDownloadURL, checkPermission, checkFeature, checkFileValid, cacheFile, uncacheFile } from './file';
+import {
+    getURL,
+    getDownloadURL,
+    checkPermission,
+    checkFeature,
+    checkFileValid,
+    cacheFile,
+    uncacheFile,
+    isWatermarked
+} from './file';
 import {
     API_HOST,
     APP_HOST,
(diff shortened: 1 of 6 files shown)


boxcla:

tonyjin:
Looking for reviews before I add tests. When I dug into this issue, two things came up:
- We cache and then immediately uncache and reload watermarked files (this was causing the bug with double images when you open a preview of a file before a watermarked image and then navigate to it and rotate)
- We don't distinguish between when we need to load a viewer for the first time (file isn't cached) and when we need to reload (cache is stale or file is newly watermarked)

This patch aims to fix both of those issues and also exposes a new public `reload()` method that can be used to reload the current preview, optionally refreshing file info from the server. Because I'd like to keep some compatibility with our existing `skipServerUpdate` option, `reload()` with no params respects `skipServerUpdate`.

src/lib/viewers/BaseViewer.js
```diff
@@ -175,7 +175,9 @@ class BaseViewer extends EventEmitter {
         }
 
         const iconWrapperEl = container.querySelector(SELECTOR_BOX_PREVIEW_ICON);
-        iconWrapperEl.innerHTML = this.fileLoadingIcon || getIconFromName('FILE_DEFAULT');
+        if (iconWrapperEl) {
+            iconWrapperEl.innerHTML = this.fileLoadingIcon || getIconFromName('FILE_DEFAULT');
```
JustinHoldstock: Do we have a constant anywhere for `'FILE_DEFAULT'`?

src/lib/Preview.js
```diff
@@ -779,35 +834,33 @@ class Preview extends EventEmitter {
...
             this.logger.setFile(file);
 
             // Keep reference to previously cached file version
             const cachedFile = this.cache.get(file.id);
 
             // Explicitly uncache watermarked files, otherwise update cache
-            const isWatermarked = file.watermark_info && file.watermark_info.is_watermarked;
-            if (isWatermarked) {
+            const isFileWatermarked = isWatermarked(file);
+            if (isFileWatermarked) {
```
JustinHoldstock: Unless you're using `isFileWatermarked` elsewhere, could you just put the `isWatermarked()` call into the if?

src/lib/Preview.js
```diff
@@ -213,6 +222,44 @@ class Preview extends EventEmitter {
...
+     *   - skipServerUpdate is true (either passed in or defined in preview options) AND cached file is not valid
+     *   - skipServerUpdate is false AND there is no cached file ID
+     *
+     * @public
+     * @param {boolean} skipServerUpdate - Whether or not to update file info from server
+     * @return {void}
+     */
+    reload(skipServerUpdate) {
+        // If not passed in, default to Preview option for skipping server update
+        if (typeof skipServerUpdate === 'undefined') {
```
JustinHoldstock: If someone passes in `null`, `this.options.skipServerUpdate` won't be respected

jeremypress: We could check if it's not a boolean

src/lib/Preview.js
tonyjin: Used a few lines below

src/lib/Preview.js
tonyjin: I think that's fine, you'd have to explicitly pass in `null` and in that case, it evaluates to falsy.

src/lib/viewers/BaseViewer.js
tonyjin: Nope, I can add it. We're not super consistent about what's defined as a constant vs what's not.

tonyjin:
@jeremypress `reload()` has the benefit of not requiring the user to pass in file ID, authentication, or any preview options.

PR #563, opened 2017-12-22
Chore: Setup viewer notifications in the correct location

pramodsum:


boxcla:

src/lib/PreviewUI.js
```diff
@@ -299,6 +310,10 @@ class PreviewUI {
      * @return {void}
      */
     showNotification(message, buttonText) {
+        if (!this.notification) {
+            this.setupNotification();
+        }
```
tonyjin: Good idea moving it here!

src/lib/PreviewUI.js
```diff
@@ -309,6 +324,10 @@ class PreviewUI {
      * @return {void}
      */
     hideNotification() {
+        if (!this.notification) {
+            this.setupNotification();
+        }
```
tonyjin: I don't think you need to setup notification here since it's only a prerequisite to showing a notification.

Something like should work:
```
if (this.notification) {
    this.notification.hide();
}
```

src/lib/PreviewUI.js
```diff
@@ -264,9 +264,6 @@ class PreviewUI {
                 // part of finishLoadingSetup in BaseViewer.js
                 crawler.classList.remove(CLASS_HIDDEN);
             }
-
-            // Setup viewer notification
-            this.notification = new Notification(this.contentContainer);
         }
```
tonyjin: Since you're in this file already, I'd move this check to be:

```
if (!this.contentContainer) {
    return;
}
```

to reduce arrow-ness

src/lib/PreviewUI.js
tonyjin: Easiest fix would be:

```
if (!this.notification) {
    this.setupNotification();
}

if (this.notification) {
     this.showNotification();
}
```

But, I'm not a huge fan of how that reads... Open to other suggestions.

pramodsum: cleaned it up a bit. Thoughts?

PR #580, opened 2018-01-11
New logger and abstraction layer

JustinHoldstock:
This PR includes a bunch of things:
1) New mechanism for saving batched performance metrics and errors to Box.
2) Printing utilities for reading what is stored in the logger (log history for the current viewer).
3) Begin deprecating the old `Logger` that dealt with file loading metrics. Now called `FileMetrics`.
4) Abstraction layer that sits between `Preview` and the `Logger` and `Resin`, called `LogLayer`, that will make it easier for us to rip out `Logger` and all of its components to be a logger that exists in the parent application.
5) Tiered console printing. Use methods like `info()/warn()/error()/metric()` to create specific logs that will only print to console if the correct tier is set, see note **6**
6) Two new options to pass to Preview.
`consoleLevel` - Silent by default. Sets current tier of console printing. Logs below the current level are not printed to the console. 
```
     silent
       ↑
     error 
       ↑
    warning
       ↑
     info
```

 `disableLogSaving` - False by default. If true, prevents saving of all logs to Box.

I want to get the ball rolling on a review for this, but outstanding work includes:
1) @jeremypress and I were talking about the `FileMetrics` class. Not a great name. Ideas?
2) Add a method to programmatically create `Resin` events.
3) Do we want viewers to directly call the `LogLayer` instance, or emit messages, as they do in this PR?


boxcla:

yarn.lock
```diff
@@ -4893,7 +4893,7 @@ loglevel-colored-level-prefix@^1.0.0:
 
 loglevel@^1.4.1:
   version "1.5.1"
```
JustinHoldstock: If anyone knows how I can get rid of the entry in the yarn file, I'd be eternally grateful.

JustinHoldstock: TIL `loglevel` is a dependency of `loglevel-colored-level-prefix`, which is a depeddency of `prettier-eslint-cli`

src/lib/Preview.js
```diff
@@ -631,6 +665,9 @@ class Preview extends EventEmitter {
             );
         }
 
+        // Set logger to use up to date file info
+        this.logger.setFile(this.file);
```
JustinHoldstock: Group this call in the same are as the one to `setupLogger()` below

src/lib/Preview.js
```diff
@@ -945,6 +996,30 @@ class Preview extends EventEmitter {
         this.viewer.addListener('viewerevent', this.handleViewerEvents);
     }
 
+    /**
+     * Handles log events and delegates to the Logger instance.
+     *
+     * @param {Object} data - Log event data.
+     * @return {void}
+     */
+    handleLogEvent(data) {
```
JustinHoldstock: This is what I mentioned in the description of the PR. Do we want viewers emitting these messages, and Preview handles it, or do we want viewers to have a `LogLayer` instance reference, and call methods on it, directly?

src/lib/logging/LogLayer.js
```diff
@@ -0,0 +1,314 @@
...
+    loggerRef;
+
+    /** @property {Object} - Reference to the global Resin object */
+    resinRef;
+
+    /**
+     * @constructor
+     *
+     * @param {Object} config - Configures level to log for console log level and network layer.
+     * @param {CONSOLE_LEVELS|string} [config.consoleLevel] - Level to set for writing to the browser console.
```
JustinHoldstock: Add param docs for network properties, for Logger

src/lib/logging/LogLayer.js
```diff
@@ -0,0 +1,314 @@
...
+     */
+    setConsoleLevel(level) {
+        this.consoleLevel = level;
+    }
+
+    /**
+     * Interface to the Logger for info messages.
+     *
+     * @public
+     * @param {*} args - Arguments to be printed to the log and cache.
```
JustinHoldstock: Remove references to cache. Same with ones below

src/lib/logging/LogLayer.js
```diff
@@ -0,0 +1,314 @@
...
+        }
+
+        this.loggerRef.setContentType(type);
+    }
+
+    /**
+     * Interface for passing through network layer settings to the Logger.
+     *
+     * @public
+     * @param {Object} config - Network related settings to pass to the Logger.
```
JustinHoldstock: Add docs for parameters that should be expected

src/lib/logging/LogLayer.js
```diff
@@ -0,0 +1,314 @@
...
+        this.loggerRef.save([LOG_TYPES.error, LOG_TYPES.metric]);
+        this.loggerRef.clearCache();
+    }
+
+    //--------------------------------------------------------
+    // PRIVATE
+    //--------------------------------------------------------
+
+    /**
+     * Maps logging type to a logging function in our third-party logger. Defaults
```
JustinHoldstock: Remove reference to third-party logger. We no longer use one

src/lib/logging/LogLayer.js
```diff
@@ -0,0 +1,314 @@
...
+     * to regular 'info' if nothing available.
+     *
+     * @private
+     * @param {LOG_TYPES|string} type - The log type to lookup a corresponding function on the logger.
+     * @return {Function} A function that can be invoked with a string, to log a message.
```
JustinHoldstock: A function that can be invoked with a string, to print a message to the console. Nothing if no correct mapping

src/lib/logging/LogLayer.js
```diff
@@ -0,0 +1,314 @@
...
+                logFunction = console.warn;
+                break;
+            case LOG_TYPES.error:
+            case LOG_TYPES.uncaught_error:
+                logFunction = console.error;
+                break;
+            case LOG_TYPES.info:
+                logFunction = console.log;
+                break;
+            default:
```
JustinHoldstock: Should default to `console.log`

src/lib/logging/LogLayer.js
```diff
@@ -0,0 +1,314 @@
...
+        /* eslint-enable no-console */
+
+        return logFunction;
+    }
+
+    /**
+     * Determine whether or not we can print to the console, based on consoleLevel.
+     *
+     * @private
+     * @param {LOG_TYPES} type - Type of log to check.
```
JustinHoldstock: {LOG_TYPES|string}

src/lib/logging/LogLayer.js
```diff
@@ -0,0 +1,314 @@
...
+    resinExists() {
+        this.resinRef = global.Resin;
+        return !!this.resinRef;
+    }
+
+    /**
+     * Print a message to the console if allowed to.
+     *
+     * @private
+     * @param {LOG_TYPE} type - Type of log to check and print to the console.
```
JustinHoldstock: {LOG_TYPES|string}

src/lib/logging/LogLayer.js
```diff
@@ -0,0 +1,314 @@
...
+
+    /**
+     * Create and attach listeners on this network layer, until we decide
+     * how the Logger will live outside of Preview.
+     *
+     * @private
+     * @param {Object} config - Configuration for the global logger
+     * @return {void}
+     */
+    setupGlobalLogger(config) {
```
JustinHoldstock: Add docs for expected params

src/lib/Preview.js
```diff
@@ -121,6 +125,9 @@ class Preview extends EventEmitter {
     /** @property {PreviewUI} - Preview's UI instance */
     ui;
 
+    /** @property {Logger} - Preview's LogLayer instance */
+    logger;
```
jeremypress: This naming is a little confusing. We call it `logger` and it's an instance of `Logger` but it's actually an instance of `LogLayer`?

src/lib/Preview.js
```diff
@@ -588,6 +596,32 @@ class Preview extends EventEmitter {
...
     }
 
+    /**
+     * Setup additional configuration for the logger.
+     *
+     * @param {Object} config - Configures log level and network layer.
+     * @param {CONSOLE_LEVELS|string} [config.consoleLevel] - Level to set for writing to the browser console.
+     * @param {boolean} [config.savingEnabled] - If true, allows saving of logs to a network layer.
+     * @param {string} [config.logURL] - Full url to save logs to. Can instead use appHost with logEndpoint (see below)
+     * @param {string} [config.appHost] - Base URL to save logs to. Is combined with logEndpoint (below)
```
jeremypress: Are we still going for the idea of allowing people to save to their own back end? That might be better for a v2/after seeing some demand

src/lib/Preview.js
```diff
@@ -693,6 +730,12 @@ class Preview extends EventEmitter {
         // Setup loading UI and progress bar
         this.ui.showLoadingIndicator();
         this.ui.startProgressBar();
+
+        this.setupLogger({
```
jeremypress: Do we have a non UI setup method we can move this to?

src/lib/Preview.js
```diff
@@ -693,6 +730,12 @@ class Preview extends EventEmitter {
...
+            consoleLevel: this.options.consoleLevel,
+            appHost: this.options.appHost,
```
jeremypress: What if your appHost and the host where you want to save your logs are different? As mentioned above I think this is too sophisticated for a v1.

src/lib/Preview.js
jeremypress: Assuming Preview is not doing anything useful with the information (just passing it to logLayer), we might as well just pass the instance into the viewers and let them log from there. Going through preview would just create another route that adds no functionality to get to the same place.

src/lib/Preview.js
```diff
@@ -1043,13 +1123,18 @@ class Preview extends EventEmitter {
...
             // Finally emit the viewer instance back with a load event
             this.emit('load', {
                 viewer: this.viewer,
-                metrics: this.logger.done(this.count),
+                metrics,
                 file: this.file
             });
 
+            // Track in interal logger
+            this.logger.metric(METRIC_FILE_PREVIEW_SUCCESS, metrics);
```
jeremypress: I know I keep bringing this up, but why not just send the `metric` stuff from the logger to Gen204 instead of maintaining both `Logger`/`LogLevel` and `FileMetrics`? It looks like we repeat a lot of function calls and here we are directly passing in our `FileMetrics` info into `LogLevel`.

src/lib/Preview.js
jeremypress: On the other hand, we could solve this third party logging problem by just broadcasting events and letting devs consume the events and do what they want with them.

src/lib/Preview.js
JustinHoldstock: I completely missed this. Updated

src/lib/logging/LogNetworkLayer.js
```diff
@@ -0,0 +1,151 @@
...
+    /** @property {string} - The locale the preview session occurs in. An estimate. */
+    locale;
+
+    /** @property {string} - Unique GUID/UUID for identifying session */
+    sessionID = uuidv4();
+
+    /**
+     * @param {Object} config - Object used to initialize the network layer.
+     * @param {string} config.logURL - The full URL to POST log events to. REQUIRED.
+     */
```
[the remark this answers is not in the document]
JustinHoldstock: done

src/lib/logging/LogNetworkLayer.js
```diff
@@ -0,0 +1,151 @@
...
+    setLocale(locale) {
+        this.locale = locale;
+    }
+
+    /**
+     * Create a properly formatted batch of logs to be saved to the network layer.
+     *
+     * @param {LOG_TYPES} type - Type of logs contained in the batch.
+     * @param {Object} logs - Object containing type and array of logs that belong to it.
```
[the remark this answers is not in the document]
JustinHoldstock: done

src/lib/logging/LogNetworkLayer.js
```diff
@@ -0,0 +1,151 @@
...
+     * @return {Object} Formatted object to be saved to the network layer.
```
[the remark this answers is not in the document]
JustinHoldstock: done

src/lib/logging/Logger.js
```diff
@@ -0,0 +1,364 @@
...
+// Filter for logs allowed to be saved to the logUrl
+const DEFAULT_ALLOWED_LOGS = {
+    info: true,
+    metric: true,
+    warning: true,
+    error: true,
+    uncaught_error: true
+};
+
+// Publicliy accessible, doesn't require auth token.
```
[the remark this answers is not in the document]
JustinHoldstock: done

src/lib/logging/Logger.js
```diff
@@ -0,0 +1,364 @@
...
+     * @param {string} [config.locale] - User's locale
+     * @return {Logger} Newly created Logger instance.
+     */
+    constructor(config = {}) {
+        this.cache = new LoggerCache();
+
+        this.onUncaughtError = this.onUncaughtError.bind(this);
+        window.addEventListener('error', this.onUncaughtError);
+
+        this.name = registerLogger(this);
```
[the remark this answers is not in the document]
JustinHoldstock: done

src/lib/Preview.js
JustinHoldstock: Agreed that this will be better suited to V2. Leaving docs and still using it bc I want to use our Preview options to configure this, as it's easier when switching between dev env and live

PR #582, opened 2018-01-12
Update: Respect 'Retry-After' header if present

tonyjin:
When fetching for file info, respect 'Retry-After' header if present. Otherwise, retry with exponential back-off.


boxcla:

PR #583, opened 2018-01-12
Fix: Ensure only loading wrapper is interactable preview container

pramodsum:
- The preview CSS selector only prevented the div directly adjacent to the .bp-loading-wrapper from being interactable.
- This selector prevents all divs within the Preview Container EXCEPT .bp-loading-wrapper from being interactable. This prevents interactions from the Preview content, find bar, controls, etc.

### DIFF
--- a/src/lib/_loading.scss
+++ b/src/lib/_loading.scss
@@ -39,8 +39,12 @@
 }
 
 // Only the loading wrapper should be interactable inside the Preview container
-.bp:not(.bp-loaded) > div:not(.bp-loading-wrapper) {
+.bp:not(.bp-loaded) {
     pointer-events: none;
+
+    & .bp-loading-wrapper {
+        pointer-events: all;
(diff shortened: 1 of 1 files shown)


boxcla:

src/lib/_loading.scss
```diff
@@ -38,8 +38,8 @@
     }
 }
 
-// Preview content should not be interactable during loading
-:not(.bp-loaded) > .bp-loading-wrapper + div {
+// Only the loading wrapper should be interactable inside the Preview container
+.bp:not(.bp-loaded) > div:not(.bp-loading-wrapper) {
```
jeremypress: Can we avoid 2 nots? Maybe add this to all and then nest for the .bp-loading-wrapper where we turn touch events back on.

pramodsum: 👍

PR #588, opened 2018-01-16
Fix: fullscreen controls not shown in IE11

DanDeMicco:
When IE11 was fullscreened it cut off the bottom controls. This was due to a class that gets added when the header is present. The class is supposed to be removed in fullscreen and preview should be shifted up, so the controls won't be cut off.

The issue occured because there was multiple event listeners which got added in the constructor of Fullscreen.js due to the fullscreen API sometimes requiring vendor prefixes [Fullscreen API](https://developer.mozilla.org/en-US/docs/Web/API/Fullscreen_API). This caused the toggle function to be called an even number of times, thus removing and re-adding the header class.

Following the suggestion from MDN, I added fscreen to resolve the cross browser differences and only add 1 event listener instead of 4.

### DIFF
--- a/src/lib/Fullscreen.js
+++ b/src/lib/Fullscreen.js
@@ -12,9 +12,42 @@ class Fullscreen extends EventEmitter {
     constructor() {
         super();
 
+        this.bindDOMListeners();
+    }
+
+    /**
+     * Binds DOM listeners for Fullscreen.
+     *
+     * @protected
+     * @return {void}
+     */
+    bindDOMListeners() {
         // as of now (1/12/18) fullscreenchange is not universally adopted, fscreen will
         // detect and add the appropriate vendor prefixed event
         fscreen.addEventListener('fullscreenchange', this.fullscreenchangeHandler);
+    }
+
+    /**
+     * Binds DOM listeners for Fullscreen.
+     *
+     * @protected
+     * @return {void}
+     */
+    unbindDOMListeners() {
+        // as of now (1/12/18) fullscreenchange is not universally adopted, fscreen will
+        // detect and add the appropriate vendor prefixed event
+        fscreen.removeEventListener('fullscreenchange', this.fullscreenchangeHandler);
+    }
+
+    /**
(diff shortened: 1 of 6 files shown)


boxcla:
Hi @DanDeMicco, thanks for the pull request. Before we can merge it, we need you to sign our Contributor License Agreement. You can do so electronically here: http://opensource.box.com/cla

DanDeMicco:
CLA signed

src/lib/Fullscreen.js
```diff
@@ -10,10 +12,9 @@ class Fullscreen extends EventEmitter {
     constructor() {
         super();
 
-        document.addEventListener('webkitfullscreenchange', this.fullscreenchangeHandler);
-        document.addEventListener('mozfullscreenchange', this.fullscreenchangeHandler);
-        document.addEventListener('MSFullscreenChange', this.fullscreenchangeHandler);
-        document.addEventListener('fullscreenchange', this.fullscreenchangeHandler);
+        // as of now (1/12/18) fullscreenchange is not universally adopted, fscreen will
+        // detect and add the appropriate vendor prefixed event
+        fscreen.addEventListener('fullscreenchange', this.fullscreenchangeHandler);
```
pramodsum: Can you add a unit test to check that only one handler is added per vendor?

src/lib/Fullscreen.js
```diff
@@ -1,4 +1,6 @@
 import EventEmitter from 'events';
+import fscreen from 'fscreen';
```
priyajeet: I would change all functions below to also use it...so that we don't have a mix of native vs lib.

src/lib/Fullscreen.js
DanDeMicco: Added a UT + some refactoring to the PR

src/lib/Fullscreen.js
DanDeMicco: @priyajeet I will create a ticket to refactor this file

PR #589, opened 2018-01-17
Event names into separate file

JustinHoldstock:
This pull request moves all viewer events into `events.js` to provide documentation for each event and reduce possibility of using the wrong event names.

### DIFF
--- a/src/lib/events.js
+++ b/src/lib/events.js
@@ -1,6 +1,6 @@
 // Events emitted by Viewers
 // eslint-disable-next-line import/prefer-default-export
-export const VIEWER_EVENTS = {
+export const VIEWER_EVENT = {
     download: 'download', // Begin downloading the file.
     reload: 'reload', // Reload preview.
     load: 'load', // Preview is finished loading.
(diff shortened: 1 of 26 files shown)


boxcla:

src/lib/events.js
```diff
@@ -0,0 +1,13 @@
+// Events emitted by Viewers
+// eslint-disable-next-line import/prefer-default-export
+export const VIEWER_EVENTS = {
```
pramodsum: could we rename it to VIEWER_EVENT rather than VIEWER_EVENTS?

src/lib/events.js
```diff
@@ -0,0 +1,13 @@
...
+    download: 'download', // Begin downloading the file.
+    reload: 'reload', // Reload preview.
+    load: 'load', // Preview is finished loading.
+    progressStart: 'progressstart', // Begin using loading indicator.
+    progressEnd: 'progressend', // Stop using loading indicator.
+    notificationShow: 'notificationshow', // Show notification modal.
+    notificationHide: 'notificationhide', // Hide notification modal.
+    mediaEndAutoplay: 'mediaendautoplay', // Media playback has completed, with autoplay enabled.
+    generic: 'viewerevent' // The generic default viewer event
```
pramodsum: default?

PR #590, opened 2018-01-18
Fix: mobile audio controls overlap with navigation

DanDeMicco:
COXP-3831


boxcla:

PR #593, opened 2018-01-18
New: Specific error messages for unsupported/tariff restricted files

jeremypress:
New Errors:
1. Generic: Only used when we  have no other message to display
2. Unsupported: We can now be sure that the file mentioned is not supported!
3. Account: In the case that Box supports the file, but your tariff does not allow preview for the given type.

### DIFF
--- a/src/i18n/en-US.properties
+++ b/src/i18n/en-US.properties
@@ -37,7 +37,11 @@ text_truncated=This file has been truncated due to size limits. Please download 
 
 # Error messages
 # Default preview error message
-error_default=We're sorry, the preview didn't load. This file type may not be supported.
+error_generic=We're sorry, the preview didn't load.
+# Default preview error message
+error_unsupported=We're sorry, the preview didn't load. {1} files are not currently supported.
+# Account doesn't have a sufficient tariff to preview the requested file type
+error_account=We're sorry, your account is unable to preview this file type.
 # No permissions preview error message
 error_permissions=We're sorry, you don't have permission to preview this file.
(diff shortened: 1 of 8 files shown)


boxcla:

src/i18n/en-US.properties
```diff
@@ -37,7 +37,11 @@ text_truncated=This file has been truncated due to size limits. Please download
 
 # Error messages
 # Default preview error message
-error_default=We're sorry, the preview didn't load. This file type may not be supported.
+error_generic=We're sorry, the preview didn't load.
```
jeremypress: Thoughts on language for the new messages?

DanDeMicco: From my limited experience refreshing the page seems to fix the issue when this error occurs. If that is usually the case maybe append something about refreshing?

src/lib/Preview.js
```diff
@@ -898,10 +898,18 @@ class Preview extends EventEmitter {
         // Determine the asset loader to use
         const loader = this.getLoader(this.file);
 
-        // If no loader then throw an unsupported error
-        // If file type specific error message, throw the generic one
+        // If no loader then check to see if any of our viewers suppor this file type.
```
JustinHoldstock: Typo: "check to see if any of our viewers **support** this file type"

src/i18n/en-US.properties
JustinHoldstock: My opinion is to leave it as is. If we put the expectation that refreshing could potentially fix the issue, it's another thing that can muddy external bug reports.

PR #594, opened 2018-01-19
Fix: Image viewer flickering in IE11 during reset zoom

DanDeMicco:
The old 'reset' case set the image width & height, then recursively called zoom, which caused the image to be resized twice and thus flickered in IE11.

Also, has a minor additional bugfix in which the image was rendered at the incorrect size by default, thus when resetting on a freshly rendered image it would slightly resize. This was due to use of clientHeight instead of offsetHeight which does not include the scrollbar (which may be on initial load)

### DIFF
--- a/src/lib/viewers/image/ImageViewer.js
+++ b/src/lib/viewers/image/ImageViewer.js
@@ -126,12 +126,26 @@ class ImageViewer extends ImageBaseViewer {
     zoom(type) {
         let newWidth;
         let newHeight;
-        const imageCurrentDimensions = this.imageEl.getBoundingClientRect(); // Getting bounding rect does not ignore transforms / rotates
+        let imageCurrentDimensions;
+        if (type === 'reset') {
+            // when it is a reset, take the original dimensions of the image
+            // rather than setting the resetting height and width and calling
+            // getBoundingClientRect as it will cause flickering on IE11
+            const height = this.imageEl.getAttribute('originalHeight');
+            const width = this.imageEl.getAttribute('originalWidth');
+            imageCurrentDimensions = {
+                height,
+                width
+            };
+        } else {
+            imageCurrentDimensions = this.imageEl.getBoundingClientRect(); // Getting bounding rect does not ignore transforms / rotates
+        }
+
         const { height, width } = imageCurrentDimensions;
         const aspect = width / height;
         const viewport = {
-            width: this.wrapperEl.clientWidth - IMAGE_PADDING,
-            height: this.wrapperEl.clientHeight - IMAGE_PADDING
+            width: this.wrapperEl.offsetWidth - 2 * IMAGE_PADDING,
+            height: this.wrapperEl.offsetHeight - 2 * IMAGE_PADDING
         };
 
         // For multi page tifs, we always modify the width, since its essentially a DIV and not IMG tag.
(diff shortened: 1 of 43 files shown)


boxcla:

src/lib/viewers/image/ImageViewer.js
```diff
@@ -126,12 +126,26 @@ class ImageViewer extends ImageBaseViewer {
     zoom(type) {
         let newWidth;
         let newHeight;
-        const imageCurrentDimensions = this.imageEl.getBoundingClientRect(); // Getting bounding rect does not ignore transforms / rotates
+        let imageCurrentDimensions;
+        if (type === 'reset') {
```
jeremypress: Can we keep this in the switch statement? You can set `imageCurrentDimensions` here and then update it in the switch statement if we are resetting. It feels a little weird to check for the 'reset' case and then switch on type right after.

src/lib/viewers/image/ImageViewer.js
```diff
@@ -116,6 +116,23 @@ class ImageViewer extends ImageBaseViewer {
...
+    /**
+     * Getter for modifyWidthInsteadOfHeight which will be used in zoom
+     * and will determine if the width or the height needs to be modified
+     *
+     * @private
+     * @param {number} [width] - The width in px
+     * @param {number} [height] - The height in px
+     * @return {boolean} true if the width will be modified instead of height
+     */
+    getModifyWidthInsteadOfHeight(width, height) {
```
JustinHoldstock: Why not go with something like `isLandscape()`, since we're essentially checking whether or not the width is greater than the height? Also, we could reduce this to `return width >= height`. Were there any other reasons behind using aspect ratio?

Note: We may already have an `isLandscape()` method hiding somewhere

src/lib/viewers/image/ImageViewer.js
```diff
@@ -126,79 +143,57 @@ class ImageViewer extends ImageBaseViewer {
...
-            case 'in':
+        if (type === 'in' || type === 'out') {
+            const imageCurrentDimensions = this.imageEl.getBoundingClientRect(); // Getting bounding rect does not ignore transforms / rotates
+            ({ height, width } = imageCurrentDimensions);
+
+            const modifyWidthInsteadOfHeight = this.getModifyWidthInsteadOfHeight(width, height);
+            if (modifyWidthInsteadOfHeight) {
+                newWidth = type === 'in' ? width * IMAGE_ZOOM_SCALE : width / IMAGE_ZOOM_SCALE;
+            } else {
+                newHeight = type === 'in' ? height * IMAGE_ZOOM_SCALE : (newHeight = height / IMAGE_ZOOM_SCALE);
```
JustinHoldstock: I think you can remove the assignment as the last component of the ternary.

`newHeight = type === 'in' ? height * IMAGE_ZOOM_SCALE : height / IMAGE_ZOOM_SCALE;`

src/lib/viewers/image/ImageViewer.js
DanDeMicco: Good point, the CR was just moving around code but I don't see the aspect ratio being used anywhere

src/lib/viewers/image/ImageViewer.js
DanDeMicco: Good catch, that was what I meant to put there to match the newWidth expression

src/lib/viewers/image/ImageViewer.js
```diff
@@ -116,6 +116,21 @@ class ImageViewer extends ImageBaseViewer {
         this.setScale(this.imageEl.offsetwidth, this.imageEl.offsetHeight);
     }
 
+    /**
+     * Getter for modifyWidthInsteadOfHeight which will be used in zoom
```
JustinHoldstock: Update docs to not refer to `modifyWidthInsteadOfHeight`

src/lib/viewers/image/ImageViewer.js
```diff
@@ -116,6 +116,21 @@ class ImageViewer extends ImageBaseViewer {
...
+     * and will determine if the width or the height needs to be modified
+     *
+     * @private
+     * @param {number} [width] - The width in px
+     * @param {number} [height] - The height in px
+     * @return {boolean} true if the width will be modified instead of height
```
JustinHoldstock: true if image is landscape layout, or something like that

src/lib/viewers/image/ImageViewer.js
```diff
@@ -116,6 +116,21 @@ class ImageViewer extends ImageBaseViewer {
...
+     */
+    isLandscape(width, height) {
+        // For multi page tifs, we always modify the width, since its essentially a DIV and not IMG tag.
```
JustinHoldstock: I'd say we can remove these two comment lines

src/lib/viewers/image/__tests__/ImageViewer-test.js
```diff
@@ -521,4 +543,16 @@ describe('lib/viewers/image/ImageViewer', () => {
             });
         });
     });
+
+    describe('getModifyWidthInsteadOfHeight()', () => {
```
JustinHoldstock: update test name to `isLandscape()`

PR #598, opened 2018-01-22
Fix: Disable annotations if user has neither annotate nor view permissions

pramodsum:


boxcla:

src/lib/viewers/BaseViewer.js
```diff
@@ -769,6 +769,15 @@ class BaseViewer extends EventEmitter {
      * @return {boolean} Whether or not viewer is annotatable
      */
     areAnnotationsEnabled() {
+        // Do not attempt to fetch annotations if the user cannot create or view annotations
+        const { permissions } = this.options.file;
+        if (
```
JustinHoldstock: Can we abstract this into a function? `hasAnnotationPermissions()`?

PR #599, opened 2018-01-23
Fix settings menu title cut off by scrollbar

JustinHoldstock:
Fixes scroll bar cutting off menu

screen shot 2018-01-22 at 4 58 54 pm

### DIFF
--- a/src/lib/viewers/media/Settings.js
+++ b/src/lib/viewers/media/Settings.js
@@ -22,6 +22,9 @@ const MEDIA_SPEEDS = ['0.5', '1.0', '1.25', '1.5', '2.0'];
 const MEDIA_QUALITY_SD = 'sd';
 const MEDIA_QUALITY_HD = 'hd';
 const MEDIA_QUALITY_AUTO = 'auto';
+const SETTINGS_MENU_PADDING = 18;
+const SETTINGS_MENU_PADDING_SCROLL = 32;
+const SETTINGS_MENU_MAX_HEIGHT = 210;
 
 const SETTINGS_TEMPLATE = `<div class="bp-media-settings">
     <div class="bp-media-settings-menu-main bp-media-settings-menu" role="menu">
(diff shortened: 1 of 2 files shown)


boxcla:

src/lib/viewers/media/Settings.js
```diff
@@ -292,11 +292,15 @@ class Settings extends EventEmitter {
      * @return {void}
      */
     setMenuContainerDimensions(menu) {
-        // NOTE: need to explicitly set the dimensions in order to get css transitions. width=auto doesn't work with css transitions
-        this.settingsEl.style.width = `${menu.offsetWidth + 18}px`;
+        const { children } = menu;
+
+        // If we have enough children to require scrolling, take into account scroll bar width.
+        const scrollPadding = children.length > 7 ? 32 : 18;
```
DanDeMicco: Extract these magic numbers to constants

src/lib/viewers/media/Settings.js
```diff
@@ -292,11 +292,15 @@ class Settings extends EventEmitter {
...
+        this.settingsEl.style.width = `${menu.offsetWidth + scrollPadding}px`;
+
         // height = n * $item-height + 2 * $padding (see Settings.scss) + 2 * border (see Settings.scss)
         // where n is the number of displayed items in the menu
-        const sumHeight = [].reduce.call(menu.children, (sum, child) => sum + child.offsetHeight, 0);
+        const sumHeight = [].reduce.call(children, (sum, child) => sum + child.offsetHeight, 0);
```
DanDeMicco: I was taking a look at this before and was trying to figure out why it iterates over the children and sums up the offsetHeight vs taking the menu.offsetHeight. Seemed to work with the menu offsetHeight. Any ideas?

src/lib/viewers/media/Settings.js
jeremypress: Would looking at scrollHeight help? 

`The Element.scrollWidth read-only property is a measurement of the width of an element's content, including content not visible on the screen due to overflow.`

https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollWidth

src/lib/viewers/media/__tests__/Settings-test.js
```diff
@@ -101,6 +101,47 @@ describe('lib/viewers/media/Settings', () => {
...
+            menuEl.appendChild(document.createElement('span'));
+            settings.setMenuContainerDimensions(menuEl);
+
+            expect(settings.settingsEl.style.width).to.equal('18px');
+        });
+
+        it('should add extra padding to settingsEl based on menu contents that require scroll bar', () => {
+            const menuEl = document.createElement('div');
+            // Greater than enough to add a scroll bar.
+            menuEl.appendChild(document.createElement('span'));
```
jeremypress: Add one element and set it's size with js?

src/lib/viewers/media/Settings.js
JustinHoldstock: I'm not too sure, but I can confirm that `menu.offsetHeight` is the way to go.

src/lib/viewers/media/__tests__/Settings-test.js
JustinHoldstock: This will work especially since we're no longer counting child heights, and using offsetHeight

PR #604, opened 2018-01-25
COXP-4005: Preview JS error in console - 'unable to get property handleEvent' of undefined or null reference

JustinHoldstock:
Looks like we missed removing the unbinding of an event listener to a function that no longer exists


boxcla:

PR #605, opened 2018-01-25
Chore: Allow notifications to persist

jeremypress:
Also overriding some styles such that the notification is absolute to the top of it's container, rather than fixed to the top of the page.


boxcla:

PR #606, opened 2018-01-26
New: Add in SinglePageViewer for pdfs

jiverson:
Implement single page viewer for pdfs.

Example: https://github.com/mozilla/pdf.js/blob/master/examples/components/singlepageviewer.html


boxcla:
Verified that @jiverson has signed the CLA. Thanks for the pull request!

src/lib/viewers/doc/DocLoader.js
```diff
@@ -22,6 +23,12 @@ const VIEWERS = [
         REP: 'pdf',
         EXT: DOCUMENT_EXTENSIONS
     },
+    {
+        NAME: 'SinglePage',
```
jeremypress: The current behavior is that if we disable the Document viewer, we fallback to the Presentation viewer. Can you put this below the next Presentation viewer entry? This means that for your viewer to take effect, you would have to disable both the Document and Presentation viewer.

We can add this to the Readme.md file in the doc viewer folder.

src/lib/viewers/doc/DocBaseViewer.js
```diff
@@ -558,6 +553,22 @@ class DocBaseViewer extends BaseViewer {
...
     }
 
+    /**
+     * Get pdf.js viewer.
+     *
+     * @protected
+     * @override
+     * @return {PDFJS.PDFViewer} PDF viewer type
+     */
+    getPdfViewer() {
```
jeremypress: since we're actually instantiating the viewer here as opposed to getting it, can you rename this to `initPdfViewer`?

src/lib/viewers/doc/DocLoader.js
jiverson: Did I rearrange this correctly I did not completely understand how this works?

jeremypress: Yep, this looks good to me. We pick the first viewer in this list that
1. Is enabled
2. Can support the passed in file extension (EXT)
3. Can support a representation that is provided by the API (REP)

jiverson: Hey thanks for that, clears it up I think I was getting stuck on the REP part I didn't completely understand that.

tonyjin:
Thanks for your patch @jiverson!

PR #608, opened 2018-01-27
New: Support previews of non-current file versions

### LINKED ISSUE #570: Support for GetVersions
Is there any way to use [getVersions](https://developer.box.com/v2.0/reference#view-versions-of-a-file) API, that user can switch between versions of a particular file?

tonyjin:
This patch adds support for previewing non-current file versions. Specify the file version like so:

```
const preview = new Box.Preview();
preview.show(FILE_ID_1, ACCESS_TOKEN, {
    fileOptions: {
    	[FILE_ID_1]: {
	    fileVersionId: FILE_VERSION_ID_1
	}
    }
});
```

In this patch, we start caching file objects by both file ID and by file version ID, distinguishing between the two with a 'file_' or 'file_version_' prefix as the key to the cache. This allows us to be backwards-compatible and show the current file version if no file version ID is passed in, while always allowing a cached entry of a current version to be overriden when you preview a non-current file version. We abstract this logic to file.js, and rely on the helper methods cacheFile(), uncacheFile(), and getCachedFile() to interact with file objects and the cache. This patch also enables prefetching of files that were previously cached by file ID or file version ID.

To support previews of file versions, we introduce a new Preview option `fileOptions`, which allows us to specific individual options by file ID. This is needed instead of just passing in a `fileVersionId` opion since we support collections. Having this mapping of file ID to file options on that file ID allows us to specify file verison IDs for any number of file IDs to preview and can also be extended to support other file-level options in the future. Example of collection + file version support together:

```
const preview = new Box.Preview();
preview.show(FILE_ID_1, ACCESS_TOKEN, {
    collection: [FILE_ID_1, FILE_ID2],
    fileOptions: {
        [FILE_ID_1]: {
            fileVersionId: FILE_VERSION_ID_1
        },
        [FILE_ID_2]: {
            fileVersionId: FILE_VERSION_ID_2
        }
    }
});
```

Caveats:
- You cannot pass in the current file version ID as an option since the Box API will fail when you make a get file version info API call for the current version
- Preview events do not distinguish between file versions, so a preview of a previous file version will be simply logged as a preview of a file ID

This resolves #570.

### DIFF
--- a/README.md
+++ b/README.md
@@ -127,10 +127,15 @@ preview.show(fileId, accessToken, {
 | collection |  | List of file IDs to iterate over for previewing |
 | header | 'light' | String value of 'none' or 'dark' or 'light' that controls header visibility and theme |
 | logoUrl |  | URL of logo to show in header |
-| showAnnotations | false | Whether annotations and annotation controls are shown. This option will be overridden by viewer-specific annotation options if they are set. See [Box Annotations](https://github.com/box/box-annotations) for more details. |
+| showAnnotations | false | Whether annotations and annotation controls are shown. This option will be overridden by viewer-specific annotation options if they are set. See [Box Annotations](https://github.com/box/box-annotations) for more details |
 | showDownload | false | Whether download button is shown |
 | useHotkeys | true | Whether hotkeys (keyboard shortcuts) are enabled |
 | pauseRequireJS | false | Temporarily disables requireJS to allow Preview's third party dependencies to load |
+| fileOptions | {} | Mapping of file ID to file-level option. See below for details |
+
+| File Option | Description |
+| --- | --- |
+| fileVersionId | File version ID to preview. This must be a valid non-current file version ID. Use [Get Versions](https://developer.box.com/reference#view-versions-of-a-file) to fetch a list of file versions |
 
 Access Token
 ------------
--- a/src/lib/Preview.js
+++ b/src/lib/Preview.js
@@ -1,8 +1,9 @@
 /* eslint-disable import/first */
 import './polyfill';
 import EventEmitter from 'events';
-import throttle from 'lodash.throttle';
 import cloneDeep from 'lodash.clonedeep';
+import getProp from 'lodash.get';
+import throttle from 'lodash.throttle';
 /* eslint-enable import/first */
 import Browser from './Browser';
 import Logger from './Logger';
@@ -29,7 +30,9 @@ import {
     checkFileValid,
     cacheFile,
     uncacheFile,
-    isWatermarked
+    isWatermarked,
+    getCachedFile,
+    normalizeFileVersion
 } from './file';
 import {
     API_HOST,
@@ -42,7 +45,8 @@ import {
     X_REP_HINT_DOC_THUMBNAIL,
     X_REP_HINT_IMAGE,
     X_REP_HINT_VIDEO_DASH,
-    X_REP_HINT_VIDEO_MP4
+    X_REP_HINT_VIDEO_MP4,
+    FILE_OPTION_FILE_VERSION_ID
 } from './constants';
 import { VIEWER_EVENT } from './events';
 import './Preview.scss';
@@ -506,7 +510,8 @@ class Preview extends EventEmitter {
      *
      * @public
      * @param {Object} options - Prefetch options
-     * @param {string} options.fileId - Box File ID
+     * @param {string} options.fileId - Box file ID (do not also pass a file version ID)
+     * @param {string} options.fileVersionId - Box file version ID (do not also pass a file ID)
      * @param {string} options.token - Access token
      * @param {string} options.sharedLink - Shared link
      * @param {string} options.sharedLinkPassword - Shared link password
@@ -514,14 +519,14 @@ class Preview extends EventEmitter {
      * @param {string} token - Access token
      * @return {void}
      */
-    prefetch({ fileId, token, sharedLink = '', sharedLinkPassword = '', preload = false }) {
+    prefetch({ fileId, fileVersionId, token, sharedLink = '', sharedLinkPassword = '', preload = false }) {
         let file;
         let loader;
         let viewer;
 
         // Determining the viewer could throw an error
         try {
-            file = this.cache.get(fileId);
+            file = getCachedFile(this.cache, { fileId, fileVersionId });
             loader = file ? this.getLoader(file) : null;
             viewer = loader ? loader.determineViewer(file) : null;
             if (!viewer) {
@@ -555,7 +560,7 @@ class Preview extends EventEmitter {
             viewerInstance.prefetch({
                 assets: true,
                 // Prefetch preload if explicitly requested or if viewer has 'preload' option set
-                preload: preload || viewerInstance.getViewerOption('preload'),
+                preload: preload || !!viewerInstance.getViewerOption('preload'),
                 // Don't prefetch file's representation content if this is for preload
                 content: !preload
             });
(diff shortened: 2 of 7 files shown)


boxcla:

src/lib/file.js
```diff
@@ -161,17 +190,47 @@ export function cacheFile(cache, file) {
...
+ * Wrapper for uncaching a file object. We uncache both the key by file ID and the key by file version ID.
  *
  * @public
  * @param {Cache} cache - Cache instance
  * @param {Object} file - Box file or simple { id: fileId } object
  * @return {void}
  */
 export function uncacheFile(cache, file) {
-    cache.unset(file.id);
+    cache.unset(`f_${file.id}`);
```
priyajeet: Instead use

file_123
file_version_123

since thats what is used by our token service, f_ is a bit ambiguous (folder vs file).

also use a getKey() function that does this appending, otherwise hard to refactor and string duplication.

src/lib/file.js
```diff
@@ -22,12 +22,17 @@ const FILE_FIELDS = [
...
+ * @param {string} fileId - Box file ID
+ * @param {string} fileVersionId - Box file version ID
  * @param {string} apiHost - Box API base url
  * @return {string} API url
  */
-export function getURL(id, apiHost) {
-    return `${apiHost}/2.0/files/${id}?fields=${FILE_FIELDS.join(',')}`;
+export function getURL(fileId, fileVersionId, apiHost) {
+    if (fileVersionId) {
+        return `${apiHost}/2.0/files/${fileId}/versions/${fileVersionId}?fields=${FILE_FIELDS.join(',')}`;
```
priyajeet:
```js
const versionFrag = fileVersionId ? `/versions/${fileVersionId}` : ''
return `${apiHost}/2.0/files/${fileId}${versionFrag}?fields=${FILE_FIELDS.join(',')}`
```

src/lib/Preview.js
```diff
@@ -613,27 +616,53 @@ class Preview extends EventEmitter {
         // Clear any existing retry timeouts
         clearTimeout(this.retryTimeout);
 
-        // Save reference to the currently shown file, if any
+        // Save reference to the currently shown file ID, if any
         const currentFileId = this.file ? this.file.id : undefined;
 
+        // Save reference to the currently shown file version ID, if any
+        const currentFileVersionId = this.file && this.file.file_version ? this.file.file_version.id : undefined;
```
jeremypress: Is there a reason that the file_version ID needs to be nested? Could we have it alongside the ID like 

```
this.file = {
    id: 1000000
    file_version_id: 1000001
}
```

jeremypress: Looking into this more, I see we get back from the file version API `sha1` and `type`. Are we planning on using those?

tonyjin: I don't think we use them now, but this is more so we don't have to maintain a file format other than what the Box API returns us.

README.md
```diff
@@ -127,10 +127,11 @@ preview.show(fileId, accessToken, {
...
 | header | 'light' | String value of 'none' or 'dark' or 'light' that controls header visibility and theme |
 | logoUrl |  | URL of logo to show in header |
-| showAnnotations | false | Whether annotations and annotation controls are shown. This option will be overridden by viewer-specific annotation options if they are set. See [Box Annotations](https://github.com/box/box-annotations) for more details. |
+| showAnnotations | false | Whether annotations and annotation controls are shown. This option will be overridden by viewer-specific annotation options if they are set. See [Box Annotations](https://github.com/box/box-annotations) for more details |
 | showDownload | false | Whether download button is shown |
 | useHotkeys | true | Whether hotkeys (keyboard shortcuts) are enabled |
 | pauseRequireJS | false | Temporarily disables requireJS to allow Preview's third party dependencies to load |
+| fileVersions | {} | Mapping of file ID to file version ID to preview. Each file ID key should map to a valid non-current file version ID for that file. If no mapping is passed, or the file being previewed has no key/value pair in this map, we assume the current file version |
```
priyajeet: I am wondering if the API here is getting confusing and complicated. Can we just re-use the `collection` attribute and document that you can pass in a `file` object and make them work with versions too?

So if you want to show a specific version, you do

const file = { id: 123, file_version: { id: 123 } }; // well formed (but incomplete)
preview.show(file, token, { ... })

Likewise collection already takes fully formed file objects.

src/lib/Preview.js
```diff
@@ -506,22 +511,23 @@ class Preview extends EventEmitter {
...
+     * @param {string} options.fileVersionId - Box file version ID (do not also pass a file ID)
      * @param {string} options.token - Access token
      * @param {string} options.sharedLink - Shared link
      * @param {string} options.sharedLinkPassword - Shared link password
      * @param {boolean} options.preload - Is this prefetch for a preload
      * @param {string} token - Access token
      * @return {void}
      */
-    prefetch({ fileId, token, sharedLink = '', sharedLinkPassword = '', preload = false }) {
+    prefetch({ fileId, fileVersionId, token, sharedLink = '', sharedLinkPassword = '', preload = false }) {
```
priyajeet: Does prefetching need file version?
Can we simplify that if you want to view an old version, you just view that.
Is there a use case to be iteratable over older versions

src/lib/file.js
```diff
@@ -143,11 +166,32 @@ function addOriginalRepresentation(file) {
...
+ *
+ * @private
+ * @param {Object} options - Cache key options
+ * @param {string} fileId - Get cache key by file ID
+ * @param {string} fileVersionId - Get cache key by file version ID
+ * @return {string} Cache key to use
+ */
+function getCacheKey({ fileId, fileVersionId }) {
+    if (fileId) {
+        return `file_${fileId}`;
```
priyajeet: Maybe this function can be reused for tokens.js
Probably export it here or in util?

README.md
tonyjin: I think the use case of an incomplete file object (or even a fully formed file object) is not going to be very common for external developers. IMO:

```
preview.show(FILE_ID, ACCESS_TOKEN, {
    fileVersions: {
        FILE_ID: FILE_VERSION_ID
    }
});
```

is easier to understand than:

```
preview.show({ id: FILE_ID, file_version: { id: FILE_VERSION_ID } }, ACCESS_TOKEN, {
    useFileVersion: true
});
```

I think most people are going to be working with file IDs and file version IDs, and having to jigger them into our internal file format isn't going to be commonplace.

Thoughts @jeremypress @pramodsum @JustinHoldstock @DanDeMicco?

src/lib/Preview.js
tonyjin: I don't think this is going to be immediately used, I just wanted to make sure our entire code base was compatible with file versions in one pass.

README.md
DanDeMicco: @tonyjin there is definitely some overlap in the design of this, and what I am working on

The question is, for every option we support that is specific to what preview renders for a given FILE_ID, do we want to add in a new property that is of this format? My opinion is that will not be ideal.

What if we take what @priyajeet mentioned and have a well defined file type such as
```
id : string (required),
version : number (optional),
foo : string (optional),
bar : string (optional),
```

and we can then use the file object anywhere a FILE_ID string is currently used, while maintaining backwards compatibility.

For example:

The current API :
```javascript
preview.show(FILE_ID_1, ACCESS_TOKEN, {})
```
and
```javascript
preview.show(FILE_ID_1, ACCESS_TOKEN, {
   collection : [FILE_ID_1, FILE_ID_2]
})
```
will continue work

but if they want more options they can use:
```javascript
const file = { id: 123, file_version: 2 }; // well formed (but incomplete)
preview.show(file, ACCESS_TOKEN, {});
```

and if they want to apply it for collections:
```javascript
const file1 = { id: 123, file_version: 2 };
const file2 = { id: 456, file_version: 3, foo: "bar" };
preview.show(file, ACCESS_TOKEN, {
   collection : [file1, file2]
});
```

they can also be mixed and matched
```javascript
const file2 = { id: 456, file_version: 3, foo: "bar" };
preview.show(FILE_ID_1, ACCESS_TOKEN, {
   collection : [FILE_ID_1, file2]
});
```

tonyjin: Ah I see what you mean now, let me think about this.

priyajeet: I think its incorrect to assume that people will be dealing with just file ids and file version ids.
They will mostly dealing with objects that our API returns (complete or incomplete) since that's where the ids will come in the first place. As such preview interface should be able to take those objects readily along with ids (string id should signify the latest).

We should also not create our own objects as shown above, because we don't want people to learn yet another object structure. They should use existing objects that our returned by the API, so maybe instead of a flag...

User gets a file object either via /file API or more likely /folder API.

```
const file = {
    "type": "file",
    "id": "5000948880",
    "file_version": {
        "type": "file_version",
        "id": "26261748416",
        "sha1": "134b65991ed521fcfe4724b7d814ab8ded5185dc"
    },
    ...
};
preview.show(file, ...) // Uses the file and version to get in this case just the latest.
```

The above shows the file with the specified latest version. But the user may only want to see another older version(s). Since they don't know which version, they use /versions API to get versions which returns the array or versions as shown below. But now they want to preview all prior versions, they would do something like


File Version
```js
const fileVersions = [
        {
            "type": "file_version",
            "id": "46313006161",
            ...
        },
        {
            "type": "file_version",
            "id": "41550732434",
            ...
    ];

const versions = fileVersions.map((version) => Object.assign({}, file, { file_version: version });
preview.show(versions[0], token, {
    collection: versions
})
```

PR #609, opened 2018-01-29
Chore: Add download reachability checks to Document and Image viewers

jeremypress:


boxcla:

src/lib/Preview.js
```diff
@@ -688,6 +688,9 @@ class Preview extends EventEmitter {
             this.throttledMousemoveHandler
         );
 
+        // Set up the notification
+        this.ui.setupNotification();
```
jeremypress: I moved this up so it was available after the 'load' event.

src/lib/util.js
```diff
@@ -111,6 +113,65 @@ function createDownloadIframe() {
...
+}
+
+/**
+ * Sets local storage to use the default download host
+ *
+ * @public
+ * @return {void}
+ */
+export function setDownloadHostFallback() {
+    sessionStorage.setItem('download_host_fallback', 'true');
```
jeremypress: Per @DanDeMicco we use session store so that we can re-check every time the session ends. This prevents the problem where the IP address does get whitelisted but the user can't see the change

src/lib/util.js
```diff
@@ -111,6 +113,65 @@ function createDownloadIframe() {
...
+}
+
+/**
+ * Determines if the degraded download notification should be shown.
+ *
+ * @public
+ * @return {boolean} Should the notification be shown
+ */
+export function shouldShowDegradedDownloadNotification() {
```
jeremypress: We will only show the download notification if:
1. we don't have the cookie already
2. our downloads are degraded.

This means at most we can see the notification once per week, and only if we are still in a degraded state.

These will be constants in the final PR :)

src/lib/util.js
```diff
@@ -158,7 +219,20 @@ export function get(url, ...rest) {
 
     return fetch(url, { headers })
         .then(checkStatus)
-        .then(parser);
+        .then(parser)
+        .catch((e) => {
+            // Make sure we are making a dl call
+            if (isNonDefaultDownloadHost(url)) {
```
jeremypress: This checks to make sure that this URL is a dl request in addition to it not being the default. This means we don't have to also check if the session storage has the flag.

src/lib/util.js
```diff
@@ -158,7 +219,20 @@ export function get(url, ...rest) {
...
+                // Retry download with default host.
+                setDownloadHostFallback();
+                return get(url, ...rest);
+            }
+
+            /* eslint-disable no-console */
+            console.error(e);
+            /* eslint-enable no-console */
+            return Promise.resolve();
```
jeremypress: Not sure what we want to do here. We aren't really ever catching this call, so I'm just logging an error for now.  If it's not caught (e.g. by the preloader) preview will fail.

src/lib/viewers/BaseViewer.js
```diff
@@ -270,6 +274,23 @@ class BaseViewer extends EventEmitter {
...
+     * Handles a download error when using a non default host.
+     *
+     * @param {Error} err - Load error
+     * @param {string} downloadURL - download URL
+     * @return {void}
+     */
+    handleDownloadError(err, downloadURL) {
+        if (isNonDefaultDownloadHost(downloadURL)) {
+            setDownloadHostFallback();
+            this.load();
```
jeremypress: This still has a few problems as some of our UI methods called in load are not item potent, investigating further.

PR #611, opened 2018-01-30
Chore: Add download reachability checks to Document and Image viewers

jeremypress:
This also will add the check to any generic download request, such as prefetching or preloading. 

Tests on the way, other viewers will be in a separate PR.

### DIFF
--- a/src/i18n/en-US.properties
+++ b/src/i18n/en-US.properties
@@ -176,6 +176,8 @@ notification_button_default_text=Okay
 notification_annotation_point_mode=Click anywhere to add a comment to the document
 # Notification message shown when user enters drawing annotation mode
 notification_annotation_draw_mode=Press down and drag the pointer to draw on the document
+# Notification message shown when the user has a degraded preview experience due to blocked download hosts
+notification_degraded_preview=Your preview experience is degraded!
 
 # File Types
 # 360 degree video file type
--- a/src/i18n/en-US.properties
+++ b/src/i18n/en-US.properties
@@ -177,7 +177,7 @@ notification_annotation_point_mode=Click anywhere to add a comment to the docume
 # Notification message shown when user enters drawing annotation mode
 notification_annotation_draw_mode=Press down and drag the pointer to draw on the document
 # Notification message shown when the user has a degraded preview experience due to blocked download hosts
-notification_degraded_preview=Your preview experience is degraded!
+notification_degraded_preview=It looks like your connection to {1} is being blocked. We think we can make file previews in Box faster for you but we will need your firewall settings to be updated. To do that, please request your Box Admin to configure firewall settings for Box.
 
 # File Types
 # 360 degree video file type
--- a/src/lib/util.js
+++ b/src/lib/util.js
@@ -1,10 +1,12 @@
 import Uri from 'jsuri';
 import 'whatwg-fetch';
+import { DEFAULT_DOWNLOAD_HOST_PREFIX } from './constants';
 
 const HEADER_CLIENT_NAME = 'X-Box-Client-Name';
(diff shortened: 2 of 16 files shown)


boxcla:

src/i18n/en-US.properties
```diff
@@ -172,6 +172,8 @@ notification_button_default_text=Okay
 notification_annotation_point_mode=Click anywhere to add a comment to the document
 # Notification message shown when user enters drawing annotation mode
 notification_annotation_draw_mode=Press down and drag the pointer to draw on the document
+# Notification message shown when the user has a degraded preview experience due to blocked download hosts
+notification_degraded_preview=Your preview experience is degraded!
```
jeremypress: Not the final text

src/lib/viewers/image/ImageBaseViewer.js
```diff
@@ -311,23 +310,27 @@ class ImageBaseViewer extends BaseViewer {
...
-    errorHandler(err) {
-        /* eslint-disable no-console */
-        console.error(err);
-        /* eslint-enable no-console */
-
-        // Display a generic error message but log the real one
-        const error = err;
-        if (err instanceof Error) {
-            error.displayMessage = __('error_refresh');
+    handleDownloadError(err, imgUrl) {
```
JustinHoldstock: This is the same as `ImageBaseViewer.handleDownloadError()`. Can you make this a function in the parent class?

jeremypress: Do you mean in docBase vs. imageBase? The doc version of this function throws a different error (error_document vs. error_refresh, I need to fix that), and other viewers will want custom behavior here as well. See how the dash-viewer handles download errors currently.

src/lib/util.js
```diff
@@ -111,6 +116,63 @@ function createDownloadIframe() {
...
 }
 
+/**
+ * Checks if the url is a download host, but not the default download host.
+ *
+ * @public
+ * @param {string} url - The URL to check
+ * @return {boolean} - HTTP response
+ */
+export function isNonDefaultDownloadHost(url) {
```
pramodsum: Could we use something like `isCustomDefaultDownloadHost`? Feels weird to have isNon.... in the method name

src/lib/util.js
```diff
@@ -111,6 +116,63 @@ function createDownloadIframe() {
...
+}
+
+/**
+ * Stores the host in an array via localstorage so that we don't show a notification for it again
+ *
+ * @public
+ * @param {string} downloadHost - Download URL host
+ * @return {void}
+ */
+export function setDownloadHostNotificationShown(downloadHost) {
```
pramodsum: `cacheDownloadHostNotificationShown`?

src/lib/viewers/BaseViewer.js
```diff
@@ -384,13 +406,27 @@ class BaseViewer extends EventEmitter {
     }
 
     /**
-     * Handles the viewer load to potentially set up Box Annotations.
```
pramodsum: 👍

src/lib/viewers/image/ImageBaseViewer.js
pramodsum: If the only difference between this and docBase is the error message, wouldn't it make sense to use this as the default and maybe pass in an error message in each viewer? The rest of the code seems to stay the same

jeremypress: Looking at it again you're right, the only significant difference are the messages, so I'll pass those in.

src/lib/util.js
jeremypress: It's a little more extreme than a cache since we are using localStorage, but I'll think some more about the name.

PR #613, opened 2018-01-31
Fix: Annotations upgrade script

pramodsum:


boxcla:

build/upgrade_annotations.sh
```diff
@@ -50,6 +50,12 @@ reset_to_master() {
         git remote add github-upstream git@github.com:box/box-content-preview.git || return 1
     fi
 
+    # Fetch latest code with tags
+    git fetch github-upstream || return 1;
+
+    # Update to latest code on GitHub master
+    git checkout master || return 1
```
DanDeMicco: Comment here is incorrect. Shouldnt this be reset --hard upstream/master ?

DanDeMicco: Also side note you can look into the -e flag instead of the || return 1

build/upgrade_annotations.sh
```diff
@@ -50,6 +50,10 @@ reset_to_master() {
         git remote add github-upstream git@github.com:box/box-content-preview.git || return 1
     fi
 
+    # Fetch latest code
+    git fetch github-upstream || return 1;
+    git checkout master || return 1
+
     # The master branch should not have any commits
     if [[ $(git log --oneline ...github-upstream/master) != "" ]] ; then
         echo "----------------------------------------------------"
```
DanDeMicco: Change this error message to say something like you have extra commits in your master branch

pramodsum: 👍

PR #615, opened 2018-01-31
Fix: Bind close method in DocFindBar

jeremypress:
Open doesn't need to be bound as it's not used as a handler callback.


boxcla:

PR #616, opened 2018-01-31
Fix: Generate i18n json files before running prod webpack build

jeremypress:


boxcla:

PR #617, opened 2018-01-31
Fix: add extra padding to settings menu to prevent scrollbar when zoomed

DanDeMicco:
Adds 1px extra to the height of the settings menu, to prevent scrollbar from appearing in devices which have zoomed in


boxcla:

PR #619, opened 2018-01-31
New: Log preview_error message from Preview

JustinHoldstock:
This PR allows Preview to emit "preview_error" messages. Also changes some logs to reflect correct error code AND translated display message.

Listen:
```
previewInstance.on('preview_error', (err) => {
  // err.error is the error object that was created.
  // err.error.message SHOULD be the error code, if created by us, otherwise the default error message
  // err.data should be any additional data we want to log, IE) FileId, invalid file to cache  
});
```

To Use, from Preview.js (already used in `triggerError()`):
```
  // create an error with the right error code
  const myError = new Error(ERROR_CODE.myFancyNewError);

  // [OPTIONAL] Attach supplementary data. Any type.
  myError.data = objectThatCausedError;

  // invoke log method
  this.logPreviewError(myError);
```

To use in a Viewer:
```
  // It's just using triggerError()!! Yay!
  this.triggerError(new Error('relevant_error_code'));
```

Action Items:
- Preview option for disabling


src/lib/Preview.js
```diff
@@ -531,6 +535,11 @@ class Preview extends EventEmitter {
             /* eslint-disable no-console */
             console.error(`Error prefetching file ID ${fileId} - ${err}`);
             /* eslint-enable no-console */
+
+            const error = new Error(ERROR_CODE.prefetchFileId);
```
JustinHoldstock: MAybe I should abstract away, this functionality?

src/lib/RepStatus.js
```diff
@@ -106,26 +106,31 @@ class RepStatus extends EventEmitter {
      */
     handleResponse() {
         const status = RepStatus.getStatus(this.representation);
-        let errorCode;
+        const errCode = RepStatus.getErrorCode(this.representation);
+        let errorMessage;
+        let error;
```
JustinHoldstock: This section changed to allow readable error codes with translated display messages

src/lib/Preview.js
```diff
@@ -1226,6 +1246,27 @@ class Preview extends EventEmitter {
...
+     */
+    logPreviewError(error) {
+        const err = error;
+        // Make sure to strip auth
+        err.message = stripAuthFromString(error.message || '');
+        err.displayMessage = stripAuthFromString(error.displayMessage || '');
+
+        const errorLog = {
+            error: err,
+            data: error.data || ''
```
jeremypress: How would we access the error type for analytics/debugging? Would that be in the error.message? Would that potentially overwrite a message we would get such as `cannot access property hasTouch of undefined`?

JustinHoldstock: For logging purposes, we specify our own error codes (see events.js). If we have not thrown the error ourselves, setting the message to one of our error codes, the `error.message` will be that of what the browser created. 

I'm thinking a single logged event will be as follows:
```
{
  event_type: <error | metric>
  logger_session_id: <string> // ID of session 
  timestamp: <string>, // ISO Format
  file_id: <string>,
  file_version_id: <string>,
  content_type: <string>, // Content type of the file previewed
  extension: <string>,
  client_version: <string>, // preview version number
  browser_name: <string>,
  locale: <string>,                        
  event_name: <string>, // Either our error code OR browser error message
  value: <any> // Must be serializable
}
```

Note: I just realized I need to update so that we're sending more session related info, per `prevew_error` event.

jeremypress: Should we separate that out so that we have
`error.code`: our custom code we made or a generic one if the browser threw it
`error.messge`: what the browser created / more info from if required

If all of that info will be needed by whatever is posting our logs, don't we need to emit that stuff in the error events? Or do we emit that once per preview and we expect our consumer to retain that information for us.

src/lib/Preview.js
DanDeMicco: @JustinHoldstock Maybe it would make sense to make a PreviewError which extends Error

src/lib/Preview.js
```diff
@@ -1202,6 +1217,9 @@ class Preview extends EventEmitter {
      * @return {void}
      */
     triggerError(err) {
+        // Always log preview errors
+        this.logPreviewError(err);
```
jeremypress: I think there's a chance of double logging here. Above you bound to the `VIEWER_EVENT.error` event, which you then call to `triggerError`. On line 966 of your patch (this is old code),  we also bind to the `error` event coming from the viewer and call `triggerError`. This would end up calling `logPreviewError` twice, one from `handleViewerEvents` and one from `triggerError`

src/lib/Preview.js
```diff
@@ -1226,6 +1244,49 @@ class Preview extends EventEmitter {
         this.viewer.load(err);
     }
 
+    /**
+     * Create a generic log Object.
+     *
+     * @private
+     * @return {Object} Log details for viewer session and current file.
+     */
+    createLog() {
```
jeremypress: 👍

src/lib/logUtils.js
```diff
@@ -0,0 +1,63 @@
...
+        const v = c == 'x' ? r : (r & 0x3) | 0x8;
+        return v.toString(16);
+    }
+    /* eslint-enable */
+
+    return 'xxxxxxxx-xxxx-4xxx-yxxx-xxxxxxxxxxxx'.replace(/[xy]/g, generateRandom);
+}
+
+// Unique ID of this Preview session. Needs to be generated AFTER function def.
+const PREVIEW_SESSION_ID = uuidv4();
```
jeremypress: Is this safe? What if the file gets cached? We're never explicitly importing PREVIEW_SESSION_ID so when does it get executed?

JustinHoldstock: This'll only be created when the file is parsed and executed by the browser, meaning that cache will have no affect on it.

src/lib/Preview.js
JustinHoldstock: Aww hell. So, as it is, node requires that you bind to the 'error' event (listener #1), then we listen for it in `handleViewerEvents` (listener #2). The huge issue here is *when* #1 is bound can be *after* any setup errors occur. 

Does it make sense that we begin listening for errors immediately on viewer creation, and not in `handleViewerEvents`?

JustinHoldstock: We've confirmed that `handleViewerEvents()` *only* handles the generic `viewerevent`, therefore our initial thoughts are invalid. Merging

JustinHoldstock:
Note that viewer errors propagate as `preview_error`, and are not handled by `handleViewerEvent()` in `Preview.js`.

```
preview: "preview_error"
        ↑
  viewer: "error"
```

src/lib/Preview.js
```diff
@@ -1051,6 +1061,10 @@ class Preview extends EventEmitter {
             case VIEWER_EVENT.mediaEndAutoplay:
                 this.navigateRight();
                 break;
+            case VIEWER_EVENT.error:
```
jeremypress: Can we remove this case entirely?

JustinHoldstock: We cannot. This prevents a double error message from being emitted, which would double up on firing off preview_error events.

PR #621, opened 2018-01-31
Fix: No longer account for header above notification wrapper

pramodsum:
The extra padding which accounts for the header is no longer needed since the notification wrapper is positioned absolutely.


boxcla:

PR #622, opened 2018-01-31
Fix: add extra padding to settings menu to prevent scrollbar when zoomed

DanDeMicco:
Adds 1px extra to the height of the settings menu, to prevent scrollbar from appearing in devices which have zoomed in


boxcla:

src/lib/viewers/media/Settings.js
```diff
@@ -295,7 +296,10 @@ class Settings extends EventEmitter {
      * @return {void}
      */
     setMenuContainerDimensions(menu) {
-        const paddedHeight = menu.offsetHeight + SETTINGS_MENU_PADDING;
+        // When the browser is zoomed in, sometimes the height can be a floating point value rather
+        // than an integer. OffsetHeight rounds to the nearest integer, so if it rounds down, there
+        // will be a vertical scrollbar. To account for this, add in some extra padding.
+        const paddedHeight = menu.offsetHeight + SETTINGS_MENU_PADDING + EXTRA_ZOOM_PADDING;
```
jeremypress: Would using `getBoundingClientRect` be too expensive? It gives fractional values.

PR #624, opened 2018-02-02
Chore: add catch for dashViewer load

DanDeMicco:


boxcla:

src/lib/viewers/media/DashViewer.js
```diff
@@ -166,7 +166,12 @@ class DashViewer extends VideoBaseViewer {
             }
         });
         this.player.getNetworkingEngine().registerRequestFilter(this.requestFilter);
-        this.player.load(this.mediaUrl);
+        return this.player.load(this.mediaUrl).catch((error) => {
```
jeremypress: Is that not caught by this `this.player.addEventListener('error', this.shakaErrorHandler);` on line 152?

DanDeMicco: @jeremypress no it doesnt appear to be. If you force an error like a 404, it throws an uncaught exception

JustinHoldstock: Since we've added the new error metrics stuff, can you see if this is causing a "preview_error" event to be emitted, by Preview? Also, would you please change `shakeErrorHandler()` to use `this.triggerError(error)` instead of `this.emit('error', error)`?

Thanks!

src/lib/viewers/media/DashViewer.js
```diff
@@ -166,9 +166,13 @@ class DashViewer extends VideoBaseViewer {
             }
         });
         this.player.getNetworkingEngine().registerRequestFilter(this.requestFilter);
-
         this.startLoadTimer();
-        this.player.load(this.mediaUrl);
+        return this.player.load(this.mediaUrl).catch((error) => {
+            // The error is of a different format than the shakaErrorHandler expects
```
jeremypress: I noticed this as well, itt's so weird that shaka formats this error differently.

PR #626, opened 2018-02-02
Chore: No longer silently handle errors in callbacks

JustinHoldstock:
Instead of "silently" handling developer error, we're now allowing unhandled errors in event callbacks to occur.

Note: Box Content Preview will misbehave if developers are not creating callbacks with care


boxcla:

jeremypress:
Can you add documentation about being careful of a circular dependency in JSON?

PR #627, opened 2018-02-03
Chore: Refactor lodash

tonyjin:
- Import lodash modules using 'lodash/throttle' instead of individual modules
- Replace lodash.get (4.8KB) with custom implementation


boxcla:

priyajeet:
Import 'lodash' full package in package.json and then reference individual packages via import 'lodash/package'

That's the better pattern to avoid adding packages as well as let parent projects (granted that preview currently doesn't have a parent) define lodash version via peer deps.

src/lib/file.js
```diff
@@ -76,7 +77,7 @@ export function isWatermarked(file) {
  * @return {boolean} Whether or not action is permitted
  */
 export function checkPermission(file, operation) {
-    return !!file && !!file.permissions && !!file.permissions[operation];
+    return getProp(file, `permissions.${operation}`, false);
```
priyajeet: Could also use getProp(file, permissions, operation) without combined string

tonyjin: Went to custom lodash.get implementation because lodash.get is 4.8KB un-gzipped.

src/lib/util.js
```diff
@@ -837,3 +837,28 @@ export function getClosestPageToPinch(x, y, visiblePages) {
...
+/**
+ * Simplified lodash.get, this returns the value of a nested property with string path `propPath`. If that property
+ * does not exist on the object, then return `defaultValue`.
+ *
+ * @param {Object} object - Object to fetch property from
+ * @param {string} propPath - String path to property, e.g. 'b.c' if you are trying to fetch a.b.c
+ * @param {*} defaultValue - Default value if property is undefined
+ * @return {*} Value of prop if defined, defaultValue otherwise
+ */
+export function getProp(object, propPath, defaultValue) {
```
jeremypress: 🔥

src/lib/viewers/BaseViewer.js
```diff
@@ -577,14 +578,12 @@ class BaseViewer extends EventEmitter {
...
      * @param {string} option - to get
-     * @return {Object} Value of a viewer option
+     * @return {Object|undefined} Value of a viewer option
      */
     getViewerOption(option) {
         const { viewers, viewer } = this.options;
-        if (viewers && viewers[viewer.NAME]) {
-            return viewers[viewer.NAME][option];
-        }
-        return null;
```
jeremypress: Any chance changing from `null` to `undefined` here will make a difference?

tonyjin: I looked through where we use `getViewerOption()` and it doesn't look like there will be a problem.

priyajeet:
I still think you guys should have just used lodash/get. Unsure if you are getting any benefit from re-inventing the wheel with the added cost of maintenance and extra code in our codebase.

Size difference seemed negligible when I tried (left one is with your changes, right one is using lodash/get):

screen shot 2018-02-06 at 10 13 36 am

PR #629, opened 2018-02-05
Chore: Prepending annotations CSS classes with ba instead of bp

pramodsum:


boxcla:

pramodsum:
Dependent on https://github.com/box/box-annotations/pull/112

tonyjin:
These buttons are still owned by Preview so I think we should keep the `bp-` prefix here.

PR #630, opened 2018-02-05
New: Option to disable `preview` event log

### LINKED ISSUE #471: Add option to suppress "preview" event logging
It's a common use case for platforms and web applications to prefetch/preload data but not have them visible to the user. For example, we want to load multiple `ContentPreview` instances ahead of time and only log an actual preview event when the user interacts/sees the preview element on the page.

For this, we need to actually suppress the current Preview logging that happens out of the box, and talking with Tony, it should be fine since content access is always logged. Furthermore, we'd need a way to log the preview event for the content after user has interacted with it in some way.

tonyjin:
Add new Preview option `disableEventLog` to disable the client-side `preview` log to the Events API. Previews with this option enabled will not increment access stats, but all content access is still logged server-side for audit purposes.

Resolves #471.

### DIFF
--- a/README.md
+++ b/README.md
@@ -125,13 +125,14 @@ preview.show(fileId, accessToken, {
 | sharedLink |  | Shared link URL |
 | sharedLinkPassword |  | Shared link password |
 | collection |  | List of file IDs to iterate over for previewing |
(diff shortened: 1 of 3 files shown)


boxcla:

PR #631, opened 2018-02-06
Update: Rename 'pauseRequireJS' option to 'fixDependencies'

tonyjin:
Rename `pauseRequireJS` option since it's not immediately clear what you'd use it for. We retain support for the older name for backwards compatibility. Also remove unnecessary patching of `require` and `requireJS`.

### DIFF
--- a/src/lib/__tests__/util-test.js
+++ b/src/lib/__tests__/util-test.js
@@ -397,20 +397,17 @@ describe('lib/util', () => {
                 assert.ok(head.querySelector('script[src="bar"]') instanceof HTMLScriptElement);
             });
 
-            it('should clear requireJS until scripts are loaded or fail to load', () => {
-                window.define = true;
-                window.require = true;
-                window.requrejs = true;
+            it('should disable AMD until scripts are loaded or fail to load', () => {
+                const func = () => {};
+                func.amd = ['jquery'];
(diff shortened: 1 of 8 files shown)


boxcla:

tonyjin:
@jeremypress can you help test this patch with your RequireJS environment? I've tested myself but could use another pair of eyes on this, thanks.

src/lib/__tests__/util-test.js
```diff
@@ -397,20 +397,17 @@ describe('lib/util', () => {
...
+                const func = () => {};
+                func.amd = ['jquery'];
+                window.define = func;
+
+                const promise = util.loadScripts(['foo', 'bar'], true);
+                expect(define).to.equal(undefined);
 
-                expect(window.define).to.equal(undefined);
-                expect(window.require).to.equal(undefined);
-                expect(window.requirejs).to.equal(undefined);
```
tonyjin: FYI @jeremypress, this test didn't check anything before. These last 3 statements are after the return and weren't checked. The expect statements within the catch block were also not checked there was there was no exception to catch.

jeremypress:
both the new option and the legacy name worked great for me testing locally.

tonyjin:
Thanks

PR #633, opened 2018-02-06
New: Add new error case for files in a bad state

tonyjin:
Do not merge yet. Depends on a back-end change.


boxcla:

PR #637, opened 2018-02-07
Chore: Removing soon to be deprecated JWT addon in travis

jeremypress:
I've added the env variables to our travis account, so only our internal team can see them. This will mean that for now we cannot run sauce-labs on PRs, as our forks (or external contributor's forks) do not have access to these variables. 

Info on the deprecation here: https://blog.travis-ci.com/2018-01-23-jwt-addon-is-deprecated


boxcla:

PR #648, opened 2018-02-09
New: Emit preview metric messages

JustinHoldstock:
The intention of this PR is to
1) Provide setup for emitting `preview_metric` events
2) Emit a `preview_metric` event with load related metrics

## **How to use**
**Timer Mechanism**
To begin timing the load events, we call `Timer.start(tag)`, the `tag` being a unique tag for that event. I've been using the name of the event, paired with the file id of the file we're tracking. ie) `Timer.start('file_info_time_1234545')`

To stop a timer, to get an accurate reading of how long it took, call `Timer.stop(tag)` with the same tag. ie) `Timer.stop('file_info_time_1234545')`

To get timer values, call `Timer.get(tag)`, and you'll receive an object with `start`, `end`, and `elapsed` values. If the `tag` has been started and not stopped, no `end` or `elapsed` will exist. If you've never started the `tag`, then you will receive either an object with `undefined` values OR `undefined`.

I've been using `Timer.start()` either **right** before a `GET` is made on the URL, or **right** before the function call that does so. I've been using `Timer.stop()` when the asset is received, parsed, and usable. IE) For `full_document_load_time` and `DocBaseViewer`, I start when we get the PDF loading task (in `initViewer()`) and I stop inside of `finishLoading()`, (in `Preview.`), when we emit 'load'.

Finally, on `Preview.destroy()` or `Preview.finishLoading()` (guarantees we log either when the document loads **or** errored out along the way), we emit the message with the relevant props.

Relies of PR #619


src/lib/Preview.js
```diff
@@ -1269,6 +1310,95 @@ class Preview extends EventEmitter {
...
+    }
+
+    /**
+     * Message, to any listeners of Preview, that an error has occurred.
+     *
+     * @private
+     * @param {Error} error - The error that occurred.
+     * @return {void}
+     */
+    logPreviewError(error) {
```
tonyjin: Maybe different names to differentiate from `logPreviewEvent()` since this isn't actually logging anything while the other function is actually making a log call?

src/lib/Preview.js
```diff
@@ -867,6 +888,8 @@ class Preview extends EventEmitter {
         const { apiHost, queryParams } = this.options;
         const fileVersionId = this.getFileOption(this.file.id, FILE_OPTION_FILE_VERSION_ID) || '';
 
+        Timer.start(`${LOAD_METRIC.fileInfoTime}_${this.file.id}`);
```
tonyjin: Let's use `file_${this.file.id}_XXX` since we're using `file_${this.file.id}` for the internal file object key. We probably don't need to worry about file versions for now but this will disambiguate if we choose to do so in the future.

src/lib/Preview.js
```diff
@@ -1269,6 +1310,95 @@ class Preview extends EventEmitter {
...
+            Timer.get(`${LOAD_METRIC.convertTime}_${this.file.id}`) || {},
+            Timer.get(`${LOAD_METRIC.downloadResponseTime}_${this.file.id}`) || {},
+            Timer.get(`${LOAD_METRIC.fullDocumentLoadTime}_${this.file.id}`) || {}
+        ];
+        const times = timerList.map((timer) => parseInt(timer.elapsed, 10) || 0);
+        const total = times.reduce((acc, current) => acc + current);
+
+        const event = {
+            event_name: 'preview_load',
+            value: total, // Sum of all available load times.
```
tonyjin: Have you tested whether this value matches the legacy total_load_time?

src/lib/RepStatus.js
```diff
@@ -80,6 +83,8 @@ class RepStatus extends EventEmitter {
             return Promise.resolve();
         }
 
+        // Using content.url_template to guarantee uniqueness
+        Timer.start(`${LOAD_METRIC.convertTime}_${this.representation.content.url_template}`);
```
tonyjin: This is going to get started multiple times since when we hit `pending` as a status, we call this function again.

src/lib/Timer.js
```diff
@@ -0,0 +1,57 @@
...
+
+    getAll() {
+        return this.times;
+    }
+
+    reset() {
+        Object.keys(this.times).forEach((key) => {
+            delete this.times[key];
+        });
+    }
```
tonyjin: JSDoc when you get a chance :)

src/lib/Preview.js
```diff
@@ -1269,6 +1310,95 @@ class Preview extends EventEmitter {
...
+     */
+    logLoadMetrics() {
+        if (!this.file || !this.file.id) {
+            Timer.reset();
+            return;
+        }
+
+        // Do nothing if there is nothing worth logging.
+        const infoTime = Timer.get(`${LOAD_METRIC.fileInfoTime}_${this.file.id}`) || {};
+        if (!infoTime.elapsed) {
```
tonyjin: Could elapsed be 0 here? e.g. if something is already cached we don't spend any time fetching file info?

src/lib/RepStatus.js
tonyjin: Also, we don't really need to key on the url template right? We just care about 'How long did this file spend converting'. Doesn't matter which representation (or does it?)

src/lib/Preview.js
```diff
@@ -163,6 +166,13 @@ class Preview extends EventEmitter {
         this.navigateLeft = this.navigateLeft.bind(this);
         this.navigateRight = this.navigateRight.bind(this);
         this.keydownHandler = this.keydownHandler.bind(this);
+
+        this.on(PREVIEW_ERROR, (msg) => {
```
DanDeMicco: Arrow function is unnecessary here

jeremypress: Does this mean we need to remove the `console.error` from everywhere else in the code base?

src/lib/Preview.js
```diff
@@ -163,6 +166,13 @@ class Preview extends EventEmitter {
...
+            console.error(msg);
+        });
+        this.on(PREVIEW_METRIC, (msg) => {
+            console.log(msg);
```
jeremypress: Do we always want to log these?

src/lib/viewers/BaseViewer.js
```diff
@@ -260,6 +262,16 @@ class BaseViewer extends EventEmitter {
         }, this.loadTimeout);
     }
 
+    /**
+     * Start the laod timer for fullDocumentLoad event.
```
jeremypress: load

src/lib/Timer.js
```diff
@@ -0,0 +1,57 @@
...
+    get(key) {
+        return this.times[key];
+    }
+
+    getAll() {
+        return this.times;
+    }
+
+    reset() {
+        Object.keys(this.times).forEach((key) => {
```
DanDeMicco: Could you just set it back to {}?

DanDeMicco: Would there ever be a case that you would want to reset only a single timer?

src/lib/__tests__/Preview-test.js
```diff
@@ -1943,6 +1948,108 @@ describe('lib/Preview', () => {
...
+            const id = '12345';
+            preview.file = {
+                id
+            };
+
+            const log = preview.createLog();
+            expect(log.timestamp).to.exist;
+            expect(log.file_id).to.equal(id);
+            expect(log.file_version_id).to.exist;
+            expect(log.content_type).to.exsit;
```
DanDeMicco: typeo?

src/lib/Preview.js
JustinHoldstock: This section is for debugging purposes only

src/lib/Preview.js
JustinHoldstock: This section is for debugging purposes only, but yes, we will always want to log PREVIEW_METRICs

src/lib/Preview.js
JustinHoldstock: This section is for debugging purposes only

src/lib/__tests__/Preview-test.js
JustinHoldstock: I want to know why these tests passed, locally, if this was here :O

src/lib/Preview.js
JustinHoldstock: From the handful of tests I've run, this total is "closer" to Chrome's Network tab

src/lib/Preview.js
JustinHoldstock: If file info is in the cache already, will the representation asset that it belongs to be loaded into browser cache? If that's the case, we don't want to log how fast the browser can load things from cache

src/lib/RepStatus.js
JustinHoldstock: I've made it so start only starts the timer once. I'm now extracting the file id from the template URL, for a similar tag name to the rest

src/lib/viewers/BaseViewer.js
JustinHoldstock: done

src/lib/Timer.js
JustinHoldstock: I've decided to just reset the values in the struct to `undefined`. Also added case to target a single timer.

src/lib/Preview.js
JustinHoldstock: DONE!

src/lib/Timer.js
JustinHoldstock: Done

src/lib/Preview.js
```diff
@@ -1269,6 +1314,100 @@ class Preview extends EventEmitter {
         this.viewer.load(err);
     }
 
+    /**
+     * Create a generic log Object.
+     *
+     * @private
+     * @return {Object} Log details for viewer session and current file.
+     */
+    createLog() {
```
jeremypress: to follow Tony's comment below, maybe this becomes `createEvent`?

src/lib/Preview.js
JustinHoldstock: I'll rename both to `emitPreviewError()` and `emitLoadMetric()`

src/lib/Preview.js
```diff
@@ -1269,6 +1314,100 @@ class Preview extends EventEmitter {
...
+
+    /**
+     * Load metrics behave slightly different than other metrics, in that they have
+     * higher level properties that do not fit into the general purpose "value" and "event_name".
+     * A value of 0 means that the load milestone was never reached.
+     *
+     * @private
+     * @return {void}
+     */
+    logLoadMetrics() {
```
jeremypress: this and the above function both actually emit as opposed to log. How do you feel about `emitLoadMetrics` and `emitPreviewError` instead?

src/lib/Preview.js
```diff
@@ -1269,6 +1314,100 @@ class Preview extends EventEmitter {
...
+        err.code = error.code || ERROR_CODE.browserError;
+        // Make sure to strip auth, if it's a string.
+        err.message = typeof error.message === 'string' ? stripAuthFromString(error.message) : error.message;
+        err.displayMessage = typeof error.displayMessage === 'string' ? stripAuthFromString(error.displayMessage) : '';
+
+        const errorLog = {
+            error: err,
+            ...this.createLog()
+        };
+
```
jeremypress: We could move all of our console.errors here whenever we call logPreview error (or maybe in triggerError), to reduce console.error clutter.

src/lib/viewers/media/MediaBaseViewer.js
```diff
@@ -129,6 +129,7 @@ class MediaBaseViewer extends BaseViewer {
         return this.getRepStatus()
             .getPromise()
             .then(() => {
+                this.startLoadTimer();
```
jeremypress: Could we move the startLoadTimer to a part of  the promise chain in RepStatus.js? We could `then` it over there once and avoid doing it everywhere else. Basically if the repstatus is success then we want to start timing the content load.

src/lib/Preview.js
JustinHoldstock: `createLogEvent()`? Since it contains relevant logging values?

src/lib/Preview.js
JustinHoldstock: See above response to Tony's comment. It's exactly that :)

src/lib/viewers/media/MediaBaseViewer.js
JustinHoldstock: This is not meant to track time RepStatus. This begins the timer for when we setup and send the GET request for the representation. Inside of RepStatus, you'll see a timer related to representation conversion time

src/lib/Preview.js
```diff
@@ -1269,6 +1306,100 @@ class Preview extends EventEmitter {
...
+    /**
+     * Message, to any listeners of Preview, that an error has occurred.
+     *
+     * @private
+     * @param {Error} error - The error that occurred.
+     * @return {void}
+     */
+    emitPreviewError(error) {
+        const err = error;
+        // If we haven't supplied a code, then it was thrown by the browser
```
pramodsum: newline before comments

src/lib/Timer.js
```diff
@@ -0,0 +1,118 @@
...
+     * Stop the timer from running, for a tag, and calculate elapsed time. If a tag has never been started, or has already
+     * been stopped, this does nothing.
+     *
+     * @public
+     * @param {string} tag - Time structure to look up, at this string.
+     * @return {void}
+     */
+    stop(tag) {
+        const time = this.get(tag);
+        // The timer has already been stopped, or hasn't started. Don't stop it again.
```
pramodsum: newline

src/lib/Timer.js
```diff
@@ -0,0 +1,118 @@
...
+     *
+     * @public
+     * @param {string} [tag] - If provided, will reset a specific time structure associated with the tag.
+     *                         If empty, resets everything.
+     * @return {void}
+     */
+    reset(tag) {
+        if (tag) {
+            const time = this.get(tag);
+            // We don't need to clean up nothin'
```
pramodsum: newline and maybe cleanup the comment :P

src/lib/Timer.js
```diff
@@ -0,0 +1,118 @@
...
+            if (!time) {
+                return;
+            }
+
+            time.start = undefined;
+            time.end = undefined;
+            time.elapsed = undefined;
+        } else {
+            Object.keys(this.times).forEach((timeTag) => {
```
pramodsum: Object.keys(this.times).forEach(this.reset);

src/lib/viewers/BaseViewer.js
```diff
@@ -718,8 +727,13 @@ class BaseViewer extends EventEmitter {
         const viewerOptions = {};
         viewerOptions[this.options.viewer.NAME] = this.viewerConfig;
 
-        /* global BoxAnnotations */
-        const boxAnnotations = this.options.boxAnnotations || new BoxAnnotations(viewerOptions);
+        if (!global.BoxAnnotations) {
+            const error = createPreviewError(ERROR_CODE.annotationsLoadFail);
```
pramodsum: 😍

src/lib/viewers/media/MediaBaseViewer.js
jeremypress: Right, but since we want to track from once rep status comes back to success until the file is done loading, why not append to the end of rep status so we only have to add it in one place?

JustinHoldstock: Oh, I see, I misunderstood your Q. We don't want to do that because we don't want to include the time to process the JS between the rep coming back and then making the request for the rep. We ONLY want to be tracking network related times, for that metric.

src/lib/Preview.js
```diff
@@ -882,6 +899,9 @@ class Preview extends EventEmitter {
      */
     handleFileInfoResponse(response) {
         let file = response;
+        // Stop timer for file info time event.
```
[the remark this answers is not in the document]
JustinHoldstock: done

src/lib/Preview.js
```diff
@@ -1096,7 +1128,6 @@ class Preview extends EventEmitter {
         } else {
             // Bump up preview count
             this.count.success += 1;
-
```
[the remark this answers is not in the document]
JustinHoldstock: done

src/lib/Timer.js
JustinHoldstock: done

src/lib/Timer.js
JustinHoldstock: done

src/lib/Timer.js
JustinHoldstock: boom, done

src/lib/viewers/BaseViewer.js
JustinHoldstock: 🎉

src/lib/Preview.js
JustinHoldstock: done

PR #650, opened 2018-02-09
Upgrade: pdf.js 2.0.338

tonyjin:
- pdf.js 2.x updated the way workers are used. We can no longer prefetch the worker and instead must preload it
- Latest patch also respects `PDFJS.workerSrc` over any internal overrides, which previously could cause issues with RequireJS


boxcla:

tonyjin:
See #659

PR #654, opened 2018-02-13
Revert "Chore: Add download reachability checks to Document and Image viewers (#611)"

jeremypress:
Issue with this commit:
We don't have a way of determining the status code of the request, thus we cannot know if Preview failed due to a blocked host vs. any regular error (404, 500, etc). This means I was indiscriminately falling back on any error that preview saw to the default host. This is not intended/desired behavior.


boxcla:

PR #659, opened 2018-02-14
Upgrade: pdf.js 2.0.363

tonyjin:
- pdf.js 2.x updated the way workers are used. We can no longer prefetch the worker and instead must preload it
- Latest patch also respects `PDFJS.workerSrc` over any internal overrides, which previously could cause issues with RequireJS
- Fix upgrade_pdfjs.sh patch to properly list last pdf.js version


boxcla:

build/upgrade_pdfjs.sh
```diff
@@ -2,7 +2,7 @@ DOC_STATIC_ASSETS_VERSION=$(./build/current_version.sh)
 DOC_STATIC_ASSETS_PATH="src/third-party/doc/$DOC_STATIC_ASSETS_VERSION"
 
 if [ ! -d $DOC_STATIC_ASSETS_PATH ]; then
-    DOC_LATEST_STATIC_ASSETS=`ls src/third-party/doc | sort -nr | head -1`
+    DOC_LATEST_STATIC_ASSETS=`ls src/third-party/doc | sort -t "." -k1,1n -k2,2n -k3,3n | tail -1`
```
pramodsum: Our of curiosity what does this do?

tonyjin: It was sorting directories alphabetically before, so we had:

```
1.9.0
1.7.0
1.17.0
```

when we really wanted 1.17.0 to be the latest version. This was a bash snippet I found online that splits on the decimal and sorts by each one individually low -> high, then I took the last one.

Snippet was from a comment here: https://gist.github.com/loisaidasam/b1e6879f3deb495c22cc

PR #660, opened 2018-02-14
Chore: pass file id to RepStatus for easier timers

JustinHoldstock:
This will go into patch release 1.30.1


boxcla:

PR #668, opened 2018-02-16
Fix: define check

tonyjin:
Global define check was causing errors, adding a window.define check and also fixing document viewer to correctly rely on minified pdf.js.

### DIFF
--- a/src/lib/__tests__/util-test.js
+++ b/src/lib/__tests__/util-test.js
@@ -93,7 +93,7 @@ describe('lib/util', () => {
 
             return util.get(url, 'any').then((response) => {
                 expect(fetchMock.called(url)).to.be.true;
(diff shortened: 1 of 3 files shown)


boxcla:

tonyjin:
@jeremypress it did, but in an odd way. I'm not exactly clear why, but this worked:

```
const amdPresent = !!window.define && typeof define === 'function' && !!define.amd;
const defineRef = amdPresent ? define : undefined;

if (disableAMD && amdPresent) {
    define = undefined;
}
```

and this didn't:

```
const amdPresent = !!window.define && typeof define === 'function' && !!define.amd;
let defineRef;

if (disableAMD && amdPresent) {
    defineRef = define;
    define = undefined;
}
```

Something to do with hoisting?

PR #669, opened 2018-02-16
Chore: Adding download reachability checks to Downloads, Document and Image Viewers

jeremypress:


src/lib/downloadReachability.js
```diff
@@ -0,0 +1,90 @@
...
+
+/**
+ * Checks if the url is a download host, but not the default download host.
+ *
+ * @public
+ * @param {string} url - The URL to check
+ * @return {boolean} - HTTP response
+ */
+export function isCustomDownloadHost(url) {
+    return url && !url.startsWith(DEFAULT_DOWNLOAD_HOST_PREFIX) && !!url.match(/^https:\/\/dl\d+\./);
```
JustinHoldstock: Move the capture statement to a constant, then you can use it here and in places like `replaceDownloadHostWithDefaultUrl()`

src/lib/viewers/BaseViewer.js
```diff
@@ -97,6 +107,9 @@ class BaseViewer extends EventEmitter {
     /** @property {boolean} - Whether viewer is being used on a touch device */
     hasTouch;
 
+    /** @property {boolean} - Has the viewer retried downloading the content */
+    haveRetried = false;
```
JustinHoldstock: Could you be a little more descriptive with this property name? `haveRetried` is pretty vague

src/lib/viewers/BaseViewer.js
```diff
@@ -284,6 +297,31 @@ class BaseViewer extends EventEmitter {
...
+     * Handles a download error when using a non default host.
+     *
+     * @param {Error} err - Load error
+     * @param {string} downloadURL - download URL
+     * @return {void}
+     */
+    handleDownloadError(err, downloadURL) {
+        if (this.haveRetried) {
+            /* eslint-disable no-console */
+            console.error(err);
```
JustinHoldstock: If you want to put `console.error(err)` inside of the code that emits `preview_error` events, we can throttle back the number of `console` statements that exist in our codebase! Up to you :)

jeremypress: I do! But we would need to remove them all in one to avoid double printing, so I'm leaving this for now and then we can go back and do a commit where we grab them all.

src/lib/viewers/BaseViewer.js
jeremypress: `haveRetriedContentDownload`?

tonyjin:
Add your diagram to the PR description please!

src/lib/Preview.js
```diff
@@ -745,6 +767,9 @@ class Preview extends EventEmitter {
             this.throttledMousemoveHandler
         );
 
+        // Set up the notification
+        this.ui.setupNotification();
```
tonyjin: Please work with @pramodsum to make sure this doesn't muck with annotations. I remember we've had a few issues with the location of `setupNotification()` before.

jeremypress:
Diagram: download reachability

src/lib/downloadReachability.js
```diff
@@ -0,0 +1,93 @@
...
+export function isCustomDownloadHost(downloadUrl) {
+    return (
+        downloadUrl && !downloadUrl.startsWith(DEFAULT_DOWNLOAD_HOST_PREFIX) && !!downloadUrl.match(DEFAULT_HOST_REGEX)
+    );
+}
+
+/**
+ * Replaces the hostname of a download URL with the default hostname, https://dl.
+ *
+ * @private
```
tonyjin: Not private if you're exporting it :)

src/lib/viewers/image/ImageBaseViewer.js
```diff
@@ -311,23 +310,20 @@ class ImageBaseViewer extends BaseViewer {
...
-     * @private
-     * @param {Error} err - Error to handle
+     * @param {Error} err - Load error
+     * @param {string} imgUrl - URL we are using as the image src
      * @return {void}
      */
-    errorHandler(err) {
-        /* eslint-disable no-console */
-        console.error(err);
-        /* eslint-enable no-console */
```
tonyjin: Did you mean to take these out? They're stripped out on the production build but are occasionally useful on dev.

src/lib/viewers/image/ImageViewer.js
```diff
@@ -52,13 +53,14 @@ class ImageViewer extends ImageBaseViewer {
 
         const { representation, viewer } = this.options;
         const template = representation.content.url_template;
+        const downloadURL = this.createContentUrlWithAuthParams(template, viewer.ASSET);
```
tonyjin: downloadUrl* to keep consistent with other `Url` in the code

src/lib/Preview.js
```diff
@@ -479,13 +485,29 @@ class Preview extends EventEmitter {
...
+                // If we know the host is blocked, or we are already using the default,
+                // use the default.
+                openUrlInsideIframe(defaultDownloadUrl);
+            } else {
+                // Try the custom host, then check reachability
+                openUrlInsideIframe(data.download_url);
+                setDownloadReachability(data.download_url).then(() => {
+                    if (isDownloadHostBlocked()) {
+                        // If download is unreachable, try again with default
+                        openUrlInsideIframe(defaultDownloadUrl);
```
tonyjin: Is there any opportunity for a double download here? e.g. can the first download succeed but then this reachability check fail?

src/lib/downloadReachability.js
```diff
@@ -0,0 +1,93 @@
...
+ *
+ * @public
+ * @param {string} downloadUrl - Content download URL, may either be a template or an actual URL
+ * @return {boolean} - HTTP response
+ */
+export function isCustomDownloadHost(downloadUrl) {
+    return (
+        downloadUrl && !downloadUrl.startsWith(DEFAULT_DOWNLOAD_HOST_PREFIX) && !!downloadUrl.match(DEFAULT_HOST_REGEX)
+    );
+}
```
tonyjin: Does your code work properly on dev with dl.jpress.inside-box.net and dl.app.jpress.inside-box.net? I don't think it's super important for the fallback logic to work on dev since there's only one download host, but please make sure we can't get stuck into some edge case.

src/lib/downloadReachability.js
```diff
@@ -0,0 +1,93 @@
...
+
+/**
+ * Replaces the hostname of a download URL with the default hostname, https://dl.
+ *
+ * @private
+ * @param {string} downloadUrl - Content download URL, may either be a template or an actual URL
+ * @return {string} - The updated download URL
+ */
+export function replaceDownloadHostWithDefault(downloadUrl) {
+    return downloadUrl.replace(DEFAULT_HOST_REGEX, DEFAULT_DOWNLOAD_HOST_PREFIX);
```
tonyjin: Some AWS or accelerated hosts have weird patterns like `https://ec2apne1.boxcloud.com`, I'm not sure your regex would capture that

src/lib/downloadReachability.js
```diff
@@ -0,0 +1,93 @@
...
+}
+
+/**
+ * Determines what notification should be shown if needed.
+ *
+ * @public
+ * @param {string} contentTemplate - Content download URL template
+ * @return {string|undefined} Should the notification be shown
+ */
+export function downloadNotificationToShow(contentTemplate) {
```
tonyjin: This function signature is a bit confusing to me. The JSDoc for returns implies it should return a {boolean} and JSDoc description implies this should return a notification ID of some kind.

src/lib/viewers/image/ImageBaseViewer.js
jeremypress: yep, I forgot these are stripped out of the prod build, thanks.

src/lib/Preview.js
jeremypress: I don't believe so. The only way the reachability check fails is if the network request itself fails. This is different from any bad status `500`, `404` etc. which will be treated as "successes" because we get a valid HTTP status.

src/lib/downloadReachability.js
jeremypress: Yes this works on dev since I only check the DL, but this will have to change due to your comment below.

src/lib/downloadReachability.js
jeremypress: This used to return a boolean, I'll update the return description.

src/lib/downloadReachability.js
jeremypress: womp, that is good information to know. Do you think its safe to check `startsWith(dl#.) || endsWith('boxcloud.com')`? Excuse the hand-wavey pseudocode.

src/lib/events.js
```diff
@@ -53,3 +54,9 @@ export const LOAD_METRIC = {
     downloadResponseTime: 'download_response_time', // Time it took for TTFB when requesting a rep.
     fullDocumentLoadTime: 'full_document_load_time' // How long it took to load the document so it could be previewed.
 };
+
+// Events around download reachability
+export const DOWNLOAD_REACHABILITY_METRICS = {
```
JustinHoldstock: 🎉

src/lib/viewers/BaseViewer.js
```diff
@@ -286,6 +300,31 @@ class BaseViewer extends EventEmitter {
...
+    /**
+     * Handles a download error when using a non default host.
+     *
+     * @param {Error} err - Load error
+     * @param {string} downloadURL - download URL
+     * @return {void}
+     */
+    handleDownloadError(err, downloadURL) {
+        if (this.hasRetriedContentDownload) {
+            this.triggerError(err);
```
JustinHoldstock: You may want to make sure that this error complies with our new `PreviewError`

src/lib/viewers/image/ImageBaseViewer.js
```diff
@@ -82,7 +81,9 @@ class ImageBaseViewer extends BaseViewer {
                 this.loaded = true;
                 this.emit(VIEWER_EVENT.load);
             })
-            .catch(this.errorHandler);
+            .catch(() => {
+                // No-op, this prmise should always resolve
```
JustinHoldstock: Seems funky to not track an error if it occurs. Can we remove this, and allow for an uncaught error, then (if it's possible for this to error)?

src/lib/__tests__/downloadReachability-test.js
```diff
@@ -0,0 +1,130 @@
...
+            const result = dr.setDownloadHostNotificationShown(blockedHost);
+
+            const shownHostsArr = JSON.parse(localStorage.getItem(DOWNLOAD_NOTIFICATION_SHOWN_KEY)) || [];
+            expect(shownHostsArr).to.contain('https://dl3.boxcloud.com');
+        });
+    });
+        
+    describe('setDownloadHostFallback()', () => {
+        it('should set the download host fallback key to be true', () => {
+            expect(sessionStorage.getItem(DOWNLOAD_HOST_FALLBACK_KEY) === 'true').to.be.false;
```
DanDeMicco: This looks a bit weird

expect(sessionStorage.getItem(DOWNLOAD_HOST_FALLBACK_KEY)).to.equal('true')

src/lib/Preview.js
tonyjin: Were you able to resolve this @jeremypress @pramodsum?

PR #674, opened 2018-02-20
Chore: Refactor preview errors

tonyjin:
Create custom `PreviewError` class and use when possible. `PreviewError` uses standardized error codes that be used to filter and group categories of errors. It also contains a display message that can be shown to the end user, an optional details object that can contain relevant properties like file ID(s), file object, representation, link information, etc., and lastly an original error message if `PreviewError` is wrapped around a native `Error`.

Misc changes:
- Switch from karma-chai-sinon to karma-sinon-chai (other package seems to be unmaintained)
- Fix some tests that used deprecated chai patterns (e.g. be.a.number, be.a.function)
- Use // eslint-disable-next-line


boxcla:

tonyjin:
Do not merge, work in progress.

src/lib/PreviewError.js
```diff
@@ -0,0 +1,29 @@
...
+     * @param {string} [message] - Error message, should be translated if displayed to the user
+     * @param {Object} [details] - Optional error details
+     * @param {Object} [details.originalError] - If error originates from a native error, this will refer to that error
+     * @return {PreviewError} PreviewError instance
+     */
+    constructor(code, message, details) {
+        super();
+
+        this.code = code;
+        this.message = message || code;
```
jeremypress: Should we use the generic message as a fallback here? This could cleanup logic in the error viewer.

tonyjin: Good idea.

src/lib/PreviewError.js
```diff
@@ -0,0 +1,29 @@
...
+    /** @property {string} - Translated error message if error is shown to user */
+    message;
+
+    /**
+     * [constructor]
+     *
+     * @param {string} code - English error code
+     * @param {string} [message] - Error message, should be translated if displayed to the user
+     * @param {Object} [details] - Optional error details
+     * @param {Object} [details.originalError] - If error originates from a native error, this will refer to that error
```
JustinHoldstock: I've started a conversation on Slack about copying the native error properties, since `PreviewError` extends `Error`

.babelrc
```diff
@@ -7,7 +6,10 @@
     ],
     "plugins": [
         "transform-class-properties",
-        "transform-object-rest-spread"
+        "transform-object-rest-spread",
+        ["babel-plugin-transform-builtin-extend", {
+            "globals": ["Error"]
+        }]
```
tonyjin: See: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error

```Babel and other transpilers will not correctly handle the following code without additional configuration.```

File: src/lib/Preview.js [hunk 4]
```diff
@@ -674,7 +676,10 @@ class Preview extends EventEmitter {
             }
             /* eslint-enable camelcase */
         } else {
-            throw new Error(
+            throw new PreviewError(
+                ERROR_CODE.BAD_INPUT,
+                __('error_generic'),
+                {},
```
tonyjin: I'm throwing errors when possible as to not call `triggerError()` manually too many times.

src/lib/Preview.js
```diff
@@ -959,7 +963,12 @@ class Preview extends EventEmitter {
                 this.reload(true); // Reload viewer without fetching updated file info from server
             }
         } catch (err) {
-            this.triggerError(err instanceof Error ? err : new Error(__('error_refresh')));
+            const error =
+                err instanceof PreviewError
+                    ? err
+                    : new PreviewError(ERROR_CODE.LOAD_VIEWER, __('error_refresh'), {}, err.message);
+
+            this.triggerError(error);
```
tonyjin: This will show the `PreviewError` if that was thrown elsewhere, otherwise it constructs a new, more generic `PreviewError`

src/lib/Preview.js
```diff
@@ -1344,20 +1355,22 @@ class Preview extends EventEmitter {
...
         // If we haven't supplied a code, then it was thrown by the browser
-        err.code = error.code || ERROR_CODE.browserError;
-        // Make sure to strip auth, if it's a string.
-        err.message = typeof error.message === 'string' ? stripAuthFromString(error.message) : error.message;
-        err.displayMessage = typeof error.displayMessage === 'string' ? stripAuthFromString(error.displayMessage) : '';
+        sanitizedError.code = error.code || ERROR_CODE.BROWSER_GENERIC;
+
+        // Strip auth from messages
+        const { displayMessage, message } = sanitizedError;
+        sanitizedError.displayMessage = stripAuthFromString(displayMessage);
```
tonyjin: I moved the typeof string check to `stripAuthFromString`

PR #675, opened 2018-02-21
Update: Modify file info retry logic

tonyjin:
Retry a maximum of 3 times and use full jitter instead of exponential backoff.


boxcla:

src/lib/Preview.js
```diff
@@ -1253,8 +1253,8 @@ class Preview extends EventEmitter {
 
         clearTimeout(this.retryTimeout);
 
-        // Respect 'Retry-After' header if present, otherwise retry using exponential backoff
-        let timeoutMs = 2 ** this.retryCount * MS_IN_S;
+        // Respect 'Retry-After' header if present, otherwise retry full jitter
+        let timeoutMs = Math.random() * (2 ** this.retryCount * MS_IN_S);
```
jeremypress: Math.random() is potentially 0, would that be a problem if we very rarely retry right away?

tonyjin: Don't think so.

PR #679, opened 2018-02-22
Chore: Allow viewers to emit metrics

jeremypress:


boxcla:

src/lib/Preview.js
```diff
@@ -1094,6 +1095,23 @@ class Preview extends EventEmitter {
...
+     * Handle metrics emitted by the viewer
+     *
+     * @private
+     * @param {Object} [data] - Viewer metric data
+     * @return {void}
+     */
+    handleViewerMetrics(data) {
+        const formattedEvent = {
+            event_name: data.event,
+            ...data.data,
```
JustinHoldstock: All loggable events require a `value` property. `value` pertains to additional info about the event. In this case, you can assign `data.data` to the value prop.

ie)
```
const formattedEvent = {
  event_name: data.event,
  value: data.data,
  ...this.createLogEvent()
}
```

src/lib/viewers/BaseViewer.js
```diff
@@ -493,6 +493,22 @@ class BaseViewer extends EventEmitter {
...
+     * @protected
+     * @emits metric
+     * @param {string} event - Event name
+     * @param {Object} data - Event data
+     * @return {void}
+     */
+    emitMetric(event, data) {
+        super.emit(VIEWER_EVENT.metric, {
+            event,
+            ...data
```
JustinHoldstock: I don't think we you want use spread syntax here. you want the `data` prop on the object emitted to equal `data`.

PR #685, opened 2018-02-26
Fix: Cleanup baseViewer.hasAnnotationPermissions() and tests

pramodsum:


boxcla:

src/lib/viewers/BaseViewer.js
```diff
@@ -811,8 +811,10 @@ class BaseViewer extends EventEmitter {
             return false;
         }
 
-        const canViewAnnotations = !!(permissions.can_view_annotations_all || permissions.can_view_annotations_self);
-        return !permissions.can_annotate && !canViewAnnotations;
+        const { can_annotate, can_view_annotations_all, can_view_annotations_self } = permissions;
+
+        // eslint-disable-next-line
+        return can_annotate || can_view_annotations_all || can_view_annotations_self;
```
jeremypress: You were forcing a boolean before with !! on line 814, is ok to return undefined now if no permissions are passed in the object?

pramodsum: Forcing it again now just to be sure :)

pramodsum:
@DanDeMicco what do you mean?

tonyjin:
This change lgtm, but I'm curious why `hasAnnotationPermissions()` worked before. It looked like it was returning true if you could not leave annotations and you could not view annotations, false otherwise.

This patch changes it to return true if you can either leave annotations, view your own, or view all annotations. I'd like some investigation into what this was doing before before we merge this.

pramodsum:
@tonyjin looks like this was always returning true because we were passing in the file object rather than the expected permissions object. There are a lot more permissions checks within the annotations codebase which is why this may method may not have actually surfaced a bug but yea the method was never working correctly 👎

PR #687, opened 2018-02-27
Fix: Downgrade pdf.js

tonyjin:
Downgrade because of https://github.com/mozilla/pdf.js/issues/9514


boxcla:

PR #690, opened 2018-02-28
Fix: move startAt from constructor to setup

DanDeMicco:
Changed to setup due to prefetch. Refactored tests and getStart* to avoid changing a bunch of test setup


boxcla:

PR #692, opened 2018-03-01
Chore: add unit test for prefetching actual viewers

DanDeMicco:
Prettier formatted this file quite a lot,  The actual changes are separating out the current ```prefetchViewers``` tests into a nested describe (line 583), and a new test which has an unstubbed ```getViewers``` (line 629).

This test fails before #690 and now passes

### DIFF
--- a/src/lib/__tests__/Preview-test.js
+++ b/src/lib/__tests__/Preview-test.js
@@ -119,7 +119,6 @@ describe('lib/Preview', () => {
             });
         });
 
-
         it('should set the preview options with function token', () => {
             const foo = () => {};
             preview.show('123', foo, { viewer: 'viewer' });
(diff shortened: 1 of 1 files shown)


boxcla:

src/lib/__tests__/Preview-test.js
```diff
@@ -552,48 +580,57 @@ describe('lib/Preview', () => {
...
-                {
-                    NAME: 'viewer2',
-                    CONSTRUCTOR: stubViewer
-                },
-                {
-                    NAME: 'viewer3',
-                    CONSTRUCTOR: stubViewer
-                }
-            ];
+            it('should prefetch no viewers if no viewer names are specified', () => {
```
pramodsum: 'should not prefetch viewers if none are specified'

DanDeMicco: This was test was not modified by me, I only added the nested describe and ```should not throw when prefetching the viewers```

PR #693, opened 2018-03-01
Create load timeout error and guard trigger error

JustinHoldstock:
Adding better guarding around empty `err` param, or string param, as well as providing a PreviewError upon viewer load timeout


boxcla:

src/lib/viewers/BaseViewer.js
```diff
@@ -300,10 +302,17 @@ class BaseViewer extends EventEmitter {
...
-            err instanceof PreviewError
-                ? err
-                : new PreviewError(ERROR_CODE.LOAD_VIEWER, __('error_refresh'), {}, err.message);
+        let error;
+
+        if (err instanceof PreviewError) {
+            error = err;
+        } else {
+            let message = err || '';
+            message = typeof message === 'string' ? message : err.message;
```
tonyjin: Do we have any code that triggers an error with just a string?

JustinHoldstock: I'm adding this in as a guard until we're 100% certain we aren't allowing `triggerError()` to accept a string. I couldn't find any instances where we do, but there are quite a few places that call triggerError(), and I don't want anything to blow up.

jeremypress: I'm fine with punishing a bad use of triggerError here. If you're not passing in a PreviewError then you get a generic error load viewer code. 

If we don't have this usage anywhere right now, we should prevent it before anyone starts doing that.

src/lib/viewers/__tests__/BaseViewer-test.js
```diff
@@ -168,6 +168,22 @@ describe('lib/viewers/BaseViewer', () => {
             // Test cleanup
             clearTimeout(base.loadTimeoutId);
         });
+
+        it('should trigger an error if it takes too long to load', () => {
```
jeremypress: should trigger an error if the viewer times out

src/lib/viewers/BaseViewer.js
JustinHoldstock: That sounds fine by me! I'll remove the string case from this, and always expect an error to be passed in

PR #695, opened 2018-03-01
Chore: support for numeric file id

DanDeMicco:
Adds support for passing fileId's as a number. It currently breaks if a number is passed instead of a string

### DIFF
--- a/README.md
+++ b/README.md
@@ -96,7 +96,7 @@ Initialization
 var preview = new Box.Preview();
 preview.show(fileId, accessToken, { options });
 ```
-Where `fileId` is a string or number `Box_File` id and `accessToken` is a Box API access token.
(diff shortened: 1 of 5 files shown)


boxcla:

README.md
```diff
@@ -96,7 +96,7 @@ Initialization
 var preview = new Box.Preview();
 preview.show(fileId, accessToken, { options });
 ```
-Where `fileId` is a string `Box_File` id and `accessToken` is a Box API access token.
+Where `fileId` is a string or number `Box_File` id and `accessToken` is a Box API access token.
```
tonyjin: Let's not change the documentation. We can add some more error recovery but we should recommend devs to still send strings.

PR #696, opened 2018-03-02
Fix: Preview error viewer in platform and IE11

tonyjin:
- Contact Us redirect to Box support page should only show up in the Box web application
- Fix non-wrapping error text in IE11
- Consolidate download checks into util function
- Move file-specific util method into file.js


boxcla:

src/lib/util.js
```diff
@@ -887,3 +875,12 @@ export function getProp(object, propPath, defaultValue) {
...
     return value !== undefined ? value : defaultValue;
 }
+
+/**
+ * Returns whether Preview is running inside the Box web application.
+ *
+ * @return {boolean} Is Preview running in the Box WebApp
+ */
+export function inBoxWebApp() {
+    return (window.location.hostname || '').indexOf('app.box.com') !== -1;
```
jeremypress: We have an 'includes' polyfill if you want to use that
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/includes

src/lib/__tests__/util-test.js
```diff
@@ -812,4 +786,12 @@ describe('lib/util', () => {
             expect(util.getProp(a, 'foo.bar', 'default')).to.equal('default');
         });
     });
+
+    describe('inBoxWebApp()', () => {
+        it('should return false when window location is not a Box domain', () => {
+            expect(util.inBoxWebApp()).to.be.false;
+        });
+
+        /* No quick way to stub window.location.hostname to test for truth case */
```
jeremypress: According to SO this works for href, maybe will work for hostname too?
```
global.window = {
       location: {
           href: {
               value: 'foo'
           }
       }
   }
```

tonyjin: I found that thread and wasted ~an hour yesterday trying to get the test to work unsuccessfully. Tried a few things including Object.defineProperty, global.window (which wouldn't work in the browser anyways I think), and a few others.

src/lib/util.js
tonyjin: Actually I'm gonna hold off adding a polyfill for now. The existing polyfill is for Array.prototype.includes. If I included the String polyfill, I'd want to update all references of `indexOf` and that looks like quite a few files.

DanDeMicco: Seems like we follow an isSomething format for boolean returning functions. Maybe isBoxWebApp, or the inverse function isPlatformApp

src/lib/__tests__/Preview-test.js
```diff
@@ -1721,64 +1690,39 @@ describe('lib/Preview', () => {
...
+        it('should show print button if print is supported', () => {
+            stubs.checkFeature.withArgs(sinon.match.any, 'print').returns(true);
             stubs.canDownload.returns(true);
-
             preview.finishLoading();
-            expect(stubs.showDownloadButton).to.be.called;
+            expect(stubs.showPrintButton).to.be.called;
+        });
 
+        it('should show not print button if print is not supported', () => {
```
DanDeMicco: should not show

src/lib/__tests__/Preview-test.js
```diff
@@ -1721,64 +1690,39 @@ describe('lib/Preview', () => {
...
-
             preview.finishLoading();
-            expect(stubs.showDownloadButton).to.be.calledWith(preview.download);
+            expect(stubs.showPrintButton).to.not.be.called;
         });
 
-        it('should show print button if print is supported', () => {
-            stubs.checkFeature.returns(false);
-
+        it('should show not print button if file can\'t be downloaded', () => {
```
DanDeMicco: should not show

PR #699, opened 2018-03-05
Fix: Removing reliance on String includes

jeremypress:
This would break on IE11 in the case that we are using a certain kind of custom download host. We'll eventually include the Polyfill and update all instances of indexOf.


boxcla:

PR #700, opened 2018-03-05
Chore: use handViewerMetrics() to log successful/failed preview

JustinHoldstock:


boxcla:

src/lib/Preview.js
```diff
@@ -1187,6 +1187,12 @@ class Preview extends EventEmitter {
                 file: this.file
             });
 
+            // Explicit preview failure
+            this.handleViewerMetrics({
```
jeremypress: I'm still not sure if we want to separate success/error from the `load` metric. If that's the direction you want to go then the change looks fine.. Also, if the old table isn't going away, why not just use that for tracking?

JustinHoldstock: The legacy table isn't going away *right now*.  

Note that `load_metric` does not signify successful preview. It can be emitted on a fatal error that occurs, when the user navigates away from preview before being loaded, etc. `load_metric` is used to track network response times. 

I'm on the fence about adding another column for `success`, as it defeats the purpose of a general data structure for events. Also note that with this method of tracking success, we're just querying for `event_name == "success" AND value == "true"`

tonyjin: Is this change so we can eventually be entirely on new events? I'm guessing that's gonna be a more long-term transition since we'd have to get the business off of `auto_log_preview` and make sure all useful data in there has been moved over.

What do you imagine we'll need the event_name = success && value = false case for?

jeremypress: yeah should we just do event name 'success' or event name 'failure'? or maybe rename it to something like load_status and then we could rely on true or false?

PR #702, opened 2018-03-07
Chore: Add util tests and fix file tests

tonyjin:
Add Babel Rewire plugin so we can properly test `inBoxWebApp()` and fix `canDownload()` tests

Depends on https://github.com/box/box-content-preview/pull/696


boxcla:

src/lib/__tests__/file-test.js
```diff
@@ -324,4 +327,64 @@ describe('lib/file', () => {
...
+        });
+
+        [
+            [false, false, false, false, false],
+            [false, false, false, true, false],
+            [false, false, true, false, false],
+            [false, true, false, false, false],
+            [true, false, false, false, false],
+            [true, true, true, true, true],
+        ].forEach(([isDownloadable, isDownloadEnabled, havePermission, isBrowserSupported, expectedResult]) => {
```
tonyjin: I didn't have the [] here at first, but the test still passed since an array for the first param and undefined for all of the others still passed the test...

src/lib/util.js
```diff
@@ -905,3 +893,23 @@ export function isValidFileId(fileId) {
...
 }
+
+/**
+ * Returns window hostname
+ *
+ * @return {string} Window hostname
+ */
+export function getHostname() {
+    return window.location.hostname || '';
+}
```
tonyjin: Now hostname is stub-able for tests

tonyjin:
Huh, looks like Rewire may be causing a bunch of other issues with our tests... Will re-visit after #696 is merged.

PR #705, opened 2018-03-07
Chore: Adding download reachability checks to other viewers

jeremypress:
Tests on the way, adding checks for media and text viewers.

### DIFF
--- a/src/lib/viewers/media/DashViewer.js
+++ b/src/lib/viewers/media/DashViewer.js
@@ -16,6 +16,9 @@ const MANIFEST = 'manifest.mpd';
 const DEFAULT_VIDEO_WIDTH_PX = 854;
 const DEFAULT_VIDEO_HEIGHT_PX = 480;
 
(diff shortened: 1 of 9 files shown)


boxcla:

src/lib/viewers/media/DashViewer.js
```diff
@@ -16,6 +16,9 @@ const MANIFEST = 'manifest.mpd';
 const DEFAULT_VIDEO_WIDTH_PX = 854;
 const DEFAULT_VIDEO_HEIGHT_PX = 480;
 
+const SHAKA_CODE_UNEXPECTED_NETWORK_FAILURE = 1002;
```
DanDeMicco: You can get the error codes from shaka.util.Error.Code. So you should be able to do shaka.util.Error.Code.HTTP_ERROR instead of hard coding the number. Might be more future proof? docs at https://shaka-player-demo.appspot.com/docs/api/lib_util_error.js.html

PR #709, opened 2018-03-08
Chore: Add util tests and fix file tests

tonyjin:
- Move hostname check to Browser function so we can test `isBoxWebApp()`
- Fix incorrect `file.canDownload()` tests, they were missing brackets
- Add missing peer dependency
- Clean up package.json and run `yarn upgrade`

### DIFF
--- a/package.json
+++ b/package.json
@@ -75,6 +75,7 @@
     "optimize-css-assets-webpack-plugin": "^3.2.0",
     "phantomjs-prebuilt": "^2.1.16",
     "postcss-loader": "^2.0.9",
+    "postcss-sass": "^0.3.0",
     "prettier": "^1.8.2",
     "prettier-eslint-cli": "^4.4.2",
     "raw-loader": "^0.5.1",
@@ -130,6 +131,5 @@
       "prettier-eslint --print-width 120 --single-quote --tab-width 4 --write",
       "git add"
     ]
-  },
-  "dependencies": {}
+  }
 }
--- a/src/lib/util.js
+++ b/src/lib/util.js
@@ -1,6 +1,7 @@
 import Uri from 'jsuri';
 import 'whatwg-fetch';
 
+import Location from './Location';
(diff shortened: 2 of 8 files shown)


package.json
```diff
@@ -75,6 +75,7 @@
     "optimize-css-assets-webpack-plugin": "^3.2.0",
     "phantomjs-prebuilt": "^2.1.16",
     "postcss-loader": "^2.0.9",
+    "postcss-sass": "^0.3.0",
```
DanDeMicco: Hmm, any reason why a dependency got added here?

src/lib/util.js
```diff
@@ -901,5 +902,6 @@ export function isValidFileId(fileId) {
  * @return {boolean} Is Preview running in the Box WebApp
  */
 export function isBoxWebApp() {
-    return (window.location.hostname || '').indexOf('app.box.com') !== -1;
+    const boxHostnameRegex = /(app|ent)\.(box\.com|boxcn\.net|boxenterprise\.net)/;
+    return !!Browser.getHostname().match(boxHostnameRegex);
```
DanDeMicco: Can you use regex.test here since you are returning a boolean?

src/lib/Browser.js
```diff
@@ -364,6 +364,15 @@ class Browser {
             }
         };
     }
+
+    /**
+     * Returns window hostname
+     *
+     * @return {string} Window hostname
+     */
+    static getHostname() {
```
DanDeMicco: This seems a bit out of place in this file to me. There is a few instances of window.location in the codebase, maybe create a Url class and we can refactor?

jeremypress: If there's nothing else needed for URL related stuff, maybe just add it to util since it's being used in a few places?

tonyjin: I specifically moved this here (instead of including in util.js) so I could test `isBoxWebApp()` without refactoring util to a class with all static functions or adding Babel Rewire. I'll move this to a new URL class per Daniel's recommendation.

PR #711, opened 2018-03-08
Fix: Avoid use of Number.isFinite()

jeremypress:
Number.isFinite() is not supported in IE, and all it does is check that the input is a number and call the global isFinite, which is supported everywhere. 

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/isFinite

### DIFF
--- a/src/lib/viewers/media/MediaBaseViewer.js
+++ b/src/lib/viewers/media/MediaBaseViewer.js
@@ -540,7 +540,7 @@ class MediaBaseViewer extends BaseViewer {
      * @return {boolean} - true if time is valid
      */
     isValidTime(time) {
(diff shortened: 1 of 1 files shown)


boxcla:

src/lib/viewers/media/MediaBaseViewer.js
```diff
@@ -540,7 +540,8 @@ class MediaBaseViewer extends BaseViewer {
      * @return {boolean} - true if time is valid
      */
     isValidTime(time) {
-        return typeof time === 'number' && Number.isFinite(time) && time >= 0 && time <= this.mediaEl.duration;
+        // eslint-disable-next-line
+        return typeof time === 'number' && isFinite(time) && time >= 0 && time <= this.mediaEl.duration;
```
priyajeet: remove isFinite, since we are doing bounds check

PR #713, opened 2018-03-12
Chore: update status for jobs in saucelabs

DanDeMicco:
This adds job status reporting to saucelabs so it will be visible on the saucelabs dashboard if jobs and builds were successful or failed


boxcla:

PR #714, opened 2018-03-12
Fix: startAt should disable document preload

DanDeMicco:
Fixes an issue where document preload was being shown when startAt was set

### DIFF
--- a/src/lib/viewers/doc/DocBaseViewer.js
+++ b/src/lib/viewers/doc/DocBaseViewer.js
@@ -228,9 +228,13 @@ class DocBaseViewer extends BaseViewer {
         const { file } = this.options;
         const isWatermarked = file && file.watermark_info && file.watermark_info.is_watermarked;
 
-        // Don't show preload if there's a cached page since preloads are only for the 1st page
+        // Don't show preload if there's a cached page or startAt is set and > 1 since preloads are only for the 1st page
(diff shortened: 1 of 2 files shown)


boxcla:

src/lib/viewers/doc/DocBaseViewer.js
```diff
@@ -228,9 +228,13 @@ class DocBaseViewer extends BaseViewer {
...
         const isWatermarked = file && file.watermark_info && file.watermark_info.is_watermarked;
 
-        // Don't show preload if there's a cached page since preloads are only for the 1st page
+        // Don't show preload if there's a cached page or startAt is set and > 1 since preloads are only for the 1st page
         // Also don't show preloads for watermarked files
-        if (!this.preloader || isWatermarked || this.getCachedPage() !== 1) {
+        if (
+            !this.preloader ||
+            isWatermarked ||
+            ((this.startPageNum && this.startPageNum !== 1) || this.getCachedPage() !== 1)
```
jeremypress: I forget, are we always initializing this? Or could the normal case (where nothing is passed in and nothing is cached, be `undefined`.

DanDeMicco: The value could be undefined if nothing is passed in

PR #719, opened 2018-03-14
Fix: Allow annotation text to be selectable

tonyjin:
Old CSS was prevent user selection of anything that wasn't a text layer div to prevent weird selection edge cases. This patch relaxes that to allow selection of `<span>`, `<p>` & `<input>` so annotation dialog text can be selected (they use `<p>`).

### DIFF
--- a/src/lib/viewers/doc/_docBase.scss
+++ b/src/lib/viewers/doc/_docBase.scss
@@ -79,6 +79,13 @@
         }
     }
 
+    // Allow other text to be selectable
+    input,
+    p,
+    span {
+        user-select: text;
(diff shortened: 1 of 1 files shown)


boxcla:

src/lib/viewers/doc/_docBase.scss
```diff
@@ -79,6 +79,12 @@
         }
     }
 
+    // Allow other text to be selectable
+    p,
+    span {
```
jeremypress: Can you add inputs too? Right now you can't highlight text that you're currently entering.

PR #721, opened 2018-03-15
New: Watermarking preferences

tonyjin:
Add support for two new watermarking preferences: `previewWMPref` and `downloadWM`. `previewWMPref` allows configuration of the preview experience for watermarked files (do you see watermarked versions of the file or not) and `downloadWM` controls whether watermarked files or the original is downloaded via the download button and download() API.

Also fixed some flaky tests.


boxcla:

src/lib/Preview.js
```diff
@@ -488,31 +484,40 @@ class Preview extends EventEmitter {
      * @return {void}
      */
     download() {
-        const { apiHost, queryParams } = this.options;
-
+        const downloadErrorMsg = __('notification_cannot_download');
         if (!canDownload(this.file, this.options)) {
+            this.ui.showNotification(downloadErrorMsg);
```
jeremypress: Why would we show a button if they can't download? If we don't know till reps come back this will get interesting...

src/lib/Preview.js
```diff
@@ -488,31 +484,40 @@ class Preview extends EventEmitter {
...
+        // the correct content disposition, and download
+        if (shouldDownloadWM(this.file, this.options)) {
+            const contentUrlTemplate = getProp(this.viewer.getRepresentation(), 'content.url_template');
+            if (!contentUrlTemplate) {
+                this.ui.showNotification(downloadErrorMsg);
+                return;
             }
-        });
+
+            queryParams.response_content_disposition_type = 'attachment';
```
jeremypress: What does setting this to attachment accomplish?

src/lib/Preview.js
```diff
@@ -857,6 +862,22 @@ class Preview extends EventEmitter {
...
+        // Sets how previews of watermarked files behave.
+        // 'all' - Forces watermarked previews of supported file types regardless of collaboration or permission level,
+        //         except for `Uploader`, which cannot preview.
+        // 'any' - The default watermarking behavior in the Box Web Application. If the file type supports
+        //         watermarking, all users except for those collaborated as an `Uploader` will see a watermarked
+        //         preview. If the file type cannot be watermarked, users will see a non-watermarked preview if they
+        //         are at least a `Viewer-Uploader` and no preview otherwise.
+        // 'none' - Forces non-watermarked previews. If the file type cannot be watermarked or the user is not at least
+        //          a `Viewer-Uploader`, no preview is shown.
+        this.options.previewWMPref = options.previewWMPref || 'any';
```
jeremypress: I don't think the "Pref" at the end adds much

src/lib/viewers/BaseViewer.js
```diff
@@ -1030,6 +1030,15 @@ class BaseViewer extends EventEmitter {
             })
         );
     }
+
+    /**
+     * Returns the representation used for Preview.
+     *
+     * @return {Object} Box representation used/to be used by Preview
+     */
+    getRepresentation() {
```
jeremypress: 👍 good call

src/lib/Preview.js
tonyjin: Wanted to distinguish some way between previewWM and downloadWM since one takes a string and one takes a boolean, but I'm open to dropping `Pref`

src/lib/Preview.js
tonyjin: This is a public method so one could call `preview.download()` directly

src/lib/Preview.js
tonyjin: You had this in your original PR to force [Content Disposition](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Disposition) to `attachment` instead of the default (`inline`?). I assume this came out of a conversation between you and Albert.

DanDeMicco: If you decide to keep this line, I think you should do an object spread / Object.assign for this. If you add this here, you will be modifying the options the user passes in, and all further uses of ```this.options.queryParams``` will have ```response_content_disposition_type``` which may not be intended

src/lib/Preview.js
jeremypress: sounds good!

src/lib/Preview.js
jeremypress: ah makes sense, make the download one a boolean like verb by adding should or force to the front?

src/lib/Preview.js
jeremypress: @tonyjin yeah I think that came from him as a way to force the browser to treat it like a DL. We should experiment if we can replace opening in the iframe with that?

src/lib/Preview.js
JustinHoldstock: +1 to prefixing the property name to help us distinguish boolean type

tonyjin: I considered `shouldDownloadWM` and `forceDownloadWM` but I'm going to keep as `downloadWM` since download is already a verb and I interpret the option as downloadWatermarked, which implies true/false to me, whereas I interpret `previewWMPref` as previewWatermark(ed/ing)Preference, which implies options other than just true/false.

src/lib/Preview.js
tonyjin: I'll be removing `response_content_disposition_type` since we don't need it, but I'll definitely be using `Object.assign` other places instead of directly modifying `queryParams`. Thanks for the reminder @DanDeMicco.

PR #722, opened 2018-03-16
Use correct property for handle viewer event

JustinHoldstock:
My misreading lead to a null event_name being used for the metric. This fixes it to use the correct one.


boxcla:

PR #724, opened 2018-03-20
Fix: document findbar no longer ignores useHotkeys

DanDeMicco:
This change refactors DocFindBar keyboard events to be called from the DocumentViewer and the normal onKeypress flow. Fixes #723

PR #725, opened 2018-03-20
Fix: parse options during show instead of after token response

DanDeMicco:
This moves the parsing of the preview options to show instead of after the token returns. This fixes an issue uncovered by #724 in which disableHotkeys was disabled immediately after show, but the hotkeys were not disabled since parseOptions overrides.


boxcla:

PR #728, opened 2018-03-20
Fix: Reduce WebGL checks to reduce instances of Rats! error

tonyjin:
We don't use metrics associated with whether the user's browser supports WebGL or 3D rendering, so this removes the WebGL checks from the browser info check for every Preview instance and isolates them to 3D and 360 files, which should reduce instances of the `Rats! WebGL hit a snag.` error.


boxcla:

src/lib/Browser.js
```diff
@@ -184,7 +184,7 @@ class Browser {
         if (!gl) {
             const canvas = document.createElement('canvas');
             // Should stop 'Rats! WebGL hit a snag' error when checking WebGL support
-            canvas.addEventListener(EVENT_WEBGL_CONEXT_LOST, (e) => {
+            canvas.addEventListener(EVENT_WEBGL_CONTEXT_LOST, (e) => {
```
jeremypress: Can we remove the supportsWebGL global variable?

tonyjin: Will still be useful if someone is viewing a folder that has multiple 3D files since that result will be cached between previews.

PR #730, opened 2018-03-21
BFMVP-7637 Use original representation only for PDF files

stariqmi:
This temporarily fixes https://bonfirehub.atlassian.net/browse/BFMVP-7637.

Some PDF files are not being rendered correctly using their PDF representation, hence we need to use only the original representation while Box fixes this issue. Instead of completely changing the `DocLoader.js` - `determineRepresentation`, I have modified it such that we can pass in an option to use only the original representation.


boxcla:
Hi @stariqmi, thanks for the pull request. Before we can merge it, we need you to sign our Contributor License Agreement. You can do so electronically here: http://opensource.box.com/cla

stariqmi:
CLA Signed

PR #733, opened 2018-03-22
Fix: display error when previewing flash and not enabled

DanDeMicco:
swfobject is no longer supported, and flash detection is broken thus a blank page shows up when previewing flash content and flash is not enabled in the browser. This will use the existing flash detection and display a previewError asking the user to enable flash


boxcla:

PR #737, opened 2018-03-26
Chore: Remove unused babel-polyfill

tonyjin:
With Node v8+ we no longer need babel-polyfill for Webpack.

### DIFF
--- a/README.md
+++ b/README.md
@@ -306,14 +306,14 @@ Development Setup
 -----------------
 1. Install Node v8.9.4 or higher.
 2. Install yarn package manager `https://yarnpkg.com/en/docs/install`. Alternatively, you can replace any `yarn` command with `npm`.
-2. Fork the upstream repo `https://github.com/box/box-content-preview`.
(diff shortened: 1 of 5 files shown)


priyajeet:
Probably add these in package.json

```
"devEngines": {
    "node": ">=8.x",
    "npm": ">=3.x",
    "yarn": ">=1.0.0"
  },
  ```

tonyjin:
@priyajeet you don't strictly need yarn right? I'll add node & npm. Where's the documentation on this?Couldn't find anything except for github PRs with a quick search of "npm devEngines" or "package.json devEngines"

tonyjin:
Decided to remove `devEngines` for now since it doesn't seem to be an official option in `npm` or `yarn`, just `engines`. `devEngines` seems to be a React-only feature and they have a custom post-install hook that checks this: https://github.com/facebook/fbjs/blob/master/packages/fbjs-scripts/node/check-dev-engines.js#L29-L61

PR #745, opened 2018-03-28
Chore: Only reset to upstream/master for major/minor releases

pramodsum:
This will prevent patch releases from resetting to upstream/master. Instead, it will reset to the previous version of Preview (if available). 

- TODO: replicate the same logic in BoxAnnotations

### DIFF
--- a/build/release.sh
+++ b/build/release.sh
@@ -37,7 +37,7 @@ reset_to_previous_version() {
     # Reset and fetch upstream with tags
     reset_tags || return 1;
 
-    # Reset to latest code and clear unstashed changes
+    # Reset to previous release version and clear unstashed changes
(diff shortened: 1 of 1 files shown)


boxcla:

build/release.sh
```diff
@@ -27,6 +23,32 @@ reset_to_master() {
...
+        echo "----------------------------------------------------------------------"
+        echo "Error while cleaning workspace!"
+        echo "----------------------------------------------------------------------"
+        return 1;
+    fi
+
+    # Reset and fetch upstream with tags
+    reset_tags || return 1;
+
+    # Reset to latest code and clear unstashed changes
```
jeremypress: "Reset to latest release"

PR #746, opened 2018-03-28
Fix: Download watermarked

tonyjin:
- Send content disposition header to force representation to be downloaded as attachment
- Expose asset path as method on viewer instead of directly accessing


boxcla:

PR #749, opened 2018-03-28
Chore: Use key decoder from box-react-ui to prevent duplication

priyajeet:


boxcla:

pramodsum:
Can you add the same thing to BoxAnnotations?

PR #750, opened 2018-03-28
Chore: Only reset to upstream/master for major/minor releases

pramodsum:
This will prevent patch releases from resetting to upstream/master. Instead, it will reset to the previous version of Preview (if available).

TODO: replicate the same logic in BoxAnnotations

Copy of https://github.com/box/box-content-preview/pull/745

### DIFF
--- a/build/release.sh
+++ b/build/release.sh
@@ -12,11 +12,7 @@ major_release=false
 minor_release=false
 patch_release=false
 
-
-reset_to_master() {
-    # Update to latest code on GitHub master
-    git checkout master || return 1
-
+reset_tags() {
     # Wipe tags
     git tag -l | xargs git tag -d || return 1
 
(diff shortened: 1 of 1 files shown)


boxcla:

build/release.sh
```diff
@@ -27,6 +23,32 @@ reset_to_master() {
...
+reset_to_previous_version() {
+    if OLD_VERSION === "XXX"; then
+        echo "----------------------------------------------------------------------"
+        echo "Error while cleaning workspace!"
+        echo "----------------------------------------------------------------------"
+        return 1;
+    fi
+
+    # Reset and fetch upstream with tags
+    reset_tags || return 1;
```
DanDeMicco: If you end up staying with this approach, I would add a message saying what you did

build/release.sh
```diff
@@ -44,6 +78,14 @@ increment_version() {
     # Old version
     OLD_VERSION=$(./build/current_version.sh)
 
+    # The master branch should not match the previous release tag
```
jeremypress: This isn't necessarily the master branch right? If we're coming from a patch this would be some detached state.

PR #753, opened 2018-03-29
Docs: Updating issue template

jeremypress:
If we like the language here I can add this to our other repos.


boxcla:

PR #758, opened 2018-04-03
Fix: Add blend mode fallbacks if not supported

jeremypress:
This allows "legacy" highlight annotations to properly render in IE11, which by default does not support blend modes. 

**Current behavior:**
screen shot 2018-04-03 at 3 45 49 pm

**with fix:**
screen shot 2018-04-03 at 3 44 15 pm

**Is this reliable?**
Normally, no. This workaround sets the opacity of the artifact in question to 0.5, whenever it encounters some kind of blending. If an opacity instruction for the given artifact appears **after** the blending instruction, the override will be replaced (this happens at an unknown frequency). 

However, the code used to generate these annotations is written as follows:
screen shot 2018-04-03 at 3 42 54 pm

This means we know that our blend instruction will come after the opacity instruction, so this override will work. This will also make some other blending issues more viewable on IE11, but not all, due to the reason described above.

Tests on the way, wanted to get this out ASAP to see how everyone feels.

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -2,6 +2,8 @@ Please fill out the following template so we can reproduce and fix your issue as
 
 **Note:** if your issue includes a potential security vulnerability, please do not file it here. Instead, email the issue to security@box.com for support.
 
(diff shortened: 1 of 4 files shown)


boxcla:

PR #759, opened 2018-04-04
Fix: Revert media duration setting to on loadeddata

jeremypress:
Fixes the bug where media files showed no duration when previewed.


boxcla:

PR #774, opened 2018-04-16
Chore: Refactor functional tests

jeremypress:
1. We only need 1 HTML file now
2. Removed use of navigate since that seemed to be flakey
3. Switched around how we wait for load to be built in
4. Switched to new app where we have upload privileges


boxcla:

PR #775, opened 2018-04-17
New: Navigation arrows now remain visible when focused

jeremypress:
video
doc


boxcla:

src/lib/_navigation.scss
```diff
@@ -26,6 +26,10 @@ $navigationBtnWidth: 50px;
     &:hover {
         opacity: 1;
     }
+
+    &:focus {
+        opacity: 1;
```
DanDeMicco: Can you do &:hover, &:focus and combine the two?

jeremypress: 👍

PR #778, opened 2018-04-19
Chore: Refactor Timer to allow a reset of multiple tags

jeremypress:
@JustinHoldstock will add tests after your thoughts.


boxcla:

PR #780, opened 2018-04-20
Chore: update timeouts for saucelabs for hanging builds

DanDeMicco:


pramodsum:
Could you add this to BoxAnnotations as well?

DanDeMicco:
@pramodsum going to let changes like this soak for a while before merging similar ones into annotations

PR #781, opened 2018-04-23
Chore: Adding preview end and download attempt metrics

jeremypress:


src/lib/Preview.js
```diff
@@ -186,6 +195,19 @@ class Preview extends EventEmitter {
 
         // Destroy viewer
         if (this.viewer && typeof this.viewer.destroy === 'function') {
+            // Log a preview end event
+            const previewDurationTag = Timer.createTag(this.file.id, DURATION_METRIC);
+            Timer.stop(previewDurationTag);
+
+            const event = {
+                event_name: PREVIEW_END_EVENT,
+                value: previewDurationTag.elapsed,
```
JustinHoldstock: Tags are just a  string. You'll need to do `Timer.get(previewDurationTag)` to get the time structure for that tag, and access its `elapsed`

PR #784, opened 2018-04-24
Fix: Guard session/local storage usage in download reachability

jeremypress:
TODO: write tests for when storage is not available


boxcla:

PR #788, opened 2018-04-30
Fix: Embedded notes in Box WebApp

tonyjin:
Use /notes_embedded instead of /notes to resolve X-Frame-Origin issue.


boxcla:

PR #790, opened 2018-05-01
Fix: Correct media settings checkmark icon color on hover

jstoffan:
Also verified that an item receiving keyboard focus still turns the background blue and the icon white, as before.


boxcla:
Hi @jstoffan, thanks for the pull request. Before we can merge it, we need you to sign our Contributor License Agreement. You can do so electronically here: http://opensource.box.com/cla

jstoffan:
CLA signed

PR #801, opened 2018-05-25
Fix: video controls are cut off

mxiao6:


boxcla:

PR #807, opened 2018-06-18
Add docs for token generator function

### LINKED ISSUE #641: publish API documentation
When I went digging a little to find out how I might update a token, I found that there's already a method to do this, complete with documented parameters and types. But that wasn't on https://developer.box.com/docs/box-content-preview or in the `preview.js` you direct people to load (since that's minified), it was only because I was stumbling around in the GitHub sources that I saw it.

Please make this API documentation (that you've already written!) more discoverable from the Box Developers site. Thanks!

tonyjin:
Resolves #641


boxcla:

PR #808, opened 2018-06-19
Fix: Retry logic

tonyjin:
- Remove incorrect short-circuit-on-retry logic. This short circuit was being called any time you called preview on the same file twice and subsequent loads would always fail because the short circuit happened before an access token was assigned
- Set up fake clock in beforeEach of tests for handleFetchError() - not resetting the timer was causing weird issues


boxcla:

PR #814, opened 2018-06-27
Chore: functional test browser start timeout/retries

DanDeMicco:
I looked into why builds have been failing recently, and it seems it mostly has to do with browser startup hanging. This is a workaround to add a timeout and retries to the browser startup


boxcla:

PR #820, opened 2018-07-10
Upgrade: Webpack 4 and Babel 7

tonyjin:
Do not merge, this is for review only.


boxcla:

jeremypress:
double check if methods (arrow functions) on the prototype on the class are not being overriden properly?

jeremypress:
Why are we doing this?

build/webpack.config.js
```diff
@@ -26,15 +26,15 @@ const thirdParty = path.resolve('src/third-party');
 const staticFolder = path.resolve('dist');
 
 const languages = isRelease
-    ? locales
+    ? ['en-US']
```
jstoffan: Can this be simplified by removing the ternary?

build/webpack.config.js
```diff
@@ -26,15 +26,15 @@ const thirdParty = path.resolve('src/third-party');
...
     : ['en-US']; // Only 1 language needed for dev
 
 /* eslint-disable key-spacing, require-jsdoc */
 function updateConfig(conf, language, index) {
     const config = Object.assign(conf, {
         entry: {
             annotations: ['box-annotations'],
-            preview: [`${lib}/Preview.js`],
+            preview_modern: [`${lib}/Preview.js`],
```
jstoffan: What's the reasoning for the name change?

.browserslistrc
```diff
@@ -0,0 +1,5 @@
+Chrome >= 60
```
jstoffan: These will need to be maintained manually now, since they're not relative like before (last two versions). Is the relative syntax not supported in the new version?

browserslist
```diff
@@ -1,7 +0,0 @@
-# Browser support list - used by CSS autoprefixer
-last 2 Chrome versions
-last 2 Firefox versions
-last 2 Safari versions
-last 2 Edge versions
-last 2 iOS versions
-ie 11
```
jstoffan: Do we still need autoprefixing for IE11?

package.json
```diff
@@ -9,20 +9,19 @@
     "url": "git@github.com:box/box-content-preview.git"
   },
   "devDependencies": {
+    "@babel/core": "7.0.0-beta.44",
```
jstoffan: Are there any concerns with using a beta release?

build/webpack.config.js
```diff
@@ -73,14 +73,7 @@ function updateConfig(conf, language, index) {
     if (isRelease) {
         // http://webpack.github.io/docs/list-of-plugins.html#uglifyjsplugin
         config.plugins.push(
-            new UglifyJsPlugin({
-                compress: {
```
jstoffan: Are these options no longer needed or supported or are they defaults?

tonyjin: Defaults with webpack -p

package.json
tonyjin: Yep, I suggest waiting for the official release

browserslist
tonyjin: I think so? Also AFAIK some moz- prefixes are needed too

.browserslistrc
tonyjin: Hmm I might have been playing around with something other than just the Webpack upgrade here... I might have been trying to play around with https://developers.google.com/web/fundamentals/primers/modules as well

build/webpack.config.js
tonyjin: See below, this isn't a proper patch. I vaguely remember I wanted to play around with native browser module support

PR #824, opened 2018-07-23
Fix: Fix scenario where skills are present with not transcript

jeremypress:


boxcla:

PR #825, opened 2018-07-25
Chore: Support auto captions in legacy and edited captions

jeremypress:
Will add tests/cleanup tomorrow


src/lib/viewers/media/DashViewer.js
```diff
@@ -460,24 +460,22 @@ class DashViewer extends VideoBaseViewer {
...
-    loadAutoGeneratedCaptions() {
-        const textCues = [];
-
-        // Convert Box Skill transcript cards to Shaka Text Cues
-        const skillsCards = getProp(this.options, 'file.metadata.global.boxSkillsCards.cards');
-        // No skills are available on the file
-        if (!skillsCards) {
-            return;
+    parseTextCues(transcriptCards) {
+        if (!transcriptCards) {
```
jstoffan: Could we use default parameter values for these functions in order to cut down on some of the guard clauses? Something like `parseTextCues(transcriptCards = [])`?

src/lib/viewers/media/DashViewer.js
```diff
@@ -460,24 +460,22 @@ class DashViewer extends VideoBaseViewer {
     }
 
     /**
-     * Loads auto-generated captions
+     * Turn Box Skills transcript cards into Shaka Player text cues
      *
-     * @return {void}
+     * @param {Object} transcriptCards - Transcript cards
```
jstoffan: Should this be `@param {Array}`? It looks like `.find` is used on it below.

src/lib/viewers/media/DashViewer.js
```diff
@@ -488,17 +486,68 @@ class DashViewer extends VideoBaseViewer {
...
-            // Update the subtitles/caption button to reflect auto-translation
-            this.mediaControls.setLabel(this.mediaControls.subtitlesButtonEl, __('media_auto_generated_captions'));
+    /**
+     * Edits auto-generated captions
+     *
+     * @public
+     * @param {Object} skillsData - updated skills data
+     * @return {void}
+     */
+    editAutoGeneratedCaptions(skillsData) {
```
jstoffan: It looks like there's some duplication between this function and `loadAutoGeneratedCaptions`. Should there be a third method that both of these call? I also think `update` might be more clear than `edit`, since no explicit user interaction is involved.

src/lib/viewers/media/DashViewer.js
```diff
@@ -488,17 +486,68 @@ class DashViewer extends VideoBaseViewer {
...
+        const textCues = this.parseTextCues(skillsData);
+
+        if (textCues.length > 0) {
+            const areAutoCaptionsVisible = this.autoCaptionDisplayer.isTextVisible();
+            this.autoCaptionDisplayer.destroy();
+            this.autoCaptionDisplayer = new shaka.text.SimpleTextDisplayer(this.mediaEl);
+            this.autoCaptionDisplayer.append(textCues);
+            this.player.configure({ textDisplayFactory: this.autoCaptionDisplayer });
+            if (areAutoCaptionsVisible) {
+                this.autoCaptionDisplayer.setTextVisibility(true);
```
jstoffan: Could we call `this.autoCaptionDisplayer.setTextVisibility(areAutoCaptionsVisible)` to clarify that we're preserving the user's preference? What do you think?

src/lib/viewers/media/DashViewer.js
jeremypress: Yep, will clean up in the next commit. Agreed that's a better name.

src/lib/viewers/media/DashViewer.js
jeremypress: I don't think there's any harm to calling it again with false if they are already off, and it'll make the code cleaner.

src/lib/viewers/media/DashViewer.js
```diff
@@ -460,24 +460,22 @@ class DashViewer extends VideoBaseViewer {
...
-        // No skills are available on the file
-        if (!skillsCards) {
-            return;
+    parseTextCues(transcriptCards) {
+        if (!transcriptCards) {
+            return [];
         }
 
-        const transcriptCard = skillsCards.find((card) => card.skill_card_type === 'transcript');
+        const textCues = [];
```
JustinHoldstock: Since `textCues` isn't being used until after the `if`, why not declare it there?

src/lib/viewers/media/DashViewer.js
```diff
@@ -488,17 +486,68 @@ class DashViewer extends VideoBaseViewer {
...
+            return;
+        }
+
+        // Save passed in skills data to avoid regenerating captions unnecessarily
+        this.skillsData = skillsData;
+
+        const textCues = this.parseTextCues(skillsCards);
+
         if (textCues.length > 0) {
             this.autoCaptionDisplayer = new shaka.text.SimpleTextDisplayer(this.mediaEl);
```
JustinHoldstock: Could you reuse this, if it's already been created beforehand? 
IE) `this.autoCaptionDisplayer = this.autoCaptionDisplayer || new SimpleTextDisplayer(...)`

src/lib/viewers/media/DashViewer.js
```diff
@@ -488,17 +486,68 @@ class DashViewer extends VideoBaseViewer {
...
+        if (!skillsData || skillsData === this.skillsData) {
+            return;
+        }
+
+        // Save passed in skills data to avoid regenerating captions unnecessarily
+        this.skillsData = skillsData;
+
+        const textCues = this.parseTextCues(skillsData);
+
+        if (textCues.length > 0) {
```
JustinHoldstock: Guard to make sure that `autoCaptionDisplayer` exists?

PR #827, opened 2018-08-02
New: add preview.load callback option

danielSONCCO:
Hi, I need to do some operations at the end of the preview html loading but preview.show ends without that html loaded.

I realized that preview.show calls preview.load that calls getTokens without waiting for the end of that promise. So i'm adding an optional callback in the options parameter of preview.show, that it's passed as a parameter for preview.load.
So now using preview.show("123", "token", { callback: onHtmlLoad }); i can move the operations depending on the html loading in the callback parameter.

Maybe you know a simpler way of listening the end of this preview html loading. I would be happy of know it.

### DIFF
--- a/src/lib/Preview.js
+++ b/src/lib/Preview.js
@@ -690,10 +690,10 @@ class Preview extends EventEmitter {
      *
      * @private
      * @param {string|Object} fileIdOrFile - Box File ID or well-formed Box File object
-     * @param {Function} callback - Optional load callback
+     * @param {Function} loadCallback - Optional load callback
      * @return {void}
      */
-    load(fileIdOrFile, callback) {
+    load(fileIdOrFile, loadCallback) {
         // Clean up any existing previews before loading
         this.destroy();
 
@@ -776,7 +776,7 @@ class Preview extends EventEmitter {
         // combine Box Elements + Preview. This could potentially break if we have Box Elements fetch the file object
         // and pass the well-formed file object directly to the preview library to render.
         const isPreviewOffline = typeof fileIdOrFile === 'object' && this.options.skipServerUpdate;
-        callback = callback || function() {};
+        const callback = loadCallback || function () {};
         if (isPreviewOffline) {
             this.handleTokenResponse({});
             callback();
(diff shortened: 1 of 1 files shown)


boxcla:
Hi @danielSONCCO, thanks for the pull request. Before we can merge it, we need you to sign our Contributor License Agreement. You can do so electronically here: http://opensource.box.com/cla

danielSONCCO:
"CLA signed". Thanks!

danielSONCCO:
UPDATE: I found that it's possible to wait until the preview display finishes with `preview.addListener("load", callback);` but it seems that this load is the preview.viewer.load and i wanted to wait only the html loading generated by getTokens(fileId, token).then(handleTokenResponse) that finishes long before the preview.viewer.load
So still i wonder if it's possible to listen an event emitted by the promise above. (how to be triggered by the html loading without callbacks)

JustinHoldstock:
Hey @danielSONCCO ,

What are you trying to accomplish with the callback and events? We may be able to assist you without having to add any additional code OR we can expose some other events for you.

Thanks!
Justin

danielSONCCO:
@JustinHoldstock I'm trying to add a new button to the preview html, so it's something like 
```
preview.show();
const buttonHeader = options.container && options.container.findFirstDescendant(".bp-header .bp-header-btns");
buttonHeader.appendChild(button);
```

JustinHoldstock:
Hey @danielSONCCO,

Is there a reason that listening to 'load' doesn't meet requirements?

We recommend  setting up your UI via:
```
preview.addListener('load',() => {
    const buttonHeader = options.container.findFirstDescendant(".bp-header .bp-header-btns");
    buttonHeader.appendChild(button);
});

preview.show();
```

danielSONCCO:
@JustinHoldstock preview.addListener("load", ) waits for the end of previewer.viewer.load that depends on the size of the file, but previewer.load ends before and there we have already the html, so having a listener for previewer.load would be an optimization

danielSONCCO:
@DanDeMicco i don't get it. do you mean something like preview.addListener("viewer", )?

JustinHoldstock:
@danielSONCCO,

Yup, that's what he's referring to!

You can see some documentation here, for more.
https://github.com/box/box-content-preview#events

danielSONCCO:
@DanDeMicco @JustinHoldstock thank you very much!

PR #828, opened 2018-08-15
Chore: Allow load metrics emission without file info

JustinHoldstock:
Allows for file info to be request external to preview, and still report timings. 

*Note: Updates are required in Box UI Elements before this can be merged, otherwise false reports will be generated.*

### DIFF
--- a/src/lib/events.js
+++ b/src/lib/events.js
@@ -44,7 +44,6 @@ export const ERROR_CODE = {
     FLASH_NOT_ENABLED: 'error_flash_not_enabled'
 };
 
-export const PREVIEW_LOAD_EVENT = '';
 // Event fired from Preview with error details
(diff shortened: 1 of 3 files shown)


boxcla:

src/lib/events.js
```diff
@@ -44,7 +44,6 @@ export const ERROR_CODE = {
     FLASH_NOT_ENABLED: 'error_flash_not_enabled'
 };
 
-export const PREVIEW_LOAD_EVENT = '';
```
jeremypress: Why are we removing this?

JustinHoldstock: It's not used anywhere AND it's an empty string

JustinHoldstock: We now use the below `LOAD_METRIC.previewLoadEvent`

PR #829, opened 2018-08-16
Fix: Display correct error when handling an asset error

jeremypress:


boxcla:

src/lib/viewers/BaseViewer.js
```diff
@@ -299,7 +299,7 @@ class BaseViewer extends EventEmitter {
      */
     handleAssetError(err) {
         const originalMessage = err ? err.message : '';
-        const error = new PreviewError(ERROR_CODE.LOAD_ASSET, '', {}, originalMessage);
+        const error = err instanceof PreviewError ? err : new PreviewError(ERROR_CODE.LOAD_ASSET, originalMessage, {});
```
jeremypress: If handleAssetError is called with a preview error, we should use that instead of creating a generic load asset error. More specific error types were replaced here.

PR #833, opened 2018-08-21
Update: Preview should allow disabling of  auto focus

priyajeet:
default autoFocus is true


boxcla:

jstoffan:
This doesn't change the current behavior, it just allows folks to disable autofocus; is that intentional? It doesn't directly solve the issue in the PR title, just supports it as an option.

priyajeet:
Fixed title

PR #834, opened 2018-08-23
New: Adding NodeJS version check to Release script

JustinHoldstock:


boxcla:

build/release.sh
```diff
@@ -236,6 +253,8 @@ while getopts "mnp" opt; do
     esac
 done
 
+# Before running release, make sure using correct version of NodeJS
+validate_node_version;
```
jeremypress: this might make more sense added to the push_new_release function like we do with other checks?

build/validateNodeVersion.js
```diff
@@ -0,0 +1,112 @@
...
+    });
+
+    if (isValidVersion) {
+        pass();
+    } else {
+        fail();
+    }
+}
+
+https.get(NODE_JS_VERSION_LIST_URL, response => {
```
jeremypress: No harm in bringing in axios as a dev dependency to accomplish this (IMO).

build/validateNodeVersion.js
```diff
@@ -0,0 +1,112 @@
...
+  response.on('data', data => {
+    body += data;
+  });
+
+  response.on('end', validateNodeVersion);
```
jeremypress: Where does this data get used? ValidateNodeVersion doesn't take in any parameters

build/validateNodeVersion.js
```diff
@@ -0,0 +1,112 @@
+const https = require('https');
+
+const SUPPORTED_LTS = "Carbon";
+const VERSION = process.versions.node;
+const NODE_JS_VERSION_LIST_URL="https://nodejs.org/dist/index.json"
+const MIN_MAJOR_SUPPORT = 8;
```
jeremypress: Are we still relying solely off the information from nodejs.org? In that case why are these constants needed?

build/validateNodeVersion.js
```diff
@@ -0,0 +1,112 @@
...
+const MIN_MINOR_SUPPORT = 9;
+const MIN_PATCH_SUPPORT = 4;
+
+let body = '';
+
+function fail() {
+    body = '';
+    console.log(false);
```
jeremypress: Do we need to throw an error at some point to get the release script to break?

build/validateNodeVersion.js
JustinHoldstock: See use of `body`.

build/validateNodeVersion.js
JustinHoldstock: console.log() is picked up by the bash script that invokes this. Do you think we also need to throw an error if the parent bash script is halting all execution after receiving a `false` signal from this?

build/release.sh
JustinHoldstock: Done!

build/validateNodeVersion.js
JustinHoldstock: Done! Really good idea, thank you

build/validateNodeVersion.js
```diff
@@ -0,0 +1,105 @@
...
+ * Returns true if greater or same version as minimum supported version
+ *
+ * @param {Number} major - Major version number
+ * @param {Number} minor - Minor version number
+ * @param {Number} patch - Patch version number
+ * @returns {boolean} True if a supported LTS version of NodeJS
+ */
+function compareVersion(major, minor, patch) {
+    // If major version is greater, it's valid
+    if (major > MIN_MAJOR_SUPPORT) {
```
DanDeMicco: Can this be a problem, lets say someone is using node 10? There could be breaking changes

JustinHoldstock: `compareVersion()` only compares against our minimum requirements of node `8.9.4`. After this is executed, we do a check against LTS versions of node, which are supported (at least) by conventional-github-releaser.

build/validateNodeVersion.js
```diff
@@ -0,0 +1,105 @@
+const axios = require('axios');
+
+const SUPPORTED_LTS = "Carbon";
+const VERSION = process.versions.node;
+const NODE_JS_VERSION_LIST_URL="https://nodejs.org/dist/index.json"
+const MIN_MAJOR_SUPPORT = 8;
+const MIN_MINOR_SUPPORT = 9;
+const MIN_PATCH_SUPPORT = 4;
+
+function fail() {
```
jeremypress: Maybe drop a doc string on these explaining what the process.exit does?

build/validateNodeVersion.js
```diff
@@ -0,0 +1,116 @@
...
+const SUPPORTED_LTS = "Carbon";
+const VERSION = process.versions.node;
+const NODE_JS_VERSION_LIST_URL="https://nodejs.org/dist/index.json"
+const MIN_MAJOR_SUPPORT = 8;
+const MIN_MINOR_SUPPORT = 9;
+const MIN_PATCH_SUPPORT = 4;
+
+
+/**
+ * Exit with code 1 so that calling script can see that current node version was NOT valid.
```
jeremypress: lgtm

PR #837, opened 2018-08-28
Fix: Navigation buttons should not cover scrollbar

ConradJChan:
Made the navigation buttons skinnier (though visibly in the same place) to allow room for the scrollbar of the preview content to be clickable to the right of the next file button


boxcla:
Hi @ConradJChan, thanks for the pull request. Before we can merge it, we need you to sign our Contributor License Agreement. You can do so electronically here: http://opensource.box.com/cla

PR #838, opened 2018-08-28
Fix: navigation buttons shouldn't cover scrollbar

ConradJChan:
Made the navigation buttons skinner, in particular the next file navigation button, in order to allow room for the preview content scrollbar to be clickable

### DIFF
--- a/src/lib/_navigation.scss
+++ b/src/lib/_navigation.scss
@@ -55,10 +55,10 @@ $navigationBtnWidth: 50px;
 
 .bp-navigate-left {
     border-radius: 0 2px 2px 0;
-    left: 15px;
+    left: 20px;
 }
 
 .bp-navigate-right {
     border-radius: 2px 0 0 2px;
-    right: 15px;
+    right: 20px;
 }
--- a/src/lib/_navigation.scss
+++ b/src/lib/_navigation.scss
@@ -1,4 +1,4 @@
-$navigationBtnWidth: 50px;
+$navigationBtnWidth: 40px;
 
 .bp-is-fullscreen .bp-navigate {
     display: none;
@@ -39,6 +39,11 @@ $navigationBtnWidth: 50px;
         box-shadow: none;
         outline: none;
     }
+
+    & > svg {
+        margin-left: -5px;
+        width: 50px;
+    }
 }
 
 .bp-is-navigation-visible .bp-navigate {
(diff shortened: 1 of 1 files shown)


boxcla:

src/lib/_navigation.scss
```diff
@@ -39,6 +39,11 @@ $navigationBtnWidth: 50px;
         box-shadow: none;
         outline: none;
     }
+
+    & > svg {
+        margin-left: -5px;
```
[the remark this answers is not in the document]
ConradJChan: This is to center the SVG within the now 40px button, stand by for screenshots

src/lib/_navigation.scss
```diff
@@ -52,9 +57,10 @@ $navigationBtnWidth: 50px;
 .bp-navigate-left {
     border-radius: 0 2px 2px 0;
     left: 0;
```
jeremypress: remove this?

src/lib/_navigation.scss
```diff
@@ -1,4 +1,4 @@
-$navigationBtnWidth: 50px;
+$navigationBtnWidth: 40px;
```
DanDeMicco: Why reduce the button width? What if you left everything the same, but changed the left and right of .bp-navigate-left and .bp-navigate-right, would that work?

ConradJChan:
Before:
before

After:
after

src/lib/_navigation.scss
ConradJChan: Yeah, you're right I could keep the button width the same, I guess I was just trying to keep the button about in the same place as it ever was (though what is 5px vs 10px really)

ConradJChan:
In addressing @DanDeMicco's comment, I'm keeping the button the same size now. I've also increased the offset to 20px because across the different browsers the scroll bar width is 17px, so just rounding up to 20 to be safe

Now looks like this:
after 20px offset

Thoughts?

src/lib/_navigation.scss
```diff
@@ -39,6 +39,10 @@ $navigationBtnWidth: 50px;
         box-shadow: none;
         outline: none;
     }
+
+    & > svg {
```
jstoffan: Is this still necessary?

ConradJChan: nope it is not, will remove

PR #839, opened 2018-08-29
Fix: add min-width to video

ConradJChan:
Issue is that on minimum sized videos, the video controls get too cramped once you max out the content to be displayed. This means that the video needs have a length in hours, not minutes, as well as close captioning as an option. This is only observed when the hover over the volume control expands out to the right. From experimenting, it appears that no matter how small the video is that is uploaded, the conversion upscales it to minimum 480p and when it's rendered is at least 420px wide.

Before:
before min width

After:
after min width

The fix was to set a min-width on the video element to be 460px which seems to be wide enough to comfortably render the volume slider, the time display in hours, and the CC button (along with the other default controls).

### DIFF
--- a/src/lib/viewers/media/Dash.scss
+++ b/src/lib/viewers/media/Dash.scss
@@ -4,9 +4,9 @@
     video {
         cursor: none;
         display: block;
-        min-width: 460px;
         max-height: 100%;
         max-width: 100%;
+        min-width: 460px;
         width: 100%;
     }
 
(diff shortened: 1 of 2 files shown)


boxcla:

jstoffan:
Just as a side note: Is the CC button not vertically centered? It looks weird.

ConradJChan:
@jstoffan let me know if your hawkeyes see something else

screen shot 2018-08-31 at 3 57 36 pm

jstoffan:
@ConradJChan, it looks like there's a bunch of unrelated changes from the past few days. Can you make sure there's no merge conflicts?

PR #841, opened 2018-09-07
Chore: remove x-box-accept-encoding header to prevent OPTIONS call

JustinHoldstock:


boxcla:

PR #842, opened 2018-09-13
Fix: construct annotator when BoxAnnotations loads

ConradJChan:
Instead of waiting for the Preview viewer to finish loading before constructing the Annotator (which fetches the annotations for the file) we construct the Annotator as soon as the BoxAnnotations static assets are loaded.

From testing on my computer, saw a slight boost of 200-300ms in loading preview with annotations with caching disabled (or about the time it takes to do a network roundtrip for the annotations call). With caching there wasn't much of a time difference since the BoxAnnotations assets were already fetched.

### DIFF
--- a/src/lib/viewers/BaseViewer.js
+++ b/src/lib/viewers/BaseViewer.js
@@ -861,6 +861,10 @@ class BaseViewer extends EventEmitter {
         const boxAnnotations = this.options.boxAnnotations || new global.BoxAnnotations(viewerOptions);
         this.annotatorConf = boxAnnotations.determineAnnotator(this.options, this.viewerConfig);
 
+        if (!this.annotatorConf) {
+            return;
+        }
+
         const annotatorOptions = this.createAnnotatorOptions({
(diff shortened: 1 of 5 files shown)


boxcla:

src/lib/viewers/BaseViewer.js
```diff
@@ -860,9 +861,11 @@ class BaseViewer extends EventEmitter {
         const boxAnnotations = this.options.boxAnnotations || new global.BoxAnnotations(viewerOptions);
         this.annotatorConf = boxAnnotations.determineAnnotator(this.options, this.viewerConfig);
 
-        if (this.annotatorConf) {
```
pramodsum: Make sure you leave this check in because `boxAnnotations.determineAnnotator()` can return null

pramodsum:
Make sure you test the case where Preview doesn't have to fetch the static assets and an instance of BoxAnnotations is passed as an option to Preview

PR #845, opened 2018-09-24
New: Add gzip encoding param to pdf reps

JustinHoldstock:


boxcla:

src/lib/constants.js
```diff
@@ -103,3 +103,9 @@ export const PREVIEW_SCRIPT_NAME = 'preview.js';
 
 export const FILE_OPTION_FILE_VERSION_ID = 'fileVersionId';
 export const FILE_OPTION_START = 'startAt';
+
+// Query parameter for requesting encoded PDF representations
+export const QUERY_PARAM_PDF_ENCODING = 'encoding';
+export const PDFEncodingTypes = {
```
jstoffan: Should this use the same `STYLE_OF_NAMING` as the other constants?

src/lib/viewers/doc/DocBaseViewer.js
```diff
@@ -585,6 +594,14 @@ class DocBaseViewer extends BaseViewer {
             };
         }
 
+        // If range requests are disabled, request the gzip compressed version of the representation
+        if (PDFJS.disableRange) {
+            const { url } = docInitParams;
+            docInitParams.url = appendQueryParams(url, {
+                [QUERY_PARAM_PDF_ENCODING]: PDFEncodingTypes.GZIP
```
jstoffan: If the key/value pair is only used once, it can be declared inline. It doesn't look like we have any other encoding types.

jeremypress: You could move this to before the docInitParams are set, then you can just pass this value in so you don't have to unpack it/modify a "const" object

src/lib/constants.js
JustinHoldstock: Good catch

PR #847, opened 2018-10-04
Fix: Adding annotations back to MultiImageViewer

ConradJChan:


boxcla:

src/lib/viewers/image/MultiImageViewer.js
```diff
@@ -92,6 +92,8 @@ class MultiImageViewer extends ImageBaseViewer {
 
                 this.wrapperEl.addEventListener('scroll', this.scrollHandler, true);
             })
+            .then(this.loadBoxAnnotations)
+            .then(this.createAnnotator)
```
pramodsum: Could this still be triggered off the viewerLoadHandler as is done in BaseViewer? I think that handler is triggered off of when the 'load' event is fired

ConradJChan: As discussed, I'll create a new `handleAssetsAndRep` method on BaseViewer which will handle loadingBoxAnnotations as well as create the annotator, and have the different viewers override it with their viewer specific logic to avoid this issue.

PR #848, opened 2018-10-04
Fix: Pass shaka data to error detail so it can be logged

jeremypress:


boxcla:

PR #853, opened 2018-10-23
New: Enable PDF Signatures in PDF.js

jeremypress:
Upgrading PDF.js and including this change is much more involved than running our script. A lot of interfaces have changed/there may be some methods that are no longer available. 

Instead I modified our current version of PDF.js, and modified our minification script to take in a version of the assets you want to minify, instead of assuming that they align with the current version of preview.


boxcla:

PR #854, opened 2018-10-24
Fix: Prevent extra multi image page padding added by preview consumers

jeremypress:


boxcla:

PR #858, opened 2018-11-05
Fix: Add tabindex to preview container

ConradJChan:


src/lib/Fullscreen.js
```diff
@@ -99,6 +99,15 @@ class Fullscreen extends EventEmitter {
 
         if (this.isSupported()) {
             if (this.isFullscreen()) {
+                // Focus on the fullscreen element so keyboard
+                // events are triggered without an extra click
+                if (el) {
```
DanDeMicco: When can el be falsy?

src/lib/Fullscreen.js
```diff
@@ -99,6 +99,13 @@ class Fullscreen extends EventEmitter {
 
         if (this.isSupported()) {
             if (this.isFullscreen()) {
+                // Focus on the fullscreen element so keyboard
+                // events are triggered without an extra click
+                // If el has target property, then it is an Event
+                // otherwise it is a HTMLElement
+                const element = el.target || el;
```
priyajeet: Wouldnt we want to do this for both cases when enter=true?
If so, then it should be in the if block below, and not here

src/lib/__tests__/Fullscreen-test.js
```diff
@@ -77,17 +82,6 @@ describe('lib/Fullscreen', () => {
 
             expect(fullscreen.emit).to.have.been.calledWith('exit');
         });
-
-        it('should be called only once when the fullscreenchange event is emitted', () => {
```
ConradJChan: @DanDeMicco I deleted this test because 1. it stopped passing for me 😢and I couldn't get it working again but also 2. it seems like it's more testing the DOM than our code? Thoughts?

ConradJChan: nvm got it working again!

src/lib/Fullscreen.js
```diff
@@ -91,27 +91,39 @@ class Fullscreen extends EventEmitter {
...
         let enter = false;
 
-        if (this.isSupported()) {
-            if (this.isFullscreen()) {
-                enter = true;
-            }
-        } else if (!this.isFullscreen(el)) {
-            enter = true;
-        }
+        enter = (this.isSupported() && this.isFullscreen()) || (!this.isSupported() && !this.isFullscreen(el));
```
DanDeMicco: We should remove this in the future, lets add it to the backlog

src/lib/Fullscreen.js
```diff
@@ -91,27 +91,39 @@ class Fullscreen extends EventEmitter {
...
     };
 
+    /**
+     * Focuses the element
+     *
+     * @private
+     * @param {HTMLElement|Event} el - Fullscreen element or event
+     * @return {void}
+     */
+    focusFullscreenElement = (el) => {
```
DanDeMicco: This seems like it is pretty generic, maybe add to util.js?

ConradJChan: It is and it isn't, taking either a HTMLElement or Event and extracting out the HTMLElement to focus on seems a bit specific to this logic, maybe if this situation comes up again in another file we can move it then

PR #860, opened 2018-11-07
Chore: Remove outline from preview container

ConradJChan:
Adding the tabindex caused the outline to appear on the preview container

screen shot 2018-11-07 at 1 34 04 pm


boxcla:

PR #861, opened 2018-11-09
Chore: Emit pause event with flag if initiated by user 

JustinHoldstock:


src/lib/viewers/media/MediaBaseViewer.js
```diff
@@ -599,7 +599,7 @@ class MediaBaseViewer extends BaseViewer {
             this.mediaEl.addEventListener('timeupdate', this.pauseListener);
         } else {
             this.mediaEl.pause();
-            this.emit('pause');
+            this.emit('pause', this.mediaEl.currentTime);
```
DanDeMicco: Would using ```time``` work?

jeremypress:
why can't they just get the currentTime of the mediaEl when this event fires?

PR #864, opened 2018-11-15
Fix: Focus on correct element when entering fullscreen in Firefox

jeremypress:


boxcla:

jeremypress:
Let me know if it's more clear when breaking the cases out like this. If it's too verbose I can move it back to a one liner.

src/lib/Fullscreen.js
```diff
@@ -111,15 +111,22 @@ class Fullscreen extends EventEmitter {
...
      */
-    focusFullscreenElement = (el) => {
+    focusFullscreenElement = (elementOrEvent) => {
         // Focus on the fullscreen element so keyboard
         // events are triggered without an extra click
         // If el has target property, then it is an Event
         // otherwise it is a HTMLElement
-        const element = el.target || el;
+        let element;
+        if (elementOrEvent.target) {
```
DanDeMicco: Could you maybe use ```getProp``` here? Something like ```getProp(elementOrEvent, 'target.activeElement', elementOrEvent)```

src/lib/Fullscreen.js
```diff
@@ -111,15 +111,22 @@ class Fullscreen extends EventEmitter {
...
+            // This is an element
+            // On certain browsers (Firefox), the actual target is on the activeElement property of the event target
+            element = elementOrEvent.target.activeElement || elementOrEvent.target;
+        } else {
```
jstoffan: You could simplify this a bit by avoiding the `else` clause here.

```
let element = elementOrEvent;

// Comment explaining why this may sometimes be an event instead of an element
if (element.target) {
    element = // etc.
}

element.focus();
```

PR #868, opened 2018-11-26
Fix: fixes bug for cancel/done not working in annotations

JustinHoldstock:


boxcla:

PR #871, opened 2018-11-28
Fix: Don't wait when rep status is none

ConradJChan:
If representation does not yet exist for a file (status is 'none') don't delay in fetching rep in order to kick off conversion asap.


boxcla:

src/lib/RepStatus.js
```diff
@@ -142,19 +142,21 @@ class RepStatus extends EventEmitter {
...
                 if (this.logger) {
                     this.logger.setUnConverted();
                 }
 
                 this.emit('conversionpending');
 
-                // Check status again after delay
+                // Check status again after delay or
+                // If status is none, request immediately since conversion
+                // won't kick of until representation is requested
```
jstoffan: `kick off`

PR #876, opened 2018-12-04
New: Thumbnails toggle sidebar

ConradJChan:
Looking for suggestions on areas to test to make sure the DOM structure changing didn't break some functionality. Areas I've been looking at are:

- prev/next file
- dark/light theme
- annotations
- sanity testing other document types
- toggling fullscreen

Will add tests as well


boxcla:

ConradJChan:
dec-04-2018 10-18-03

src/lib/_common.scss
```diff
@@ -117,34 +118,51 @@ $header-height: 48px;
...
+        flex: 1;
+        flex-direction: column;
+        justify-content: center;
         left: 0;
-        position: fixed;
+        margin: 0;
+        outline: none;
+        -webkit-overflow-scrolling: touch;
+        padding: 0;
+        position: relative;
```
jstoffan: Is it necessary for the core preview container to be positioned?

src/lib/_common.scss
```diff
@@ -117,34 +118,51 @@ $header-height: 48px;
...
+        width: 180px;
+    }
+
+    .bp {
+        align-items: center;
+        background-color: $ffive;
+        border: 0 none;
         bottom: 0;
+        display: flex;
+        flex: 1;
```
jstoffan: I usually try to avoid shorthand flex notation, since it can cause issues in older versions of Safari and IE. Did you want `flex: 1 1 auto` or `flex: 1 1 100%`?

src/lib/_common.scss
```diff
@@ -117,34 +118,51 @@ $header-height: 48px;
...
 
-    // Pseudo fullscreen for iOS Safari which doesn't support the fullscreen API
-    &.bp-fullscreen-unsupported {
+    .bp-thumbnails-container {
+        background-color: $ffive;
+        border-right: solid 1px $seesee;
+        width: 180px;
+    }
+
+    .bp {
```
jstoffan: Is changing the markup structure here considered a breaking change? If so, we may want to keep the outer container as `.bp` and rename the inner container to `.bp-preview` or something for clarity. Some CSS frameworks advocate for using hyphens as descendent indicators, so it'd be `.bp` -> `.bp-preview`.

src/lib/_common.scss
```diff
@@ -117,34 +118,51 @@ $header-height: 48px;
...
     position: absolute;
     right: 0;
     top: 0;
 
-    // Pseudo fullscreen for iOS Safari which doesn't support the fullscreen API
-    &.bp-fullscreen-unsupported {
+    .bp-thumbnails-container {
+        background-color: $ffive;
+        border-right: solid 1px $seesee;
+        width: 180px;
```
jstoffan: It seems appropriate to use `flex: 0 1 180px` here and add `max-width: 180px` if necessary.

src/lib/_common.scss
```diff
@@ -117,34 +118,51 @@ $header-height: 48px;
...
-    -webkit-overflow-scrolling: touch;
-    padding: 0;
     position: absolute;
     right: 0;
     top: 0;
 
-    // Pseudo fullscreen for iOS Safari which doesn't support the fullscreen API
-    &.bp-fullscreen-unsupported {
+    .bp-thumbnails-container {
+        background-color: $ffive;
```
jstoffan: Should we move the background color to the root element and have both children be transparent?

src/lib/_common.scss
```diff
@@ -117,34 +118,51 @@ $header-height: 48px;
...
+        padding: 0;
+        position: relative;
         right: 0;
         top: 0;
-    }
 
-    &.bp-dark {
-        background-color: $black;
+        // Pseudo fullscreen for iOS Safari which doesn't support the fullscreen API
+        &.bp-fullscreen-unsupported {
```
jstoffan: Does all this still work properly? Does the new thumbnail navigation work in this case?

src/lib/_common.scss
```diff
@@ -117,34 +118,51 @@ $header-height: 48px;
...
+            bottom: 0;
+            left: 0;
+            position: fixed;
+            right: 0;
+            top: 0;
+        }
+
+        &.bp-dark {
```
jstoffan: Seems like this should still live on the root container.

src/lib/_navigation.scss
```diff
@@ -22,6 +22,7 @@ $navigationBtnWidth: 50px;
     transition-property: opacity;
     white-space: nowrap;
     width: $navigationBtnWidth;
+    z-index: 1;
```
jstoffan: What's the reason for this change?

src/lib/viewers/doc/DocBaseViewer.js
```diff
@@ -99,6 +99,8 @@ class DocBaseViewer extends BaseViewer {
         // Call super() to set up common layout
         super.setup();
 
+        this.thumbnailsSidebarEl = this.options.container.querySelector('.bp-thumbnails-container');
```
jstoffan: Super minor, but it seems like this should happen last. The core preview experience is the most important thing going on in this method.

src/lib/viewers/doc/DocBaseViewer.js
```diff
@@ -1235,7 +1237,11 @@ class DocBaseViewer extends BaseViewer {
      * @protected
      * @return {void}
      */
-    toggleThumbnails() {}
+    toggleThumbnails() {
```
JustinHoldstock: Guard in here, incase the query selector above, fails?

src/lib/_common.scss
ConradJChan: I think `flex: 1 1 100%` makes sense, i'll make that change

src/lib/_common.scss
ConradJChan: I think I can remove all the top, right, bottom, left, but it still needs to be relative positioned so that the buttons, controls, and findbar can be absolutely positioned

src/lib/_common.scss
ConradJChan: Actually it seems like it's only applied for the video player, in which case I think we can leave it as is

src/lib/_common.scss
ConradJChan: Ok, as discussed, I'll introduce a Upgrade.md describing the breaking change. Also I'll make the following changes:
`.bp-body` -> `.bp`
`.bp` -> `.bp-content`

src/lib/_navigation.scss
jstoffan: Hm, can we take a look at this tomorrow? It doesn't seem like the buttons should need to change to support the navigation bar.

src/lib/_common.scss
```diff
@@ -117,34 +122,46 @@ $header-height: 48px;
...
     top: 0;
 
-    // Pseudo fullscreen for iOS Safari which doesn't support the fullscreen API
-    &.bp-fullscreen-unsupported {
-        bottom: 0;
-        left: 0;
-        position: fixed;
-        right: 0;
-        top: 0;
+    .bp-thumbnails-container {
```
DanDeMicco: Are thumbnails specific to documents? If so, should this be in _docBase.scss? Or better yet in the thumbnails scss file

src/lib/viewers/doc/DocBaseViewer.js
```diff
@@ -1238,6 +1239,10 @@ class DocBaseViewer extends BaseViewer {
      * @return {void}
      */
     toggleThumbnails() {
+        if (!this.thumbnailsSidebarEl) {
+            return;
```
JustinHoldstock: This will prevent `resize()` from occurring if the thumbnails are missing. IIRC, this guarantees that Preview scales properly inside of the parent container.

src/lib/shell.html
```diff
@@ -33,34 +33,37 @@
...
-        <div class="bp-loading-wrapper">
-            <div class="bp-loading">
-                <div class="bp-icon bp-icon-file"></div>
-                <div class="bp-crawler-wrapper">
-                    <div class="bp-crawler">
-                        <div></div>
-                        <div></div>
-                        <div></div>
+    <div class="bp-body">
+        <div class="bp-thumbnails-container bp-is-hidden"></div>
```
DanDeMicco: Does this need to be in the shell? Can it be appended to the dom dynamically?

src/lib/_common.scss
```diff
@@ -117,34 +122,46 @@ $header-height: 48px;
...
+        border-right: solid 1px $seesee;
+        flex: 0 0 180px;
     }
 
-    &.bp-dark {
-        background-color: $black;
+    .bp {
+        align-items: center;
+        border: 0 none;
```
DanDeMicco: Is there some other way to do this rather than refactoring the DOM and css for this? This will affect all viewers. Possibly by adding a class to .bp, and overriding the css accordingly

src/lib/_common.scss
jeremypress: There's a small chance we want to expand this to Tif files one day, but that's unlikely

src/lib/_common.scss
jeremypress: I think setting the option `header` to `dark` will also trigger this ```String value of none or dark or light that controls header visibility and theme```.

src/lib/shell.html
jeremypress: Our pattern for most big pieces of UI has been to add them hidden and toggle as needed. Not saying it's a good thing, but it's consistent.

src/lib/_common.scss
DanDeMicco: (related to my comment below)

src/lib/shell.html
jeremypress: Talked to Daniel about this. He's right, we dynamically generate viewer specific things like the print popup.

ConradJChan:
@JustinHoldstock Yeah I think I will expose the thumbnails as an option to be turned on or off, thanks for the input!

src/lib/_common.scss
ConradJChan: yeah I tried that but didn't see this css class applied, only when using the video player

src/lib/_common.scss
ConradJChan: moved to _docBase.scss

src/lib/viewers/BaseViewer.js
```diff
@@ -1094,6 +1095,29 @@ class BaseViewer extends EventEmitter {
...
         this.loadBoxAnnotations().then(this.createAnnotator);
     }
+
+    /**
+     * Method to insert the viewer wrapper before any other children
+     *
+     * @param {Element} element Element to be inserted into the DOM
+     * @return {Element} inserted element
+     */
+    insertViewerWrapper(element) {
```
jeremypress: Maybe generalize this to an insertBefore method that can live as util? You pass in the element and the container div that you want to stick it into.

src/lib/viewers/doc/DocBaseViewer.js
```diff
@@ -99,7 +100,7 @@ class DocBaseViewer extends BaseViewer {
         // Call super() to set up common layout
         super.setup();
 
-        this.docEl = this.containerEl.appendChild(document.createElement('div'));
+        this.docEl = this.insertViewerWrapper(document.createElement('div'));
```
jeremypress: Could you do the creation in the base setup and just do the assignment here? That would make it more similar to if the div already existed in the shell by the time we got to any specific viewer.

src/lib/viewers/BaseViewer.js
```diff
@@ -23,13 +23,13 @@ import {
...
     FILE_OPTION_START,
-    SELECTOR_BOX_PREVIEW,
     SELECTOR_BOX_PREVIEW_BTN_ANNOTATE_POINT,
     SELECTOR_BOX_PREVIEW_BTN_ANNOTATE_DRAW,
     SELECTOR_BOX_PREVIEW_CRAWLER_WRAPPER,
     SELECTOR_BOX_PREVIEW_ICON,
     STATUS_SUCCESS,
-    STATUS_VIEWABLE
+    STATUS_VIEWABLE,
+    SELECTOR_BOX_PREVIEW_CONTENT
```
jstoffan: Minor: It looks like the imports are alphabetized. Can we maintain that?

src/lib/viewers/doc/DocBaseViewer.js
```diff
@@ -120,6 +121,11 @@ class DocBaseViewer extends BaseViewer {
         this.loadTimeout = LOAD_TIMEOUT_MS;
 
         this.startPageNum = this.getStartPage(this.startAt);
+
+        const thumbnailsSidebarEl = document.createElement('div');
+        thumbnailsSidebarEl.classList.add(CLASS_BOX_PREVIEW_THUMBNAILS_CONTAINER, CLASS_HIDDEN);
+        this.containerEl.parentNode.insertBefore(thumbnailsSidebarEl, this.containerEl);
+        this.thumbnailsSidebarEl = thumbnailsSidebarEl;
```
jstoffan: It looks like the existing convention is to assign the variable to the class instance and work with that rather than a local variable.

```
this.thumbnailsSidebarEl.classList.add(CLASS_BOX_PREVIEW_THUMBNAILS_CONTAINER, CLASS_HIDDEN);
// etc
```

src/lib/PreviewUI.js
```diff
@@ -21,9 +21,9 @@ import {
     SELECTOR_BOX_PREVIEW_LOADING_WRAPPER,
     SELECTOR_BOX_PREVIEW_LOGO_CUSTOM,
     SELECTOR_BOX_PREVIEW_LOGO_DEFAULT,
-    SELECTOR_BOX_PREVIEW,
     SELECTOR_NAVIGATION_LEFT,
-    SELECTOR_NAVIGATION_RIGHT
+    SELECTOR_NAVIGATION_RIGHT,
+    SELECTOR_BOX_PREVIEW
```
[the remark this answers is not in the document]
ConradJChan: will remove

src/lib/_common.scss
```diff
@@ -146,6 +145,13 @@ $header-height: 48px;
     &.bp-dark {
         background-color: $black;
     }
+
+    .bp-content {
+        align-items: center;
+        display: flex;
+        flex: 1 1 auto;
+        position: relative;
```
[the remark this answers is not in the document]
ConradJChan: It seems ok in Chrome, my testing in IE and Edge will probably inform this

src/lib/viewers/BaseViewer.js
ConradJChan: ok I'll create a util

src/lib/viewers/doc/DocBaseViewer.js
ConradJChan: Well each viewer has it's own concept of what the viewer wrapper is, whether it is a 'div' or 'pre' or the classes applied to it. Also, it only applies to documents currently so doing it in BaseViewer doesn't seem correct?

jeremypress: good point, not worth passing in the Element we want into the super call of setup.

PR #880, opened 2018-12-13
New: Virtual scroll for the thumbnails sidebar

ConradJChan:
virtual scroll

Unit tests to follow in separate PR

### DIFF
--- a/src/lib/Preview.js
+++ b/src/lib/Preview.js
@@ -931,7 +931,8 @@ class Preview extends EventEmitter {
         this.options.fileOptions = options.fileOptions || {};
 
         // Option to enable use of thumbnails sidebar for document types
-        this.options.enableThumbnailsSidebar = !!options.enableThumbnailsSidebar;
+        // this.options.enableThumbnailsSidebar = !!options.enableThumbnailsSidebar;
+        this.options.enableThumbnailsSidebar = true;
 
         // Prefix any user created loaders before our default ones
         this.loaders = (options.loaders || []).concat(loaderList);
--- a/src/lib/Preview.js
+++ b/src/lib/Preview.js
@@ -932,7 +932,7 @@ class Preview extends EventEmitter {
 
         // Option to enable use of thumbnails sidebar for document types
         // this.options.enableThumbnailsSidebar = !!options.enableThumbnailsSidebar;
-        this.options.enableThumbnailsSidebar = true;
+        this.options.enableThumbnailsSidebar = !!options.enableThumbnailsSidebar;
 
         // Prefix any user created loaders before our default ones
         this.loaders = (options.loaders || []).concat(loaderList);
--- a/src/lib/VirtualScroller.js
+++ b/src/lib/VirtualScroller.js
@@ -80,9 +80,9 @@ class VirtualScroller {
         this.containerHeight = config.containerHeight;
         this.renderItemFn = config.renderItemFn;
         this.margin = config.margin || 0;
-        this.totalViewItems = Math.ceil(this.containerHeight / (this.itemHeight + this.margin));
+        this.totalViewItems = Math.floor(this.containerHeight / (this.itemHeight + this.margin));
         this.maxBufferHeight = this.totalViewItems * this.itemHeight;
-        this.maxRenderedItems = this.totalViewItems * BUFFERED_ITEM_MULTIPLIER;
+        this.maxRenderedItems = (this.totalViewItems + 1) * BUFFERED_ITEM_MULTIPLIER;
 
         // Create the scrolling container element
         this.scrollingEl = document.createElement('div');
--- a/src/lib/VirtualScroller.js
+++ b/src/lib/VirtualScroller.js
@@ -1,0 +1,183 @@
+import isFinite from 'lodash/isFinite';
+
+class VirtualScroller {
+    constructor(container) {
+        /*
+            config {
+                container:          HTMLElement
+                totalNumItems:      Number
+                itemHeight:         Number
+                maxItemsRendered:   Number
+                containerHeight:    Number
+                renderItemFn:       Function
+            }
+         */
+        this.container = container;
+        this.previousScrollTop = 0;
+        this.renderedItems = {};
+
+        this.getPositionOfHighestItem = this.getPositionOfHighestItem.bind(this);
+        this.getPositionOfLowestItem = this.getPositionOfLowestItem.bind(this);
+        this.handleOnScroll = this.handleOnScroll.bind(this);
+        this.pruneRows = this.pruneRows.bind(this);
+        this.renderRows = this.renderRows.bind(this);
+    }
+
+    destroy() {
+        if (this.containerEl) {
+            this.containerEl.remove();
+        }
+
+        this.renderedItems = {};
+
+        this.containerEl = null;
+        this.contentEl = null;
+    }
+
+    init(config) {
+        if (!config.totalNumItems || !config.itemHeight) {
+            throw new Error('Need to provide totalNumItems and itemHeight');
+        }
+
+        this.totalNumItems = config.totalNumItems;
+        this.itemHeight = config.itemHeight;
+        this.containerHeight = config.containerHeight;
+        this.pageSize = Math.ceil(this.containerHeight / this.itemHeight);
+        this.maxItemsRendered = config.maxItemsRendered || Math.ceil(this.pageSize * 1.5);
+        this.renderItemFn = config.renderItemFn;
+        this.marginTop = config.marginTop || 0;
+        this.marginBottom = config.marginBottom || 0;
+
+        this.containerEl = document.createElement('div');
+        this.containerEl.className = 'vs-container';
+
+        this.contentEl = document.createElement('div');
+        this.contentEl.className = 'vs-content-container';
+        this.contentEl.style.height = `${this.totalNumItems * this.itemHeight + this.marginTop + this.marginBottom}px`;
+
+        this.containerEl.appendChild(this.contentEl);
+        this.container.appendChild(this.containerEl);
+
+        this.renderRows(0, Math.min(this.maxItemsRendered, this.totalNumItems));
+
+        this.bindDOMListeners();
+    }
+
+    bindDOMListeners() {
+        this.containerEl.addEventListener('scroll', this.handleOnScroll);
+    }
+
+    handleOnScroll(e) {
+        const { scrollTop } = e.target;
+        console.log(scrollTop);
+        const direction = scrollTop - this.prevScrollTop > 0 ? 'down' : 'up';
+        this.prevScrollTop = scrollTop;
+
+        if (direction === 'down') {
+            const lowestTop = this.getPositionOfLowestItem();
+            if (lowestTop - (scrollTop + this.containerHeight) < this.itemHeight) {
+                // add next pages
+                const newRowsOffset = Math.floor(lowestTop / this.itemHeight) + 1;
+                const rowsToRender =
+                    newRowsOffset + this.pageSize > this.totalNumItems
+                        ? this.totalNumItems - newRowsOffset
+                        : this.pageSize;
+
+                if (rowsToRender > 0) {
+                    this.renderRows(newRowsOffset, rowsToRender);
+
+                    // prune previous pages
+                    const removeRowsOffset =
+                        Math.floor(Math.max(scrollTop - this.containerHeight, 0) / this.itemHeight) - 1;
+                    const startIndex = parseInt(this.contentEl.firstElementChild.dataset.item, 10);
+                    this.pruneRows(startIndex, removeRowsOffset);
+                }
+            }
+        } else {
+            const highestTop = this.getPositionOfHighestItem();
+            if (scrollTop - highestTop < this.itemHeight) {
+                // add next pages
+                const nextRowOffset = Math.floor(highestTop / this.itemHeight);
+                const newRowsOffset = Math.max(nextRowOffset - this.pageSize, 0);
+                const rowsToRender = nextRowOffset - this.pageSize > 0 ? this.pageSize : nextRowOffset;
+
+                if (rowsToRender > 0) {
+                    this.renderRows(newRowsOffset, rowsToRender, true);
+
+                    // prune unnecessary pages
+                    const removeRowsOffset = Math.ceil((scrollTop + 2 * this.containerHeight) / this.itemHeight);
+                    const removeEndOffset = parseInt(this.contentEl.lastElementChild.dataset.item, 10);
+                    this.pruneRows(removeRowsOffset, removeEndOffset);
+                }
+            }
+        }
+    }
+
+    getPositionOfLowestItem() {
+        const bottomStyle = this.contentEl.lastElementChild.style.top;
+        return bottomStyle.substr(0, bottomStyle.indexOf('px'));
+    }
+
+    getPositionOfHighestItem() {
+        const topStyle = this.contentEl.firstElementChild.style.top;
+        return topStyle.substr(0, topStyle.indexOf('px'));
+    }
+
+    pruneRows(startIndex, endIndex) {
+        if (!isFinite(startIndex) || !isFinite(endIndex) || endIndex < startIndex) {
+            return;
+        }
+
+        console.log(`pruning rows starting from ${startIndex} to ${endIndex}`);
+        for (let i = startIndex; i <= endIndex; i++) {
+            if (this.renderedItems[i]) {
+                this.renderedItems[i].remove();
+                this.renderedItems[i] = null;
+            }
+        }
+    }
+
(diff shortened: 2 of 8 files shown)


boxcla:

src/lib/VirtualScroller.js
```diff
@@ -0,0 +1,183 @@
...
+    }
+
+    /**
+     * Initializes the virtual scroller
+     *
+     * @param {Object} config - The config
+     * @return {void}
+     */
+    init(config) {
+        if (!config.totalItems || !config.itemHeight) {
```
jstoffan: Are there other required properties? It looks like `containerHeight` and `renderItemFn` are also needed.

src/lib/VirtualScroller.js
```diff
@@ -0,0 +1,183 @@
...
+        this.marginBottom = config.marginBottom || 0;
+
+        // The number of items that can fit in view
+        this.totalViewItems = Math.ceil(this.containerHeight / this.itemHeight);
+
+        // The height of the buffer before virtual scroll renders the next set
+        this.maxBufferHeight = this.totalViewItems * this.itemHeight;
+
+        // The max number of items to render at any one given time
+        this.maxRenderedItems = Math.ceil(this.totalViewItems * 3);
```
jstoffan: Can you add a variable somewhere for the buffer multiple, just to clarify its intention?

src/lib/VirtualScroller.js
```diff
@@ -0,0 +1,183 @@
...
+
+        // Create the scrolling container element
+        this.containerEl = document.createElement('div');
+        this.containerEl.className = 'vs-container';
+
+        // Create the true height content container
+        this.contentEl = document.createElement('div');
+        this.contentEl.className = 'vs-content-container';
+        this.contentEl.style.height = `${this.totalItems * this.itemHeight + this.marginTop + this.marginBottom}px`;
```
jstoffan: It looks like we're recalculating here. Can we use `this.maxBufferHeight`, instead?

src/lib/VirtualScroller.js
```diff
@@ -0,0 +1,183 @@
...
+        // The max number of items to render at any one given time
+        this.maxRenderedItems = Math.ceil(this.totalViewItems * 3);
+
+        // Create the scrolling container element
+        this.containerEl = document.createElement('div');
+        this.containerEl.className = 'vs-container';
+
+        // Create the true height content container
+        this.contentEl = document.createElement('div');
+        this.contentEl.className = 'vs-content-container';
```
jstoffan: Try to avoid `-container` class names when you can.

src/lib/VirtualScroller.js
```diff
@@ -0,0 +1,183 @@
...
+        this.totalViewItems = Math.ceil(this.containerHeight / this.itemHeight);
+
+        // The height of the buffer before virtual scroll renders the next set
+        this.maxBufferHeight = this.totalViewItems * this.itemHeight;
+
+        // The max number of items to render at any one given time
+        this.maxRenderedItems = Math.ceil(this.totalViewItems * 3);
+
+        // Create the scrolling container element
+        this.containerEl = document.createElement('div');
```
jstoffan: Should this be a `nav` tag?

src/lib/VirtualScroller.js
```diff
@@ -0,0 +1,183 @@
...
+        this.containerEl.className = 'vs-container';
+
+        // Create the true height content container
+        this.contentEl = document.createElement('div');
```
jstoffan: The architecture is going to make semantic markup a little difficult. The thumbnail list should probalby be a an `ol`, thumbnails `li` containing a button or an anchor for the click target.

src/lib/VirtualScroller.js
```diff
@@ -0,0 +1,183 @@
...
+     * @return {void}
+     */
+    onScrollHandler(e) {
+        const { scrollTop } = e.target;
+
+        if (Math.abs(scrollTop - this.previousScrollTop) > this.maxBufferHeight) {
+            // The first item to be re-rendered will be a totalViewItems height up from the
+            // item at the current location
+            const firstIndex = Math.floor(scrollTop / this.itemHeight) - this.totalViewItems;
+            this.renderItems(firstIndex < 0 ? 0 : firstIndex);
```
jstoffan: Consider using `Math.max(firstIndex, 0)` rather than a ternary.

src/lib/VirtualScroller.js
```diff
@@ -0,0 +1,183 @@
...
+
+        let renderedThumbnail;
+        try {
+            renderedThumbnail = this.renderItemFn.call(this, rowIndex);
+        } catch (err) {
+            // eslint-disable-next-line
+            console.error(`Error rendering thumbnail - ${err}`);
+        }
+
+        rowEl.style.top = `${topPosition}px`;
```
jstoffan: Consider moving the row "component" to its own file. Also, if you want to get fancy, you could use template strings for these elements; it'd give you a JSX-ish syntax.

src/lib/VirtualScroller.js
```diff
@@ -0,0 +1,183 @@
...
+        let numItemsRendered = 0;
+        const fragment = document.createDocumentFragment();
+
+        while (numItemsRendered < count) {
+            const rowEl = this.renderItem(offset + numItemsRendered, this.itemHeight);
+            fragment.appendChild(rowEl);
+            numItemsRendered += 1;
+        }
+
+        this.contentEl.innerHTML = '';
```
jstoffan: I'm finding conflicting information on  the performance difference between clearing content via innerHtml vs. looping/removing DOM nodes. In most cases it wouldn't matter, but here it might make a difference if the latter is significantly faster in most browsers.

Take a look at https://stackoverflow.com/a/3955238

src/lib/VirtualScroller.js
```diff
@@ -0,0 +1,183 @@
...
+            renderedThumbnail = this.renderItemFn.call(this, rowIndex);
+        } catch (err) {
+            // eslint-disable-next-line
+            console.error(`Error rendering thumbnail - ${err}`);
+        }
+
+        rowEl.style.top = `${topPosition}px`;
+        rowEl.style.height = `${this.itemHeight}px`;
+        rowEl.classList.add('vs-content-item');
+        rowEl.dataset.item = rowIndex;
```
jstoffan: Unused?

src/lib/VirtualScroller.js
```diff
@@ -0,0 +1,183 @@
...
+        // then just render the difference
+        if (count + offset > this.totalItems) {
+            count = this.totalItems - offset;
+        }
+
+        let numItemsRendered = 0;
+        const fragment = document.createDocumentFragment();
+
+        while (numItemsRendered < count) {
+            const rowEl = this.renderItem(offset + numItemsRendered, this.itemHeight);
```
jstoffan: It looks like the second param here isn't being used by `renderItem`.

src/lib/VirtualScroller.scss
```diff
@@ -0,0 +1,16 @@
+.vs-container {
+    flex: 1 0 auto;
+    overflow-y: auto;
+
+    .vs-content-container {
```
jstoffan: Try to avoid nesting in SCSS when possible. Keeping specificity low almost always makes things easier to maintain.

src/lib/viewers/doc/DocBaseViewer.js
```diff
@@ -126,6 +127,8 @@ class DocBaseViewer extends BaseViewer {
             this.thumbnailsSidebarEl = document.createElement('div');
             this.thumbnailsSidebarEl.className = `${CLASS_BOX_PREVIEW_THUMBNAILS_CONTAINER} ${CLASS_HIDDEN}`;
             this.containerEl.parentNode.insertBefore(this.thumbnailsSidebarEl, this.containerEl);
+
+            this.thumbnailsSidebar = new VirtualScroller(this.thumbnailsSidebarEl);
```
jstoffan: Consider grouping all the thumbnail-related statements together.

src/lib/viewers/doc/DocBaseViewer.js
```diff
@@ -1000,6 +1007,24 @@ class DocBaseViewer extends BaseViewer {
...
             this.setupPageIds();
         }
+
+        this.initThumbnails();
+    }
+
+    initThumbnails() {
+        this.thumbnailsSidebar.init({
+            totalItems: this.pdfViewer.pagesCount,
+            itemHeight: 115,
```
jstoffan: Can we use the height of the first page with some defined lower limit as our starting value rather than hardcoding it?

src/lib/viewers/doc/DocBaseViewer.js
```diff
@@ -1000,6 +1007,24 @@ class DocBaseViewer extends BaseViewer {
...
+            containerHeight: this.docEl.clientHeight,
+            marginTop: 7.5,
```
jstoffan: Is this the pixel value in the design? Can you pass in a class and allow these to live in CSS, instead?

src/lib/viewers/doc/DocBaseViewer.js
```diff
@@ -1000,6 +1007,24 @@ class DocBaseViewer extends BaseViewer {
...
+            marginBottom: 7.5,
+            renderItemFn: (page) => {
+                const thumbnail = document.createElement('button');
```
jstoffan: We're creating/trashing a lot of objects, but given the low number of living objects at any given time, it should be okay.

src/lib/viewers/doc/DocBaseViewer.js
```diff
@@ -1000,6 +1007,24 @@ class DocBaseViewer extends BaseViewer {
...
+                thumbnail.className = 'bp-thumbnail';
+                thumbnail.textContent = `${page}`;
```
jstoffan: Does this work properly in IE11?

>Unlike textContent, altering innerText in Internet Explorer (version 11 and below) removes child nodes from the element and permanently destroys all descendant text nodes.

https://developer.mozilla.org/en-US/docs/Web/API/Node/textContent#Differences_from_innerText

src/lib/viewers/doc/_docBase.scss
```diff
@@ -3,8 +3,19 @@
...
     }
+
+    .bp-thumbnail {
+        align-items: center;
+        background-color: rgba(0, 0, 255, .5);
+        border-radius: 4px;
+        display: flex;
+        flex: 1 1 auto;
+        justify-content: center;
+        margin: 7.5px 15px;
```
jstoffan: Try to avoid fractional pixels when possible.

src/lib/viewers/doc/_docBase.scss
```diff
@@ -3,8 +3,19 @@
 .bp {
     .bp-thumbnails-container {
         border-right: solid 1px $seesee;
+        display: flex;
         flex: 0 0 180px;
     }
+
+    .bp-thumbnail {
```
jstoffan: You may want to look into adding an `object-fit` rule when images come into play here, depending on what the design looks like.

src/lib/VirtualScroller.js
```diff
@@ -0,0 +1,183 @@
...
+    /**
+     * Handler for 'scroll' event
+     *
+     * @param {Event} e - The scroll event
+     * @return {void}
+     */
+    onScrollHandler(e) {
+        const { scrollTop } = e.target;
+
+        if (Math.abs(scrollTop - this.previousScrollTop) > this.maxBufferHeight) {
```
jeremypress: We're doing a decent amount of math here and scroll events fire extremely often. Do you think the frequency of scroll events could cause a perf problem here?

src/lib/VirtualScroller.js
```diff
@@ -0,0 +1,183 @@
...
+
+        // The height of the buffer before virtual scroll renders the next set
+        this.maxBufferHeight = this.totalViewItems * this.itemHeight;
+
+        // The max number of items to render at any one given time
+        this.maxRenderedItems = Math.ceil(this.totalViewItems * 3);
+
+        // Create the scrolling container element
+        this.containerEl = document.createElement('div');
+        this.containerEl.className = 'vs-container';
```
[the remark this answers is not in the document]
ConradJChan: will change to `.bp-vs` (removing container like @jstoffan suggested in another comment)

src/lib/VirtualScroller.js
ConradJChan: In this case, it makes sense to be a `nav`, but the VirtualScroller is built a bit generically and if used elsewhere `nav` might not be semantically correct. Thoughts?

src/lib/VirtualScroller.js
```diff
@@ -0,0 +1,183 @@
...
+
+        // Create the true height content container
+        this.contentEl = document.createElement('div');
+        this.contentEl.className = 'vs-content-container';
+        this.contentEl.style.height = `${this.totalItems * this.itemHeight + this.marginTop + this.marginBottom}px`;
+
+        this.containerEl.appendChild(this.contentEl);
+        this.container.appendChild(this.containerEl);
```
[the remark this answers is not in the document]
ConradJChan: Yeah let me rename the variables to make more sense

src/lib/VirtualScroller.js
ConradJChan: This is the true height of all the items in the list, so it needs to be calculated for the first time

jstoffan: Ah, gotcha. Reading comprehension fail on my part.

src/lib/viewers/doc/DocBaseViewer.js
ConradJChan: This is just to create the temporary thumbnails for this PR, I'll keep that in mind for when I do a PR to cut the thumbnail images

src/lib/VirtualScroller.js
ConradJChan: definitely room for improvement, let me address that in a follow up PR

src/lib/Preview.js
```diff
@@ -931,7 +931,8 @@ class Preview extends EventEmitter {
         this.options.fileOptions = options.fileOptions || {};
 
         // Option to enable use of thumbnails sidebar for document types
-        this.options.enableThumbnailsSidebar = !!options.enableThumbnailsSidebar;
+        // this.options.enableThumbnailsSidebar = !!options.enableThumbnailsSidebar;
+        this.options.enableThumbnailsSidebar = true;
```
jstoffan: Is this intentional?

src/lib/VirtualScroller.js
```diff
@@ -0,0 +1,239 @@
...
+     * @return {VirtualScroller} Instance of VirtualScroller
+     */
+    constructor(anchor) {
+        this.anchorEl = anchor;
+
+        this.previousScrollTop = 0;
+
+        this.createListElement = this.createListElement.bind(this);
+        this.onScrollHandler = this.onScrollHandler.bind(this);
+        this.throttledOnScrollHandler = throttle(this.onScrollHandler, 50);
```
jstoffan: May as well just do this all inline:
```
this.onScrollHandler = throttle(this.onScrollHandler.bind(this), 50);

File: src/lib/VirtualScroller.js [hunk 25]
```diff
@@ -0,0 +1,239 @@
...
+
+        // Create a new list element to be swapped out for the existing one
+        const newListEl = this.createListElement();
+        while (numItemsRendered < count) {
+            const rowEl = this.renderItem(offset + numItemsRendered);
+            newListEl.appendChild(rowEl);
+            numItemsRendered += 1;
+        }
+
+        this.scrollingEl.replaceChild(newListEl, this.listEl);
```
jstoffan: Nice.

src/lib/viewers/doc/DocBaseViewer.js
```diff
@@ -1000,6 +1006,36 @@ class DocBaseViewer extends BaseViewer {
...
+            this.initThumbnails();
+        }
+    }
+
+    initThumbnails() {
+        this.thumbnailsSidebar = new VirtualScroller(this.thumbnailsSidebarEl);
+
+        // Get the first page of the document, and use its dimensions
+        // to set the thumbnails size of the thumbnails sidebar
+        this.pdfViewer.pdfDocument.getPage(1).then((page) => {
```
jstoffan: Consider moving this logic to a ThumbnailSidebar-specific component file.

src/lib/viewers/doc/DocBaseViewer.js
```diff
@@ -1000,6 +1006,36 @@ class DocBaseViewer extends BaseViewer {
...
+            const desiredWidth = DEFAULT_THUMBNAILS_SIDEBAR_WIDTH;
+            const viewport = page.getViewport(1);
+            const scale = desiredWidth / viewport.width;
```
jstoffan: Is it possible for this to return undefined or zero? Can `viewport` be undefined?

PR #883, opened 2018-12-27
Update: Continuous progress bar through retries

ConradJChan:
continuous progress bar

### DIFF
--- a/src/lib/Preview.js
+++ b/src/lib/Preview.js
@@ -795,7 +795,10 @@ class Preview extends EventEmitter {
         // Set the authorization token
         this.options.token = tokenMap[this.file.id];
 
-        this.setupUI();
+        // Do nothing if container is already setup and in the middle of retrying
+        if (!(this.ui.isSetup() && this.retryCount > 0)) {
+            this.setupUI();
+        }
 
         // Load from cache if the current file is valid, otherwise load file info from server
         if (checkFileValid(this.file)) {
@@ -814,11 +817,6 @@ class Preview extends EventEmitter {
      * @return {void}
      */
     setupUI() {
-        // Do nothing if container is already setup and in the middle of retrying
-        if (this.container && this.retryCount > 0) {
-            return;
-        }
-
         // Setup the shell
         this.container = this.ui.setup(
             this.options,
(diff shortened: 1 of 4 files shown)


src/lib/Preview.js
```diff
@@ -814,6 +814,11 @@ class Preview extends EventEmitter {
      * @return {void}
      */
     setupUI() {
+        // Do nothing if container is already setup and in the middle of retrying
+        if (this.container && this.retryCount > 0) {
```
jeremypress: would this.open be helpful here?

jstoffan: This seems like a somewhat risky place for this change, since the `this.ui.setup` method is intended to destroy/replace the existing UI. Does this cover edge cases where a particular file is being retried and the user navigates to another file?

ConradJChan: Looking at the code for navigating prev/next files, maybe it would be better to use `retryTimeout` instead of relying on `retryCount` since I don't see that getting reset in the navigation flow.

The other approach I was thinking was to pass in a retry boolean to be more explicit when calling `load` via `handleFetchError` but I would need to propagate that value through a few functions and result in a bigger change, let me take another look

ConradJChan: I'm not sure, it seems like `this.open` is set when `load` is first called and is no guarantee of the preview UI having been setup already

DanDeMicco: It doesn't look like `this.container` gets cleared out in the destroy or navigate, so this may cause issues

ConradJChan: created the method `PreviewUI.isSetup()` instead

src/lib/Preview.js
```diff
@@ -795,7 +795,10 @@ class Preview extends EventEmitter {
         // Set the authorization token
         this.options.token = tokenMap[this.file.id];
 
-        this.setupUI();
+        // Do nothing if container is already setup and in the middle of retrying
+        if (!(this.ui.isSetup() && this.retryCount > 0)) {
+            this.setupUI();
```
jeremypress: Do we want to add this to the `reload` case? We call it a few times throughout preview.js

ConradJChan: I'm not sure, IMO the reload intention is to wipe everything clean and start over from scratch in which case, it seems to make sense to always re setup the UI

PR #888, opened 2019-01-10
Chore: unit tests for ThumbnailsSidebar and VirtualScroller

ConradJChan:


boxcla:

src/lib/VirtualScroller.js
```diff
@@ -204,12 +204,24 @@ class VirtualScroller {
      * @return {Object} - info object
      */
     getCurrentListInfo() {
-        const { firstElementChild, lastElementChild } = this.listEl;
+        const { firstElementChild, lastElementChild, children } = this.listEl;
 
-        const curStartOffset = firstElementChild ? Number.parseInt(firstElementChild.dataset.bpVsRowIndex, 10) : -1;
-        const curEndOffset = lastElementChild ? Number.parseInt(lastElementChild.dataset.bpVsRowIndex, 10) : -1;
+        // Parse the row index from the data-attribute
+        let curStartOffset = firstElementChild ? Number.parseInt(firstElementChild.dataset.bpVsRowIndex, 10) : -1;
```
jeremypress: Might want to double check support for dataset on Safari. MDN has it as `?`. https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/dataset

src/lib/VirtualScroller.js
```diff
@@ -290,12 +302,25 @@ class VirtualScroller {
      * @return {void}
      */
     cloneItems(newListEl, oldListEl, start, end) {
-        if (!newListEl || !oldListEl || start < 0 || end < 0) {
+        if (
+            !newListEl ||
+            !oldListEl ||
+            start === null ||
```
jeremypress: would `!start` and `!end` suffice here instead of explicitly checking `null` and `undefined`?

src/lib/__tests__/ThumbnailsSidebar-test.js
```diff
@@ -0,0 +1,224 @@
...
+
+            return createImagePromise.then(() => {
+                expect(stubs.appendChild).to.be.called;
+                expect(stubs.addClass).to.be.called;
+            });
+        });
+    });
+
+    describe('createThumbnailImage()', () => {
+        it('should resolve immediately if is in cache', () => {
```
jeremypress: *if the image is in cache

src/lib/__tests__/VirtualScroller-test.js
```diff
@@ -235,4 +251,177 @@ describe('VirtualScroller', () => {
...
+
+    describe('cloneItems()', () => {
+        let newListEl;
+
+        beforeEach(() => {
+            stubs.appendChild = sandbox.stub();
+            newListEl = { appendChild: stubs.appendChild };
+        });
+
+        const paramaterizedTests = [
```
jeremypress: 👍

PR #889, opened 2019-01-11
Update: Thumbnails navigation

ConradJChan:
Tests to follow 

thumbnail-navigation

### DIFF
--- a/src/lib/ThumbnailsSidebar.js
+++ b/src/lib/ThumbnailsSidebar.js
@@ -101,7 +101,7 @@ class ThumbnailsSidebar {
     /**
      * Initializes the Thumbnails Sidebar
      *
-     * @param {Object} options - options for the Thumbnails Sidebar
+     * @param {Object} [options] - options for the Thumbnails Sidebar
      * @return {void}
      */
     init(options) {
--- a/src/lib/ThumbnailsSidebar.js
+++ b/src/lib/ThumbnailsSidebar.js
@@ -2,8 +2,10 @@ import isFinite from 'lodash/isFinite';
 import VirtualScroller from './VirtualScroller';
 
 const CLASS_BOX_PREVIEW_THUMBNAIL = 'bp-thumbnail';
+const CLASS_BOX_PREVIEW_THUMBNAIL_IMAGE = 'bp-thumbnail-image';
 const CLASS_BOX_PREVIEW_THUMBNAIL_IMAGE_LOADED = 'bp-thumbnail-image-loaded';
 const CLASS_BOX_PREVIEW_THUMBNAIL_IS_SELECTED = 'bp-thumbnail-is-selected';
+const CLASS_BOX_PREVIEW_THUMBNAIL_PAGE_NUMBER = 'bp-thumbnail-page-number';
 const DEFAULT_THUMBNAILS_SIDEBAR_WIDTH = 150;
 const THUMBNAIL_WIDTH_MAX = 210;
 const THUMBNAIL_MARGIN = 15;
@@ -182,8 +184,9 @@ class ThumbnailsSidebar {
         thumbnailEl.appendChild(this.createPageNumber(pageNum));
 
         if (pageNum === this.currentPage) {
-            thumbnailEl.classList.add('bp-thumbnail-is-selected');
+            thumbnailEl.classList.add(CLASS_BOX_PREVIEW_THUMBNAIL_IS_SELECTED);
         }
+
         return thumbnailEl;
     }
 
@@ -231,7 +234,7 @@ class ThumbnailsSidebar {
             })
             .then(() => {
                 const imageEl = document.createElement('img');
-                imageEl.classList.add('bp-thumbnail-image');
+                imageEl.classList.add(CLASS_BOX_PREVIEW_THUMBNAIL_IMAGE);
                 imageEl.src = canvas.toDataURL();
 
                 // Cache this image element for future use
@@ -249,7 +252,7 @@ class ThumbnailsSidebar {
      */
     createPageNumber(pageNumber) {
         const pageNumberEl = document.createElement('div');
-        pageNumberEl.className = 'bp-thumbnail-page-number';
+        pageNumberEl.className = CLASS_BOX_PREVIEW_THUMBNAIL_PAGE_NUMBER;
         pageNumberEl.textContent = `${pageNumber}`;
         return pageNumberEl;
     }
--- a/src/lib/ThumbnailsSidebar.js
+++ b/src/lib/ThumbnailsSidebar.js
@@ -1,6 +1,9 @@
 import isFinite from 'lodash/isFinite';
 import VirtualScroller from './VirtualScroller';
 
+const CLASS_BOX_PREVIEW_THUMBNAIL = 'bp-thumbnail';
+const CLASS_BOX_PREVIEW_THUMBNAIL_IMAGE_LOADED = 'bp-thumbnail-image-loaded';
+const CLASS_BOX_PREVIEW_THUMBNAIL_IS_SELECTED = 'bp-thumbnail-is-selected';
 const DEFAULT_THUMBNAILS_SIDEBAR_WIDTH = 150;
 const THUMBNAIL_WIDTH_MAX = 210;
 const THUMBNAIL_MARGIN = 15;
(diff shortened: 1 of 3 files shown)


boxcla:

src/lib/ThumbnailsSidebar.js
```diff
@@ -120,9 +174,16 @@ class ThumbnailsSidebar {
...
-        thumbnailEl.className = 'bp-thumbnail';
+        const pageNum = itemIndex + 1;
+
+        thumbnailEl.className = CLASS_BOX_PREVIEW_THUMBNAIL;
         thumbnailEl.setAttribute('type', 'button');
-        thumbnailEl.appendChild(this.createPageNumber(itemIndex + 1));
+        thumbnailEl.dataset.bpPageNum = pageNum;
+        thumbnailEl.appendChild(this.createPageNumber(pageNum));
+
+        if (pageNum === this.currentPage) {
```
jeremypress: I feel like the setting of selected thumbnail should be handled by the caller, not inside this method itself. Do we call this in many places?

src/lib/ThumbnailsSidebar.js
```diff
@@ -120,9 +174,16 @@ class ThumbnailsSidebar {
...
+            thumbnailEl.classList.add('bp-thumbnail-is-selected');
```
jeremypress: You're using constants for selectors like this in other places

src/lib/ThumbnailsSidebar.js
```diff
@@ -170,8 +231,8 @@ class ThumbnailsSidebar {
             })
             .then(() => {
                 const imageEl = document.createElement('img');
+                imageEl.classList.add('bp-thumbnail-image');
```
jeremypress: same here on the constants

src/lib/viewers/doc/DocBaseViewer.js
```diff
@@ -444,6 +444,10 @@ class DocBaseViewer extends BaseViewer {
 
         this.pdfViewer.currentPageNumber = parsedPageNumber;
         this.cachePage(this.pdfViewer.currentPageNumber);
+
+        if (this.thumbnailsSidebar) {
+            this.thumbnailsSidebar.setCurrentPage(parsedPageNumber);
```
jeremypress: Why do we need to do this here? The data goes from sidebar -> doc base, so why are we using doc base to report back?

src/lib/viewers/doc/_docBase.scss
```diff
@@ -1,36 +1,55 @@
 @import '../../boxuiVariables';
 
+$thumbnail-border-radius: 4px;
+
 .bp {
     .bp-thumbnails-container {
         border-right: solid 1px $seesee;
         display: flex;
-        flex: 0 0 180px;
+        flex: 0 0 181px;
```
jeremypress: why +1?

src/lib/viewers/doc/_docBase.scss
```diff
@@ -1,36 +1,55 @@
...
+    .bp-thumbnail-image {
+        border-radius: $thumbnail-border-radius;
+        pointer-events: none;
+        width: 100%;
+    }
+
+    .bp-thumbnail.bp-thumbnail-is-selected::after {
+        border: 4px solid $box-blue;
+        border-radius: $thumbnail-border-radius;
+        bottom: -2px;
```
jeremypress: This smells a bit weird, is the size changing which requires these negative values?

src/lib/ThumbnailsSidebar.js
```diff
@@ -192,6 +253,40 @@ class ThumbnailsSidebar {
...
+
+    /**
+     * Sets the currently selected page
+     *
+     * @param {number} pageNumber - The page number to set to selected
+     * @return {void}
+     */
+    setCurrentPage(pageNumber) {
+        const parsedPageNumber = parseInt(pageNumber, 10);
+        if (!parsedPageNumber || parsedPageNumber < 1 || parsedPageNumber > this.pdfViewer.pagesCount) {
```
jeremypress: How generic to we want to make this? This wouldn't work for multi page images.

src/lib/ThumbnailsSidebar.js
ConradJChan: I just made constants for reoccurring strings, but I can make them all constants

src/lib/ThumbnailsSidebar.js
ConradJChan: yeah, I was debating this as well, but I figured that would be a refactor for later, unless you feel this would be better served to be done upfront

src/lib/viewers/doc/DocBaseViewer.js
ConradJChan: This is for the case where the page controls are used directly to change page (up, down, text input) in which case I need to communicate back to the thumbnails sidebar to also update the current page

src/lib/viewers/doc/_docBase.scss
ConradJChan: To account for the 1px border on the right, so that the remaining width is actually 180

src/lib/viewers/doc/_docBase.scss
ConradJChan: no this is a trick I found which allows me to put the selected border on the *outside* of the thumbnail button rather than taking the border width and height away from the thumbnail button itself. What I want to avoid is the thumbnail image resizing whenever the selection is applied or not

src/lib/ThumbnailsSidebar.js
ConradJChan: let me see how to move this out

src/lib/viewers/doc/_docBase.scss
jeremypress: cool, can you drop a comment explaining?

src/lib/ThumbnailsSidebar.js
jeremypress: fine by me!

src/lib/ThumbnailsSidebar.js
ConradJChan: As we discussed, I'm going to leave this in since the performance cost should be minimal and what this does is allow the placeholder thumbnail to have the selected class & appearance while you're scrolling through

src/lib/viewers/doc/_docBase.scss
ConradJChan: will do

src/lib/ThumbnailsSidebar.js
```diff
@@ -51,16 +92,29 @@ class ThumbnailsSidebar {
...
+        this.currentThumbnails = [];
+        this.currentPage = null;
+
+        this.anchorEl.removeEventListener('click', this.thumbnailClickHandler);
     }
 
     /**
      * Initializes the Thumbnails Sidebar
      *
+     * @param {Object} options - options for the Thumbnails Sidebar
```
JustinHoldstock: If options is optional, as seen by the guard below, you can wrap the comment docs in square braces.

`@param {Object} [options] - ...`

src/lib/viewers/doc/_docBase.scss
JustinHoldstock: Can you add a comment beside this, to clarify?

src/lib/ThumbnailsSidebar.js
```diff
@@ -31,11 +42,41 @@ class ThumbnailsSidebar {
...
+     */
+    thumbnailClickHandler(evt) {
+        const { target } = evt;
+
+        // Only care about clicks on the thumbnail element itself.
+        // The image and page number have pointer-events: none so
+        // any click should be the thumbnail element itself.
+        if (target.classList.contains(CLASS_BOX_PREVIEW_THUMBNAIL)) {
+            // Get the page number
+            const { bpPageNum: pageNumStr } = target.dataset;
```
jeremypress: By binding to the anchor el, is there a chance we click on something that's not the thumbnail container? If `dataset` is undefined then trying to grab the page number would throw an exception.

ConradJChan: shouldn't be, but that's the if statement above so that only the element with the `bp-thumbnail` class will be processed as a click which should have the data attribute set for page number. Any other element would just be a `noop`

jeremypress: good point 👍

PR #892, opened 2019-01-15
Update: Emit metrics for Thumbnail Sidebar usage

ConradJChan:
Tests to follow

### DIFF
--- a/src/lib/viewers/doc/DocBaseViewer.js
+++ b/src/lib/viewers/doc/DocBaseViewer.js
@@ -41,7 +41,7 @@ import {
     ICON_THUMBNAILS_TOGGLE
 } from '../../icons/icons';
 import { JS, PRELOAD_JS, CSS } from './docAssets';
-import { ERROR_CODE, VIEWER_EVENT } from '../../events';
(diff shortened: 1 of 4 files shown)


boxcla:

src/lib/viewers/doc/DocBaseViewer.js
```diff
@@ -572,8 +586,17 @@ class DocBaseViewer extends BaseViewer {
      * @param {Object} event - Event object
      * @return {void}
      */
-    emitMetric(event) {
-        super.emitMetric(event.name, event.data);
+    emitMetric({ name, data }) {
+        // If this metric has been emitted already and is on the whitelist of metrics
+        // to be emitted only once per session, then do nothing
+        if (this.emittedMetrics[name] && METRICS_WHITELIST.includes(name)) {
```
jeremypress: I think this logic would be useful at the base layer if we want to prevent multiple emits for other events. Thoughts @JustinHoldstock ?

JustinHoldstock: I think that's a great idea. At this level we could then add in `unique`, or some other property, to the data object that the base layer can check against and add a logging utility method to do the uniqueness check/storage of unique ids

PR #893, opened 2019-01-15
Chore: Upgrade deps and scripts to work with node 10

jstoffan:
I also added an `.npmrc` file and package.json `engines` declaration to replace the current version-checking script.


boxcla:

jstoffan:
Good catch, @jeremypress. I ran the script and didn't see any output. It looks like it's checking the minimum version, which shouldn't conflict with upgrading to node 10. Seems like everything's okay.

.npmrc
```diff
@@ -0,0 +1 @@
+registry=https://registry.npmjs.org/
```
priyajeet: weird that this was missing before :O

priyajeet: in other projects we have both yarnrc and npmrc with this

PR #896, opened 2019-01-18
Update: Handle non uniform thumbnail sizes

ConradJChan:
thumbnails non uniform page sizes

### DIFF
--- a/src/lib/ThumbnailsSidebar.js
+++ b/src/lib/ThumbnailsSidebar.js
@@ -223,10 +223,32 @@ class ThumbnailsSidebar {
         return this.pdfViewer.pdfDocument
             .getPage(itemIndex + 1)
             .then((page) => {
-                const viewport = page.getViewport(1);
-                canvas.width = THUMBNAIL_WIDTH_MAX;
-                canvas.height = THUMBNAIL_WIDTH_MAX / this.pageRatio;
-                const scale = THUMBNAIL_WIDTH_MAX / viewport.width;
+                const { width, height } = page.getViewport(1);
+                // Get the current page w:h ratio in case it differs from the first page
+                const curPageRatio = width / height;
+
+                let canvasWidth;
+                // Handle the case where the current page's w:h ratio is less than the
+                // `pageRatio` which means that this page is probably more portrait than
+                // landscape
+                if (curPageRatio < this.pageRatio) {
+                    // Set the canvas height to that of the thumbnail max height
+                    canvas.height = THUMBNAIL_WIDTH_MAX / this.pageRatio;
+                    // Find the canvas width based on the curent page ratio
+                    canvasWidth = canvas.height * curPageRatio;
+                    canvas.width = canvasWidth;
+                } else {
+                    // In case the current page ratio is same as the first page
+                    // or in case it's larger (which means that it's wider), keep
+                    // the width at the max thumbnail width
+                    canvasWidth = THUMBNAIL_WIDTH_MAX;
+                    canvas.width = canvasWidth;
+                    // Find the height based on the current page ratio
+                    canvas.height = THUMBNAIL_WIDTH_MAX / curPageRatio;
+                }
+
+                // The amount for which to scale down the current page
+                const scale = canvasWidth / width;
                 return page.render({
                     canvasContext: canvas.getContext('2d'),
                     viewport: page.getViewport(scale)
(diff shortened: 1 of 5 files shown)


boxcla:

src/lib/ThumbnailsSidebar.js
```diff
@@ -218,30 +220,77 @@ class ThumbnailsSidebar {
...
+                    // Set the canvas height to that of the thumbnail max height
+                    canvas.height = THUMBNAIL_WIDTH_MAX / this.pageRatio;
+                    // Find the canvas width based on the curent page ratio
+                    canvasWidth = canvas.height * curPageRatio;
+                    canvas.width = canvasWidth;
+                } else {
+                    // In case the current page ratio is same as the first page
+                    // or in case it's larger (which means that it's wider), keep
+                    // the width at the max thumbnail width
+                    canvasWidth = THUMBNAIL_WIDTH_MAX;
```
jstoffan: Could this live up with `let canvasWidth = THUMBNAIL_WIDTH_MAX` as a default value?

src/lib/ThumbnailsSidebar.js
```diff
@@ -218,30 +220,77 @@ class ThumbnailsSidebar {
...
+                    canvas.width = canvasWidth;
+                    // Find the height based on the current page ratio
+                    canvas.height = THUMBNAIL_WIDTH_MAX / curPageRatio;
+                }
+
+                // The amount for which to scale down the current page
+                const scale = canvasWidth / width;
```
jstoffan: Are we caching `canvasWidth` because calling `canvas.width` has a performance penalty? Or can we simplify and just pull the value directly?

src/lib/ThumbnailsSidebar.js
```diff
@@ -218,30 +220,77 @@ class ThumbnailsSidebar {
...
+     * @param {number} pageNum  - The page number of the document
+     * @return {string} The data URL of the page image
+     */
+    getThumbnailDataURL(pageNum) {
         const canvas = document.createElement('canvas');
 
         return this.pdfViewer.pdfDocument
-            .getPage(itemIndex + 1)
+            .getPage(pageNum)
             .then((page) => {
```
jstoffan: Is it worth moving these to named class methods to avoid recreating the functions each time? It usually wouldn't matter, but I think we're doing this many times a second on scroll, right?

src/lib/ThumbnailsSidebar.js
```diff
@@ -218,30 +220,77 @@ class ThumbnailsSidebar {
...
+        const imageEl = document.createElement('img');
+        imageEl.classList.add(CLASS_BOX_PREVIEW_THUMBNAIL_IMAGE);
+        imageEl.src = dataUrl;
 
-                return imageEl;
-            });
+        // Add the height and width to the image to be the same as the thumbnail
+        // so that the css `object-fit` rule will work
+        imageEl.style.width = `${DEFAULT_THUMBNAILS_SIDEBAR_WIDTH}px`;
+        imageEl.style.height = `${DEFAULT_THUMBNAILS_SIDEBAR_WIDTH / this.pageRatio}px`;
```
jstoffan: Any rounding required to avoid fractional pixels?

src/lib/viewers/doc/_docBase.scss
```diff
@@ -36,9 +36,8 @@ $thumbnail-border-radius: 4px;
     }
 
     .bp-thumbnail-image {
-        border-radius: $thumbnail-border-radius;
+        object-fit: contain;
```
jstoffan: What's the experience like in IE11? https://developer.mozilla.org/en-US/docs/Web/CSS/object-fit#Browser_compatibility

src/lib/ThumbnailsSidebar.js
```diff
@@ -49,6 +49,8 @@ class ThumbnailsSidebar {
         this.createThumbnailImage = this.createThumbnailImage.bind(this);
         this.generateThumbnailImages = this.generateThumbnailImages.bind(this);
         this.thumbnailClickHandler = this.thumbnailClickHandler.bind(this);
+        this.getThumbnailDataURL = this.getThumbnailDataURL.bind(this);
+        this.createImageEl = this.createImageEl.bind(this);
```
jstoffan: How much would I need to bribe you to alphabetize these?

src/lib/ThumbnailsSidebar.js
ConradJChan: It seems like the modern desktop browsers all support sub pixel rendering right now, i'll investigate and maybe do a follow up PR to see if I can get away from fractional values

src/lib/viewers/doc/_docBase.scss
ConradJChan: updating PR to use `background-image`

PR #900, opened 2019-01-23
Update: Change VirtualScroller.renderItems strategy

ConradJChan:
Changed the strategy from re-creating the `ol` on each render, to manipulating the individual `li` (creating new ones and deleting the old ones outside the "window" into the list). 

This is because Safari wasn't smart about the cloned elements in the previous strategy and would cause a flicker of the thumbnails on each `renderItems` call

### DIFF
--- a/src/lib/VirtualScroller.js
+++ b/src/lib/VirtualScroller.js
@@ -253,7 +253,7 @@ class VirtualScroller {
             newEndOffset = this.totalItems - 1;
         }
 
-        // Create a new list element to be swapped out for the existing one
+        // Creates a document fragment for the new list items to be appended into the list
         const fragment = document.createDocumentFragment();
 
         if (curStartOffset <= offset && offset <= curEndOffset) {
--- a/src/lib/VirtualScroller.js
+++ b/src/lib/VirtualScroller.js
@@ -254,7 +254,7 @@ class VirtualScroller {
         }
 
         // Create a new list element to be swapped out for the existing one
-        const newListEl = document.createDocumentFragment();
+        const fragment = document.createDocumentFragment();
 
         if (curStartOffset <= offset && offset <= curEndOffset) {
             // Scenario #1: New start offset falls within the current range of items rendered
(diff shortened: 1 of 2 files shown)


boxcla:

src/lib/VirtualScroller.js
```diff
@@ -263,77 +263,70 @@ class VirtualScroller {
...
 
     /**
-     * Creates new HTMLElements appended to the newList
+     * Deletes elements of the 'ol'
      *
-     * @param {HTMLElement} newListEl - the new `li` element
+     * @param {HTMLElement} listEl - the `ol` element
      * @param {number} start - start index
-     * @param {number} end  - end index
+     * @param {number} [end] - end index
```
jstoffan: Should `[end]` be in brackets?

src/lib/VirtualScroller.js
```diff
@@ -254,7 +254,7 @@ class VirtualScroller {
         }
 
         // Create a new list element to be swapped out for the existing one
-        const newListEl = this.createListElement();
+        const newListEl = document.createDocumentFragment();
```
jstoffan: The naming here seems a little confusing, since later we append this to `this.listEl`.

src/lib/VirtualScroller.js
```diff
@@ -263,77 +263,70 @@ class VirtualScroller {
...
+        if (!listEl || start < 0 || end < 0) {
             return;
         }
 
-        for (let i = start; i <= end; i++) {
-            const newEl = this.renderItem(i);
-            newListEl.appendChild(newEl);
-        }
+        const listItems = Array.prototype.slice.call(listEl.children, start, end);
+        listItems.forEach((listItem) => listEl.removeChild(listItem));
```
jstoffan: Do we incur a performance penalty here, since `this.listEl` is currently mounted? Each `removeChild` call will force a reflow, right?

src/lib/VirtualScroller.js
```diff
@@ -254,7 +254,7 @@ class VirtualScroller {
         }
 
         // Create a new list element to be swapped out for the existing one
```
jeremypress: Is this comment still relevant?

PR #902, opened 2019-01-24
Chore: Add webpack-dev-server to support local development

jstoffan:
Note: The test file has to live inside `/src` so it can access the `/src/third-party` assets.


build/webpack.config.js
```diff
@@ -67,7 +73,7 @@ function updateConfig(conf, language, index) {
         }
 
         // Add inline source map
```
jeremypress: can remove/update this comment if you'd like

PR #903, opened 2019-01-25
Chore: Improve fullscreen logic to use fscreen for compatibility

jstoffan:


src/lib/Fullscreen.js
```diff
@@ -73,33 +67,21 @@ class Fullscreen extends EventEmitter {
...
         if (this.isSupported()) {
-            fullscreen = !!(
-                document.fullscreenElement ||
-                document.mozFullScreenElement ||
-                document.webkitFullscreenElement ||
-                document.msFullscreenElement
-            );
-        } else {
-            fullscreen = element instanceof HTMLElement && element.classList.contains(CLASS_FULLSCREEN);
+            return !!fscreen.fullscreenElement;
```
jeremypress: Could we simplify this more?
From the fscreen docs
```null if not in fullscreen mode, or the DOM element that's in fullscreen mode
// replacement for: document.fullscreenElement
// mapped to: document.vendorMappedFullsceenElement
// note that fscreen.fullscreenElement uses a getter to retrieve the element
// each time the property is accessed.```

We know that this will either be null (not supported or not in fullscreen), or the element that's in fullscreen.

File: src/lib/viewers/media/MediaControls.js [hunk 2]
```diff
@@ -84,6 +83,8 @@ class MediaControls extends EventEmitter {
         this.handleAutoplay = this.handleAutoplay.bind(this);
         this.handleSubtitle = this.handleSubtitle.bind(this);
         this.handleAudioTrack = this.handleAudioTrack.bind(this);
+        this.handleFullscreenEnter = this.handleFullscreenEnter.bind(this);
```
jeremypress: No changes were needed for the regular controls?

src/lib/Fullscreen.js
DanDeMicco: Yeah, maybe swap it up and check if the fullscreen class is present, else return !!fscreen.fullscreenElement

src/lib/Fullscreen.js
```diff
@@ -73,33 +67,21 @@ class Fullscreen extends EventEmitter {
...
      * @return {void}
      */
-    fullscreenchangeHandler = (el) => {
-        let enter = false;
-
-        enter = (this.isSupported() && this.isFullscreen()) || (!this.isSupported() && !this.isFullscreen(el));
-
-        if (enter) {
+    fullscreenchangeHandler = () => {
+        if (this.isFullscreen()) {
```
DanDeMicco: Before it was checking `isFullscreen(el)`. This for sure isn't needed anymore?

DanDeMicco:
Double check ios

src/lib/Fullscreen.js
jstoffan: Correct. This method is only called as a handler for fscreen's `fullscreenchange` event, which gets its element from `fscreen.fullscreenElement`.

src/lib/Fullscreen.js
jstoffan: I'm not sure I understand these suggestions. This method is intended to check if the page is currently in fullscreen mode. If native fullscreen is supported, it coerces the `fscreen.fullscreenElement` property to return a boolean. If not, it does the same using the passed-in element's classlist property.

src/lib/viewers/media/MediaControls.js
jstoffan: I updated all locations where `fullscreen` was being used. It looks like the fullscreen button for all other viewers is added via `BaseViewer.js`.

jstoffan:
@DanDeMicco, thanks for calling it out. I tested iOS yesterday and it seemed to be working correctly.

PR #904, opened 2019-01-25
Fix: Prevent Annotations errors from triggering a Preview failure

pramodsum:
- Adds ability to emit silent preview errors without triggering the Preview error viewer. This is useful for cases such as annotations, where failures don't necessarily mean that the entire preview experience needs to fail

- [x] Update tests


boxcla:

jeremypress:
Why not just directly call `this.emit('error', error)` in the annotations error listener and not call triggerError?

pramodsum:
preview.triggerError() generates a more thorough log event which i figured might be useful to also have. 

```
    /**
     * Create a generic log Object.
     *
     * @private
     * @return {Object} Log details for viewer session and current file.
     */
    createLogEvent() {
        const file = this.file || {};
        const log = {
            timestamp: getISOTime(),
            file_id: getProp(file, 'id', ''),
            file_version_id: getProp(file, 'file_version.id', ''),
            content_type: getProp(this.viewer, 'options.viewer.NAME', ''),
            extension: file.extension || '',
            locale: getProp(this.location, 'locale', ''),
            rep_type: getProp(this.viewer, 'options.representation.representation', '').toLowerCase(),
            ...getClientLogDetails()
        };

        return log;
    }
```

PR #906, opened 2019-01-28
Update: Scroll into view

ConradJChan:
TODO:
- [x] Test in IE & Edge
- [x] Add Unit tests

scroll into view

### DIFF
--- a/src/lib/ThumbnailsSidebar.js
+++ b/src/lib/ThumbnailsSidebar.js
@@ -50,9 +50,13 @@ class ThumbnailsSidebar {
         this.createThumbnailImage = this.createThumbnailImage.bind(this);
         this.generateThumbnailImages = this.generateThumbnailImages.bind(this);
         this.getThumbnailDataURL = this.getThumbnailDataURL.bind(this);
+        this.isOpen = this.isOpen.bind(this);
         this.renderNextThumbnailImage = this.renderNextThumbnailImage.bind(this);
         this.requestThumbnailImage = this.requestThumbnailImage.bind(this);
         this.thumbnailClickHandler = this.thumbnailClickHandler.bind(this);
+        this.toggle = this.toggle.bind(this);
+        this.toggleClose = this.toggleClose.bind(this);
+        this.toggleOpen = this.toggleOpen.bind(this);
 
         this.anchorEl.addEventListener('click', this.thumbnailClickHandler);
     }
--- a/src/lib/ThumbnailsSidebar.js
+++ b/src/lib/ThumbnailsSidebar.js
@@ -1,5 +1,6 @@
 import isFinite from 'lodash/isFinite';
 import VirtualScroller from './VirtualScroller';
+import { CLASS_HIDDEN } from './constants';
 
 const CLASS_BOX_PREVIEW_THUMBNAIL = 'bp-thumbnail';
 const CLASS_BOX_PREVIEW_THUMBNAIL_IMAGE = 'bp-thumbnail-image';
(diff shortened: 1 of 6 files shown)


boxcla:

src/lib/VirtualScroller.js
```diff
@@ -372,6 +374,53 @@ class VirtualScroller {
...
+        const foundItem = Array.prototype.slice.call(this.listEl.children).find((listItem) => {
+            const { bpVsRowIndex } = listItem.dataset;
+            const parsedRowIndex = parseInt(bpVsRowIndex, 10);
+            return parsedRowIndex === rowIndex;
+        });
+
+        if (foundItem) {
+            // If it is already present and visible, do nothing, but if not visible
+            // then scroll it into view
+            if (!this.isVisible(foundItem)) {
```
jeremypress: Is it that bad to call scrollIntoView even the thumbnail is already visible? that would allow you to get rid of isVisible and it could be kind of nice to top align the thumbnail even if it's already visible.

ConradJChan: it's not necessarily bad, but since cross browser support for the "smooth" `scrollIntoView` is not there, I was trying to avoid the slight jumpiness it causes IMO. I can show you tomorrow to see what you think

src/lib/ThumbnailsSidebar.js
```diff
@@ -49,9 +50,13 @@ class ThumbnailsSidebar {
         this.createThumbnailImage = this.createThumbnailImage.bind(this);
         this.generateThumbnailImages = this.generateThumbnailImages.bind(this);
         this.getThumbnailDataURL = this.getThumbnailDataURL.bind(this);
+        this.isOpen = this.isOpen.bind(this);
         this.renderNextThumbnailImage = this.renderNextThumbnailImage.bind(this);
         this.requestThumbnailImage = this.requestThumbnailImage.bind(this);
         this.thumbnailClickHandler = this.thumbnailClickHandler.bind(this);
+        this.toggle = this.toggle.bind(this);
```
jeremypress: I don't think you need to bind these

src/lib/VirtualScroller.js
```diff
@@ -372,6 +374,53 @@ class VirtualScroller {
...
+     */
+    isVisible(listItemEl) {
+        if (!this.scrollingEl || !listItemEl) {
+            return false;
+        }
+
+        const { scrollTop } = this.scrollingEl;
+        const { offsetTop } = listItemEl;
+
+        return scrollTop <= offsetTop && offsetTop <= scrollTop + this.containerHeight;
```
jeremypress: Can you put a () around `scrollTop + this.containerHeight`? Reading this I saw
`scrollTop <= offsetTop && offsetTop <= scrollTop` first 😮

src/lib/ThumbnailsSidebar.js
```diff
@@ -361,6 +368,56 @@ class ThumbnailsSidebar {
             }
         });
     }
+
+    /**
+     * Toggles the thumbnails sidebar
+     * @return {void}
+     */
+    toggle() {
+        if (!this.anchorEl) {
```
JustinHoldstock: Since you're already checking for `anchorEl` in each function below you could remove this , if you want

src/lib/ThumbnailsSidebar.js
```diff
@@ -361,6 +364,56 @@ class ThumbnailsSidebar {
...
+            return;
+        }
+
+        if (!this.isOpen()) {
```
DanDeMicco: You could probably just use `classList.toggle`

PR #810, opened 2018-06-20
Upgrade: pdf.js v2.0.550

tonyjin:
- Upgrade to the official pdf.js 2.0 pre-release (see mozilla/pdf.js#9521)
- Upgrade minifier to Google Closure Compiler release from 2018-06-10
- Minify pdf.js assets
- Update upgrade pdf.js script to support passing in a version tag, e.g. `yarn upgrade-pdfjs v2.0.550`
- Update document viewer code to reference pdf.js the v2.0 way
- Enable rendering of e-signatures without validation

TODO:
- Fix tests

PR #908, opened 2019-01-29
Upgrade: Update pdf.js to v2.0.943 and closure compiler to v20190121

jstoffan:
**Summary**
This changeset is partially based on and supersedes #810. I've migrated our calls to pdf.js to match the associated API changes from 1.x -> 2.x. However, there is no official migration guide, so further manual testing on document variations will be required.

**Known Issues**
- [ ] **Blocker**: Find controls do not work properly and do not highlight search results
- [x] **Blocker**: Missing glyphs and console warning for documents with invalid fonts: `Warning: Failed to load font "g_d0_f2": SyntaxError: Invalid font data in ArrayBuffer`
- [x] **Minor**: Console error on preview load for documents only in DocumentViewer.pagesinitHandler: `Uncaught TypeError: Cannot read property 'div' of undefined`

**Features Tested**
- Documents with digital signatures
- Documents with single and multiple pages
- Presentations with single and multiple pages
- Fullscreen mode via on-screen controls
- Page-up/down via on-screen controls and scroll wheel
- Text layer highlighting via click-drag and double-click
- Zoom-in/out via on-screen controls


boxcla:

src/lib/util.js
```diff
@@ -508,28 +521,33 @@ export function loadScripts(urls, disableAMD = false) {
     }
 
     urls.forEach((url) => {
-        if (!head.querySelector(`script[src="${url}"]`)) {
+        if (!promiseMap[url]) {
```
DanDeMicco: Would this function be cleaner as a map?

src/lib/util.js
```diff
@@ -13,6 +13,8 @@ const CLIENT_NAME = __NAME__;
 export const CLIENT_VERSION = __VERSION__;
 /* eslint-enable no-undef */
 
+const promiseMap = {};
```
DanDeMicco: Can you use an actual `Map`? This way you get built in methods such as clear for free (which you below)

src/lib/util.js
```diff
@@ -508,28 +521,33 @@ export function loadScripts(urls, disableAMD = false) {
...
+        promises.push(promiseMap[url]);
     });
 
     return Promise.all(promises)
         .then(() => {
             if (disableAMD && amdPresent) {
                 define = defineRef;
             }
+
+            clearPromises();
```
DanDeMicco: can this go into a `.finally`?

PR #909, opened 2019-01-30
Chore: Add support for integration testing with Cypress

ConradJChan:


boxcla:

.travis.yml
```diff
@@ -18,4 +18,4 @@ jobs:
           sudo: required
           addons:
               chrome: beta
-          script: travis_retry yarn run functional-tests-ci
+          script: travis_retry yarn run functional-tests-ci; yarn test:e2e
```
jstoffan: I think we'll want to use `&&` rather than `;` here. The latter causes the next command to execute regardless of the exit code from the prior command.

package.json
```diff
@@ -112,6 +116,9 @@
     "ci": "yarn install && yarn run clean && yarn run build-rb && yarn run lint && yarn run test && yarn run build-ci",
     "clean": "rm -rf dist && rm -rf reports/coverage && rm -rf src/i18n/json",
     "commitmsg": "commitlint -e",
+    "cy:open": "yarn cy:wait; yarn cypress open",
+    "cy:run": "yarn cy:wait; yarn cypress run --spec test/integration/**/*.test.js",
+    "cy:wait": "wait-on tcp:8000",
```
jstoffan: The associated box-ui-elements script was recently improved. Probably want to take a look and use the same concept here.

test/support/commands.js
```diff
@@ -0,0 +1,8 @@
+Cypress.Commands.add('getByTestId', (testId) => cy.get(`[data-testid="${testId}"]`));
+Cypress.Commands.add('getByTitle', (title) => cy.get(`[title="${title}"]`));
+Cypress.Commands.add('showPreview', (token, fileId) => {
+    cy.get('#token-set').type(token);
```
jstoffan: I realize these are inputs in the test harness, but they should ideally not use CSS selectors, regardless.

ConradJChan:
@jstoffan Yeah, this was just to get the groundwork in place. Was thinking of creating an on call task to migrate the existing functional tests over to cypress, then we can get rid of the old functional tests.

test/integration/sanity/Sanity.e2e.test.js
```diff
@@ -0,0 +1,15 @@
...
+
+    beforeEach(() => {
+        token = Cypress.env('ACCESS_TOKEN');
+        cy.visit('/');
+    });
+
+    it('Should load a document preview', () => {
+        const fileId = Cypress.env('FILE_ID_DOC');
+        cy.showPreview(token, fileId);
+        cy.getByTestId('preview-container').contains('The Content Platform for Your Apps', { timeout: 10000 });
```
jstoffan: Is the timeout needed to account for fetching the file/rep? You should be able to use a combination of `cy.route` and `cy.wait` to avoid an arbitrary timeout.

ConradJChan: I tried the load event approach, but couldn't find a way to have Cypress wait for additional assertions within the 'load' event handler. Instead I wait for the `.bp` container to have the `.bp-loaded` class before attempting to assert the content

test/integration/sanity/Sanity.e2e.test.js
```diff
@@ -0,0 +1,20 @@
+// <reference types="Cypress" />
+describe('Preview Sanity', () => {
+    let token;
+
+    beforeEach(() => {
+        token = Cypress.env('ACCESS_TOKEN');
```
jstoffan: Seems like this could be made a const and done just inside `describe` above unless token is modified in tests.

test/integration/sanity/Sanity.e2e.test.js
```diff
@@ -0,0 +1,20 @@
...
+        cy.visit('/');
+    });
+
+    it('Should load a document preview', () => {
+        const fileId = Cypress.env('FILE_ID_DOC');
+
+        // Show the preview
+        cy.showPreview(token, fileId);
+        // Wait for .bp to load viewer
+        cy.getByTestId('bp').should('have.class', 'bp-loaded')
```
jstoffan: Should this live in the `showPreview` helper?

test/support/commands.js
```diff
@@ -0,0 +1,8 @@
+Cypress.Commands.add('getByTestId', (testId) => cy.get(`[data-testid="${testId}"]`));
+Cypress.Commands.add('getByTitle', (title) => cy.get(`[title="${title}"]`));
+Cypress.Commands.add('showPreview', (token, fileId) => {
+    cy.get('[data-testid="token"]').type(token);
```
jstoffan: These should be able to use `cy.getByTestId` since it's declared above.

PR #911, opened 2019-02-04
Chore: track encoding type for documents

JustinHoldstock:


boxcla:

src/lib/viewers/doc/DocBaseViewer.js
```diff
@@ -583,9 +589,12 @@ class DocBaseViewer extends BaseViewer {
 
         // If range requests are disabled, request the gzip compressed version of the representation
         if (PDFJS.disableRange) {
+            const encodingType = ENCODING_TYPES.GZIP;
             url = appendQueryParams(url, {
-                [QUERY_PARAM_ENCODING]: ENCODING_TYPES.GZIP
+                [QUERY_PARAM_ENCODING]: encodingType
```
jstoffan: Can `this.encoding` be assigned once above, depending on if `PDFJS.diableRange` is enabled, then used here? That way we only assign it one time and also avoid an explicit `undefined` assignment.

JustinHoldstock: Yes, that sounds reasonable

JustinHoldstock:
> Does this need a test?

Yup, just wanted to make sure this was fine, first

src/lib/Preview.js
```diff
@@ -1524,7 +1524,7 @@ class Preview extends EventEmitter {
      * @private
      * @return {void}
      */
-    emitLoadMetrics() {
+    emitLoadMetrics({ encoding }) {
```
ConradJChan: It feels a little strange IMO to pass in an empty object to `emitLoadMetrics` to avoid an error destructuring `encoding` from the argument. Would it make sense to instead have logic to handle an undefined argument?

oh and jsdoc for the new method argument.

PR #915, opened 2019-02-05
Chore: adding e2e tests for Thumbnails

ConradJChan:
Modified the test fixture `index.html` to have an explicit load preview button instead of loading it when submitting the token or file id. This is to make it easier when using it during e2e Cypress testing to invoke it with preview options.

### DIFF
--- a/test/support/commands.js
+++ b/test/support/commands.js
@@ -4,8 +4,8 @@ Cypress.Commands.add('getPreviewPage', (pageNum) => {
     cy
         .get(`.page[data-page-number=${pageNum}]`)
         .as('previewPage')
-        // Adding 10s timeout here because sometimes it takes more than the Cypress
-        // default 5s to render the preview
+        // Adding timeout here because sometimes it takes more than the Cypress
+        // default timeout to render the preview
         .find('[data-testid="page-loading-indicator"]', { timeout: 15000 })
         .should('not.exist');
 


src/lib/shell.html
```diff
@@ -50,7 +50,7 @@
                 <button class="bp-btn-plain bp-btn-loading-download bp-is-invisible"></button>
             </div>
         </div>
-        <div class="bp-content">
+        <div class="bp-content" data-testid="bp-content">
```
jeremypress: Are we stripping out data-testid like in elements?

ConradJChan: Yeah we are, so this will get stripped out but html elements created via javascript won't. But when talking to @jstoffan this wasn't an issue because stripping these out is mainly to decrease bloat in the bundle.

test/integration/document/Thumbnails.e2e.test.js
```diff
@@ -0,0 +1,99 @@
+// <reference types="Cypress" />
+describe('Preview Document Thumbnails', () => {
+    const token = Cypress.env('ACCESS_TOKEN');
+    const fileId = Cypress.env('FILE_ID_DOC');
+    const THUMBNAIL_SELECTED_CLASS = 'bp-thumbnail-is-selected';
+
+    /* eslint-disable */
+    const getThumbnail = (pageNum) => cy.get(`.bp-thumbnail[data-bp-page-num=${pageNum}]`);
+    const showDocumentPreview = (enableThumbnailsSidebar) => {
```
jstoffan: Passing in an object with a named `enableThumbnailsSidebar` property makes this a bit easier to read.

test/support/commands.js
```diff
@@ -4,11 +4,17 @@ Cypress.Commands.add('getPreviewPage', (pageNum) => {
     cy
         .get(`.page[data-page-number=${pageNum}]`)
         .as('previewPage')
-        .find('[data-testid="page-loading-indicator"]')
+        // Adding 10s timeout here because there's no reliable way to wait for
```
jstoffan: Using the loading indicator seems reliable, but the document just might take longer to load/render than the default timeout allows.

test/support/commands.js
```diff
@@ -19,6 +25,8 @@ Cypress.Commands.add('showPreview', (token, fileId, options) => {
         cy.window().then((win) => {
             win.loadPreview(options);
         });
+    } else {
+        cy.getByTestId('load-preview').click();
```
jstoffan: Is this needed or can we just remove the `if (options)` check above and allow `win.loadPreview(options)` to be used for all `showPreview` invocations. It seems like it would be the same effect.

ConradJChan: Yeah that's cleaner let me change that

test/support/commands.js
```diff
@@ -4,22 +4,26 @@ Cypress.Commands.add('getPreviewPage', (pageNum) => {
     cy
         .get(`.page[data-page-number=${pageNum}]`)
         .as('previewPage')
-        .find('[data-testid="page-loading-indicator"]')
+        // Adding 10s timeout here because sometimes it takes more than the Cypress
```
jstoffan: Looks like the duration in the comment is outdated. Probably easier to just remove it?

PR #919, opened 2019-02-11
Chore: Remove fractional thumbnail heights

ConradJChan:


boxcla:

src/lib/viewers/doc/_docBase.scss
```diff
@@ -18,6 +18,7 @@ $thumbnail-border-radius: 4px;
         flex: 1 0 auto;
         justify-content: center;
         margin: 0 25px;
+        outline: none;
```
jstoffan: We should leave enabled this unless we have an alternative focus affordance.

PR #921, opened 2019-02-12
Chore(metrics): Update load event 'value' calculation

JustinHoldstock:
This PR allows for the load time to be calculated from the point of starting preview, to the point of it presenting to the user, giving us a more accurate load time to base Apdex off of. Previously, value was the sum of all network response times.

TODO:
- [ ] Tests
- [ ] Handling in EUA


boxcla:

PR #922, opened 2019-02-12
Chore(metrics): add new timer for full preview load time tracking

JustinHoldstock:
This puts our apdex and preview load timings at parity with our legacy metrics system.

TODO:
- [x] tests


boxcla:

src/lib/Preview.js
```diff
@@ -1245,8 +1250,10 @@ class Preview extends EventEmitter {
      */
     finishLoading(data = {}) {
         if (this.file && this.file.id) {
-            const tag = Timer.createTag(this.file.id, LOAD_METRIC.fullDocumentLoadTime);
-            Timer.stop(tag);
+            const contentLoadTag = Timer.createTag(this.file.id, LOAD_METRIC.fullDocumentLoadTime);
```
jeremypress: Should this move to finshLoading of docBase?

src/lib/Preview.js
```diff
@@ -1530,34 +1537,34 @@ class Preview extends EventEmitter {
...
+            [LOAD_METRIC.convertTime]: convertTime.elapsed || 0,
+            [LOAD_METRIC.downloadResponseTime]: downloadTime.elapsed || 0,
+            [LOAD_METRIC.fullDocumentLoadTime]: contentLoadTime.elapsed || 0,
             ...this.createLogEvent()
         };
 
         this.emit(PREVIEW_METRIC, event);
 
-        Timer.reset([infoTag, convertTag, downloadTag, fullLoadTag]);
+        Timer.reset([infoTag, convertTag, downloadTag, contentLoadTag, previewLoadTag]);
```
jeremypress: why not just rename the timer fullDocumentLoadTime to contentLoadTime?

src/lib/Preview.js
```diff
@@ -753,6 +754,10 @@ class Preview extends EventEmitter {
             );
         }
 
+        // Start the preview duration timer when the user starts to perceive preview's load
+        const tag = Timer.createTag(this.file.id, PREVIEW_LOAD_TIME_TAG);
```
ConradJChan: Do we care about this metric if `reload` is called with `skipServerUpdate = true`? That code path seems to bypass `load`

src/lib/Preview.js
JustinHoldstock: We want to track load times across all file types, not just documents

src/lib/Preview.js
JustinHoldstock: That would be a lot more consistent. Can do!

src/lib/Preview.js
JustinHoldstock: For us, our most important metrics are going to be related to our network response times, and served content times. If it's coming from a "reload" most of the content will already be in the browser cache, which we don't care so much about.

src/lib/Preview.js
jeremypress: sounds good after rename

src/lib/Preview.js
```diff
@@ -1530,34 +1537,34 @@ class Preview extends EventEmitter {
...
-        const convertTag = Timer.createTag(this.file.id, LOAD_METRIC.convertTime);
-        const downloadTag = Timer.createTag(this.file.id, LOAD_METRIC.downloadResponseTime);
-        const fullLoadTag = Timer.createTag(this.file.id, LOAD_METRIC.fullDocumentLoadTime);
+        const { id } = this.file;
+
+        const infoTag = Timer.createTag(id, LOAD_METRIC.fileInfoTime);
+        const convertTag = Timer.createTag(id, LOAD_METRIC.convertTime);
+        const downloadTag = Timer.createTag(id, LOAD_METRIC.downloadResponseTime);
+        const contentLoadTag = Timer.createTag(id, LOAD_METRIC.contentLoadTime);
+        const previewLoadTag = Timer.createTag(id, PREVIEW_LOAD_TIME_TAG);
```
jstoffan: Seems like this tag should use the same convention as the other values. Something like `LOAD_METRIC.previewLoadTime`.

PR #923, opened 2019-02-12
Chore: Replace loading doc loading animation with a GIF

jeremypress:
Performance diff on IE11
Before:
capture2

After:
capture1

Todo:
- [x] perf / memory comparison with change
- [X] fix GIF on non-white backgrounds
screen shot 2019-02-20 at 1 59 30 pm

- [x] optimize GIF (currently ~800bytes)
- [x] tests


boxcla:

src/lib/_loading.scss
```diff
@@ -104,45 +106,24 @@
...
 
-@keyframes bp-spinner {
-    0% {
-        transform: translate(-50%, -50%) rotate(0deg);
-    }
-
-    100% {
-        transform: translate(-50%, -50%) rotate(360deg);
+.bp .bp-doc.bp-doc-document {
+    .pdfViewer .page .loadingIcon {
```
jstoffan: Is this level of specificity required? It also seems like the nesting is kind of arbitrary.

src/lib/_loading.scss
```diff
@@ -104,45 +106,24 @@
...
+        background: url('icons/loading.gif') center no-repeat;
+        background-size: $spinner-size $spinner-size;
```
jstoffan: This can be folded into `background` above.

src/lib/_loading.scss
```diff
@@ -104,45 +106,24 @@
...
-    transform: translate(-15px, -15px) scale(.15) translate(15px, 15px);
-    width: 30px;
+.bp-document-preload-wrapper {
+    .bp-preload-spinner {
+        background: url('icons/loading.gif') center no-repeat;
+        background-size: $spinner-size $spinner-size;
+        bottom: 0;
+        height: $spinner-size;
+        left: 0;
+        margin: auto;
```
jstoffan: Is this needed?

src/lib/_loading.scss
```diff
@@ -104,45 +106,24 @@
...
-    left: 0;
-    margin: auto;
-    position: absolute;
-    right: 0;
-    top: 0;
-    transform: translate(-15px, -15px) scale(.15) translate(15px, 15px);
-    width: 30px;
+.bp-document-preload-wrapper {
+    .bp-preload-spinner {
+        background: url('icons/loading.gif') center no-repeat;
```
jstoffan: You can use SCSS mixins to avoid repeating styles. Something like:

```
@mixin spinner() {
    background: url('icons/loading.gif') center $spinner-size $spinner-size no-repeat;
}

.bp .bp-doc .pdfViewer .page .loadingIcon {
    @include spinner;
}

.bp-preload-spinner {
    @include spinner;
    // other rules
}
```

jeremypress:
Before:
capture2

After:
capture1

src/lib/_loading.scss
jeremypress: I didn't think so, but the div is still appearing in the top left without it. The old styles for the pre load spinner included margin as well.

src/lib/_loading.scss
jeremypress: the first level of nesting is our styles, the second level are the styles within PDFj.s' viewer that we are overriding. We could skip the first part but that if someone is using pdf.js directly elsewhere it would override their styles as well.

ConradJChan:
Can we remove the styles related to the animation of the div?

PR #924, opened 2019-02-12
Chore: Render the visible thumbnails first

ConradJChan:


boxcla:

src/lib/VirtualScroller.js
```diff
@@ -425,6 +425,21 @@ class VirtualScroller {
...
+    /**
+     * Gets the currently visible list items
+     * @return {Array<HTMLElement>} - the list of current visible list items
+     */
+    getVisibleItems() {
+        if (!this.listEl) {
+            return [];
+        }
+
+        return Array.prototype.slice
```
jeremypress: iirc you do this a few types, might be worth abstracting the conversion of the listEl into an array?

src/lib/VirtualScroller.js
```diff
@@ -425,6 +425,21 @@ class VirtualScroller {
...
+            .call(this.listEl.children)
+            .filter((itemEl) => this.isVisible(itemEl))
+            .map((itemEl) => itemEl && itemEl.children && itemEl.children[0]);
```
jeremypress: Why do we need to look at the first child of the item?

ConradJChan: From a `VirtualScroller` perspective, the first child of the `itemEl` will be what the user of the `VirtualScroller` generated, or in this case, the generated thumbnail. Basically, it's stripping away the `VirtualScroller` shell and returning just the items

jeremypress: 👍 , just wanted to make sure it wasn't looking at something thumbnail sidebar specific.

PR #925, opened 2019-02-13
Chore: Handle Thumbnails Sidebar resize event

ConradJChan:


boxcla:

src/lib/VirtualScroller.js
```diff
@@ -123,6 +120,23 @@ class VirtualScroller {
         }
     }
 
+    /**
+     * Given the container height, calculate the virtual window properties
+     * @param {number} containerHeight - the available container height of the virtual scroller
```
jeremypress: due to the fact that getContainerHeight may return null I think we should also expect null for the param here.

ConradJChan: Seeing as null would be an invalid input and the method would fail fast, does it make sense to document null as a valid param?

PR #926, opened 2019-02-13
Chore: Fix fullscreen nav and hide toggle thumbnail button

ConradJChan:
* Re-enables keyboard nav while in fullscreen
* Hides the toggle thumbnail button from the document controls while in fullscreen because thumbnail sidebar is not visible anyway
* applies a `min-width` to the `.bp-content` div so that fullscreen works in Safari

### DIFF
--- a/src/lib/_common.scss
+++ b/src/lib/_common.scss
@@ -149,6 +149,7 @@ $header-height: 48px;
         align-items: center;
         display: flex;
         flex: 1 1 auto;
+        min-width: 100%;
         position: relative;
     }
 }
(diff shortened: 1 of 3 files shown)


boxcla:

src/lib/_common.scss
```diff
@@ -149,6 +149,7 @@ $header-height: 48px;
         align-items: center;
         display: flex;
         flex: 1 1 auto;
```
jstoffan: Does it make sense to set the flex-basis to 100%, instead or as well?

ConradJChan: I tried that but Safari doesn't seem to care about that value either

PR #929, opened 2019-02-14
Breaking(logger): Remove legacy Logger from Preview

JustinHoldstock:
This removes the legacy logging system as part of the move to the "new" event driven metrics and error logging


boxcla:

PR #930, opened 2019-02-14
Chore: Safari fullscreen

ConradJChan:
The previous attempt to fix this broke the adaptability of the `.bp-content` div from shrinking when the thumbnails sidebar was open.

In this fix, only when the `.bp-is-fullscreen` class is applied, do we set the width to 100% as well as set the background-color to inherit because Safari by default does not do this for flex items in a flex container that request fullscreen. In this way, it handles both the default #f5f5f5 as well as #000 when in a media viewer's "lowered lights" mode


boxcla:

src/lib/_common.scss
```diff
@@ -149,9 +149,13 @@ $header-height: 48px;
         align-items: center;
         display: flex;
         flex: 1 1 auto;
-        min-width: 100%;
         position: relative;
     }
+
+    .bp-content.bp-is-fullscreen {
+        background-color: $black;
```
jstoffan: I don't think this exists today. Should it only apply to the media player?

ConradJChan: Oh you're right. Today it just keeps the preview background which is #f5f5f5 I think. But in Safari, the background changed to white. I'll update it with the #f5f5f5 value and make sure it's the same experience across browsers

PR #931, opened 2019-02-15
Chore: Move loading wrapper inside bp-content

ConradJChan:
Fixes the issue where the file navigation buttons were not clickable while the viewer was loading.

The issue is that before the viewer is loaded `pointer-events: none` is applied to `.bp` div and is inherited by all the descendants except for the `.bp-loading-wrapper` child which sets `pointer-events: all` and `z-index: 1` (to appear above the viewer div). 

The solution here is to move the `.bp-loading-wrapper` inside of `.bp-content` instead of being a peer and override the `pointer-events` property on the navigation buttons so that they're always clickable

### DIFF
--- a/src/lib/_loading.scss
+++ b/src/lib/_loading.scss
@@ -57,7 +57,6 @@
     color: $twos;
     position: relative;
     text-align: center;
-    z-index: 1;
 }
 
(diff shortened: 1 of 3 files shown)


boxcla:

PR #932, opened 2019-02-19
New: Thumbnails Sidebar

ConradJChan:
Now supports the thumbnails sidebar for document types.

Click the toggle button to open and close the sidebar
screen shot 2019-02-19 at 10 59 54 am

screen shot 2019-02-19 at 10 56 01 am


boxcla:

src/lib/BoundedCache.js
```diff
@@ -0,0 +1,62 @@
...
+     * [constructor]
+     *
+     * @param {number} [maxEntries] - Override the maximum number of cache entries
+     */
+    constructor(maxEntries) {
+        super();
+
+        this.maxEntries = maxEntries || 500;
+        this.cache = {};
+        this.cacheQueue = [];
```
JustinHoldstock: You could initialize `cacheQueue` as a class field, above the constructor.

No need to set `this.cache` since that will happen in the `super` call

Also, give a default value to `maxEntries` parameter, then you can remove the `||`

PR #937, opened 2019-02-20
Chore(compression): Test for disabling gzip for Japan

JustinHoldstock:
Note that we're allowing for watermarked content to still be served as usual.

### DIFF
--- a/src/lib/viewers/doc/DocBaseViewer.js
+++ b/src/lib/viewers/doc/DocBaseViewer.js
@@ -798,7 +798,9 @@ class DocBaseViewer extends BaseViewer {
         // more than the additional time for a continuous request. This also overrides any range request
         // disabling that may be set by pdf.js's compatibility checking since the browsers we support
         // should all be able to properly handle range requests.
-        PDFJS.disableRange = location.locale !== 'en-US' && size < MINIMUM_RANGE_REQUEST_FILE_SIZE_NON_US;
+        // TEST - Also disabling for Japan to compare GZIP compression results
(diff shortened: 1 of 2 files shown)


boxcla:

src/lib/viewers/doc/DocBaseViewer.js
```diff
@@ -798,7 +798,9 @@ class DocBaseViewer extends BaseViewer {
         // more than the additional time for a continuous request. This also overrides any range request
         // disabling that may be set by pdf.js's compatibility checking since the browsers we support
         // should all be able to properly handle range requests.
-        PDFJS.disableRange = location.locale !== 'en-US' && size < MINIMUM_RANGE_REQUEST_FILE_SIZE_NON_US;
+        // TEST - Also disabling for Japan to compare GZIP compression results
+        PDFJS.disableRange =
```
jstoffan: I would think changing more than one variable at a time (disabling range requests and disabling gzip compression) will make it more difficult to tell what impact (or not) the latter is having.

JustinHoldstock: I was trying to think of a way that we won't affect watermarked files, but I can just modify our `this.encoding = foo` check, instead

JustinHoldstock:
@jstoffan We're going to merge this in AFTER we can track TTFB via your work with using Axios in Preview. @jeremypress and I have discussed that measuring TTFB will be extremely valuable in determining usefulness of GZIP compression

PR #939, opened 2019-02-21
Upgrade: Migrate from whatwg-fetch to axios for all requests

jstoffan:


boxcla:

src/lib/api.js
```diff
@@ -0,0 +1,150 @@
...
+
+    /**
+     * Wrapper function for XHR post put and delete
+     *
+     * @private
+     * @param {string} url - URL for XHR
+     * @param {Object} options - Request options
+     * @return {Promise} XHR promise
+     */
+    xhr(url, { data, headers, method, responseType } = {}) {
```
JustinHoldstock: Can you please also allow for all requests (event just `get` if it's too much of a hassle) to use `onDownloadProgress`?

This'll allow for us to easily track TTFB

src/lib/api.js
```diff
@@ -0,0 +1,114 @@
+import axios from 'axios';
+import pickBy from 'lodash/pickBy';
```
jeremypress: this is 4.5kb minfied and gzipped, https://bundlephobia.com/result?p=lodash.pickby@4.6.0, is it a big deal to avoid it?

src/lib/api.js
```diff
@@ -0,0 +1,114 @@
...
+
+const api = {
```
jeremypress: This looks more like a more efficient way to structure a util file. we should probably do this to `utils` at some point as well.

src/lib/api.js
```diff
@@ -0,0 +1,114 @@
...
+    /**
+     * HTTP PUTs a URL with JSON data
+     *
+     * @public
+     * @param {string} url - The URL to fetch
+     * @param {Object} data - JS Object representation of JSON data to send
+     * @param {Object} options - The request options
+     * @return {Promise} HTTP response
+     */
+    del(url, data, options = {}) {
```
jeremypress: We don't abbreviate this elements, can we keep the mapping 1 - 1 between method and HTTP verb?

src/lib/api.js
jeremypress: this is 4.5k minified and gzipped. Can we avoid it? https://bundlephobia.com/result?p=lodash.pickby@4.6.0

src/lib/api.js
jstoffan: Agreed.

src/lib/Preview.js
```diff
@@ -1349,15 +1349,17 @@ class Preview extends EventEmitter {
...
-        post(`${apiHost}/2.0/events`, headers, {
+        const data = {
             event_type: 'preview',
             source: {
                 type: 'file',
                 id: fileId
             }
-        })
+        };
+        const headers = getHeaders({}, token, sharedLink, sharedLinkPassword);
```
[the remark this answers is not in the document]
jstoffan: It looks to be used only used in two places. I wouldn't want to move the URL generators, either.

src/lib/api.js
jstoffan: Done.

src/lib/api.js
```diff
@@ -0,0 +1,132 @@
...
+     *
+     * @private
+     * @param {Object} options - The request options
+     * @return {Object} The cleaned request options
+     */
+    filterOptions(options = {}) {
+        const result = {};
+
+        Object.keys(options).forEach((key) => {
+            if (options[key] !== undefined && options[key] !== null && options[key] !== '') {
```
jeremypress: Using filter might clean this up a bit

PR #940, opened 2019-02-22
Fix: Manually dispatch scroll event

ConradJChan:
Firefox and IE11 didn't fire the scroll event when setting `scrollTop`, so this fix is to manually dispatch the scroll event in `VirtualScroller` so that the items at that scroll point will get rendered


boxcla:

src/lib/VirtualScroller.js
```diff
@@ -421,6 +421,10 @@ class VirtualScroller {
             // will get rendered.
             const topPosition = (this.itemHeight + this.margin) * rowIndex;
             this.scrollingEl.scrollTop = topPosition;
+            // Some browsers don't fire the scroll event when setting scrollTop
+            // (IE11 & Firefox) so we need to manually dispatch the event
+            // in order to trigger `onScrollHandler` to render the items
+            this.scrollingEl.dispatchEvent(new Event('scroll'));
```
jeremypress: Any issues with the event firing twice on browsers that automatically trigger it?

ConradJChan: I didn't see it firing twice, but I'll double check

ConradJChan: Yeah you're right it fires twice, but the `scrollTop` is the same so the logic to render items won't get triggered the second invocation

PR #943, opened 2019-02-23
Update: Moving thumbnail page numbers

ConradJChan:
* Moved the page numbers from overlaying inside the thumbnail image to outside the thumbnail image. 
* Also converted the thumbnail button to an anchor

Old:
screen shot 2019-02-22 at 5 22 49 pm

New:
screen shot 2019-02-22 at 5 20 31 pm


boxcla:

src/lib/ThumbnailsSidebar.js
```diff
@@ -231,7 +245,8 @@ class ThumbnailsSidebar {
             this.createThumbnailImage(itemIndex).then((imageEl) => {
                 // Promise will resolve with null if create image request was already in progress
                 if (imageEl) {
-                    thumbnailEl.appendChild(imageEl);
+                    // Appends to the lastChild which should be the thumbnail button element
```
jeremypress: looks like we're not using a button anymore.

jeremypress:
nit: I think it might look nice if whichever page is selected has the page number on the left get that same black border that the old design had.

PR #944, opened 2019-02-25
chore(deps): Upgrade to Webpack 4

jstoffan:
Todo
- [x] Manual testing of regression folder


boxcla:

build/webpack.config.js
```diff
@@ -24,90 +23,74 @@ if (fs.existsSync('build/rsync.json')) {
...
         entry: {
             annotations: ['box-annotations'],
             preview: [`${lib}/Preview.js`],
-            csv: [`${lib}/viewers/text/BoxCSV.js`]
+            csv: [`${lib}/viewers/text/BoxCSV.js`],
+        },
+        mode: isRelease ? 'production' : 'development',
+        optimization: {
+            minimizer: [
+                new UglifyJsPlugin({
```
ConradJChan: I had read that TerserPlugin (which I think comes by default) is preferred. Should we switch to that instead?

jstoffan: I've seen that as well, but it seemed beyond the scope of what I was trying to achieve. It seems to work fine and other repos all still use UglifyJsPlugin, so I'm going to leave it as-is for now.

ConradJChan: 👍

PR #945, opened 2019-02-26
Chore: Address Security updates

ConradJChan:
Addressing the security alerts for `cryptiles` and `merge`.

`cryptiles` is a nested dependency of `webdriverio` which was for codeceptjs which was already removed so this can be removed as well.

`merge` is a nested dependency of `mojito-rb-gen`


boxcla:

PR #946, opened 2019-02-28
Fix: Fix internal presentation links

jeremypress:


boxcla:

src/lib/viewers/doc/PresentationViewer.js
```diff
@@ -276,8 +276,18 @@ class PresentationViewer extends DocBaseViewer {
      */
     overwritePdfViewerBehavior() {
         // Overwrite scrollPageIntoView for presentations since we have custom pagination behavior
```
jstoffan: Can we update the comment to advise future editors that this is used for inter-page navigation within presentations?

PR #948, opened 2019-03-04
Update: Animate Thumbnails Sidebar transition

ConradJChan:
animated thumbnail sidebar transition

This PR takes the approach of transitioning the `left` property of the thumbnails sidebar off to the left of the `bp-container` out of sight instead of adjusting the width of the thumbnails sidebar as it is toggled.


boxcla:

src/lib/viewers/doc/_docBase.scss
```diff
@@ -49,6 +59,10 @@ $thumbnail-border-radius: 4px;
     .bp-thumbnail.bp-thumbnail-is-selected .bp-thumbnail-nav {
         border-color: $fours;
     }
+
+    &.bp-thumbnails-open .bp-content {
+        left: 201px;
```
JustinHoldstock: Would it be easier to manage if this was stored in a variable, just in case anything about the sidebar width changes?

src/lib/viewers/doc/_docBase.scss
```diff
@@ -5,8 +5,18 @@ $thumbnail-border-radius: 4px;
...
     .bp-thumbnails-container {
         border-right: solid 1px $seesee;
+        bottom: 0;
         display: flex;
-        flex: 0 0 201px; // Accounts for the 1px border on the right so the remaining width is actually 180
+        flex: 0 0 auto;
+        left: -201px;
+        position: absolute;
+        top: 0;
+        transition: left 300ms cubic-bezier(.4, 0, .2, 1);
```
jstoffan: I believe it's usually more efficient to animate `translate` vs `left/right/top/bottom`. Is that an option here?

src/lib/viewers/doc/_docBase.scss
```diff
@@ -49,6 +59,10 @@ $thumbnail-border-radius: 4px;
     .bp-thumbnail.bp-thumbnail-is-selected .bp-thumbnail-nav {
         border-color: $fours;
     }
+
+    &.bp-thumbnails-open .bp-content {
```
jstoffan: Can this modifier (`&.bp-thumbnails-open`) live up with its twin on line 18?

src/lib/_common.scss
```diff
@@ -159,6 +163,10 @@ $header-height: 48px;
     }
 }
 
+.bp.bp-loaded .bp-content {
+    transition: left 300ms cubic-bezier(.4, 0, .2, 1);
```
jeremypress: Could this affect anything when the thumbnail sidebar is off?

ConradJChan: Yeah, that's why I added the `.bp-loaded` specificity because it affected the loading view, but I'll hunt around to check again

src/lib/viewers/doc/_docBase.scss
```diff
@@ -1,12 +1,30 @@
...
+        bottom: 0;
         display: flex;
-        flex: 0 0 201px; // Accounts for the 1px border on the right so the remaining width is actually 180
+        flex: 0 0 auto;
+        left: 0;
+        position: absolute;
+        top: 0;
+        transform: translateX(-$thumbnail-sidebar-width);
+        transition: transform 300ms cubic-bezier(.4, 0, .2, 1);
+        width: $thumbnail-sidebar-width; // Accounts for the 1px border on the right so the remaining width is actually 180
```
jstoffan: Should the comment live with the variable? Is the `180` part correct or should it be `200`?

src/lib/viewers/doc/_docBase.scss
```diff
@@ -1,12 +1,30 @@
...
+    }
+
+    &.bp-thumbnails-open {
+        .bp-thumbnails-container {
+            transform: translateX(0);
```
JustinHoldstock: Is this necessary? It's essentially a no-op

ConradJChan: Need this for the css transition

PR #950, opened 2019-03-05
Fix: DownloadReachability regex

ConradJChan:


boxcla:

src/lib/DownloadReachability.js
```diff
@@ -6,7 +6,7 @@ const PROD_CUSTOM_HOST_SUFFIX = 'boxcloud.com';
 const DOWNLOAD_NOTIFICATION_SHOWN_KEY = 'download_host_notification_shown';
 const DOWNLOAD_HOST_FALLBACK_KEY = 'download_host_fallback';
 const NUMBERED_HOST_PREFIX_REGEX = /^https:\/\/dl\d+\./;
```
jstoffan: Would we want to update this regex to include `dl-las` as a valid host?

ConradJChan: Yeah that's a good point, let me update that

src/lib/DownloadReachability.js
```diff
@@ -5,8 +5,8 @@ const DEFAULT_DOWNLOAD_HOST_PREFIX = 'https://dl.';
 const PROD_CUSTOM_HOST_SUFFIX = 'boxcloud.com';
 const DOWNLOAD_NOTIFICATION_SHOWN_KEY = 'download_host_notification_shown';
 const DOWNLOAD_HOST_FALLBACK_KEY = 'download_host_fallback';
-const NUMBERED_HOST_PREFIX_REGEX = /^https:\/\/dl\d+\./;
-const CUSTOM_HOST_PREFIX_REGEX = /^https:\/\/[A-Za-z0-9]+./;
+const VALID_CUSTOM_HOST_PREFIX_REGEX = /^https:\/\/dl\d+|^https:\/\/dl-las\./;
+const CUSTOM_HOST_PREFIX_REGEX = /^https:\/\/.+?\./;
```
jeremypress: This is also still valid too right? This handles the public cloud case

src/lib/DownloadReachability.js
```diff
@@ -55,7 +55,7 @@ class DownloadReachability {
         // 2. or starts with a custom prefix and ends with boxcloud.com
         return (
             !downloadUrl.startsWith(DEFAULT_DOWNLOAD_HOST_PREFIX) &&
-            (!!downloadUrl.match(NUMBERED_HOST_PREFIX_REGEX) || downloadUrl.indexOf(PROD_CUSTOM_HOST_SUFFIX) !== -1)
+            (!!downloadUrl.match(VALID_CUSTOM_HOST_PREFIX_REGEX) || downloadUrl.indexOf(PROD_CUSTOM_HOST_SUFFIX) !== -1)
```
jeremypress: Looking at this again, maybe we could simplify to say that it just shouldn't start with the default host prefix and ends with our boxcloud suffix? This will guarantee it comes from box and still validates that it's custom.

ConradJChan: Sounds good let me change that

PR #952, opened 2019-03-06
Update: Thumbnails Sidebar dark mode

ConradJChan:
screen shot 2019-03-05 at 4 12 16 pm


boxcla:

src/lib/viewers/doc/_docBase.scss
```diff
@@ -49,6 +49,23 @@ $thumbnail-border-radius: 4px;
     .bp-thumbnail.bp-thumbnail-is-selected .bp-thumbnail-nav {
         border-color: $fours;
     }
+
+    .bp-thumbnails-dark {
```
jstoffan: May want to consider using BEM-ish syntax here to make it more clear that it's a modifier/override. Something like `.bp-thumbnail-nav--dark`?

PR #955, opened 2019-03-11
Fix: Only execute viewer setup if not already setup

ConradJChan:
Fixes an issue where if any viewer's attempt to download the content first fails and then succeeds the second attempt would run `setup()` twice and duplicate any DOM elements added to the `bp-content` div. 

This fix decouples `setup` from `load` in the viewers and also checks an instance variable `isSetup` to determine whether setup needs to be executed again.

TODO:
- [x] unit tests
- [x] e2e tests

### DIFF
--- a/src/lib/viewers/BaseViewer.js
+++ b/src/lib/viewers/BaseViewer.js
@@ -18,6 +18,7 @@ import {
     replacePlaceholders
 } from '../util';
 import {
+    ANNOTATOR_EVENT,
     CLASS_BOX_PREVIEW_MOBILE,
     CLASS_HIDDEN,
     FILE_OPTION_START,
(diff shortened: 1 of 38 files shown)


boxcla:

src/lib/viewers/BaseViewer.js
```diff
@@ -195,6 +200,14 @@ class BaseViewer extends EventEmitter {
         }
     }
 
+    /**
+     * Checks to see whether the viewer has been setup already
+     * @return {boolean} - Boolean whether the viewer has been set up yet
+     */
+    isSetup() {
+        return this.containerEl && this.containerEl.classList.contains(CLASS_BOX_PREVIEW_HAS_VIEWER);
```
jstoffan: May want to use an instance property, if possible, so we can avoid tracking this state via the DOM.

ConradJChan: Moved to instance property as you suggested. Also decoupled the `setup()` from `load()` and invoke it separately. Couldn't do it in the constructor because it interferes with the `prefetch` logic

PR #957, opened 2019-03-12
Fix: Get response as text from axios

ConradJChan:
For some reason, when specifying `responseType: 'text'` axios still returns the response data as parsed json, so adding a `transformResponse` option to get the text back for PlainTextViewer content

### DIFF
--- a/src/lib/api.js
+++ b/src/lib/api.js
@@ -51,6 +51,8 @@ const api = {
         return response.data;
     },
 
+    transformTextResponse: (data) => data,
+
     /**
      * Wrapper function for XHR post put and delete
      *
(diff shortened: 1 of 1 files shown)


boxcla:

src/lib/api.js
```diff
@@ -60,7 +62,13 @@ const api = {
      * @return {Promise} - XHR promise
      */
     xhr(url, options = {}) {
-        return axios(url, api.filterOptions(options))
+        let transformResponse;
+
+        if (options.responseType === 'text') {
+            transformResponse = api.transformTextResponse;
```
jstoffan: Looks like passing `undefined` also works.

src/lib/api.js
```diff
@@ -60,7 +62,13 @@ const api = {
      * @return {Promise} - XHR promise
      */
     xhr(url, options = {}) {
-        return axios(url, api.filterOptions(options))
+        let transformResponse;
```
jstoffan: Does having this `undefined` affect any other response types?

ConradJChan: should get filtered out on L70

PR #959, opened 2019-03-13
Chore: Add e2e Thumbnail tests for large doc

ConradJChan:


boxcla:

test/integration/document/Thumbnails.e2e.test.js
```diff
@@ -21,7 +33,13 @@ describe('Preview Document Thumbnails', () => {
 
         cy.wait(301); // Wait for toggle animation to complete
 
-        return cy.getByTestId('thumbnails-sidebar');
+        const thumbnailSubject = cy.getByTestId('thumbnails-sidebar');
+
+        return {
```
ConradJChan: I wanted to expose methods to assert the visibility instead of having to copy paste the absurd `have.css...` assertion. Open to better ideas

jstoffan: The current solution looks okay, although spreading the thumbnailsSubject into the return looks a bit odd. If you're looking for an alternative, the `.should` command can accept a function as an argument. You could do something like this:

```
const beVisible = ($thumbnailsSidebar) => {
    expect($thumbnailsSidebar).to.have.css(...etc.);
}

cy.getByTestId('thumbnails-sidebar').should(beVisible);
```

test/integration/document/Thumbnails.e2e.test.js
```diff
@@ -1,16 +1,28 @@
...
+    const getThumbnail = (pageNum) => {
+        return cy
+            .get(`.bp-thumbnail[data-bp-page-num=${pageNum}]`, { timeout: 5000 })
+            .as('foundThumbnail')
+    };
+
+    const getThumbnailWithRenderedImage = (pageNum) => {
+        getThumbnail(pageNum).find('.bp-thumbnail-image').should('exist');
+
+        return cy.get('@foundThumbnail');
```
jstoffan: Sharing aliases across helpers feel a bit awkward. Is the `.should` assertion needed everywhere that we want to get the thumbnail? I believe any call to `.find` or `.get` carries an implicit `.should('exist')` automatically.

test/integration/document/Thumbnails.e2e.test.js
```diff
@@ -1,16 +1,28 @@
...
     const token = Cypress.env('ACCESS_TOKEN');
-    const fileId = Cypress.env('FILE_ID_DOC');
+    const fileId = Cypress.env('FILE_ID_DOC_LARGE');
     const THUMBNAIL_SELECTED_CLASS = 'bp-thumbnail-is-selected';
 
     /* eslint-disable */
-    const getThumbnail = (pageNum) => cy.get(`.bp-thumbnail[data-bp-page-num=${pageNum}]`);
+    const getThumbnail = (pageNum) => {
+        return cy
+            .get(`.bp-thumbnail[data-bp-page-num=${pageNum}]`, { timeout: 5000 })
```
jstoffan: What's the default timeout? May be worth raising it to this value if it's below. We have several timeouts declared already in other tests.

test/integration/document/Thumbnails.e2e.test.js
```diff
@@ -1,16 +1,28 @@
...
+
+    const getThumbnailWithRenderedImage = (pageNum) => {
+        getThumbnail(pageNum).find('.bp-thumbnail-image').should('exist');
+
+        return cy.get('@foundThumbnail');
+    };
+
     const showDocumentPreview = ({ enableThumbnailsSidebar } = {}) => {
         cy.showPreview(token, fileId, { enableThumbnailsSidebar });
         cy.getPreviewPage(1);
```
jstoffan: Is this not already the default behavior of Preview if no page number options are passed in?

test/integration/document/Thumbnails.e2e.test.js
```diff
@@ -1,16 +1,28 @@
...
-        cy.contains('The Content Platform for Your Apps');
+        cy.contains('IN  THE  HOUSE  OF  REPRESENTATIVES');
```
jstoffan: Not sure if this needs to be checked each time we load the document. Maybe just in a single test?

test/integration/document/Thumbnails.e2e.test.js
ConradJChan: I like that, thanks for the suggestion! I'll try it out

test/integration/document/Thumbnails.e2e.test.js
ConradJChan: Default command timeout is 4s, I'll try overriding this

jstoffan: It looks like the `defaultCommandTimeout` in Elements is 8 seconds, so you could definitely bump this one as well.

test/integration/document/Thumbnails.e2e.test.js
```diff
@@ -1,16 +1,27 @@
...
 describe('Preview Document Thumbnails', () => {
     const token = Cypress.env('ACCESS_TOKEN');
-    const fileId = Cypress.env('FILE_ID_DOC');
+    const fileId = Cypress.env('FILE_ID_DOC_LARGE');
     const THUMBNAIL_SELECTED_CLASS = 'bp-thumbnail-is-selected';
 
     /* eslint-disable */
-    const getThumbnail = (pageNum) => cy.get(`.bp-thumbnail[data-bp-page-num=${pageNum}]`);
+    const getThumbnail = (pageNum) => {
+        return cy.get(`.bp-thumbnail[data-bp-page-num=${pageNum}]`);
```
jstoffan: It's not super important, but both this and line 13 can be implicit returns.

test/integration/document/Thumbnails.e2e.test.js
```diff
@@ -1,16 +1,27 @@
 // <reference types="Cypress" />
 describe('Preview Document Thumbnails', () => {
     const token = Cypress.env('ACCESS_TOKEN');
-    const fileId = Cypress.env('FILE_ID_DOC');
+    const fileId = Cypress.env('FILE_ID_DOC_LARGE');
     const THUMBNAIL_SELECTED_CLASS = 'bp-thumbnail-is-selected';
 
     /* eslint-disable */
```
jstoffan: Do you recall why this is necessary? It's wrapping more functions now.

ConradJChan: Mainly to avoid the eslint errors around not having jsdoc for these helper functions. but I'm thinking it would be helpful to treat the test code as important as production code, so I'll remove these and add some jsdoc

PR #963, opened 2019-03-25
Update: Wider thumbnails

ConradJChan:
Increases the thumbnails sidebar width by 25px. Also decreases the selected state border to 3px

Before:
Screen Shot 2019-03-25 at 3 08 32 PM

Now:
Screen Shot 2019-03-25 at 3 07 52 PM

### DIFF
--- a/src/lib/viewers/doc/_docBase.scss
+++ b/src/lib/viewers/doc/_docBase.scss
@@ -1,8 +1,8 @@
 @import '../../boxuiVariables';
 
 $thumbnail-border-radius: 4px;
-// Accounts for the 1px border on the right so the remaining width is actually 200
-$thumbnail-sidebar-width: 201px;
+// Accounts for the 1px border on the right so the remaining width is actually 225
+$thumbnail-sidebar-width: 226px;
 
 .bp {
     .bp-thumbnails-container {
@@ -49,13 +49,12 @@ $thumbnail-sidebar-width: 201px;
 
     .bp-thumbnail-nav {
         background-color: $d-eight;
-        border: $thumbnail-border-radius solid $ffive;
+        border: 3px solid $ffive;
         border-radius: $thumbnail-border-radius;
         cursor: pointer;
         flex: 1 0 auto;
         overflow: hidden;
         padding: 0;
-        width: 134px;
     }
 
     .bp-thumbnail-image {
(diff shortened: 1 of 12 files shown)


boxcla:

test/integration/document/Thumbnails.e2e.test.js
```diff
@@ -64,7 +64,7 @@ describe('Preview Document Thumbnails', () => {
      * @return {Assertion} Chai Assertion
      */
     const notBeVisible = ($thumbnailsSidebar) =>
-        expect($thumbnailsSidebar).to.have.css('transform', 'matrix(1, 0, 0, 1, -201, 0)'); // translateX(-201px)
+        expect($thumbnailsSidebar).to.have.css('transform', 'matrix(1, 0, 0, 1, -226, 0)'); // translateX(-226px)
```
jeremypress: Should we just check for transform? Then we don't need to worry about a change in value.

src/lib/viewers/doc/_docBase.scss
```diff
@@ -49,13 +49,12 @@ $thumbnail-sidebar-width: 201px;
 
     .bp-thumbnail-nav {
         background-color: $d-eight;
-        border: $thumbnail-border-radius solid $ffive;
+        border: 3px solid $ffive;
```
jeremypress: Update the variable and use it?

test/integration/document/Thumbnails.e2e.test.js
ConradJChan: What I want to verify is that the transform value changed from 0 to 226 or vice versa, to reflect the state of the sidebar toggling

src/lib/viewers/doc/_docBase.scss
```diff
@@ -55,7 +55,6 @@ $thumbnail-sidebar-width: 201px;
         flex: 1 0 auto;
         overflow: hidden;
         padding: 0;
-        width: 134px;
```
jstoffan: Was this not needed?

ConradJChan: Yeah I don't think so because the flexbox width seems to take precedence

ConradJChan: I take it back, ie11 needs this, but I'll set it to the flex-basis instead

PR #964, opened 2019-03-26
Fix: Fail intentionally when preview image files with deleted reps

jeremypress:
This download strategy can be applied to other "simple" viewers. We're now able to determine the http status in a 202 case, and caching still works as expected. 

todo:
- [x] tests


boxcla:

src/lib/util.js
```diff
@@ -294,6 +297,24 @@ export function prefetchAssets(urls, preload = false) {
...
 }
 
+/**
+ * Method to fetch a given representation as a blob via an authenticated content URL.
+ * Useful for determining the true state of a "successful" representation
+ *
+ * @param {string} contentUrl - a representation's authenticated content URL
+ * @return {Promise} - Resolves with the representation data, rejects with a deleted rep error
+ */
+export function fetchRepresentationAsBlob(contentUrl) {
```
jeremypress: I added this to util because the DocPreloader will need it when fetching first page images. That class lives outside of the normal viewer inheritance hierarchy.

src/lib/viewers/image/MultiImageViewer.js
```diff
@@ -148,7 +148,14 @@ class MultiImageViewer extends ImageBaseViewer {
         this.singleImageEls[index].setAttribute('data-page-number', index + 1);
         this.singleImageEls[index].classList.add(CLASS_MULTI_IMAGE_PAGE);
 
-        this.singleImageEls[index].src = imageUrl;
+        this.downloadUrl = imageUrl;
+
+        fetchRepresentationAsBlob(imageUrl)
+            .then((blob) => {
+                const srcUrl = URL.createObjectURL(blob);
+                this.singleImageEls[index].src = srcUrl;
```
jeremypress: Now that we're using axios, we can optimize here in the future by cancelling failed requests after we detect an initial failure.

src/lib/viewers/image/ImageViewer.js
```diff
@@ -57,12 +57,12 @@ class ImageViewer extends ImageBaseViewer {
...
         const { representation, viewer } = this.options;
         const template = representation.content.url_template;
-        const downloadUrl = this.createContentUrlWithAuthParams(template, viewer.ASSET);
+        this.downloadUrl = this.createContentUrlWithAuthParams(template, viewer.ASSET);
 
         this.bindDOMListeners();
         return this.getRepStatus()
             .getPromise()
-            .then(() => this.handleAssetAndRepLoad(downloadUrl))
+            .then(() => this.handleAssetAndRepLoad())
```
jstoffan: You should be able to remove the wrapping arrow function.

src/lib/util.js
jstoffan: Is it possible for us to share the response handler function only, rather than the api call itself? That seems to fit more as a util, to me.

jeremypress: It's possible, but I like the idea of keeping the response type in one place so that it's not repeated every time we use it.

src/lib/viewers/image/ImageBaseViewer.js
```diff
@@ -320,9 +323,8 @@ class ImageBaseViewer extends BaseViewer {
         // eslint-disable-next-line
         console.error(err);
 
-        // Display a generic error message but log the real one
-        const error = new PreviewError(ERROR_CODE.CONTENT_DOWNLOAD, __('error_refresh'), {}, err.message);
-        super.handleDownloadError(error, imgUrl);
+        const genericDownloadError = new PreviewError(ERROR_CODE.CONTENT_DOWNLOAD, __('error_refresh'));
```
ConradJChan: do we not need the extra params passed into the `PreviewError` constructor?

src/lib/util.js
```diff
@@ -294,6 +297,24 @@ export function prefetchAssets(urls, preload = false) {
...
+    return api.get(contentUrl, { type: 'blob' }).then((response) => {
```
JustinHoldstock: Can you pull out the callback into a method? It'll allow for easy testing

src/lib/viewers/image/ImageBaseViewer.js
JustinHoldstock: Nope, they're optional past the first param (error code)

jeremypress: true, but this one should stay since it was there previously. I had modified this function before but it's not necessary anymore.

PR #967, opened 2019-03-27
Fix: Hide thumbnails in mobile portrait

ConradJChan:
When in a CSS breakpoint of less than 600px, hide the toggle thumbnails button & the sidebar

Todo:
- [x] e2e tests


boxcla:

src/lib/viewers/doc/_docBase.scss
```diff
@@ -16,6 +20,10 @@ $thumbnail-sidebar-width: 226px;
         transform: translateX(-$thumbnail-sidebar-width);
         transition: transform 300ms cubic-bezier(.4, 0, .2, 1);
         width: $thumbnail-sidebar-width;
+
+        @include for-phone-only {
```
jstoffan: Rather than overriding styles, could we wrap all the affected classes in a min-width media query? That way we don't need to maintain the base/overrides separately. Something like this:

```
@mixin breakpoint-medium {
   @media (min-width: 600px) { @content; }
}

.bp {
    @include breakpoint-medium {
        .bp-thumbnails-container { // etc }
    }
}
```

test/integration/document/Thumbnails.e2e.test.js
```diff
@@ -210,4 +210,31 @@ describe('Preview Document Thumbnails', () => {
 
         getThumbnailWithRenderedImage(1).should('have.class', THUMBNAIL_SELECTED_CLASS);
     });
+
+    it('Should not show the toggle thumbnails button on a small viewport', () => {
+        cy.viewport(375, 667);
```
jstoffan: Can we use a constant or a preset for these?

PR #968, opened 2019-03-27
Fix: Handle failed document preload for deleted representations

jeremypress:


boxcla:

PR #969, opened 2019-03-28
Fix: Allow scrollable viewers to accept keyboard input when focused

jeremypress:


boxcla:

src/lib/viewers/image/MultiImage.scss
```diff
@@ -33,7 +33,8 @@
 .bp-images {
     border: 0 none;
     margin: 0 auto;
-    padding: 0;
+    padding: 7px;
```
jstoffan: Can you add a comment explaining that this is to allow the focus treatment to be visible?

src/lib/_common.scss
```diff
@@ -155,6 +155,11 @@ $header-height: 48px;
         position: absolute;
         right: 0;
         top: 0;
+
+        :first-child:focus {
```
jstoffan: Is this generating `*:first-child:focus`? If so, can we put the asterisk in here explicitly? You also should consider scoping it more tightly:

```
> *[tab-index="0"]:first-child:focus {
    box-shadow: inset 0 0 7px 0 fade-out($twos, .75);
    outline: 0;
}
```

jeremypress: per our discussion offline I'm going to switch to a class to account for CSVs

src/lib/constants.js
```diff
@@ -49,6 +49,7 @@ export const CLASS_IS_SELECTABLE = 'bp-is-selectable';
 export const CLASS_IS_BUFFERING = 'bp-is-buffering';
 export const CLASS_DARK = 'bp-dark';
 export const CLASS_CRAWLER = 'bp-crawler';
+export const CLASS_BOX_PREVIEW_CUSTOM_FOCUS = 'bp-custom-focus';
```
jstoffan: Worth grouping this with the other `CLASS_BOX_PREVIEW` prefixes?

src/lib/viewers/text/PlainTextViewer.js
```diff
@@ -113,8 +113,8 @@ class PlainTextViewer extends TextBaseViewer {
         super.setup();
 
         this.textEl = this.createViewer(document.createElement('pre'));
-        this.textEl.className = 'bp-text bp-text-plain hljs';
-        this.textEl.classList.add(CLASS_HIDDEN);
+        this.textEl.classList.add('bp-text', 'bp-text-plain', 'hjls', CLASS_BOX_PREVIEW_CUSTOM_FOCUS, CLASS_HIDDEN);
```
jstoffan: I don't think multiple arguments to `classList.add` are supported by IE11. You could use the `classnames` helper, instead?

https://developer.mozilla.org/en-US/docs/Web/API/Element/classList#Browser_compatibility

src/lib/viewers/image/MultiImage.scss
```diff
@@ -33,7 +33,10 @@
 .bp-images {
     border: 0 none;
     margin: 0 auto;
-    padding: 0;
+    // Since we use a box-shadow to denote focus, we need to add padding as images appear on top of box-shadow
+    padding: 7px;
+    // Allows focus to apply to the scrollable content on click
+    pointer-events: none;
```
jstoffan: To ask the obvious question: Do we have any cases where images are expecting click events to work? Image maps or something? 😄

src/lib/viewers/text/PlainTextViewer.js
```diff
@@ -113,8 +113,8 @@ class PlainTextViewer extends TextBaseViewer {
         super.setup();
 
         this.textEl = this.createViewer(document.createElement('pre'));
-        this.textEl.className = 'bp-text bp-text-plain hljs';
-        this.textEl.classList.add(CLASS_HIDDEN);
+        this.textEl.className = `bp-text bp-text-plain hljs ${CLASS_BOX_PREVIEW_CONTENT_BODY} ${CLASS_HIDDEN}`;
```
ConradJChan: Could the setting of tabIndex & classes be something we extend the `createViewer` method with?

jeremypress: tabIndex probably not since we don't want certain viewer's content div to be focusable. I'll see how big the change would be for classes

src/lib/viewers/BaseViewer.js
```diff
@@ -1161,6 +1162,7 @@ class BaseViewer extends EventEmitter {
             addedElement = this.containerEl.insertBefore(element, firstChildEl);
         }
 
+        addedElement.className = classes;
```
jstoffan: This feels a little shoehorned. Having each viewer just assign `className` in the same way as they do `tabIndex` seemed fine to me.

src/lib/_common.scss
```diff
@@ -212,6 +212,16 @@ $header-height: 48px;
     }
 }
 
+.bp-content-body {
+    border: 2px solid transparent;
+}
+
+.bp-content-body:focus {
```
jstoffan: This could use `&:focus` above to avoid duplicating the class name.

PR #970, opened 2019-03-30
Fix: Keyboard navigation should only focus meaningful elements

jstoffan:


boxcla:

src/lib/shell.html
```diff
@@ -33,8 +33,8 @@
             </div>
         </div>
     </div>
-    <div class="bp" tabindex="0" data-testid="bp">
-        <div class="bp-content" tabindex="0" data-testid="bp-content">
+    <div class="bp" data-testid="bp" tabindex="-1">
```
ConradJChan: Do both of these need to be programmatically focusable? It looks like `BaseViewer` focuses on `.bp-content` when `autoFocus` is enabled.

src/lib/viewers/doc/DocBaseViewer.js
```diff
@@ -1347,21 +1347,27 @@ class DocBaseViewer extends BaseViewer {
...
         }
 
         this.emitMetric({ name: metricName, data: pagesCount });
         this.emit(eventName);
 
         // Resize after the CSS animation to toggle the sidebar is complete
-        setTimeout(() => this.resize(), THUMBNAILS_SIDEBAR_TRANSITION_TIME);
+        setTimeout(() => {
+            this.resize();
+            this.thumbnailsSidebarEl.style = style;
```
ConradJChan: Seems like setting `style` directly as a string is discouraged (https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/style#Setting_styles).

Also, does this affect the animation transition at all?

src/lib/ThumbnailsSidebar.js
```diff
@@ -232,9 +232,9 @@ class ThumbnailsSidebar {
      * @return {HTMLElement} - thumbnail anchor element
      */
     createThumbnailNav() {
-        const thumbnailNav = document.createElement('a');
+        const thumbnailNav = document.createElement('button');
```
ConradJChan: Does `button` work better than `a` in this case?

ConradJChan:
By adding the `display: none`, we can update the e2e helper functions to to check sidebar visibility normally (`.should('be.visible')` instead of checking via the presence of the classname

src/lib/viewers/doc/DocBaseViewer.js
jstoffan: It does appear to affect the hidden -> visible transition. I'll update it to remove the `display: none` immediately rather than via setTimeout for that case. Good catch!

src/lib/ThumbnailsSidebar.js
jstoffan: Anchors without an `href` attribute are considered placeholders rather than links and aren't included in the tab order ([source](https://www.w3.org/TR/2011/WD-html5-20110525/text-level-semantics.html#the-a-element)). Since we just want to response to a user action rather than navigating, a button seems more semantically correct and also obviates the need for `tabindex="0"`

src/lib/shell.html
jstoffan: The `ContentPreview` Element auto-focuses on `.bp` and the Preview SDK auto-focuses on `.bp-content`.

PR #971, opened 2019-04-01
Update: Open Preview Thumbnails by default

ConradJChan:
Will open the Preview Thumbnails by default for document viewers. Will also remember on a per file basis (via `localStorage`) the toggle state of the thumbnails sidebar.

Todo:
- [x] unit tests
- [x] e2e tests
- [x] verify UX with design

### DIFF
--- a/src/lib/viewers/doc/DocBaseViewer.js
+++ b/src/lib/viewers/doc/DocBaseViewer.js
@@ -141,7 +141,7 @@ class DocBaseViewer extends BaseViewer {
             this.thumbnailsSidebarEl.setAttribute('data-testid', 'thumbnails-sidebar');
             this.rootEl.insertBefore(this.thumbnailsSidebarEl, this.containerEl);
 
-            if (this.getCachedThumbnailsToggledState()) {
+            if (this.shouldThumbnailsBeToggled()) {
                 this.rootEl.classList.add(CLASS_BOX_PREVIEW_THUMBNAILS_OPEN);
             }
         }
@@ -1094,7 +1094,7 @@ class DocBaseViewer extends BaseViewer {
         this.thumbnailsSidebar = new ThumbnailsSidebar(this.thumbnailsSidebarEl, this.pdfViewer);
         this.thumbnailsSidebar.init({
             currentPage: this.pdfViewer.currentPageNumber,
-            isOpen: this.getCachedThumbnailsToggledState(),
+            isOpen: this.shouldThumbnailsBeToggled(),
             onClick: this.onThumbnailClickHandler
         });
     }
@@ -1420,22 +1420,11 @@ class DocBaseViewer extends BaseViewer {
 
     /**
      * Gets the cached thumbnails toggled state based on file id. Will retrieve from
-     * localStorage if not cached. Defaults to true if not found.
-     * @return {boolean} Whether thumbnails is toggled open or not
+     * localStorage if not cached.
+     * @return {boolean} Whether thumbnails is toggled open or not from the cache
      */
     getCachedThumbnailsToggledState() {
-        let toggledOpen = true;
-
-        if (this.cache.has(THUMBNAILS_SIDEBAR_TOGGLED_MAP_KEY)) {
-            const thumbnailsToggledMap = this.cache.get(THUMBNAILS_SIDEBAR_TOGGLED_MAP_KEY);
-            toggledOpen = thumbnailsToggledMap[this.options.file.id];
-
-            // If cached toggled state is anything other than false, set it to true
-            if (toggledOpen !== false) {
-                toggledOpen = true;
-            }
-        }
-
+        const { [this.options.file.id]: toggledOpen } = this.cache.get(THUMBNAILS_SIDEBAR_TOGGLED_MAP_KEY) || {};
         return toggledOpen;
     }
 
@@ -1445,13 +1434,28 @@ class DocBaseViewer extends BaseViewer {
      * @return {void}
      */
     cacheThumbnailsToggledState(isOpen) {
-        let thumbnailsToggledMap = {};
-        if (this.cache.has(THUMBNAILS_SIDEBAR_TOGGLED_MAP_KEY)) {
-            thumbnailsToggledMap = this.cache.get(THUMBNAILS_SIDEBAR_TOGGLED_MAP_KEY);
+        const thumbnailsToggledMap = this.cache.get(THUMBNAILS_SIDEBAR_TOGGLED_MAP_KEY) || {};
+        const newThumbnailsToggledMap = { ...thumbnailsToggledMap, [this.options.file.id]: !!isOpen };
+
+        this.cache.set(THUMBNAILS_SIDEBAR_TOGGLED_MAP_KEY, newThumbnailsToggledMap, true /* useLocalStorage */);
+    }
+
+    /**
+     * Determines if the thumbnails sidebar is toggled based on the value from the cache
+     * as well as defaulting to true if there is no cache entry for this file id.
+     * @return {boolean} Whether thumbnails is toggled open or not
+     */
+    shouldThumbnailsBeToggled() {
+        const cachedToggledState = this.getCachedThumbnailsToggledState();
+        let toggledState = cachedToggledState;
+
+        // If cached toggled state is anything other than false, set it to true
+        // because we want the default state to be true
+        if (toggledState !== false) {
+            toggledState = true;
         }
 
-        thumbnailsToggledMap[this.options.file.id] = !!isOpen;
-        this.cache.set(THUMBNAILS_SIDEBAR_TOGGLED_MAP_KEY, thumbnailsToggledMap, true /* useLocalStorage */);
+        return toggledState;
     }
 }
 
--- a/src/lib/viewers/doc/DocBaseViewer.js
+++ b/src/lib/viewers/doc/DocBaseViewer.js
@@ -62,6 +62,7 @@ const SAFARI_PRINT_TIMEOUT_MS = 1000; // Wait 1s before trying to print
 const SCROLL_END_TIMEOUT = this.isMobile ? 500 : 250;
 const SCROLL_EVENT_THROTTLE_INTERVAL = 200;
 const THUMBNAILS_SIDEBAR_TRANSITION_TIME = 301; // 301ms
+const THUMBNAILS_SIDEBAR_TOGGLED_MAP_KEY = 'doc-thumbnails-toggled-map';
 // List of metrics to be emitted only once per session
 const METRICS_WHITELIST = [
     USER_DOCUMENT_THUMBNAIL_EVENTS.CLOSE,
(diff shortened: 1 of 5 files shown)


boxcla:

src/lib/viewers/doc/DocBaseViewer.js
```diff
@@ -1396,6 +1404,42 @@ class DocBaseViewer extends BaseViewer {
...
+    }
+
+    /**
+     * Caches the toggled state of the thumbnails sidebar, also saving to localStorage
+     * @param {boolean} isOpen Toggled state of the sidebar
+     * @return {void}
+     */
+    cacheThumbnailsToggledState(isOpen) {
+        let thumbnailsToggledMap = {};
+        if (this.cache.has(THUMBNAILS_SIDEBAR_TOGGLED_MAP_KEY)) {
```
jeremypress: Do we even need to check this before? We're wiping it whether it exists or not.

src/lib/viewers/doc/DocBaseViewer.js
```diff
@@ -135,7 +136,11 @@ class DocBaseViewer extends BaseViewer {
             this.thumbnailsSidebarEl = document.createElement('div');
             this.thumbnailsSidebarEl.className = `${CLASS_BOX_PREVIEW_THUMBNAILS_CONTAINER}`;
             this.thumbnailsSidebarEl.setAttribute('data-testid', 'thumbnails-sidebar');
-            this.containerEl.parentNode.insertBefore(this.thumbnailsSidebarEl, this.containerEl);
+            this.rootEl.insertBefore(this.thumbnailsSidebarEl, this.containerEl);
```
jeremypress: Does this change fix something?

src/lib/viewers/doc/DocBaseViewer.js
```diff
@@ -59,6 +59,7 @@ const SAFARI_PRINT_TIMEOUT_MS = 1000; // Wait 1s before trying to print
 const SCROLL_END_TIMEOUT = this.isMobile ? 500 : 250;
 const SCROLL_EVENT_THROTTLE_INTERVAL = 200;
 const THUMBNAILS_SIDEBAR_TRANSITION_TIME = 301; // 301ms
+const THUMBNAILS_SIDEBAR_TOGGLED_MAP_KEY = 'doc-thumbnails-toggled-map';
```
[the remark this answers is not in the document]
ConradJChan: I believe bp gets auto prefixed

src/lib/viewers/doc/DocBaseViewer.js
ConradJChan: Just simplifies it, since `rootEl` exists now

src/lib/viewers/doc/DocBaseViewer.js
ConradJChan: The cache entry is an object where each key is a file id, just like the current page map cache entry. So we want to get the existing cache entry and then extend it for this fileId

src/lib/viewers/doc/DocBaseViewer.js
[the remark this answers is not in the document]
jeremypress: you're correct 👍

src/lib/viewers/doc/__tests__/DocBaseViewer-test.js
```diff
@@ -158,1678 +160,1795 @@ describe('src/lib/viewers/doc/DocBaseViewer', () => {
...
+                },
+                enableThumbnailsSidebar: true
+            });
+            Object.defineProperty(BaseViewer.prototype, 'setup', { value: sandbox.stub() });
 
-            docBase.destroy();
-            expect(docBase.findBar.destroy).to.be.called;
-            expect(docBase.findBar.removeListener).to.be.called;
-        });
+            // rootEl = document.querySelector(SELECTOR_BOX_PREVIEW);
```
jstoffan: Stray commented code here.

src/lib/viewers/doc/DocBaseViewer.js
```diff
@@ -135,7 +136,11 @@ class DocBaseViewer extends BaseViewer {
             this.thumbnailsSidebarEl = document.createElement('div');
             this.thumbnailsSidebarEl.className = `${CLASS_BOX_PREVIEW_THUMBNAILS_CONTAINER}`;
             this.thumbnailsSidebarEl.setAttribute('data-testid', 'thumbnails-sidebar');
-            this.containerEl.parentNode.insertBefore(this.thumbnailsSidebarEl, this.containerEl);
+            this.rootEl.insertBefore(this.thumbnailsSidebarEl, this.containerEl);
+
+            if (this.getCachedThumbnailsToggledState()) {
+                this.rootEl.classList.add(CLASS_BOX_PREVIEW_THUMBNAILS_OPEN);
```
jstoffan: Strongly recommend moving this to its own method. The sidebar open class logic is going to be in multiple places and will require a `OPEN_ACTIVE` class now, as well.

src/lib/viewers/doc/DocBaseViewer.js
```diff
@@ -135,7 +136,11 @@ class DocBaseViewer extends BaseViewer {
             this.thumbnailsSidebarEl = document.createElement('div');
             this.thumbnailsSidebarEl.className = `${CLASS_BOX_PREVIEW_THUMBNAILS_CONTAINER}`;
             this.thumbnailsSidebarEl.setAttribute('data-testid', 'thumbnails-sidebar');
-            this.containerEl.parentNode.insertBefore(this.thumbnailsSidebarEl, this.containerEl);
+            this.rootEl.insertBefore(this.thumbnailsSidebarEl, this.containerEl);
+
+            if (this.getCachedThumbnailsToggledState()) {
```
jstoffan: Should this call live in `initThumbnails`, since the same value is used there as well?

src/lib/viewers/doc/DocBaseViewer.js
```diff
@@ -1345,6 +1351,8 @@ class DocBaseViewer extends BaseViewer {
 
         const { pagesCount } = this.pdfViewer;
 
+        this.cacheThumbnailsToggledState(this.thumbnailsSidebar.isOpen);
```
jstoffan: If the user toggles the sidebar open and closed quickly, does it behave as expected?

src/lib/viewers/doc/DocBaseViewer.js
```diff
@@ -1396,6 +1404,42 @@ class DocBaseViewer extends BaseViewer {
             default:
         }
     }
+
+    /**
+     * Gets the cached thumbnails toggled state based on file id. Will retrieve from
+     * localStorage if not cached. Defaults to true if not found.
+     * @return {boolean} Whether thumbnails is toggled open or not
+     */
+    getCachedThumbnailsToggledState() {
```
jstoffan: Should this live in the Thumbnails Sidebar class?

src/lib/viewers/doc/DocBaseViewer.js
```diff
@@ -1396,6 +1404,42 @@ class DocBaseViewer extends BaseViewer {
...
+        let toggledOpen = true;
+
+        if (this.cache.has(THUMBNAILS_SIDEBAR_TOGGLED_MAP_KEY)) {
+            const thumbnailsToggledMap = this.cache.get(THUMBNAILS_SIDEBAR_TOGGLED_MAP_KEY);
+            toggledOpen = thumbnailsToggledMap[this.options.file.id];
+
+            // If cached toggled state is anything other than false, set it to true
+            if (toggledOpen !== false) {
+                toggledOpen = true;
```
jstoffan: It may be worth separating the "get cached value" logic from the default behavior if it's not present.

src/lib/viewers/doc/DocBaseViewer.js
jstoffan: It might read a bit better as below, since it looks like `Cache` checks for key presence already:

```
const thumbnailsCacheMap = this.cache.get(THUMBNAILS_SIDEBAR_TOGGLED_MAP_KEY) || {};
```

src/lib/viewers/doc/DocBaseViewer.js
```diff
@@ -1396,6 +1404,42 @@ class DocBaseViewer extends BaseViewer {
...
+     * @param {boolean} isOpen Toggled state of the sidebar
+     * @return {void}
+     */
+    cacheThumbnailsToggledState(isOpen) {
+        let thumbnailsToggledMap = {};
+        if (this.cache.has(THUMBNAILS_SIDEBAR_TOGGLED_MAP_KEY)) {
+            thumbnailsToggledMap = this.cache.get(THUMBNAILS_SIDEBAR_TOGGLED_MAP_KEY);
+        }
+
+        thumbnailsToggledMap[this.options.file.id] = !!isOpen;
```
jstoffan: Consider creating a new object, as this may be modifying the in-memory cache object.

src/lib/viewers/doc/DocBaseViewer.js
ConradJChan: The thinking behind this was to add the thumbnails open class as soon as possible so that the layout of the viewer is set while the pdf gets rendered. If we wait until `initThumbnails` to add the open class, then the UI will jump a little after the PDF content is fetched which is what design wanted to avoid

src/lib/viewers/doc/DocBaseViewer.js
ConradJChan: Hard to reproduce because the position of the button changes as the sidebar is toggled. But even triggering via keyboard, it behaves as expected

src/lib/viewers/doc/DocBaseViewer.js
ConradJChan: I'll do a follow on PR to see how I can clean this up

src/lib/viewers/doc/DocBaseViewer.js
```diff
@@ -1409,6 +1417,46 @@ class DocBaseViewer extends BaseViewer {
...
+    /**
+     * Caches the toggled state of the thumbnails sidebar, also saving to localStorage
+     * @param {boolean} isOpen Toggled state of the sidebar
+     * @return {void}
+     */
+    cacheThumbnailsToggledState(isOpen) {
+        const thumbnailsToggledMap = this.cache.get(THUMBNAILS_SIDEBAR_TOGGLED_MAP_KEY) || {};
+        const newThumbnailsToggledMap = { ...thumbnailsToggledMap, [this.options.file.id]: !!isOpen };
+
+        this.cache.set(THUMBNAILS_SIDEBAR_TOGGLED_MAP_KEY, newThumbnailsToggledMap, true /* useLocalStorage */);
```
jstoffan: Is the `/* useLocalStorage */` comment needed?

ConradJChan: Nah not needed, was a copy pasta

PR #972, opened 2019-04-01
Fix: Thumbnail navigation now animates correctly in IE11

jstoffan:
TIL that animations defined within a media query don't work in IE11 ([source - known issues tab](https://caniuse.com/#search=keyframe)).

>IE10 and IE11 do not support CSS keyframe blocks inside media queries. Must be defined outside of media query definitions


boxcla:

PR #975, opened 2019-04-03
Fix: On Thumbnails initial render, scroll item into view

ConradJChan:
Also updated `VirtualScroller.isVisible` logic to account for the whole height of the item instead of just the `offsetTop`.


boxcla:

PR #976, opened 2019-04-04
Fix: Cleanup the thumbnail sidebar when preview errors

ConradJChan:
If preview fails for a document, make sure to clean up the thumbnails sidebar element as well as the class that was added to the `.bp` div so that the content is not shifted left as if the sidebar is there


boxcla:

PR #979, opened 2019-04-05
Fix: Position bp-content left in fullscreen

ConradJChan:
This fixes an issue in ie11 where a black section appears on the left in fullscreen while the thumbnails sidebar is toggled open


boxcla:

PR #982, opened 2019-04-09
Update: Higher resolution thumbnails

ConradJChan:
Before:
Screen Shot 2019-04-09 at 3 37 34 PM

After:
Screen Shot 2019-04-09 at 3 37 58 PM


boxcla:

src/lib/ThumbnailsSidebar.js
```diff
@@ -9,6 +9,7 @@ const CLASS_BOX_PREVIEW_THUMBNAIL_IMAGE_LOADED = 'bp-thumbnail-image-loaded';
 const CLASS_BOX_PREVIEW_THUMBNAIL_IS_SELECTED = 'bp-thumbnail-is-selected';
 const CLASS_BOX_PREVIEW_THUMBNAIL_PAGE_NUMBER = 'bp-thumbnail-page-number';
 export const DEFAULT_THUMBNAILS_SIDEBAR_WIDTH = 154; // 225px sidebar width - 25px margin right, - 40px for page number - 6px for border
+const HIGHER_RES_WIDTH = DEFAULT_THUMBNAILS_SIDEBAR_WIDTH * 4;
```
jstoffan: I don't usually ask this, but can we add a comment or move `4` to a constant to indicate that it's an upscale multiplier? The purpose of it may get lost with time, otherwise.

jeremypress: nit: once this change goes in this will be our default resolution, should we call it something like thumbnail_image_resolution?

src/lib/ThumbnailsSidebar.js
```diff
@@ -316,16 +317,16 @@ class ThumbnailsSidebar {
                 // landscape
                 if (curPageRatio < this.pageRatio) {
                     // Set the canvas height to that of the thumbnail max height
-                    canvas.height = Math.ceil(DEFAULT_THUMBNAILS_SIDEBAR_WIDTH / this.pageRatio);
+                    canvas.height = Math.floor(HIGHER_RES_WIDTH / this.pageRatio);
```
jstoffan: What was the reason for moving to `Math.floor` vs `Math.ceil`? Were we seeing overflow previously?

ConradJChan: Was trying to fix an issue where an extra line of whitespace was getting generated as part of the image

src/lib/ThumbnailsSidebar.js
ConradJChan: I'll change it to `THUMBNAIL_IMAGE_WIDTH`

PR #986, opened 2019-04-15
Fix: Temporarily remove focus outline from scrollable content

jeremypress:


boxcla:

src/lib/viewers/text/__tests__/PlainTextViewer-test.js
```diff
@@ -72,7 +72,7 @@ describe('lib/viewers/text/PlainTextViewer', () => {
 
             text.setup();
 
-            expect(text.textEl.className).to.equal('bp-text bp-text-plain hljs bp-is-hidden');
+            expect(text.textEl.className).to.equal('bp-text bp-text-plain hljs bp-is-hidden bp-is-scrollable');
```
jstoffan: Classes probably need to be in the same order as the source?

PR #987, opened 2019-04-16
Fix: Media controls now visible for videos in narrow window

jstoffan:
It looks like vertical videos are affected due to a min-width set on the video element which maintains its aspect ratio on narrow screens. Unfortunately, this causes the media controls to overflow past the edge of the viewport.

### DIFF
--- a/src/lib/viewers/media/Dash.scss
+++ b/src/lib/viewers/media/Dash.scss
@@ -1,12 +1,17 @@
 @import 'mediaBase';
 
+.bp-media-container {
+    display: flex;
+    justify-content: center;
+    min-width: 320px; // Prevent media controls from overflowing on small screens
+}
+
 .bp-media-dash {
     video {
         cursor: none;
         display: block;
         max-height: 100%;
         max-width: 100%;
-        min-width: 460px;
         width: 100%;
     }
 
--- a/src/lib/viewers/media/Dash.scss
+++ b/src/lib/viewers/media/Dash.scss
@@ -1,11 +1,5 @@
 @import 'mediaBase';
 
-.bp-media-container {
-    display: flex;
-    justify-content: center;
-    min-width: 320px; // Prevent media controls from overflowing on small screens
-}
-
 .bp-media-dash {
     video {
         cursor: none;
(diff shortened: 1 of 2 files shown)


boxcla:

src/lib/viewers/media/Dash.scss
```diff
@@ -1,12 +1,17 @@
 @import 'mediaBase';
 
+.bp-media-container {
+    display: flex;
+    justify-content: center;
+    min-width: 320px; // Prevent media controls from overflowing on small screens
```
ConradJChan: Should this be in  `_mediaBase.scss` instead?

src/lib/viewers/media/Dash.scss
```diff
@@ -1,12 +1,17 @@
 @import 'mediaBase';
 
+.bp-media-container {
+    display: flex;
```
ConradJChan: Are the flex container and flex styles needed? It looks like everything is absolutely positioned inside the .bp-media-container besides the actual media element

src/lib/viewers/media/Dash.scss
jstoffan: I'll check to see if the issue affects MP4 files. I didn't actually realize they had a separate viewer until now.

src/lib/viewers/media/Dash.scss
jstoffan: It was necessary to center the media element, if I remember correctly. Will verify.

jstoffan: Yeah, it's used to center the `video` element within the container.

src/lib/viewers/media/_mediaBase.scss
```diff
@@ -24,9 +24,12 @@
 }
 
 .bp-media-container {
+    display: flex;
+    justify-content: center;
     outline: 0 none;
     position: relative;
     user-select: none;  // Prevents copy paste dialog from appearing on mobile
+    width: 100%; // Prevent media controls from overflowing on small screens
```
jstoffan: This actually won't work, since it expands the media controls to the full screen width even on small videos.

PR #988, opened 2019-04-18
Fix: Load images via img elements to avoid CORS issues with XHRs

jstoffan:
This change reverts portions of https://github.com/box/box-content-preview/pull/964 due to an issue where XHRs to cached image files are throwing CORS errors in the Embed Widget.


boxcla:

PR #989, opened 2019-04-18
Fix: Show next/previous buttons for file collections in error viewer

jstoffan:
The next/previous buttons currently don't show at all when previewing a collection with files that produce an error (such as unsupported, password protected, etc.).


boxcla:

PR #990, opened 2019-04-18
Fix: Avoid adding thumbnail sidebar classes if option is disabled

jstoffan:


boxcla:

PR #992, opened 2019-04-22
Update: Support .dwg files in the DocumentViewer

ikerr:


boxcla:
Hi @ikerr, thanks for the pull request. Before we can merge it, we need you to sign our Contributor License Agreement. You can do so electronically here: http://opensource.box.com/cla

src/lib/extensions.js
```diff
@@ -15,6 +15,7 @@ export const CODE_EXTENSIONS = [
     'css',
     'cxx',
     'diff',
+    'dwg',
```
jeremypress: CODE_EXTENSIONS are also used by the TextLoader, I think it's possible that the wrong viewer might try to load a dwg depending on order of an array.

ikerr: Ah, so should this go in DOCUMENT_EXTENSIONS instead?

jeremypress: sounds good. Can DWGs have multiple pages or are they single page file types?

ikerr: They can have multiple pages.

PR #994, opened 2019-04-26
Fix: Hide thumbnails if container is small

ConradJChan:
TODO:
- [x] unit tests
- [x] e2e test

### DIFF
--- a/src/lib/viewers/BaseViewer.js
+++ b/src/lib/viewers/BaseViewer.js
@@ -29,7 +29,8 @@ import {
     SELECTOR_BOX_PREVIEW_ICON,
     SELECTOR_BOX_PREVIEW,
     STATUS_SUCCESS,
-    STATUS_VIEWABLE
+    STATUS_VIEWABLE,
+    CLASS_BOX_PREVIEW_IS_SMALL
 } from '../constants';
 import { getIconFromExtension, getIconFromName } from '../icons/icons';
 import { VIEWER_EVENT, ERROR_CODE, LOAD_METRIC, DOWNLOAD_REACHABILITY_METRICS } from '../events';
@@ -61,6 +62,7 @@ const ANNOTATION_BUTTONS = {
 };
 
 const DEFAULT_FILE_ICON_NAME = 'FILE_DEFAULT';
+const BREAKPOINT_MEDIUM = 600; // pixels
 
 class BaseViewer extends EventEmitter {
     /** @property {Controls} - UI used to interact with the document in the viewer */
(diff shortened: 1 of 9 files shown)


boxcla:

src/lib/viewers/BaseViewer.js
```diff
@@ -560,12 +568,28 @@ class BaseViewer extends EventEmitter {
...
             height: document.documentElement.clientHeight
         });
     }
 
+    /**
+     * Determines if the preview container is a small viewport
+     * @return {boolean} Whether the container is small
+     */
+    isSmallContainer() {
+        return this.options.container.clientWidth < BREAKPOINT_MEDIUM;
```
jstoffan: Is `this.options.container` always defined and does it always include `clientWidth`?

src/lib/viewers/BaseViewer.js
```diff
@@ -560,12 +568,28 @@ class BaseViewer extends EventEmitter {
      * @return {void}
      */
     resize() {
+        this.isSmall = this.isSmallContainer();
```
jstoffan: It would be nice to have this be a bit more extensible, where it could support a map of breakpoints to css class names.

src/lib/viewers/doc/_docBase.scss
```diff
@@ -165,6 +165,19 @@ $thumbnail-sidebar-width: 226px;
             display: block;
         }
     }
+
+    &.bp-is-small {
```
jstoffan: Is this meant to override the styles in the breakpoint? If so, two levels of overrides can start to get confusing; if not, can it live up at the top of the `.bp` declaration so it's more clear what we're modifying?

src/lib/viewers/BaseViewer.js
ConradJChan: Yeah this appears to always be defined, if the user doesn't provide it, it defaults to `document.body`

src/lib/viewers/doc/_docBase.scss
ConradJChan: I removed the breakpoint CSS rules so hopefully this simplifies thing in terms of only having 1 level of override now

jstoffan: It's still three levels of CSS, base and two overrides, since the base state is for the components to be hidden. If we turn the CSS class to a positive condition (maybe `.bp-thumbnails-capable`?), then it'd let you keep it to two levels.

ConradJChan:
Still would be good to fix but no rush since the Feed issue is already resolved

PR #995, opened 2019-05-02
Fix: Center video element and maintain its size and ratio

jstoffan:
The previous change caused videos to render at their natural height, even in fullscreen, rather than allowing them to expand. This change will center the video element without affecting its displayed size or ability to expand while still limiting it and its controls to a minimum width of 300 pixels.


boxcla:

PR #997, opened 2019-05-03
Fix: Keyboard shortcuts for thumbnails sidebar

jeremypress:
- [x] unit tests
- [x] e2e tests

### DIFF
--- a/src/lib/ThumbnailsSidebar.js
+++ b/src/lib/ThumbnailsSidebar.js
@@ -90,6 +90,7 @@ class ThumbnailsSidebar {
         }
 
         this.anchorEl.focus();
+
         event.preventDefault();
         event.stopImmediatePropagation();
     }
@@ -102,17 +103,24 @@ class ThumbnailsSidebar {
      */
     onKeydown(event) {
         const key = decodeKeydown(event);
+        let consumed = false;
         let nextSelectedPage = this.currentPage;
+
         if (key === 'ArrowUp') {
             nextSelectedPage -= 1;
+            consumed = true;
         } else if (key === 'ArrowDown') {
             nextSelectedPage += 1;
+            consumed = true;
+        }
+
+        if (consumed) {
+            event.preventDefault();
+            event.stopImmediatePropagation();
         }
 
         if (this.onThumbnailSelect) {
             this.onThumbnailSelect(nextSelectedPage);
-            event.preventDefault();
-            event.stopImmediatePropagation();
         }
     }
 
--- a/src/lib/ThumbnailsSidebar.js
+++ b/src/lib/ThumbnailsSidebar.js
@@ -69,11 +69,11 @@ class ThumbnailsSidebar {
     /**
      * Method to handle the click events in the Thumbnails Sidebar
      *
-     * @param {Event} evt - Mouse click event
+     * @param {Event} event - Mouse click event
      * @return {void}
      */
-    thumbnailClickHandler(evt) {
-        const { target } = evt;
+    thumbnailClickHandler(event) {
+        const { target } = event;
 
         // Only care about clicks on the thumbnail element itself.
         // The image and page number have pointer-events: none so
@@ -90,18 +90,18 @@ class ThumbnailsSidebar {
         }
 
         this.anchorEl.focus();
-        evt.preventDefault();
-        evt.stopImmediatePropagation();
+        event.preventDefault();
+        event.stopImmediatePropagation();
     }
 
     /**
      * Method to handle keyboard events in the Thumbnails Sidebar
      *
-     * @param {Object} evt - Key event
+     * @param {Object} event - Key event
      * @return {void}
      */
-    onKeydown(evt) {
-        const key = decodeKeydown(evt);
+    onKeydown(event) {
+        const key = decodeKeydown(event);
         let nextSelectedPage = this.currentPage;
         if (key === 'ArrowUp') {
             nextSelectedPage -= 1;
@@ -111,8 +111,8 @@ class ThumbnailsSidebar {
 
         if (this.onThumbnailSelect && nextSelectedPage !== this.currentPage) {
             this.onThumbnailSelect(nextSelectedPage);
-            evt.preventDefault();
-            evt.stopImmediatePropagation();
+            event.preventDefault();
+            event.stopImmediatePropagation();
         }
     }
 
@@ -240,7 +240,6 @@ class ThumbnailsSidebar {
         thumbnailEl.appendChild(this.createPageNumber(pageNum));
 
         const thumbnailNav = this.createThumbnailNav();
-        thumbnailNav.tabIndex = -1;
         thumbnailEl.appendChild(thumbnailNav);
 
         if (pageNum === this.currentPage) {
@@ -266,6 +265,7 @@ class ThumbnailsSidebar {
         const thumbnailNav = document.createElement('button');
         thumbnailNav.className = CLASS_BOX_PREVIEW_THUMBNAIL_NAV;
         thumbnailNav.type = 'button';
+        thumbnailNav.tabIndex = -1;
         return thumbnailNav;
     }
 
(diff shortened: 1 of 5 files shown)


boxcla:

src/lib/ThumbnailsSidebar.js
```diff
@@ -81,15 +84,38 @@ class ThumbnailsSidebar {
...
+    /**
+     * Method to handle keyboard events in the Thumbnails Sidebar
+     *
+     * @param {Object} evt - Key event
+     * @return {void}
+     */
+    onKeydown(evt) {
+        const key = decodeKeydown(evt);
+        let nextSelectedPage = this.currentPage;
+        if (key === 'ArrowUp') {
```
ConradJChan: What happens if we're on the first or last page and hit up arrow or down arrow, respectively?

src/lib/ThumbnailsSidebar.js
```diff
@@ -59,8 +60,10 @@ class ThumbnailsSidebar {
         this.renderNextThumbnailImage = this.renderNextThumbnailImage.bind(this);
         this.requestThumbnailImage = this.requestThumbnailImage.bind(this);
         this.thumbnailClickHandler = this.thumbnailClickHandler.bind(this);
+        this.onKeydown = this.onKeydown.bind(this);
 
         this.anchorEl.addEventListener('click', this.thumbnailClickHandler);
+        this.anchorEl.addEventListener('keydown', this.onKeydown);
```
ConradJChan: What's the behavior if we hold down the button?

src/lib/viewers/doc/DocBaseViewer.js
```diff
@@ -1113,7 +1114,7 @@ class DocBaseViewer extends BaseViewer {
      * @param {number} pageNum - the page number
      * @return {void}
      */
-    onThumbnailClickHandler(pageNum) {
+    onThumbnailSelectHandler(pageNum) {
```
ConradJChan: Update the doc to select too?

ConradJChan:
Could you also add some e2e test coverage?

src/lib/ThumbnailsSidebar.js
```diff
@@ -81,15 +84,38 @@ class ThumbnailsSidebar {
...
             const pageNum = parseInt(pageNumStr, 10);
 
-            if (this.onClickHandler) {
-                this.onClickHandler(pageNum);
+            if (this.onThumbnailSelect) {
+                this.onThumbnailSelect(pageNum);
             }
         }
 
+        this.anchorEl.focus();
```
jstoffan: This seems risky. We've seen complaints from customers about Preview autofocusing when it shouldn't.

src/lib/ThumbnailsSidebar.js
```diff
@@ -81,15 +84,38 @@ class ThumbnailsSidebar {
...
         evt.stopImmediatePropagation();
     }
 
+    /**
+     * Method to handle keyboard events in the Thumbnails Sidebar
+     *
+     * @param {Object} evt - Key event
+     * @return {void}
+     */
+    onKeydown(evt) {
```
jstoffan: Can we use `event` here? We've been trying to avoid abbreviations where possible.

src/lib/ThumbnailsSidebar.js
```diff
@@ -214,6 +240,7 @@ class ThumbnailsSidebar {
         thumbnailEl.appendChild(this.createPageNumber(pageNum));
 
         const thumbnailNav = this.createThumbnailNav();
+        thumbnailNav.tabIndex = -1;
```
jstoffan: Should this live in `this.createThumbnailNav`?

src/lib/ThumbnailsSidebar.js
jeremypress: Nothing will happen because we check `nextSelectedPage !== this.currentPage`

src/lib/ThumbnailsSidebar.js
jeremypress: You will experience some lag due to all of the pdf.js renders. We don't currently throttle page changes when using the doc shortcuts, but it's a good optimization going forward.

src/lib/ThumbnailsSidebar.js
jeremypress: This only fires in the clickhandler of the thumbnail button, so you know that the user has focused here intentionally. We need to move focus to the container so that it's not lost when the button element is deleted by the virtual scroller.

src/lib/ThumbnailsSidebar.js
jeremypress: setPage in docBase Viewer will prevent the keydown from changing anything.

test/integration/document/Thumbnails.e2e.test.js
```diff
@@ -280,4 +280,17 @@ describe('Preview Document Thumbnails', () => {
...
+        showDocumentPreview({ enableThumbnailsSidebar: true });
+
+        getThumbnailWithRenderedImage(2).click();
+        cy.focused().type('{downarrow}');
+        getThumbnailWithRenderedImage(3).should('have.class', THUMBNAIL_SELECTED_CLASS);
+
+        getThumbnailWithRenderedImage(3).click();
+        cy.focused().type('{uparrow}');
+
+        getThumbnailWithRenderedImage(2).should('have.class', THUMBNAIL_SELECTED_CLASS);
```
ConradJChan: should we also verify the behavior that the page being previewed changes with the keyboard button presses?

src/lib/ThumbnailsSidebar.js
```diff
@@ -239,6 +276,7 @@ class ThumbnailsSidebar {
         const thumbnailNav = document.createElement('button');
```
jstoffan: Does a `div` with `role="button"` or `role="option"` serve us better here if we want to avoid allowing focus?

test/integration/document/Thumbnails.e2e.test.js
jeremypress: added

src/lib/ThumbnailsSidebar.js
```diff
@@ -59,18 +60,20 @@ class ThumbnailsSidebar {
         this.renderNextThumbnailImage = this.renderNextThumbnailImage.bind(this);
         this.requestThumbnailImage = this.requestThumbnailImage.bind(this);
         this.thumbnailClickHandler = this.thumbnailClickHandler.bind(this);
+        this.onKeydown = this.onKeydown.bind(this);
```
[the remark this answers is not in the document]
jeremypress: I think L62 needs specificity because there are potentially multiple things you could click in this sidebar. 

`onKeydown` is used as the name of the keydown handler in other places in Preview, from viewers to the find bar.

src/lib/ThumbnailsSidebar.js
```diff
@@ -228,6 +262,9 @@ class ThumbnailsSidebar {
             thumbnailEl.classList.add(CLASS_BOX_PREVIEW_THUMBNAIL_IMAGE_LOADED);
         }
 
+        // Add placeholder items to our list of current thumbnails so they can be selected
+        this.currentThumbnails.push(thumbnailEl);
```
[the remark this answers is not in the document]
jeremypress: `currentThumbnails` is a construct that lives only in the thumbnails sidebar, and the caller of this method is the `VirtualScroller`, with this method passed in as the `renderItemFn`.

I'd prefer to keep this update logic in the thumbnails sidebar, as it's used as temporary state until we update from our source of truth, which is the `VS`

src/lib/ThumbnailsSidebar.js
jeremypress: Using a div with either of those roles still grabbed focus when clicked. The only ways I could prevent focus (pointer-events: none, or the disabled attribute) also prevented click events to the needed target.

PR #999, opened 2019-05-15
Fix: Prevent focus loss in IE when navigating thumbnails

jeremypress:


boxcla:

src/lib/ThumbnailsSidebar.js
```diff
@@ -89,6 +89,10 @@ class ThumbnailsSidebar {
             }
         }
 
+        // IE 11 will focus a div when it's parent has a tabindex, so we focus the anchorEl to avoid
+        // a loss of focus when elements are deleted by the Virtual Scroller.
+        this.anchorEl.focus();
```
ConradJChan: Do we know if adding this adversely affects the screen reader? If the screen reader behavior is different, maybe it would be worth it to check if it is IE before doing this manual focus?

jeremypress: There are already issues between interacting with the screen reader and the thumbnails sidebar. I noted these in an accessibility ticket that we can resolve when we do our general pass.

PR #1000, opened 2019-05-16
Fix: Fix resize logic during preload and opening of thumbnails

jeremypress:
- [x] unit tests

### DIFF
--- a/src/lib/viewers/doc/DocBaseViewer.js
+++ b/src/lib/viewers/doc/DocBaseViewer.js
@@ -1089,10 +1089,6 @@ class DocBaseViewer extends BaseViewer {
             // Add page IDs to each page after page structure is available
             this.setupPageIds();
         }
-
-        if (this.options.enableThumbnailsSidebar) {
-            this.initThumbnails();
-        }
     }
 
     /**
@@ -1145,6 +1141,10 @@ class DocBaseViewer extends BaseViewer {
                 this.hidePreload();
                 this.emit(VIEWER_EVENT.progressEnd);
                 this.somePageRendered = true;
+
+                if (this.options.enableThumbnailsSidebar) {
+                    this.initThumbnails();
+                }
             }
         }
     }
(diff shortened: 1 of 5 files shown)


boxcla:

src/lib/viewers/doc/DocBaseViewer.js
```diff
@@ -667,6 +667,7 @@ class DocBaseViewer extends BaseViewer {
                 if (this.shouldThumbnailsBeToggled()) {
                     this.rootEl.classList.add(CLASS_BOX_PREVIEW_THUMBNAILS_OPEN);
                     this.emit(VIEWER_EVENT.thumbnailsOpen);
+                    this.resize();
```
jeremypress: Now that we aren't transitioning on initial load, a resize without a timeout will sufficiently resize the preload or the doc depending on what's ready.

src/lib/viewers/doc/DocBaseViewer.js
```diff
@@ -720,7 +721,7 @@ class DocBaseViewer extends BaseViewer {
      * @return {void}
      */
     resize() {
-        if (!this.pdfViewer || !this.pdfViewer.pageViewsReady) {
+        if (!this.pdfViewer || !this.somePageRendered) {
```
jeremypress: `this.pdfViewer.pageViewsReady` was updating way after a page was rendered. This new check waits for a rendered event from PDF.js so we know the pdfViewer is inter-actable.

src/lib/viewers/doc/DocPreloader.js
```diff
@@ -268,11 +268,19 @@ class DocPreloader extends EventEmitter {
      * @return {void}
      */
     resize() {
-        if (!this.pdfData || !this.preloadEl) {
+        if (!this.preloadEl || (!this.pdfData && !this.imageEl)) {
```
jeremypress: This was causing the resize issue with the preview sidebar. If an image didn't have exif data burnt into it, we wouldn't be able to resize and would see that overlap.

src/lib/viewers/doc/_docBase.scss
```diff
@@ -113,7 +113,8 @@ $thumbnail-sidebar-width: 226px;
     @include bp-breakpoint-medium {
         $bp-thumbnails-animation: 300ms cubic-bezier(.4, 0, .2, 1);
 
-        &.bp-loaded {
+        &.bp-loaded.bp-thumbnails-open-active,
```
jeremypress: Now we'll only apply the transition to bp content if the active classes are present. Those classes only get added in the user generated toggle case.

PR #1006, opened 2019-06-03
Fix: Resize preview when thumbnails open

jeremypress:
The initial resize may happen before the document is ready, so resizing when we decleare the document 'loaded' will ensure that the PDF is resized, not the preload image. The initial resize is preserved so that preload image will be resized when the thumbnails open.


boxcla:

PR #1010, opened 2019-06-05
New: metadataAPI.js

ConradJChan:
Todo:
- [x] Unit tests

### DIFF
--- a/src/lib/MetadataAPI.js
+++ b/src/lib/MetadataAPI.js
@@ -1,44 +1,34 @@
 import api from './api';
 import { getHeaders } from './util';
 
-// Map of extensions to global metadata template
-const EXTENSIONS_TO_TEMPLATES = {
-    dwg: 'autocad'
-};
-
-const MetadataAPI = {
+const metadataAPI = {
     /**
-     * Gets the global metadata template for the specified file
-     * @param {string} file.id - The file id
-     * @param {string} file.extension - The file extension
+     * Gets the global metadata xrefs template for the specified file
+     * @param {string} id - The file id
+     * @param {string} template - The global metadata template
      * @param {Object} options - options object
      * @return {Promise} Promise is resolved or rejected based on response
      */
-    get({ id, extension }, options = {}) {
-        const fileType = EXTENSIONS_TO_TEMPLATES[extension];
-
-        if (!fileType) {
-            return Promise.reject(new Error(`${extension} is not a supported extension`));
+    getXrefsMetadata(id, template, options = {}) {
+        if (!id || !template) {
+            return Promise.reject(new Error('id and template are required parameters'));
         }
 
         const { apiHost, token } = options;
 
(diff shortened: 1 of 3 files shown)


boxcla:

src/lib/MetadataAPI.js
```diff
@@ -0,0 +1,56 @@
...
+     * @param {string} template - Metadata template
+     * @param {string} apiHost - API host
+     * @return {string} The metadata url
+     */
+    getMetadataURL(fileId, template, apiHost) {
+        return `${apiHost}/2.0/files/${fileId}/metadata/global/${template}`;
+    }
+};
+
+export default MetadataAPI;
```
jstoffan: Should the export be lower-case, since it doesn't require instantiation?

src/lib/MetadataAPI.js
```diff
@@ -0,0 +1,56 @@
+import api from './api';
+import { getHeaders } from './util';
+
+// Map of extensions to global metadata template
+const EXTENSIONS_TO_TEMPLATES = {
```
jstoffan: Extensions could load multiple templates, in theory. Should the template(s) needed just be passed in by the calling code, rather than living in a map within the API layer?

src/lib/MetadataAPI.js
```diff
@@ -0,0 +1,56 @@
...
+    get({ id, extension }, options = {}) {
+        const fileType = EXTENSIONS_TO_TEMPLATES[extension];
+
+        if (!fileType) {
+            return Promise.reject(new Error(`${extension} is not a supported extension`));
+        }
+
+        const { apiHost, token } = options;
+
+        return new Promise((resolve, reject) => {
```
jstoffan: Looks like we can just `return api.get` rather than wrapping in a new promise. You can re-throw the error in `.catch` if you want the parent promise to hear it as well.

ConradJChan: I wanted to abstract away the logic of http status 404 being ok so that calling code doesn't need to worry about it. Is there another way to resolve the promise in the catch logic after detecting the 404 situation?

src/lib/MetadataAPI.js
jeremypress: Should we also pass in the scope in case we want to access templates at other levels in the future?

src/lib/MetadataAPI.js
```diff
@@ -0,0 +1,46 @@
...
+     */
+    getXrefsMetadata(id, template, options = {}) {
+        if (!id || !template) {
+            return Promise.reject(new Error('id and template are required parameters'));
+        }
+
+        const { apiHost, token } = options;
+
+        return api
+            .get(metadataAPI.getMetadataURL(id, template, apiHost), { headers: getHeaders({}, token) })
```
jstoffan: It seems worth it to separate out the `getMetadata` call to its own method.

PR #1015, opened 2019-06-10
New: Emit metric when missing xrefs

ConradJChan:
When dwg file has missing external references, will emit an event `missing_x_refs` with file extension as the data


boxcla:

PR #1019, opened 2019-06-18
Fix: Move notification to be child of preview content

ConradJChan:
In PreviewUI.js, introduced `previewContainer` which points to `.bp` div. Existing `contentContainer` (which used to point to `.bp`) is now `.bp-content` which is only used for Notification.js

Also changed the default "Okay" to use the `close.svg`

Centered banner

### DIFF
--- a/src/lib/Notification.js
+++ b/src/lib/Notification.js
@@ -1,4 +1,5 @@
 import { CLASS_HIDDEN, CLASS_BOX_PREVIEW_NOTIFICATION, CLASS_BOX_PREVIEW_NOTIFICATION_WRAPPER } from './constants';
+import { ICON_CLOSE } from './icons/icons';
 
 const HIDE_TIMEOUT_MS = 5000; // 5s
 
@@ -23,7 +24,7 @@ class Notification {
 
         this.notificationEl.innerHTML = `
             <span id="${uniqueLabel}"></span>
-            <button class="close-btn" type="button">✕</button>
+            <button class="close-btn" type="button">${ICON_CLOSE}</button>
         `.trim();
 
         // Save references to message and button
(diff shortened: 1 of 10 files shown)


boxcla:

src/lib/Notification.js
```diff
@@ -23,7 +23,7 @@ class Notification {
 
         this.notificationEl.innerHTML = `
             <span id="${uniqueLabel}"></span>
-            <button class="close-btn" type="button">${__('notification_button_default_text')}</button>
+            <button class="close-btn" type="button">✕</button>
```
jstoffan: Should we use an icon here rather than a unicode character? Should there be an aria-label or something for accessibility?

src/lib/PreviewUI.js
```diff
@@ -109,7 +113,8 @@ class PreviewUI {
         insertTemplate(this.container, shellTemplate);
 
         this.container = this.container.querySelector(SELECTOR_BOX_PREVIEW_CONTAINER);
-        this.contentContainer = this.container.querySelector(SELECTOR_BOX_PREVIEW);
```
jstoffan: Do any other classes extend PreviewUI and use this property? Do they work properly, if so?

ConradJChan: LOL yes referenced by `DocPreloader.checkDocumentLoaded` good catch!

src/lib/Notification.js
ConradJChan: Will introduce an SVG icon thanks

src/lib/viewers/doc/DocPreloader.js
```diff
@@ -390,7 +390,7 @@ class DocPreloader extends EventEmitter {
      */
     checkDocumentLoaded() {
         // If document is already loaded, hide the preload and short circuit
-        if (this.previewUI.contentContainer.classList.contains(CLASS_PREVIEW_LOADED)) {
+        if (this.previewUI.previewContainer.classList.contains(CLASS_PREVIEW_LOADED)) {
```
jstoffan: We'll probably want to triple-check that preload is dismissing on load. 😄

PR #1025, opened 2019-06-21
Fix: Remove bg color for cc button on hover

manishPh:
1. Adding bg color on hover for CC button doesn't play nicely on mobile devices.
2. Other buttons in media player don't have bg color on hover.
3. When CC is turned off, user doesn't know because the button is still blue because of the hover effect.

Removing `:hover` style solves all the above issues.


boxcla:
Hi @manishPh, thanks for the pull request. Before we can merge it, we need you to sign our Contributor License Agreement. You can do so electronically here: http://opensource.box.com/cla

manishPh:
CLA signed

manishPh:
@ConradJChan yes that experience won't change.

PR #1028, opened 2019-06-24
Fix: OfficeViewer no pdfUrl for .xlsb files

ConradJChan:


boxcla:

src/lib/viewers/office/OfficeViewer.js
```diff
@@ -161,8 +161,12 @@ class OfficeViewer extends BaseViewer {
     setupPDFUrl() {
         const { file } = this.options;
         const pdfRep = getRepresentation(file, 'pdf');
-        const { url_template: template } = pdfRep.content;
-        this.pdfUrl = this.createContentUrlWithAuthParams(template);
+
+        // This occurs in the case of .xlsb files where no pdf rep exists
+        if (pdfRep) {
+            const { url_template: template } = pdfRep.content;
```
manishPh: @ConradJChan, can `pdfRep.content` be `undefined`?

ConradJChan: according to the docs, it looks like it will always be returned if X-Rep-Hints is provided, which we do request. But I can put an extra defensive check here

manishPh: Ohterwise, LGTM!

src/lib/viewers/office/OfficeViewer.js
```diff
@@ -161,8 +161,12 @@ class OfficeViewer extends BaseViewer {
     setupPDFUrl() {
         const { file } = this.options;
         const pdfRep = getRepresentation(file, 'pdf');
-        const { url_template: template } = pdfRep.content;
-        this.pdfUrl = this.createContentUrlWithAuthParams(template);
+
+        // This occurs in the case of .xlsb files where no pdf rep exists
+        if (pdfRep && pdfRep.content) {
```
jeremypress: We have a method that's basically a clone of lodash/get called `getProp` that could help avoid these checks.

ConradJChan: Updated

PR #1030, opened 2019-07-01
Fix: Chrome Print Preview Fix

mickr:
* Add listener that is fired once the iframe has loaded it's content for printing.
  This enables chrome to show an accurate print preview.

### DIFF
--- a/src/lib/viewers/image/ImageViewer.js
+++ b/src/lib/viewers/image/ImageViewer.js
@@ -291,6 +291,22 @@ class ImageViewer extends ImageBaseViewer {
      * @return {void}
      */
     print() {
+        const browserName = Browser.getName();
+        const printEvent = 'printsuccess';
+
+        /**
+         * Called async to ensure resource is loaded for print preview. Then removes listener to prevent
+         * multiple handlers.
+         *
+         * @return {void}
+         */
+        const defaultPrintHandler = () => {
+            this.printframe.contentWindow.print();
+            this.printframe.removeEventListener('load', defaultPrintHandler);
+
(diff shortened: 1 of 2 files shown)


boxcla:

src/lib/viewers/image/ImageViewer.js
```diff
@@ -291,20 +291,36 @@ class ImageViewer extends ImageBaseViewer {
...
         this.printImage.style.margin = '0 auto';
 
-        if (Browser.getName() === 'Explorer' || Browser.getName() === 'Edge') {
+        if (browserName === 'Explorer' || browserName === 'Edge') {
             this.printframe.contentWindow.document.execCommand('print', false, null);
+            this.emit(printEvent);
         } else {
-            this.printframe.contentWindow.print();
+            // Chrome will show preview from iframe before the content is loaded.
+            this.printframe.addEventListener('load', defaultPrintHandler);
```
jstoffan: Can there be a situation where the frame has already fired its `load` event by the time this code executes? Such as if the content is cached? If so, does print still work?

ConradJChan: This code is from a few years ago, but I wonder if we could streamline the code to wait for the 'load' event for all browsers before attempting the print?

mickr: In my testing I haven't created that race condition with caching in place and printing works. I could attach the listener earlier to non-explorer  and non-edge browsers.

mickr: I am updating the code as @ConradJChan suggested for all browsers.

PR #1032, opened 2019-07-04
Fix: Set dimensions for close button svg

ConradJChan:


boxcla:

src/lib/_boxui.scss
```diff
@@ -376,11 +376,15 @@
...
 
-            &.close-btn.default-close-btn {
-                width: 24px;
-
-                path.icon {
-                    fill: $white;
+            &.default-close-btn {
+                svg {
+                    height: 24px;
+                    pointer-events: none;
```
jstoffan: I'm assuming the event is handled by the parent?

ConradJChan: Yeah

PR #1035, opened 2019-07-09
Fix: Add sharedLink header for autocad metadata

ConradJChan:


boxcla:

src/lib/metadataAPI.js
```diff
@@ -35,9 +35,11 @@ const metadataAPI = {
      * @return {Promise} XHR promise
      */
     getMetadata(id, scope, template, options = {}) {
-        const { apiHost, token } = options;
+        const { apiHost, token, sharedLink, sharedLinkPassword } = options;
 
-        return api.get(metadataAPI.getMetadataURL(id, scope, template, apiHost), { headers: getHeaders({}, token) });
+        return api.get(metadataAPI.getMetadataURL(id, scope, template, apiHost), {
+            headers: getHeaders({}, token, sharedLink, sharedLinkPassword)
```
jstoffan: It looks like `BaseViewer` has an `appendAuthHeader` helper that does the same. Could we pass in the headers directly rather than passing in all the Preview options?

PR #1038, opened 2019-07-15
use-http-interceptors-in-preview

mickr:
Add support for using axios http interceptors


boxcla:

PR #1039, opened 2019-07-15
fix(api): Add support for http interceptors

mickr:


boxcla:

src/lib/Preview.js
```diff
@@ -803,6 +806,15 @@ class Preview extends EventEmitter {
             this.setupUI();
         }
 
+        // Check to see if there are http interceptors and load them
+        if (this.options.responseInterceptor) {
```
jstoffan: It doesn't seem like these necessarily belong in `handleTokenResponse`. Can they be added earlier, maybe in the constructor?

src/lib/api.js
```diff
@@ -1,6 +1,47 @@
...
+    /**
+     * Adds a function to Axios that intercepts a response
+
+     * @public
+     * @param {Function} responseInterceptor - Function that gets called on each response
+     * @return {void}
+     */
+    addResponseInterceptor(responseInterceptor) {
+        if (responseInterceptor && typeof responseInterceptor === 'function') {
+            this.responseInterceptorId = axios.interceptors.response.use(responseInterceptor);
```
jstoffan: It seems like this would add the interceptor to all future instances of Preview. Should we create a private instance of axios instead of operating on the global instance?

src/lib/api.js
```diff
@@ -1,6 +1,47 @@
 import axios from 'axios';
 
 const api = {
+    responseInterceptorId: null,
+    requestInterceptorId: null,
+    /**
+     * Adds a function to Axios that intercepts a response
+
```
ConradJChan: extra line?

src/lib/api.js
```diff
@@ -1,6 +1,47 @@
...
+     * @public
+     * @param {Function} responseInterceptor - Function that gets called on each response
+     * @return {void}
+     */
+    addResponseInterceptor(responseInterceptor) {
+        if (responseInterceptor && typeof responseInterceptor === 'function') {
```
jstoffan: Are both of these checks needed?

src/lib/api.js
```diff
@@ -1,6 +1,47 @@
 import axios from 'axios';
 
 const api = {
+    responseInterceptorId: null,
+    requestInterceptorId: null,
+    /**
+     * Adds a function to Axios that intercepts a response
```
jstoffan: We should probably avoid surfacing the underlying XHR library in the docstrings. We could end up swapping it out for something else.

src/lib/__tests__/api-test.js
```diff
@@ -182,4 +183,40 @@ describe('API helper', () => {
...
+            api.addRequestInterceptor(reqInterceptor);
+
+            expect(axios.interceptors.request.handlers[0].fulfilled).to.equal(reqInterceptor);
+            api.ejectInterceptors();
+        });
+    });
+
+    describe('ejectInterceptors', () => {
+        it('should remove all interceptors', () => {
+            const reqInterceptor = sinon.stub();
```
jstoffan: We usually try to avoid abbreviations, if possible.

src/lib/api.js
ConradJChan: should we add logic to eject any preexisting interceptors in case `this.*InterceptorId` is not null?

jstoffan: It looks like `.use` pushes the interceptor onto the array rather than replacing it. It looks like there's a `.forEach` that we may be able to use to avoid tracking these IDs, though.

src/lib/Preview.js
mickr: We can add them early, the interceptors are added to preview during the `show()` call. I can refactor to add it on instantiation if that is something we want to do.

src/lib/api.js
mickr: Yeah, we can just iterate through the `Interceptor manager` and flush all of them. I'll implement that tomorrow.

src/lib/Preview.js
jstoffan: I think having them in `.show` makes sense, too. It looks like they're in the mostly-unrelated `handleTokenResponse` method right now, which was my concern.

mickr: Sounds good, I agree that handleTokenResponse isn't the best place either. I'll update accordingly

src/lib/api.js
mickr: If they are separate instances of `ContentPreview` they can have their own interceptors, if they are operating on the same instance with multiple `.show()` call they will have an issue. We can move to separate instance of Axios for each `show()` invocation, but it will be a larger refactor.

jstoffan: Right. I'm not sure that refactor can be avoided as part of this feature enhancement, though. It sounds like we need to have `lib/api` return a class that creates its own private `axios` instance. Then each instance of Preview can have its own instance of `lib/api` and axios.

mickr:
I have updated the other feedback on this PR and I'm going to start working on the api layer refactor.

src/lib/api.js
mickr: Not sure why this doesn't show as outdated.

src/lib/api.js
```diff
@@ -133,8 +170,28 @@ const api = {
...
+            transformResponse = transformTextResponse;
+        }
+
+        return this.client(url, filterOptions({ transformResponse, ...options }))
+            .then(parseResponse)
+            .catch(handleError);
+    }
+}
+
+export default new Api();
```
jstoffan: I believe this will create a singleton shared by all instances of Preview. We should probably let Preview.js instantiate its api, instead.

src/lib/api.js
```diff
@@ -1,77 +1,114 @@
...
+     * @return {void}
      */
-    xhr(url, options = {}) {
-        let transformResponse;
-
-        if (options.responseType === 'text') {
-            transformResponse = api.transformTextResponse;
-        }
+    ejectInterceptors() {
+        this.client.interceptors.response.eject(this.responseInterceptorId);
```
jstoffan: Axios supports multiple interceptors and this will track only the last one added. We discussed previously that we'd iterate over the interceptors and remove them all.

src/lib/api.js
```diff
@@ -1,77 +1,114 @@
...
-    filterOptions(options = {}) {
-        const result = {};
+/**
+ * Retrieves JSON from response.
+ *
+ * @private
+ * @param {Response} response - Response to parse
+ * @return {Promise|Response} Response if 204 or 202, otherwise promise that resolves with JSON
+ */
+export const parseResponse = (response) => {
```
jstoffan: Is this used outside of the tests? If not, can we avoid exporting and test its consuming methods, instead?

mickr: Yeah, it is only used in the preview tests.

src/lib/api.js
mickr: I think we want all calls to use the interceptors, but not pollute the global Axios instance. There are quite a few places that use the api outside of the main `Preview.js`, if we we stop sharing the instance, all calls will have their own Axios instance and need to have the interceptors passed to them. I would say this is not too far from what Preview previously had, without polluting the Axios instance.

src/lib/api.js
```diff
@@ -82,8 +116,8 @@ const api = {
      * @return {Promise} - HTTP response
      */
     get(url, { type: responseType = 'json', ...options } = {}) {
-        return api.xhr(url, { method: 'get', responseType, ...options });
-    },
+        return this.xhr(url, { method: 'get', responseType, ...options });
```
jstoffan: I don't think we use `call` or `apply` anywhere for API methods, but probably a good idea to confirm the context here is always what we expect.

src/lib/Preview.js
```diff
@@ -265,6 +265,9 @@ class Preview extends EventEmitter {
         // Clean the UI
         this.ui.cleanup();
 
+        // Eject http interceptors
+        api.ejectInterceptors();
```
jstoffan: If all Preview instances are sharing the same interceptors, destroying them when one is hidden seems counter-intuitive. Is that the behavior users would expect?

mickr: Seems like the right move is to move instantiation further up the chain like the constructor and tear down in the de-constructor. That will require a small change to UI elements

mickr:
The tests have not been updated, I want to go over the changes in the PR before updating the test implementation.

src/lib/metadataAPI.js
```diff
@@ -37,7 +36,7 @@ const metadataAPI = {
      * @param {Object} [options.sharedLinkPassword] - shared link password if any
      * @return {Promise} XHR promise
      */
-    getMetadata(id, scope, template, { apiHost, token, sharedLink, sharedLinkPassword }) {
+    getMetadata(id, scope, template, { api, apiHost, token, sharedLink, sharedLinkPassword }) {
```
jstoffan: Should this module extend `API` somehow?

src/lib/viewers/BaseViewer.js
```diff
@@ -169,6 +170,8 @@ class BaseViewer extends EventEmitter {
             this.startAt = getProp(this.options, `fileOptions.${this.options.file.id}.${FILE_OPTION_START}`, {});
         }
 
+        this.api = this.options.api || new Api();
```
jstoffan: It looks like most other files handle this in their constructor. Is the difference here meaningful?

src/lib/viewers/box3d/Box3DRenderer.js
```diff
@@ -69,10 +68,11 @@ class Box3DRenderer extends EventEmitter {
      * @param {HTMLElement} containerEl - the container element
      * @param {BoxSDK} [boxSdk] - Box SDK instance, used for requests to Box
      * @return {Image360Renderer} Image360Renderer instance
+     * @return {Object} options.api - Api layer for http calls
      */
-    constructor(containerEl, boxSdk) {
+    constructor(containerEl, boxSdk, options = {}) {
         super();
-
+        this.api = options.api;
```
jstoffan: Can this be destructured in the function declaration?

src/lib/util.js
```diff
@@ -742,8 +741,9 @@ export function handleRepresentationBlobFetch(response) {
  * Useful for determining the true state of a "successful" representation
  *
  * @param {string} contentUrl - a representation's authenticated content URL
+ * @param {Axios} api - Instance of API layer
  * @return {Promise} - Response from axios
  */
-export function fetchRepresentationAsBlob(contentUrl) {
+export function fetchRepresentationAsBlob(contentUrl, api) {
```
jstoffan: This seems a little janky. I think this helper is only used in one place. Maybe it'd be simpler to inline it where it's used?

src/lib/Preview.js
```diff
@@ -4,7 +4,7 @@ import EventEmitter from 'events';
 import cloneDeep from 'lodash/cloneDeep';
 import throttle from 'lodash/throttle';
 /* eslint-enable import/first */
-import api from './api';
+import Api from './api';
```
ConradJChan: I think the preference is that acronyms be uppercased

src/lib/RepStatus.js
```diff
@@ -46,8 +47,9 @@ class RepStatus extends EventEmitter {
      * @param {Object} [options.logger] - Optional logger instance
      * @return {RepStatus} RepStatus instance
      */
-    constructor({ representation, token, sharedLink, sharedLinkPassword, logger, fileId }) {
+    constructor({ api, representation, token, sharedLink, sharedLinkPassword, logger, fileId }) {
         super();
+        this.api = api || new Api();
```
ConradJChan: Is the `new Api()` necessary here?

src/lib/Preview.js
```diff
@@ -773,6 +778,15 @@ class Preview extends EventEmitter {
             this.retryCount = 0;
         }
 
+        // Check to see if there are http interceptors and load them
+        if (this.options.responseInterceptor) {
+            this.api.addResponseInterceptor(this.options.responseInterceptor);
```
ConradJChan: Should we eject any existing interceptors before adding the new ones?

src/lib/Preview.js
mickr: What if we are trying to denote that it is a constructor?

src/lib/metadataAPI.js
mickr: I don't think extending will help in this instance, as we still wont have access to the Preview API instance so won't have access to the loaded interceptors.

mickr: Thanks for the tip on how it is done in the Node SDK makes sense once I saw the example.

src/lib/viewers/doc/DocPreloader.js
```diff
@@ -60,10 +64,13 @@ class DocPreloader extends EventEmitter {
      * [constructor]
      *
      * @param {PreviewUI} previewUI - UI instance
+     * @param {Object} options - Preloader options
+     * @param {Api} options.api - API Instance
      * @return {DocPreloader} DocPreloader instance
      */
-    constructor(previewUI) {
+    constructor(previewUI, { api } = {}) {
```
jstoffan: It looks like this is being passed in as a top-level param rather than an options object.

src/lib/Preview.js
```diff
@@ -538,14 +543,14 @@ class Preview extends EventEmitter {
...
             );
 
-            DownloadReachability.downloadWithReachabilityCheck(downloadUrl);
+            this.api.reachability.downloadWithReachabilityCheck(downloadUrl);
 
             // Otherwise, get the content download URL of the original file and download
         } else {
             const getDownloadUrl = appendQueryParams(getDownloadURL(this.file.id, apiHost), queryParams);
-            api.get(getDownloadUrl, { headers: this.getRequestHeaders() }).then(data => {
+            this.api.get(getDownloadUrl, { headers: this.getRequestHeaders() }).then(data => {
```
jstoffan: In a follow-up, we could probably generalize this to `this.api.file.getDownloadUrl` and hide the reachability checks from the consumer entirely.

src/lib/__tests__/RepStatus-test.js
```diff
@@ -107,7 +109,7 @@ describe('lib/RepStatus', () => {
 
         it('should fetch latest status', () => {
             sandbox
-                .mock(api)
+                .mock(Api.prototype)
```
jstoffan: Curious if we can mock `repStatus.api`, rather than the prototype?

src/lib/__tests__/api-test.js
```diff
@@ -1,23 +1,32 @@
...
         it('should return the full response when the status is 202 or 204', () => {
             const response = {
                 status: 202,
                 data: 'foo',
             };
 
-            expect(api.parseResponse(response)).to.equal(response);
+            sandbox.stub(api, 'client').resolves(response);
 
-            response.status = 204;
```
jstoffan: I think we wanted to test the 204 case here explicitly, as well.

src/lib/__tests__/api-test.js
```diff
@@ -27,7 +36,11 @@ describe('API helper', () => {
                 data,
             };
 
-            expect(api.parseResponse(response)).to.equal(data);
+            sandbox.stub(api, 'client').resolves(response);
+
+            return api.get(url).then(returnedData => {
```
jstoffan: Minor, but we could standardize these to `.then(result =>` or something similar.

src/lib/__tests__/api-test.js
```diff
@@ -182,4 +195,41 @@ describe('API helper', () => {
...
         });
     });
+
+    describe('addResponseInterceptor', () => {
+        it('should add an http response interceptor', () => {
+            const responseInterceptor = sinon.stub();
+            api.addResponseInterceptor(responseInterceptor);
+
+            expect(api.client.interceptors.response.handlers[0].fulfilled).to.equal(responseInterceptor);
+            api.ejectInterceptors();
```
jstoffan: We should probably be creating a new `api` instance for each test rather than modifying a single shared instance.

PR #1045, opened 2019-07-19
fix: prevent printing via browser

priyajeet:
when there is no download permission. This still does not prevent other ways to print/get data.

### DIFF
--- a/src/lib/PreviewUI.js
+++ b/src/lib/PreviewUI.js
@@ -426,6 +426,30 @@ class PreviewUI {
         const loadingDownloadButtonEl = loadingWrapperEl.querySelector(SELECTOR_BOX_PREVIEW_BTN_LOADING_DOWNLOAD);
         loadingDownloadButtonEl.textContent = __('download_file');
     }
+
+    /**
+     * Attempts to prevent printing using the browser.
+     *
+     * @private
+     * @return {void}
+     */
+    preventBrowserPrinting() {
+        const sheet = document.styleSheets[0];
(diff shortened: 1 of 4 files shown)


boxcla:

src/lib/PreviewUI.js
```diff
@@ -426,6 +426,30 @@ class PreviewUI {
...
+        // Doesn't matter which stylesheet is picked up.
+        // Under most circumstances there should be a stylesheet.
+        // However a user may remove all of them manually.
+        if (sheet) {
+            sheet.insertRule(`
+                @media print {
+                    ${SELECTOR_BOX_PREVIEW},
+                    ${SELECTOR_BOX_PREVIEW_CONTAINER},
+                    ${SELECTOR_BOX_PREVIEW_CONTENT} {
+                        display: none !important;
```
jstoffan: Could we include this rule in a class in the base stylesheet and add the class to the root element as needed?

src/lib/PreviewUI.js
```diff
@@ -426,6 +426,30 @@ class PreviewUI {
...
     }
+
+    /**
+     * Attempts to prevent printing using the browser.
+     *
+     * @private
+     * @return {void}
+     */
+    preventBrowserPrinting() {
+        const sheet = document.styleSheets[0];
```
ConradJChan: Could we create a style element and insert as the first stylesheet?

jstoffan:
Closing for now, as I think this is partially addressed by #1076. Feel free to re-open if needed.

PR #1047, opened 2019-07-24
 chore(deps): Upgrade to Babel 7

jstoffan:
I had to modify the PreviewError class slightly because `this.message` didn't seem to be getting set properly after being transpiled. It looks like modifying built-ins the way we have can be a little tricky.


boxcla:

src/lib/PreviewError.js
```diff
@@ -20,9 +20,9 @@ class PreviewError extends Error {
      * @param {...Object} [params] - Additional params, this may contain the original error message
      * @return {PreviewError} PreviewError instance
      */
-    constructor(code, displayMessage = '', details = {}, ...params) {
+    constructor(code, displayMessage = '', details = {}, message, ...params) {
```
mickr: Does the JSDoc need to be updated? 🍹

jstoffan: Done.

src/lib/PreviewError.js
```diff
@@ -17,12 +17,13 @@ class PreviewError extends Error {
...
-     * @param {...Object} [params] - Additional params, this may contain the original error message
+     * @param {string} [message] - Original error message
+     * @param {...Object} [params] - Additional params
      * @return {PreviewError} PreviewError instance
      */
-    constructor(code, displayMessage = '', details = {}, ...params) {
+    constructor(code, displayMessage = '', details = {}, message, ...params) {
         // Pass remaining arguments to parent constructor
-        super(...params);
+        super(message, ...params);
```
priyajeet: Since the parent constructor wants [message, fileName, lineNumber], you can probably get rid of ...params since latter two are optional and not passed in for PreviewErrors usage.

jstoffan: Yeah, I didn't see them passed in anywhere, but I wanted to stay as close to the original implementation as possible.

priyajeet: Also PreviewError's 2nd arg is the message, aka `displayMessage`, so shouldn't it be super(displayMessage)?

jstoffan: Nope. If both are passed in, `this.message` should still be the original message, not the display message.

PR #1052, opened 2019-08-15
fix(find): Prevent loss of focus on content when find bar is closed

jstoffan:
Currently, users are forced to click the document pane after closing the find bar to re-open it or scroll through the content with the keyboard. We also prevent the browser find bar from opening if the user doesn't have download permissions, which is unnecessary, as the text layer doesn't exist.


boxcla:

PR #1059, opened 2019-08-29
fix(pdf): Re-enable support for unverified signatures in PDF files

jstoffan:
I accidentally tested this feature using a file with an image signature rather than an e-signature, which led to a false positive. I believe these assets will be deployed to the CDN with the next box-content-preview release, which means 2.17.0 should still be viable, as it's using the the 2.16.0 assets.


boxcla:

PR #1060, opened 2019-08-29
fix(images): Show UI earlier for large multi-page images

mickr:


boxcla:

PR #1064, opened 2019-08-30
chore: Re-enable running karma tests for a single test file

ConradJChan:
Re-enable the ability to do `yarn test -- --src=viewers/text/CSVViewer` and have it run just the CSVViewer-test.js instead of all the unit tests


boxcla:

PR #1066, opened 2019-09-04
fix(viewer): Update volume icon after UI is loaded

mxiao6:


boxcla:

src/lib/viewers/media/MediaBaseViewer.js
```diff
@@ -209,6 +209,9 @@ class MediaBaseViewer extends BaseViewer {
         this.emit(VIEWER_EVENT.load);
 
         this.loadUI();
+        // If media is muted because of auto-play,
+        // the volume icon should be updated after the UI is loaded
+        this.updateVolumeIcon();
```
jstoffan: Should this call live within `handleVolume`?

jstoffan:
Does this work for Dash videos? It looks like `loadeddataHandler` is overridden in DashViewer.

src/lib/viewers/media/MediaBaseViewer.js
mxiao6: yeah, but `handleVolume` is called within `autoplay`, which happens before `loadUI`, so we have to call `updateVolumnIcon` explicitly again.

mxiao6: Another way to do this is to move `autoplay` call after `loadUI`. What do you think?

mxiao6:
> Does this work for Dash videos? It looks like `loadeddataHandler` is overridden in DashViewer.

DashViewer has no problem because `handleVolume` in `autoplay` happens after `loadUI` in `loadeddataHandler` in DashViewer.

src/lib/viewers/media/MediaBaseViewer.js
jstoffan: Sure It seems like DashViewer and MediaBaseViewer should be as similar as possible in this regard.

PR #1068, opened 2019-09-10
fix(viewer): Pause audio instead of muting it if autoplay is prevented

mxiao6:


boxcla:

src/lib/viewers/media/MediaBaseViewer.js
```diff
@@ -584,10 +562,13 @@ class MediaBaseViewer extends BaseViewer {
             this.setMediaTime(start);
         }
         if (arguments.length === 0 || hasValidStart) {
-            this.mediaEl.play();
+            const autoPlayPromise = this.mediaEl.play();
```
ConradJChan: does `handleRate` and `handleVolume` need to wait for play to resolve if a promise?

ConradJChan: probably doesn't need to be named `autoPlayPromise`

mxiao6: No, because rate and volume are properties of `mediaEl` which can be set before or after playback.

mxiao6: ok

src/lib/viewers/media/MediaBaseViewer.js
```diff
@@ -303,30 +303,28 @@ class MediaBaseViewer extends BaseViewer {
...
      * @return {Promise}
      */
     autoplay() {
         // Play may return a promise depending on browser support. This promise
-        // will resolve when playback starts. If it fails, we mute the video
-        // and try to play again.
+        // will resolve when playback starts.
         // https://webkit.org/blog/7734/auto-play-policy-changes-for-macos/
-        const autoPlayPromise = this.mediaEl.play();
+        const autoPlayPromise = this.play();
```
ConradJChan: This is legacy, but can we rename this to be `playPromise` or something like that? `autoPlayPromise` seems like a misnomer since the promise is from `play` itself

src/lib/viewers/media/MediaBaseViewer.js
```diff
@@ -570,7 +568,7 @@ class MediaBaseViewer extends BaseViewer {
      * @param {number} start - start time in seconds
      * @param {number} end - end time in seconds
      * @emits play
-     * @return {void}
+     * @return {Promise}
```
ConradJChan: This is misleading because this method can return undefined as a result of `this.mediaEl.play` in older browsers. Could we make it consistently return a promise? Likely that would mean moving the check to see if `this.mediaEl.play` returns a promise in here. @jstoffan had a suggestion to throw an error if it's not a promise that `autoplay` can catch and handle to do the fallback logic of `this.mediaEl.autoplay = true;`

Would probably need to differentiate this situation vs the `NotAllowedError` due to not muting or user interaction in the `catch` callback in `autoplay`

src/lib/viewers/media/MediaBaseViewer.js
```diff
@@ -303,35 +304,31 @@ class MediaBaseViewer extends BaseViewer {
...
-                    // Auto-play was prevented, try muted play
-                    this.setVolume(0);
-                    this.mediaEl.play();
-                });
-        }
-
-        // Fallback to traditional autoplay tag if play does not return a promise
-        this.mediaEl.autoplay = true;
-        return Promise.resolve();
+    async autoplay() {
```
jstoffan: Await/async isn't currently used or supported in Preview, I'm afraid. You should be able to replace this with native promises pretty easily, though.

src/lib/viewers/media/MediaBaseViewer.js
```diff
@@ -19,6 +19,7 @@ const TIMESTAMP_UNIT_NAME = 'timestamp';
 const INITIAL_TIME_IN_SECONDS = 0;
 const ONE_MINUTE_IN_SECONDS = 60;
 const ONE_HOUR_IN_SECONDS = 60 * ONE_MINUTE_IN_SECONDS;
+const ERROR_BROWSER_NOT_SUPPORT = 'browsernotsupport';
```
jstoffan: Should this live alongside our other error constants?

src/lib/viewers/media/MediaBaseViewer.js
```diff
@@ -303,35 +304,29 @@ class MediaBaseViewer extends BaseViewer {
...
+     * @protected
+     */
+    handleAutoplayFail = () => {};
+
+    /**
+     * Autoplay the media
      *
      * @private
      * @emits volume
      * @return {Promise}
```
ConradJChan: Is this jsdoc still accurate?

src/lib/viewers/media/MediaBaseViewer.js
ConradJChan: maybe instead `PLAY_PROMISE_NOT_SUPPORTED = 'play_promise_not_supported'`?

src/lib/viewers/media/MediaBaseViewer.js
```diff
@@ -584,10 +579,20 @@ class MediaBaseViewer extends BaseViewer {
...
         if (arguments.length === 0 || hasValidStart) {
-            this.mediaEl.play();
+            // Play may return a promise depending on browser support. This promise
+            // will resolve when playback starts.
+            // https://webkit.org/blog/7734/auto-play-policy-changes-for-macos/
+            const playPromise = this.mediaEl.play();
             this.handleRate();
             this.handleVolume();
+
+            if (playPromise && typeof playPromise.then === 'function') {
```
ConradJChan: Could condense this to `return (playPromise && typeof playPromise.then === 'function') ? playPromise : Promise.reject(new Error(PLAY_PROMISE_NOT_SUPPORTED))`?

PR #1069, opened 2019-09-10
fix(preview): remove auto-preview experience

mxiao6:


boxcla:

src/lib/viewers/media/MediaBaseViewer.js
```diff
@@ -44,7 +44,6 @@ class MediaBaseViewer extends BaseViewer {
         this.toggleMute = this.toggleMute.bind(this);
         this.handleRate = this.handleRate.bind(this);
         this.handleAutoplay = this.handleAutoplay.bind(this);
-        this.mediaendHandler = this.mediaendHandler.bind(this);
```
ConradJChan: I think leaving in this event is still fine, we just don't need to react to it internally in the preview sdk

mxiao6: I see

src/lib/Preview.js
```diff
@@ -1239,7 +1239,6 @@ class Preview extends EventEmitter {
                 this.ui.finishProgressBar();
                 break;
             case VIEWER_EVENT.mediaEndAutoplay:
```
jstoffan: This entire condition can be removed, as it's now a no-op.

mxiao6: ok

jstoffan: Actually, it looks like it's preventing the event from being forwarded by the `default` condition.

mxiao6: Then the test for it should be removed entirely too, right?

mxiao6: or change it to test the default condition

PR #1071, opened 2019-09-13
feat(dashviewer): Video performance instrumentation

ConradJChan:
Introduces two new metric events:
-  `media_metric_buffer_fill` which represents the initial buffering before playback can begin
- `media_metric_end_playback` which contains metrics around total buffering time, lag/watch ratio, etc...

`media_metric_buffer_fill` will be emitted as soon as it is measured.

`media_metric_end_playback` will be emitted in 3 scenarios:
1. when playback is finished (`ended` event is fired)
2. when the viewer is destroyed (when preview is closed)
3. when the page unloads

As part of this PR, only the `DashViewer` will emit these metrics.

TODO:

- [x] fix unit tests
- [x] add new unit tests

### DIFF
--- a/src/lib/viewers/media/DashViewer.js
+++ b/src/lib/viewers/media/DashViewer.js
@@ -107,8 +107,7 @@ class DashViewer extends VideoBaseViewer {
      * @return {void}
      */
     load() {
-        // Add event listeners for the media element
-        this.addEventListenersForMediaElement();
+        this.loadUI();
 
         this.mediaUrl = this.options.representation.content.url_template;
         this.watermarkCacheBust = Date.now();
@@ -661,7 +660,6 @@ class DashViewer extends VideoBaseViewer {
         }
 
         this.calculateVideoDimensions();
-        this.loadUI();
 
         if (this.isAutoplayEnabled()) {
             this.autoplay();
--- a/src/lib/viewers/media/DashViewer.js
+++ b/src/lib/viewers/media/DashViewer.js
@@ -648,9 +648,8 @@ class DashViewer extends VideoBaseViewer {
             return;
         }
 
-        this.loadUI();
-
         this.calculateVideoDimensions();
+        this.loadUI();
 
         if (this.isAutoplayEnabled()) {
             this.autoplay();
--- a/src/lib/viewers/media/DashViewer.js
+++ b/src/lib/viewers/media/DashViewer.js
@@ -32,7 +32,7 @@ class DashViewer extends VideoBaseViewer {
         this.api = options.api;
         // Bind context for callbacks
         this.adaptationHandler = this.adaptationHandler.bind(this);
-        this.bufferingHandler = this.bufferingHandler.bind(this);
+        this.handleBuffering = this.handleBuffering.bind(this);
         this.getBandwidthInterval = this.getBandwidthInterval.bind(this);
         this.handleAudioTrack = this.handleAudioTrack.bind(this);
         this.handleQuality = this.handleQuality.bind(this);
@@ -164,7 +164,7 @@ class DashViewer extends VideoBaseViewer {
         this.player = new shaka.Player(this.mediaEl);
         this.player.addEventListener('adaptation', this.adaptationHandler);
         this.player.addEventListener('error', this.shakaErrorHandler);
-        this.player.addEventListener('buffering', this.bufferingHandler);
+        this.player.addEventListener('buffering', this.handleBuffering);
         this.player.configure({
             abr: {
                 enabled: false,
@@ -193,14 +193,13 @@ class DashViewer extends VideoBaseViewer {
      * @param {Object} object - BufferingEvent object
      * @param {boolean} object.buffering - Indicates whether the player is buffering or not
      */
-    bufferingHandler({ buffering }) {
+    handleBuffering({ buffering }) {
         const tag = Timer.createTag(this.options.file.id, MEDIA_METRIC.totalBufferLag);
 
         if (buffering) {
             Timer.start(tag);
         } else {
             Timer.stop(tag);
-            this.metrics[MEDIA_METRIC.totalBufferLag] = this.metrics[MEDIA_METRIC.totalBufferLag];
             this.metrics[MEDIA_METRIC.totalBufferLag] += Timer.get(tag).elapsed;
             Timer.reset(tag);
         }
@@ -227,6 +226,10 @@ class DashViewer extends VideoBaseViewer {
      * @return {void}
      */
     processMetrics() {
+        if (!this.loaded) {
+            return;
+        }
+
         const lagLength = this.metrics[MEDIA_METRIC.totalBufferLag];
         const playLength = this.determinePlayLength();
 
--- a/src/lib/viewers/media/DashViewer.js
+++ b/src/lib/viewers/media/DashViewer.js
@@ -1,11 +1,13 @@
+import isFinite from 'lodash/isFinite';
 import VideoBaseViewer from './VideoBaseViewer';
 import PreviewError from '../../PreviewError';
 import fullscreen from '../../Fullscreen';
+import Timer from '../../Timer';
 import { appendQueryParams, getProp } from '../../util';
 import { getRepresentation } from '../../file';
 import { MEDIA_STATIC_ASSETS_VERSION } from '../../constants';
 import getLanguageName from '../../lang';
-import { ERROR_CODE, VIEWER_EVENT } from '../../events';
+import { ERROR_CODE, VIEWER_EVENT, MEDIA_METRIC, MEDIA_METRIC_EVENTS } from '../../events';
 import './Dash.scss';
 
 const CSS_CLASS_DASH = 'bp-media-dash';
@@ -30,14 +32,15 @@ class DashViewer extends VideoBaseViewer {
 
         this.api = options.api;
         // Bind context for callbacks
-        this.loadeddataHandler = this.loadeddataHandler.bind(this);
         this.adaptationHandler = this.adaptationHandler.bind(this);
-        this.shakaErrorHandler = this.shakaErrorHandler.bind(this);
-        this.requestFilter = this.requestFilter.bind(this);
+        this.bufferingHandler = this.bufferingHandler.bind(this);
+        this.getBandwidthInterval = this.getBandwidthInterval.bind(this);
+        this.handleAudioTrack = this.handleAudioTrack.bind(this);
         this.handleQuality = this.handleQuality.bind(this);
         this.handleSubtitle = this.handleSubtitle.bind(this);
-        this.handleAudioTrack = this.handleAudioTrack.bind(this);
-        this.getBandwidthInterval = this.getBandwidthInterval.bind(this);
+        this.loadeddataHandler = this.loadeddataHandler.bind(this);
+        this.requestFilter = this.requestFilter.bind(this);
+        this.shakaErrorHandler = this.shakaErrorHandler.bind(this);
     }
 
     /**
@@ -104,6 +107,9 @@ class DashViewer extends VideoBaseViewer {
      * @return {void}
      */
     load() {
+        // Add event listeners for the media element
+        this.addEventListenersForMediaElement();
+
         this.mediaUrl = this.options.representation.content.url_template;
         this.watermarkCacheBust = Date.now();
         this.mediaEl.addEventListener('loadeddata', this.loadeddataHandler);
@@ -161,6 +167,7 @@ class DashViewer extends VideoBaseViewer {
         this.player = new shaka.Player(this.mediaEl);
         this.player.addEventListener('adaptation', this.adaptationHandler);
         this.player.addEventListener('error', this.shakaErrorHandler);
+        this.player.addEventListener('buffering', this.bufferingHandler);
         this.player.configure({
             abr: {
                 enabled: false,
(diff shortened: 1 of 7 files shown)


boxcla:

src/lib/viewers/media/MediaBaseViewer.js
```diff
@@ -405,6 +419,10 @@ class MediaBaseViewer extends BaseViewer {
      * @return {void}
      */
     setTimeCode() {
+        if (!this.mediaControls) {
```
mickr: Maybe could have a helper function for `hasMediaControls` or instantiate the `this.mediaControls` with an object in the interim until `loadUi` has been called?

mickr:
Seems like this is more than a chore to me

src/lib/viewers/media/MediaBaseViewer.js
ConradJChan: I attempted to move `loadUI` around instead so that the mediaEl callbacks can depend on mediaControls existing

src/lib/viewers/media/DashViewer.js
```diff
@@ -182,6 +188,87 @@ class DashViewer extends VideoBaseViewer {
...
+
+        this.emitMetric(MEDIA_METRIC_EVENTS.endPlayback, { ...this.metrics });
+    }
+
+    /**
+     * Determines the play length, or how much of the media was consumed
+     * @return {number} - The play length in milliseconds
+     */
+    determinePlayLength() {
+        if (!this.mediaEl) {
```
jstoffan: I know it would lead to property accessor duplication, but can we combine these guard clauses?

```
if (!this.mediaEl || !this.mediaEl.played) {
    return -1;
}
```

src/lib/viewers/media/DashViewer.js
```diff
@@ -182,6 +188,87 @@ class DashViewer extends VideoBaseViewer {
...
+
+        const playedParts = this.mediaEl.played;
+
+        if (!playedParts) {
+            return -1;
+        }
+
+        let playLength = 0;
+        for (let i = 0; i < playedParts.length; i += 1) {
+            const start = playedParts.start(i);
```
jstoffan: Just want to confirm that these APIs are supported in all browsers.

src/lib/viewers/media/DashViewer.js
```diff
@@ -182,6 +188,87 @@ class DashViewer extends VideoBaseViewer {
...
+     * @param {boolean} buffering - Indicates whether the player is buffering or not
+     */
+    bufferingHandler({ buffering }) {
+        const tag = Timer.createTag(this.options.file.id, MEDIA_METRIC.totalBufferLag);
+
+        if (buffering) {
+            Timer.start(tag);
+        } else {
+            Timer.stop(tag);
+            this.metrics[MEDIA_METRIC.totalBufferLag] = this.metrics[MEDIA_METRIC.totalBufferLag] || 0;
```
jstoffan: Could we seed `this.metrics` with zero-filled data when the viewer is constructed, instead?

src/lib/viewers/media/DashViewer.js
```diff
@@ -182,6 +188,87 @@ class DashViewer extends VideoBaseViewer {
         this.player.load(this.mediaUrl, this.startTimeInSeconds).catch(this.shakaErrorHandler);
     }
 
+    /**
+     * Handles the buffering events from shaka player
+     *
+     * @see {@link https://shaka-player-demo.appspot.com/docs/api/shaka.Player.html#.event:BufferingEvent}
+     * @param {boolean} buffering - Indicates whether the player is buffering or not
```
jstoffan: Is this still correct?

src/lib/viewers/media/DashViewer.js
```diff
@@ -182,6 +188,87 @@ class DashViewer extends VideoBaseViewer {
...
+     * Processes the media playback metrics
+     * @override
+     * @emits MEDIA_METRIC_EVENTS.endPlayback
+     * @return {void}
+     */
+    processMetrics() {
+        const lagLength = this.metrics[MEDIA_METRIC.totalBufferLag];
+        const playLength = this.determinePlayLength();
+
+        if (!isFinite(lagLength) || playLength <= 0) {
```
jstoffan: Do we only want to process metrics if the user has started watching the video? It seems like we'll miss out on a cohort of behaviors if so.

src/lib/viewers/media/DashViewer.js
```diff
@@ -182,6 +188,87 @@ class DashViewer extends VideoBaseViewer {
...
+
+        if (!playedParts) {
+            return -1;
+        }
+
+        let playLength = 0;
+        for (let i = 0; i < playedParts.length; i += 1) {
+            const start = playedParts.start(i);
+            const end = playedParts.end(i);
+            const duration = end - start;
```
jstoffan: Minor, but this can be inlined since it's only used once.

src/lib/viewers/media/DashViewer.js
```diff
@@ -104,6 +107,8 @@ class DashViewer extends VideoBaseViewer {
      * @return {void}
      */
     load() {
+        this.loadUI();
```
jstoffan: Moving major calls like this can be a somewhat scary change. Does everything work as expected? Does this get called more or less often than before?

src/lib/viewers/media/MediaBaseViewer.js
```diff
@@ -703,6 +780,7 @@ class MediaBaseViewer extends BaseViewer {
         // Make sure it's within bounds
         newTime = Math.max(0, Math.min(newTime, this.mediaEl.duration));
         this.setMediaTime(newTime);
+        this.metrics[MEDIA_METRIC.seeked] = true;
```
jstoffan: Can we centralize our handling of this to the `seeked` handler?

src/lib/viewers/media/MediaBaseViewer.js
```diff
@@ -680,16 +707,66 @@ class MediaBaseViewer extends BaseViewer {
...
+
+        this.processBufferFillMetric();
+    }
+
+    /**
+     * Abstract. Processes the buffer fill metric which represents the initial buffer time before playback begins
+     * @emits MEDIA_METRIC_EVENTS.bufferFill
+     * @return {void}
+     */
+    processBufferFillMetric() {}
```
jstoffan: Will we be able to share this logic between Dash/MP4 viewers? If not, should they instead override the handlers and each add their own additional method calls?

ConradJChan: They should be able to share the logic, but right now I only want to enable it for DashViewer

src/lib/viewers/media/MediaBaseViewer.js
ConradJChan: Seems reasonable. I'll double check that the `seekHandler` gets called during `quickSeek`

src/lib/viewers/media/DashViewer.js
ConradJChan: Yeah that makes sense, I'll remove this guard block

src/lib/viewers/media/DashViewer.js
```diff
@@ -30,14 +31,15 @@ class DashViewer extends VideoBaseViewer {
 
         this.api = options.api;
         // Bind context for callbacks
-        this.loadeddataHandler = this.loadeddataHandler.bind(this);
         this.adaptationHandler = this.adaptationHandler.bind(this);
-        this.shakaErrorHandler = this.shakaErrorHandler.bind(this);
-        this.requestFilter = this.requestFilter.bind(this);
+        this.bufferingHandler = this.bufferingHandler.bind(this);
```
jstoffan: What do you think about using `handleX` for all new handlers? So `handleBuffering` in this case?

src/lib/viewers/media/DashViewer.js
```diff
@@ -182,6 +186,78 @@ class DashViewer extends VideoBaseViewer {
...
+     * @param {boolean} object.buffering - Indicates whether the player is buffering or not
+     */
+    bufferingHandler({ buffering }) {
+        const tag = Timer.createTag(this.options.file.id, MEDIA_METRIC.totalBufferLag);
+
+        if (buffering) {
+            Timer.start(tag);
+        } else {
+            Timer.stop(tag);
+            this.metrics[MEDIA_METRIC.totalBufferLag] = this.metrics[MEDIA_METRIC.totalBufferLag];
```
jstoffan: Is this reassignment needed?

src/lib/viewers/media/DashViewer.js
```diff
@@ -182,6 +186,78 @@ class DashViewer extends VideoBaseViewer {
...
 
+    /**
+     * Handles the buffering events from shaka player
+     *
+     * @see {@link https://shaka-player-demo.appspot.com/docs/api/shaka.Player.html#.event:BufferingEvent}
+     * @param {Object} object - BufferingEvent object
+     * @param {boolean} object.buffering - Indicates whether the player is buffering or not
+     */
+    bufferingHandler({ buffering }) {
+        const tag = Timer.createTag(this.options.file.id, MEDIA_METRIC.totalBufferLag);
```
jstoffan: Should we pull out `totalBufferLag` value into a variable, since it's used so much?

src/lib/viewers/media/DashViewer.js
```diff
@@ -182,6 +186,78 @@ class DashViewer extends VideoBaseViewer {
...
+    /**
+     * Determines the play length, or how much of the media was consumed
+     * @return {number} - The play length in milliseconds
+     */
+    determinePlayLength() {
+        if (!this.mediaEl || !this.mediaEl.played) {
+            return -1;
+        }
+
+        const playedParts = this.mediaEl.played;
```
jstoffan: Is this information easier to get at via [player.getBufferedInfo()](https://shaka-player-demo.appspot.com/docs/api/shaka.extern.html#.BufferedInfo)?

src/lib/viewers/media/MediaBaseViewer.js
```diff
@@ -122,16 +142,15 @@ class MediaBaseViewer extends BaseViewer {
...
-                this.mediaEl.removeEventListener('ended', this.resetPlayIcon);
-                this.mediaEl.removeEventListener('seeked', this.seekHandler);
+                this.removeEventListenersForMediaElement();
+
                 this.mediaEl.removeEventListener('loadeddata', this.loadeddataHandler);
                 this.mediaEl.removeEventListener('error', this.errorHandler);
 
+                // Attempt to process the playback metrics at whatever point of playback has occurred
+                // before we destroy the viewer
+                this.processMetrics();
```
jstoffan: Should we do this at the very top of `destroy`? Also, is it ever going to be possible for metrics to be emitted twice, once by `beforeunload` and once here?

src/lib/viewers/media/MediaBaseViewer.js
```diff
@@ -179,6 +200,16 @@ class MediaBaseViewer extends BaseViewer {
             .catch(this.handleAssetError);
     }
 
+    /**
+     * Add event listeners to the media element related to loading of data
+     * @return {void}
+     */
+    addEventListenersForMediaLoad() {
+        this.mediaEl.addEventListener('canplay', this.stopBufferFillTimer);
```
jstoffan: Should this be named something like `handleCanPlay`?

src/lib/viewers/media/DashViewer.js
ConradJChan: Ah yes, this is useless now that the defaulting to 0 condition has been removed

src/lib/viewers/media/DashViewer.js
ConradJChan: My understanding is that is the buffered ranges, not the actual played ranges

src/lib/viewers/media/MediaBaseViewer.js
ConradJChan: yeah I can do that, but I think what I'll do is also add a guard clause in `processMetrics` to check `this.loaded` to make sure data has loaded first instead of emitting metrics in cases where the viewer hasn't even loaded any data

Also, yeah it should only be emitted once since it's whitelisted in `getMetricsWhitelist`

src/lib/viewers/media/MediaBaseViewer.js
ConradJChan: 👍

src/lib/viewers/media/MediaBaseViewer.js
```diff
@@ -122,13 +146,8 @@ class MediaBaseViewer extends BaseViewer {
...
-                this.mediaEl.removeEventListener('timeupdate', this.setTimeCode);
-                this.mediaEl.removeEventListener('progress', this.progressHandler);
-                this.mediaEl.removeEventListener('volumechange', this.updateVolumeIcon);
-                this.mediaEl.removeEventListener('playing', this.playingHandler);
-                this.mediaEl.removeEventListener('pause', this.pauseHandler);
-                this.mediaEl.removeEventListener('ended', this.resetPlayIcon);
-                this.mediaEl.removeEventListener('seeked', this.seekHandler);
+                this.removeEventListenersForMediaElement();
+
                 this.mediaEl.removeEventListener('loadeddata', this.loadeddataHandler);
```
jstoffan: Should these also live in `removeEventListenersForMediaElement`?

src/lib/viewers/media/MediaBaseViewer.js
```diff
@@ -157,9 +176,11 @@ class MediaBaseViewer extends BaseViewer {
     load() {
         super.load();
 
+        this.addEventListenersForMediaLoad();
+
         const template = this.options.representation.content.url_template;
         this.mediaUrl = this.createContentUrlWithAuthParams(template);
-        this.mediaEl.addEventListener('loadeddata', this.loadeddataHandler);
+
         this.mediaEl.addEventListener('error', this.errorHandler);
```
jstoffan: Should this live in `addEventListenersForMediaLoad`?

src/lib/viewers/media/MediaBaseViewer.js
```diff
@@ -122,13 +146,8 @@ class MediaBaseViewer extends BaseViewer {
...
         try {
             if (this.mediaEl) {
-                this.mediaEl.removeEventListener('timeupdate', this.setTimeCode);
-                this.mediaEl.removeEventListener('progress', this.progressHandler);
-                this.mediaEl.removeEventListener('volumechange', this.updateVolumeIcon);
-                this.mediaEl.removeEventListener('playing', this.playingHandler);
-                this.mediaEl.removeEventListener('pause', this.pauseHandler);
-                this.mediaEl.removeEventListener('ended', this.resetPlayIcon);
-                this.mediaEl.removeEventListener('seeked', this.seekHandler);
+                this.removeEventListenersForMediaElement();
```
jstoffan: What's the reasoning for moving these out? It doesn't seem to be used outside of `destroy`.

src/lib/viewers/media/DashViewer.js
```diff
@@ -182,6 +186,81 @@ class DashViewer extends VideoBaseViewer {
...
+            return;
+        }
+
+        const lagLength = this.metrics[MEDIA_METRIC.totalBufferLag];
+        const playLength = this.determinePlayLength();
+
+        this.metrics[MEDIA_METRIC.totalBufferLag] = lagLength;
+        this.metrics[MEDIA_METRIC.lagRatio] = lagLength / playLength;
+        this.metrics[MEDIA_METRIC.duration] = this.mediaEl ? this.mediaEl.duration * 1000 : 0;
+        this.metrics[MEDIA_METRIC.watchLength] = playLength;
```
jstoffan: Can we keep the metric names consistent with their variable and method names?

src/lib/viewers/media/DashViewer.js
```diff
@@ -182,6 +186,81 @@ class DashViewer extends VideoBaseViewer {
...
+    processMetrics() {
+        if (!this.loaded) {
+            return;
+        }
+
+        const lagLength = this.metrics[MEDIA_METRIC.totalBufferLag];
+        const playLength = this.determinePlayLength();
+
+        this.metrics[MEDIA_METRIC.totalBufferLag] = lagLength;
+        this.metrics[MEDIA_METRIC.lagRatio] = lagLength / playLength;
```
jstoffan: Does this derived value need to be reported separately? We're already reporting both constituent metrics separately.

src/lib/viewers/media/DashViewer.js
```diff
@@ -182,6 +186,81 @@ class DashViewer extends VideoBaseViewer {
...
+     * @return {number} - The play length in milliseconds
+     */
+    determinePlayLength() {
+        if (!this.mediaEl || !this.mediaEl.played) {
+            return -1;
+        }
+
+        const playedParts = this.mediaEl.played;
+        let playLength = 0;
+        for (let i = 0; i < playedParts.length; i += 1) {
```
jstoffan: Might look a bit cleaner via `playedParts.reduce`?

src/lib/viewers/media/MediaBaseViewer.js
```diff
@@ -680,16 +724,64 @@ class MediaBaseViewer extends BaseViewer {
...
+        this.mediaEl.removeEventListener('timeupdate', this.setTimeCode);
+        this.mediaEl.removeEventListener('volumechange', this.updateVolumeIcon);
+    }
+
+    /**
+     * Callback from the 'loadstart' event from the media element. Triggers a timer to measure the initial buffer fill.
+     * @return {void}
+     */
+    handleLoadStart() {
+        const tag = Timer.createTag(this.options.file.id, MEDIA_METRIC.bufferFill);
```
jstoffan: Is it worth creating a super basic helper method for this, like:

```
createTag(tagName) {
    return Timer.createTag(this.options.file.id, tagName);
}

src/lib/viewers/media/DashViewer.js
ConradJChan: At this point, I don't really see a difference since it's just another field in the JSON blob that's being logged

src/lib/viewers/media/DashViewer.js
ConradJChan: 👍

src/lib/viewers/media/DashViewer.js
ConradJChan: It's not actually an array so I don't think we can reduce it

src/lib/viewers/media/MediaBaseViewer.js
ConradJChan: Mainly just to consolidate, I'll add `loaddata` and `error` as well

src/lib/viewers/media/MediaBaseViewer.js
ConradJChan: This one can't because it isn't used in the `DashViewer`

src/lib/viewers/media/MediaBaseViewer.js
ConradJChan: 👍

src/lib/viewers/media/MediaBaseViewer.js
ConradJChan: 👍

PR #1072, opened 2019-09-16
fix(viewer): Re-enable support for range requests for most users

mxiao6:


boxcla:

PR #1073, opened 2019-09-16
fix(viewer): Re-enable support for range requests for most users

jstoffan:


boxcla:

jstoffan:
Looking again, I think #1072 does the same. Closing in favor of that.

PR #1076, opened 2019-09-24
fix(text): Hide print content for users without download permission

jstoffan:
I also tried to consolidate and improve the logic for text selection.


boxcla:

PR #1078, opened 2019-09-25
fix(viewer): Fix 3D viewer menu layout too narrow

mxiao6:


boxcla:

src/lib/viewers/box3d/model3d/Model3D.scss
```diff
@@ -18,6 +18,7 @@ $arrow-top-bottom: 4px solid transparent;
 
     .bp-pullup {
         bottom: 100%;
+        width: fit-content;
```
mickr: Does this only affect Chrome? IE 11 doesn't support this AFAIK

mxiao6: I tested IE 11 and Edge, and it looks fine even before the change.

jstoffan: I don't think we should use experimental CSS values. Does setting `white-space: nowrap;` on the parent work?

PR #1081, opened 2019-10-08
fix(download): show appropriate download error

priyajeet:
for when download might get disabled due to an access policy while the user is already previewing the file without browser refresh.

Also fixes the apostrophes

### DIFF
--- a/src/i18n/en-AU.properties
+++ b/src/i18n/en-AU.properties
@@ -63,7 +63,7 @@ error_password_protected=We’re sorry, but the preview didn't load. This docume
 # Preview error message shown when conversion was unable to process the file at the given time.
 error_try_again_later=We’re sorry, but the preview didn't load. Please try again later.
 # Preview error message shown when conversion failed due to file contents
-error_bad_file=We’re sorry the preview didn't load. This file could not be converted.
+error_bad_file=We’re sorry the preview didn’t load. This file could not be converted.
 # Preview error message shown when the file cannot be downloaded
 error_not_downloadable=Oops! It looks as if something is wrong with this file. We apologise for the inconvenience and recommend that you upload a new version of this file or roll back to a previous version. Please contact us for more details.
 # Preview error message shown when flash is not enabled on their browser
--- a/src/i18n/en-AU.properties
+++ b/src/i18n/en-AU.properties
@@ -41,11 +41,11 @@ has_x_refs=This preview has references you cannot view. Open the file in its nat
 
 # Error messages
 # Default preview error message
-error_generic=We're sorry, the preview didn't load.
+error_generic=We’re sorry, the preview didn't load.
 # Default preview error message
-error_unsupported=We're sorry, we can't preview this file type in your web browser. To view this content, please download and open it on your device.
+error_unsupported=We’re sorry, we can't preview this file type in your web browser. To view this content, please download and open it on your device.
 # Account doesn't have a sufficient tariff to preview the requested file type
-error_account=We're sorry, your account is unable to preview this file type.
+error_account=We’re sorry, your account is unable to preview this file type.
 # No permissions preview error message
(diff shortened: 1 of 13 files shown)


boxcla:

src/lib/Preview.js
```diff
@@ -513,25 +513,27 @@ class Preview extends EventEmitter {
...
      * @public
-     * @return {void}
+     * @return {Promise} download promise
      */
     download() {
         const downloadErrorMsg = __('notification_cannot_download');
         if (!canDownload(this.file, this.options)) {
             this.ui.showNotification(downloadErrorMsg);
-            return;
+            return Promise.reject();
```
ConradJChan: Should we reject with a reason?

src/lib/Preview.js
```diff
@@ -513,25 +513,27 @@ class Preview extends EventEmitter {
      * Downloads the file being previewed.
      *
      * @public
-     * @return {void}
+     * @return {Promise} download promise
```
ConradJChan: This is a public API for the preview instance, right? Is this considered a breaking change?

src/lib/Preview.js
```diff
@@ -541,15 +543,25 @@ class Preview extends EventEmitter {
...
             const getDownloadUrl = appendQueryParams(getDownloadURL(this.file.id, apiHost), queryParams);
-            this.api.get(getDownloadUrl, { headers: this.getRequestHeaders() }).then(data => {
-                const downloadUrl = appendQueryParams(data.download_url, queryParams);
-                this.api.reachability.downloadWithReachabilityCheck(downloadUrl);
-            });
+            downloadPromise = this.api
+                .get(getDownloadUrl, { headers: this.getRequestHeaders() })
+                .then(data => {
+                    const downloadUrl = appendQueryParams(data.download_url, queryParams);
+                    this.api.reachability.downloadWithReachabilityCheck(downloadUrl);
```
ConradJChan: Should this return the promise result from `downloadWithReachabilityCheck`?

src/lib/Preview.js
```diff
@@ -541,15 +543,25 @@ class Preview extends EventEmitter {
...
+                })
+                .catch(error => {
+                    const code = getProp(error, 'response.data.code');
+                    let msg = downloadErrorMsg;
+                    if (code === 'forbidden_by_policy') {
+                        msg = __('notification_cannot_download_due_to_policy');
+                    }
+                    this.ui.showNotification(msg);
```
ConradJChan: Should we do a `Promise.reject` here?

src/lib/Preview.js
```diff
@@ -541,15 +543,25 @@ class Preview extends EventEmitter {
                 params,
             );
 
-            this.api.reachability.downloadWithReachabilityCheck(downloadUrl);
+            downloadPromise = this.api.reachability.downloadWithReachabilityCheck(downloadUrl);
```
ConradJChan: It doesn't look like `downloadWithReachabilityCheck` actually returns a promise?

src/i18n/en-AU.properties
```diff
@@ -41,11 +41,11 @@ has_x_refs=This preview has references you cannot view. Open the file in its nat
 
 # Error messages
 # Default preview error message
-error_generic=We're sorry, the preview didn't load.
+error_generic=We’re sorry, the preview didn't load.
```
ConradJChan: What was wrong with the apostrophes?

src/lib/Preview.js
priyajeet: I am actually gonna remove the return promise. Only downside is that this function is not really being tested properly, specially for errors.

src/i18n/en-AU.properties
priyajeet: Normally for user facing strings one uses the accented versions. It also helps with not having the need to escape.

PR #1084, opened 2019-10-15
feat(printing): Move print() to imageBase

mickr:
Moving the print function to imageBase will allow
MultiImageViewer to inherit the functionality.

PR #1085, opened 2019-10-15
Add printing to imagebase

mickr:
Move printing to the imageBase class to share with MultiImageViewer so TIFF files can print as well.


boxcla:

mickr:
Closed https://github.com/box/box-content-preview/pull/1084 so that I am not PR'ing from the canonical repo.

src/lib/viewers/image/ImageBaseViewer.js
```diff
@@ -434,6 +435,42 @@ class ImageBaseViewer extends BaseViewer {
...
+
+        this.printframe = util.openContentInsideIframe(this.imageEl.outerHTML);
+        this.printImages = this.printframe.contentDocument.querySelectorAll('img');
+        this.printImages.forEach(element => {
+            element.setAttribute('style', 'display: block; margin: 0 auto; width: 100%');
+        });
+
+        this.printframe.contentWindow.focus();
+
+        this.printframe.addEventListener('load', defaultPrintHandler);
```
sahiga: I have the same question about the reordering of the function calls -- does it affect the behavior in any way? Is it necessary because of the `querySelector` -> `querySelectorAll` change?

mickr: Sorry, meant to come back and address this question. I adjusted it because it seems more readable but all in all should act the same way.

src/lib/viewers/image/ImageBaseViewer.js
```diff
@@ -6,6 +6,7 @@ import { ICON_ZOOM_IN, ICON_ZOOM_OUT } from '../../icons/icons';
 
 import { CLASS_INVISIBLE } from '../../constants';
 import { ERROR_CODE, VIEWER_EVENT } from '../../events';
+import * as util from '../../util';
```
mxiao6: Should we only import `openContentInsideIframe` instead of `*`?

src/lib/viewers/image/ImageBaseViewer.js
sahiga: No worries. That makes sense. Thanks for the explanation!

src/lib/viewers/image/ImageBaseViewer.js
```diff
@@ -434,6 +435,42 @@ class ImageBaseViewer extends BaseViewer {
...
+
+            this.printframe.removeEventListener('load', defaultPrintHandler);
+            this.emit('printsuccess');
+        };
+
+        this.printframe = util.openContentInsideIframe(this.imageEl.outerHTML);
+        this.printImages = this.printframe.contentDocument.querySelectorAll('img');
+        this.printImages.forEach(element => {
+            element.setAttribute('style', 'display: block; margin: 0 auto; width: 100%');
+        });
```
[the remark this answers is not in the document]
mickr: I don't think we need to since the preview will already have a border around the image that contains the printing information.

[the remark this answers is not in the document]
ConradJChan: Can you add a screenshot of how printing TIF files look?

src/lib/viewers/image/ImageBaseViewer.js
```diff
@@ -434,6 +435,42 @@ class ImageBaseViewer extends BaseViewer {
...
+        const browserName = Browser.getName();
+
+        /**
+         * Called async to ensure resource is loaded for print preview. Then removes listener to prevent
+         * multiple handlers.
+         *
+         * @return {void}
+         */
+        const defaultPrintHandler = () => {
+            if (browserName === BROWSER_INTERNET_EXPLORER || browserName === BROWSER_EDGE) {
```
jstoffan: It looks like this logic is duplicated line-for-line in PlainTextViewer. Not sure it's worth refactoring, though, since the viewers aren't closely related.

src/lib/constants.js
```diff
@@ -137,5 +137,8 @@ export const METADATA = {
     TEMPLATE_AUTOCAD: 'autocad',
 };
 
+export const BROWSER_EDGE = 'Edge';
```
jstoffan: Do these really _need_ to be constants? If so, can we replace any other hardcoded usages of the strings, as well? Can we namespace the constant, like so?

```
const BROWSERS = {
    EDGE: 'Edge',
    EXPLORER: 'Explorer,
};
```

src/lib/viewers/image/__tests__/ImageBaseViewer-test.js
```diff
@@ -648,4 +649,74 @@ describe('lib/viewers/image/ImageBaseViewer', () => {
...
+            };
+
+            stubs.openContentInsideIframe = sandbox.stub(util, 'openContentInsideIframe').returns(stubs.mockIframe);
+            stubs.getName = sandbox.stub(Browser, 'getName');
+        });
+
+        it('should open the content inside an iframe, center, and focus', () => {
+            imageBase.print();
+            expect(stubs.openContentInsideIframe).to.be.called;
+            expect(imageBase.printImages[0].style.display).to.equal('block');
```
jstoffan: Can we match the `style` attribute to the string we expect rather than each sub-attribute?

mickr:
Screen Shot 2019-10-22 at 10 36 01 AM

src/lib/viewers/image/ImageBaseViewer.js
mickr: I can follow up on that sometime as a cleanup task.

src/lib/__tests__/PageControls-test.js
```diff
@@ -3,6 +3,7 @@ import PageControls from '../PageControls';
 import Controls from '../Controls';
 import fullscreen from '../Fullscreen';
 import Browser from '../Browser';
+import { BROWSERS } from '../constants';
```
jstoffan: Minor, but it looks like all these classes already import `Browser`. If we move the constant there, we could do something like the following to avoid the new dependency:

```
import Browser, { Browsers } from '.../../Browser';
```

src/lib/viewers/image/ImageBaseViewer.js
```diff
@@ -434,6 +435,43 @@ class ImageBaseViewer extends BaseViewer {
...
+            }
+
+            this.printframe.removeEventListener('load', defaultPrintHandler);
+            this.emit(VIEWER_EVENT.printSuccess);
+        };
+
+        this.printframe = openContentInsideIframe(this.imageEl.outerHTML);
+        this.printImages = this.printframe.contentDocument.querySelectorAll('img');
+
+        for (let i = 0; i < this.printImages.length; i += 1) {
```
jstoffan: Minor, but if the index isn't needed, iterators are usually preferred:

```
this.printImages.forEach(printImage => {
    printImage.setAttribute('style', 'display: block; margin: 0 auto; width: 100%');
});
```

mickr: IE11 doesn't support foreach from a NodeList, so instead of using something like `Array.prototype.slice.call(this.printImages).forEach(ele =>...` or different way of converting the nodelist to an array I just went with a simpler for loop.

jstoffan: Ah, I didn't consider that. Sounds good!

PR #1087, opened 2019-10-23
fix(mediaviewer): Refresh token if token expired

mxiao6:
Will add tests later. The related BUIE change: https://github.com/box/box-ui-elements/pull/1689

### DIFF
--- a/src/lib/Preview.js
+++ b/src/lib/Preview.js
@@ -109,9 +109,6 @@ class Preview extends EventEmitter {
     /** @property {Object} - Map of disabled viewer names */
     disabledViewers = {};
 
-    /** @property {string|Function} - Access token */
-    token = '';
-
     /** @property {Object} - Current viewer instance */
     viewer;
 
@@ -974,9 +971,6 @@ class Preview extends EventEmitter {
         // Add the response interceptor to the preview instance
         this.options.responseInterceptor = options.responseInterceptor;
 
-        // Add the token generator to refresh the token if necessary
-        this.options.tokenGenerator = options.token;
-
         // Disable or enable viewers based on viewer options
         Object.keys(this.options.viewers).forEach(viewerName => {
             const isDisabled = this.options.viewers[viewerName].disabled;
@@ -1005,6 +999,7 @@ class Preview extends EventEmitter {
             location: this.location,
             cache: this.cache,
             ui: this.ui,
+            refreshToken: this.refreshToken,
         });
     }
 
@@ -1885,6 +1880,21 @@ class Preview extends EventEmitter {
         const fileId = typeof fileIdOrFile === 'string' ? fileIdOrFile : fileIdOrFile.id;
         return getProp(this.previewOptions, `fileOptions.${fileId}.${optionName}`);
     }
+
+    /**
+     * Refresh the access token
+     *
+     * @private
+     * @return {Promise<string|Error>}
+     */
+    refreshToken = () => {
+        if (typeof this.previewOptions.token !== 'function') {
+            return Promise.reject(new Error('Token is not a function and cannot be refreshed.'));
+        }
+        return getTokens(this.file.id, this.previewOptions.token).then(
+            tokenOrTokenMap => tokenOrTokenMap[this.file.id],
+        );
+    };
 }
 
 global.Box = global.Box || {};
--- a/src/lib/Preview.js
+++ b/src/lib/Preview.js
@@ -109,7 +109,7 @@ class Preview extends EventEmitter {
     /** @property {Object} - Map of disabled viewer names */
     disabledViewers = {};
 
-    /** @property {string} - Access token */
+    /** @property {string|Function} - Access token */
     token = '';
 
     /** @property {Object} - Current viewer instance */
@@ -973,6 +973,9 @@ class Preview extends EventEmitter {
 
         // Add the response interceptor to the preview instance
         this.options.responseInterceptor = options.responseInterceptor;
+
+        // Add the token generator to refresh the token if necessary
+        this.options.tokenGenerator = options.token;
 
         // Disable or enable viewers based on viewer options
         Object.keys(this.options.viewers).forEach(viewerName => {
--- a/src/lib/viewers/media/DashViewer.js
+++ b/src/lib/viewers/media/DashViewer.js
@@ -505,6 +505,18 @@ class DashViewer extends VideoBaseViewer {
                 this.handleDownloadError(error, downloadURL);
                 return;
             }
+
+            if (
+                normalizedShakaError.code === shaka.util.Error.Code.BAD_HTTP_STATUS &&
+                normalizedShakaError.data[1] === 401 // token expired
+            ) {
+                this.refreshToken().then(newToken => {
+                    this.options.token = newToken;
+                    this.player.retryStreaming();
+                });
+                return;
+            }
+
             // critical error
             this.triggerError(error);
         }
--- a/src/lib/viewers/media/DashViewer.js
+++ b/src/lib/viewers/media/DashViewer.js
@@ -40,6 +40,7 @@ class DashViewer extends VideoBaseViewer {
         this.loadeddataHandler = this.loadeddataHandler.bind(this);
         this.requestFilter = this.requestFilter.bind(this);
         this.shakaErrorHandler = this.shakaErrorHandler.bind(this);
+        this.restartPlayback = this.restartPlayback.bind(this);
     }
 
     /**
@@ -478,6 +479,33 @@ class DashViewer extends VideoBaseViewer {
     }
 
     /**
+     * Determain whether is an expired token error
+     *
+     * @private
+     * @param {PreviewError} error
+     * @return {bool}
+     */
+    isExpiredTokenError(error) {
+        const errorDetails = error.details;
+        // unauthorized error may be cuased by token expired
+        return errorDetails.code === shaka.util.Error.Code.BAD_HTTP_STATUS && errorDetails.data[1] === 401;
+    }
+
+    /**
+     * Restart playback using new token
+     *
+     * @private
+     * @param {string} newToken - new token
+     * @return {void}
+     */
+    restartPlayback(newToken) {
+        this.options.token = newToken;
+        if (this.player.retryStreaming()) {
+            this.retryTokenCount = 0;
+        }
+    }
+
+    /**
      * Handles errors thrown by shaka player. See https://shaka-player-demo.appspot.com/docs/api/shaka.util.Error.html
      *
      * @private
@@ -491,6 +519,7 @@ class DashViewer extends VideoBaseViewer {
             __('error_refresh'),
             {
                 code: normalizedShakaError.code,
+                data: normalizedShakaError.data,
                 severity: normalizedShakaError.severity,
             },
             `Shaka error. Code = ${normalizedShakaError.code}, Category = ${
@@ -506,14 +535,7 @@ class DashViewer extends VideoBaseViewer {
                 return;
             }
 
-            if (
-                normalizedShakaError.code === shaka.util.Error.Code.BAD_HTTP_STATUS &&
-                normalizedShakaError.data[1] === 401 // token expired
-            ) {
-                this.refreshToken().then(newToken => {
-                    this.options.token = newToken;
-                    this.player.retryStreaming();
-                });
+            if (this.handleExpiredTokenError(error)) {
                 return;
             }
 
--- a/src/lib/viewers/media/DashViewer.js
+++ b/src/lib/viewers/media/DashViewer.js
@@ -482,13 +482,12 @@ class DashViewer extends VideoBaseViewer {
      * Determain whether is an expired token error
      *
      * @private
-     * @param {PreviewError} error
+     * @param {Object} details - error details
      * @return {bool}
      */
-    isExpiredTokenError(error) {
-        const errorDetails = error.details;
-        // unauthorized error may be cuased by token expired
-        return errorDetails.code === shaka.util.Error.Code.BAD_HTTP_STATUS && errorDetails.data[1] === 401;
+    isExpiredTokenError({ details }) {
+        // unauthorized error may be caused by token expired
+        return details.code === shaka.util.Error.Code.BAD_HTTP_STATUS && details.data[1] === 401;
     }
 
     /**
--- a/src/lib/viewers/media/DashViewer.js
+++ b/src/lib/viewers/media/DashViewer.js
@@ -527,15 +527,15 @@ class DashViewer extends VideoBaseViewer {
             }, Severity = ${normalizedShakaError.severity}, Data = ${normalizedShakaError.data.toString()}`,
         );
 
+        if (this.handleExpiredTokenError(error)) {
+            return;
+        }
+
         if (normalizedShakaError.severity > SHAKA_CODE_ERROR_RECOVERABLE) {
             // Anything greater than a recoverable error should be critical
             if (normalizedShakaError.code === shaka.util.Error.Code.HTTP_ERROR) {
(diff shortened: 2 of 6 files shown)


boxcla:

src/lib/viewers/media/DashViewer.js
```diff
@@ -505,6 +505,18 @@ class DashViewer extends VideoBaseViewer {
                 this.handleDownloadError(error, downloadURL);
                 return;
             }
+
+            if (
+                normalizedShakaError.code === shaka.util.Error.Code.BAD_HTTP_STATUS &&
+                normalizedShakaError.data[1] === 401 // token expired
+            ) {
+                this.refreshToken().then(newToken => {
```
ConradJChan: Does this handle the case where there is no tokenGenerator function? Should we add a exit condition so we don't retry infinitely?

src/lib/Preview.js
```diff
@@ -109,7 +109,7 @@ class Preview extends EventEmitter {
     /** @property {Object} - Map of disabled viewer names */
     disabledViewers = {};
 
-    /** @property {string} - Access token */
+    /** @property {string|Function} - Access token */
```
ConradJChan: Hmm... do we actually store the token on the Preview instance? If not, maybe we can remove this

src/lib/Preview.js
```diff
@@ -974,6 +974,9 @@ class Preview extends EventEmitter {
         // Add the response interceptor to the preview instance
         this.options.responseInterceptor = options.responseInterceptor;
 
+        // Add the token generator to refresh the token if necessary
+        this.options.tokenGenerator = options.token;
```
ConradJChan: It seems like `previewOptions.token` is can already be the token or token generator, can we re-use that?

src/lib/viewers/media/MediaBaseViewer.js
```diff
@@ -260,6 +267,17 @@ class MediaBaseViewer extends BaseViewer {
...
     }
 
+    /**
+     * Refresh the access token
+     *
+     * @private
+     * @return {Promise<Object>}
+     */
+    refreshToken = async () => {
+        const tokenMap = await getTokens(this.options.file.id, this.options.tokenGenerator);
```
ConradJChan: Could we instead have `Preview.js` be the only place to set/generate tokens? Perhaps when we `createViewerOptions` we could pass in an callback to `refreshToken`

src/lib/viewers/media/MediaBaseViewer.js
```diff
@@ -229,6 +231,11 @@ class MediaBaseViewer extends BaseViewer {
             return;
         }
 
+        if (this.loaded) {
+            this.play(this.currentTime);
+            return;
+        }
+
```
ConradJChan: Does this need to make its way into `DashViewer`'s `loadeddataHandler` as well?

src/lib/viewers/media/MediaBaseViewer.js
```diff
@@ -273,7 +291,20 @@ class MediaBaseViewer extends BaseViewer {
         console.error(err);
 
         const errorCode = getProp(err, 'target.error.code');
-        const errorDetails = errorCode ? { error_code: errorCode } : {};
+        const errorMessage = getProp(err, 'target.error.message');
+        const errorDetails = errorCode ? { error_code: errorCode, error_message: errorMessage } : {};
```
ConradJChan: Does this override the `errorMessage` defined above on L294?

src/lib/viewers/media/MediaBaseViewer.js
```diff
@@ -273,7 +291,20 @@ class MediaBaseViewer extends BaseViewer {
...
+
+        // refresh the token if token expired
+        if (errorCode === MediaError.MEDIA_ERR_NETWORK && errorMessage.includes(MEDIA_TOKEN_EXPIRE_ERROR)) {
+            this.refreshToken().then(newToken => {
```
ConradJChan: Similar comment about max retries and bailing out to avoid infinite retries

src/lib/viewers/media/MediaBaseViewer.js
mxiao6: `DashViewer` loads data in another way. Calling `this.player.retryStreaming()` is enough for `DashViewer`.

src/lib/Preview.js
mxiao6: `previewOptions.token` is the token (string) generated by token generator and it (the string) will be passed into viewer to generate urls and headers, etc. So we cannot merge token and generator into one variable.

src/lib/Preview.js
```diff
@@ -127,6 +125,9 @@ class Preview extends EventEmitter {
     /** @property {number} - Number of times a particular preview has been retried */
     retryCount = 0;
 
+    /** @property {number} - Number of times refreshing token has been retried */
+    retryTokenCount = 0;
```
ConradJChan: I'm not sure that this retry count is needed

src/lib/Preview.js
```diff
@@ -1882,6 +1884,28 @@ class Preview extends EventEmitter {
...
+
+    /**
+     * Refresh the access token
+     *
+     * @private
+     * @return {Promise<string|Error>}
+     */
+    refreshToken = () => {
+        if (typeof this.previewOptions.token !== 'function') {
+            return Promise.reject(new Error('Token expired and cannot refresh token.'));
```
ConradJChan: Maybe instead the message should be "Token is not a function and cannot be refreshed"?

src/lib/Preview.js
```diff
@@ -1882,6 +1884,28 @@ class Preview extends EventEmitter {
...
+        }
+        if (this.retryTokenCount >= RETRY_TOKEN_COUNT) {
+            return Promise.reject(new Error('Reach refreshing token limit.'));
+        }
+        this.retryTokenCount += 1;
+        return getTokens(this.file.id, this.previewOptions.token).then(tokenOrTokenMap => {
+            if (!tokenOrTokenMap || typeof tokenOrTokenMap === 'string') {
```
ConradJChan: It seems like `getTokens` resolves to a tokenMap, even if the token generator only returns a string, if so, then we can get rid of L1903-L1905

src/lib/viewers/media/DashViewer.js
```diff
@@ -505,6 +505,23 @@ class DashViewer extends VideoBaseViewer {
                 this.handleDownloadError(error, downloadURL);
                 return;
             }
+
+            if (
+                normalizedShakaError.code === shaka.util.Error.Code.BAD_HTTP_STATUS &&
+                normalizedShakaError.data[1] === 401 // token expired
+            ) {
+                this.options
```
ConradJChan: I think the retry logic should be in the viewers itself, because the likelihood is that the streaming itself will fail, not the refreshing of the token

src/lib/viewers/media/DashViewer.js
```diff
@@ -505,6 +506,28 @@ class DashViewer extends VideoBaseViewer {
...
+                    error.details.error_message = 'Reach refreshing token limit for unauthorized error.';
+                } else {
+                    this.options
+                        .refreshToken()
+                        .then(newToken => {
+                            this.options.token = newToken;
+                            this.player.retryStreaming();
+                        })
+                        .catch(tokenError => {
+                            this.triggerError(tokenError);
```
ConradJChan: What is the expected UX when this error is triggered? Should this instead re-use the `error` defined on L490 so it's not logged as a generic viewer load error?

src/lib/viewers/media/DashViewer.js
```diff
@@ -505,6 +506,28 @@ class DashViewer extends VideoBaseViewer {
                 this.handleDownloadError(error, downloadURL);
                 return;
             }
+
+            if (
+                normalizedShakaError.code === shaka.util.Error.Code.BAD_HTTP_STATUS &&
+                normalizedShakaError.data[1] === 401 // unauthorized error may be cuased by token expired
+            ) {
+                if (this.retryTokenCount >= MAX_RETRY_TOKEN) {
```
ConradJChan: Do we want 3 retries or 4?

src/lib/viewers/media/DashViewer.js
```diff
@@ -505,6 +506,28 @@ class DashViewer extends VideoBaseViewer {
...
+                    error.details.error_message = 'Reach refreshing token limit for unauthorized error.';
+                } else {
+                    this.options
+                        .refreshToken()
+                        .then(newToken => {
+                            this.options.token = newToken;
+                            this.player.retryStreaming();
```
ConradJChan: Should we reset `retryTokenCount` on successful re-load?

src/lib/viewers/media/MediaBaseViewer.js
```diff
@@ -273,7 +285,32 @@ class MediaBaseViewer extends BaseViewer {
...
 
         const errorCode = getProp(err, 'target.error.code');
-        const errorDetails = errorCode ? { error_code: errorCode } : {};
+        const errorMessage = getProp(err, 'target.error.message');
+        const errorDetails = errorCode ? { error_code: errorCode, error_message: errorMessage } : {};
+
+        // refresh the token if token expired
+        if (errorCode === MediaError.MEDIA_ERR_NETWORK && errorMessage.includes(MEDIA_TOKEN_EXPIRE_ERROR)) {
+            if (this.retryTokenCount >= MAX_RETRY_TOKEN) {
+                errorDetails.error_message += 'Reach refreshing token limit for unauthorized error.';
```
ConradJChan: If we're concatenating to an existing message, should we add a space?

src/lib/viewers/media/MediaBaseViewer.js
```diff
@@ -273,7 +285,32 @@ class MediaBaseViewer extends BaseViewer {
         console.error(err);
 
         const errorCode = getProp(err, 'target.error.code');
-        const errorDetails = errorCode ? { error_code: errorCode } : {};
+        const errorMessage = getProp(err, 'target.error.message');
+        const errorDetails = errorCode ? { error_code: errorCode, error_message: errorMessage } : {};
+
+        // refresh the token if token expired
+        if (errorCode === MediaError.MEDIA_ERR_NETWORK && errorMessage.includes(MEDIA_TOKEN_EXPIRE_ERROR)) {
```
ConradJChan: What do you think of abstracting this to a common method on `MediaBaseViewer` that can be reused across the different subclasses?

Something like:
```
handleExpiredTokenError(error) {
    if (this.isExpiredTokenError(error)) { ... }
    else {
        this.options.refreshToken().then(this.restartPlayback).catch(error => this.triggerError(error))
    }
}
```

And then the classes that want to use it can implement `isExpiredTokenError` and `restartPlayback`

src/lib/viewers/media/DashViewer.js
mxiao6: Successfully generating a new token doesn't mean that the new token is valid. If a user gives a problematic token generator function, we can always get a new invalid token, and it's still 401 error. Then we'll refresh it infinitely.

src/lib/viewers/media/MediaBaseViewer.js
```diff
@@ -273,14 +355,18 @@ class MediaBaseViewer extends BaseViewer {
...
+        const errorDetails = errorCode ? { error_code: errorCode, error_message: errorMessage } : {};
         const error = new PreviewError(ERROR_CODE.LOAD_MEDIA, __('error_refresh'), errorDetails);
 
-        if (!this.isLoaded()) {
-            this.handleDownloadError(error, this.mediaUrl);
-        } else {
-            this.triggerError(error);
+        const isHandled = this.handleExpiredTokenError(error);
+
+        if (!isHandled) {
```
ConradJChan: Instead of further nesting, perhaps do an early exit?
```
if (this.handleExpiredTokenError(error)) {
    return;
}
```

src/lib/Preview.js
```diff
@@ -1882,6 +1880,21 @@ class Preview extends EventEmitter {
...
+
+    /**
+     * Refresh the access token
+     *
+     * @private
+     * @return {Promise<string|Error>}
+     */
+    refreshToken = () => {
+        if (typeof this.previewOptions.token !== 'function') {
+            return Promise.reject(new Error('Token is not a function and cannot be refreshed.'));
```
mickr: Was the token originally a string and you're converting it to a function that has to be passed in?

src/lib/Preview.js
```diff
@@ -109,9 +109,6 @@ class Preview extends EventEmitter {
     /** @property {Object} - Map of disabled viewer names */
     disabledViewers = {};
 
-    /** @property {string} - Access token */
-    token = '';
-
```
mickr: Was the token originally a string and you're converting it to a function that has to be passed in?

src/lib/viewers/media/MediaBaseViewer.js
```diff
@@ -260,6 +275,74 @@ class MediaBaseViewer extends BaseViewer {
...
+     * @return {boolean} True if it is a token error and is handled
+     */
+    handleExpiredTokenError(error) {
+        if (this.isExpiredTokenError(error)) {
+            if (this.retryTokenCount >= MAX_RETRY_TOKEN) {
+                const tokenError = new PreviewError(
+                    ERROR_CODE.TOKEN_NOT_VALID,
+                    null,
+                    { silent: true },
+                    'Reach refreshing token limit for unauthorized error.',
```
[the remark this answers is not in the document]
mxiao6: This is the error message for the developers not the displayed message.

src/lib/viewers/media/MediaBaseViewer.js
```diff
@@ -260,6 +275,74 @@ class MediaBaseViewer extends BaseViewer {
...
+     *
+     * @protected
+     * @param {string} newToken - new token
+     * @return {void}
+     */
+    restartPlayback(newToken) {
+        const { currentTime } = this.mediaEl;
+        this.currentTime = currentTime;
+        this.options.token = newToken;
+        this.mediaUrl = this.createContentUrlWithAuthParams(this.options.representation.content.url_template);
```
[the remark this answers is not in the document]
mxiao6: will change in the `restartPlayback()`

src/lib/Preview.js
mxiao6: User can pass a token or a string. In BUIE, a string was passed in Preview originally and I converted the string to a generation function in this ticket: https://github.com/box/box-ui-elements/pull/1689.

src/lib/Preview.js
mxiao6: We're currently not using this property at all. And the token is consumed by the viewer, so we actually don't need to store it in Preview instance.

src/lib/viewers/media/DashViewer.js
```diff
@@ -477,6 +478,33 @@ class DashViewer extends VideoBaseViewer {
...
+    /**
+     * Determain whether is an expired token error
+     *
+     * @private
+     * @param {PreviewError} error
+     * @return {bool}
+     */
+    isExpiredTokenError(error) {
+        const errorDetails = error.details;
+        // unauthorized error may be cuased by token expired
```
ConradJChan: spelling error (caused)

src/lib/viewers/media/MediaBaseViewer.js
```diff
@@ -260,6 +275,74 @@ class MediaBaseViewer extends BaseViewer {
...
     }
 
+    /**
+     * Determain whether is an expired token error
+     *
+     * @protected
+     * @param {PreviewError} error
+     * @return {bool}
+     */
+    isExpiredTokenError(error) {
```
ConradJChan: If details is all we care about, we could destructure that out

mxiao6:
> Where did we land on the UX of when the token expires during playback?

The video/audio will stuck just like before.

PR #1088, opened 2019-10-24
feat(docviewer): Toggle find bar control

ConradJChan:
Screen Shot 2019-10-24 at 4 22 01 PM

And don't worry about the consistency with the zoom in/out icons because they'll be changing


boxcla:

src/lib/viewers/doc/DocBaseViewer.js
```diff
@@ -771,16 +772,26 @@ class DocBaseViewer extends BaseViewer {
...
-        this.findBar = new DocFindBar(this.findBarEl, this.pdfFindController, this.pdfEventBus, canDownload);
+        this.findBar = new DocFindBar(this.findBarEl, this.pdfFindController, this.pdfEventBus);
         this.findBar.addListener(VIEWER_EVENT.metric, this.emitMetric);
         this.findBar.addListener('close', this.handleFindBarClose);
     }
 
+    /**
+     * Determines if findbar is disabled
+     *
+     * @protected
```
mxiao6: Should this be `@private` since we only use this function within this class?

mickr:
I think this may need to be rebased

src/lib/icons/search_16px.svg
```diff
@@ -0,0 +1,6 @@
+<svg xmlns="http://www.w3.org/2000/svg" width="18" height="18" viewBox="0 0 15 14">
+    <g fill="none" fill-rule="evenodd">
```
jstoffan: Is this group needed?

src/i18n/en-US.properties
```diff
@@ -38,6 +38,8 @@ text_truncated=This file has been truncated due to size limits. Please download
 toggle_thumbnails=Toggle thumbnails
 # Message when file has inaccessible xrefs
 has_x_refs=This preview has references you cannot view. Open the file in its native application.
+# Button tooltip for toggling find bar
+toggle_findbar=Toggle findbar
```
jstoffan: Minor, but can this live up next to `toggle_thumbnails`?

src/lib/viewers/doc/DocFindBar.js
```diff
@@ -401,6 +401,19 @@ class DocFindBar extends EventEmitter {
...
         this.bar.classList.add(CLASS_HIDDEN);
     }
+
+    /**
+     * Toggles the Find Bar open and closed
+     *
+     * @return {void}
+     */
+    toggle() {
+        if (!this.opened) {
```
jstoffan: We usually want to make the main condition positive, so `if (this.opened)` in this case.

src/lib/icons/search_16px.svg
ConradJChan: Yeah, a black background appears behind the search icon without it

src/lib/viewers/doc/DocBaseViewer.js
```diff
@@ -774,16 +775,26 @@ class DocBaseViewer extends BaseViewer {
...
     }
 
+    /**
+     * Determines if findbar is disabled
+     *
+     * @private
+     * @return {boolean}
+     */
+    isFindDisabled() {
+        const canDownload = checkPermission(this.options.file, PERMISSION_DOWNLOAD);
```
mickr: very minor: I think the intention with this `const` is to be more readable so I would probably add the negation in the const creation and name it as such.

PR #1092, opened 2019-10-30
feat(zoom controls): Restyle zoom controls

ConradJChan:
Updated the zoom controls across Documents, Images, and Plain Text viewers.

TODO:
- [x] Unit Tests

Document Viewer:
Screen Shot 2019-10-30 at 11 22 50 AM

Plain Text Viewer:
Screen Shot 2019-10-30 at 11 21 56 AM

Image Viewer:
Screen Shot 2019-10-30 at 11 20 43 AM

MultiImage Viewer:
Screen Shot 2019-10-30 at 11 19 26 AM


boxcla:

src/lib/ZoomControls.js
```diff
@@ -0,0 +1,134 @@
+import isFinite from 'lodash/isFinite';
+import EventEmitter from 'events';
+import { ICON_ZOOM_IN, ICON_ZOOM_OUT } from './icons/icons';
+
+const CLASS_ZOOM_CURRENT_SCALE = 'bp-zoom-current-scale';
```
mickr: minor: Seems like there could be a method that is like `getClasses` that would generate the proper classes for the different type of viewer (if those are required to be different per viewer)

src/lib/ZoomControls.js
```diff
@@ -0,0 +1,134 @@
...
+        this.setCurrentScale(currentScale);
+    }
+
+    /**
+     * Sets the current scale
+     *
+     * @param {number} scale - New scale to be set as current, range 0-1
+     * @return {void}
+     */
+    setCurrentScale(scale) {
```
mickr: Should this have a default of 0?

src/lib/ZoomControls.js
```diff
@@ -0,0 +1,134 @@
...
+
+        this.controls.add(
+            __('zoom_out'),
+            this.handleZoomOut,
+            `${CLASS_ZOOM_OUT_BUTTON} ${zoomOutClassName}`,
+            ICON_ZOOM_OUT,
+        );
+        this.controls.add(
+            __('zoom_current_scale'),
+            undefined,
```
mickr: Will this not reset the scale to 100% or some known factor?

src/lib/ZoomControls.js
ConradJChan: I tried to keep this as a reusable component that is agnostic of viewers, so in that sense, I allow the viewers to specify what classnames they want to use

src/lib/ZoomControls.js
```diff
@@ -0,0 +1,134 @@
+import isFinite from 'lodash/isFinite';
+import EventEmitter from 'events';
+import { ICON_ZOOM_IN, ICON_ZOOM_OUT } from './icons/icons';
+
+const CLASS_ZOOM_CURRENT_SCALE = 'bp-zoom-current-scale';
+const CLASS_ZOOM_IN_BUTTON = 'bp-zoom-in-btn';
+const CLASS_ZOOM_OUT_BUTTON = 'bp-zoom-out-btn';
+
+class ZoomControls extends EventEmitter {
```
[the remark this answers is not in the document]
ConradJChan: I tried to pattern it off of `PageControls.js`. I don't think we want to extend `Controls.js` because it's more of a composition relationship than an inheritance relationship since we leverage controls.js to manage the buttons

src/lib/ZoomControls.js
ConradJChan: Not for now, the decision is to let it be read only for now and then we can iterate on it

src/lib/ZoomControls.js
ConradJChan: I don't think so because that would mean if the calling code omits the scale argument we want it to be a noop rather than showing an incorrect value

src/lib/ZoomControls.js
```diff
@@ -0,0 +1,134 @@
...
+        }
+    }
+
+    /**
+     * Handles the zoom in button click
+     *
+     * @emits zoomin
+     */
+    handleZoomIn() {
+        this.emit('zoomin');
```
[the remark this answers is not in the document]
ConradJChan: Hmm, I'm not sure. The event strings that live in `events.js` appear to be events that are emitted externally outside of the preview sdk, whereas this events are just emitted from the `ZoomControls` to the consuming Viewer and handled there. But since I am copying the strings around in the different viewers, I'll create constants for them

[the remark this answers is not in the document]
mickr: 👍 Yeah, that is what I was going for, I don't have a preference where they live.

src/lib/ZoomControls.js
```diff
@@ -0,0 +1,140 @@
...
+     *
+     * @param {number} scale - New scale to be set as current, range 0-1
+     * @return {void}
+     */
+    setCurrentScale(scale) {
+        if (!isFinite(scale)) {
+            return;
+        }
+
+        this.currentScale = Math.round(scale * 100);
```
manishPh: I see uses of both `Math.round()` and `Math.ceil()` for converting ranges of 0-1 to integers. Intentional?

src/lib/ZoomControls.js
```diff
@@ -0,0 +1,140 @@
...
+        const zoomInElement = this.controlsElement.querySelector(`.${CLASS_ZOOM_IN_BUTTON}`);
+
+        if (zoomOutElement) {
+            zoomOutElement.disabled = this.currentScale <= this.minZoom;
+        }
+
+        if (zoomInElement) {
+            zoomInElement.disabled = this.currentScale >= this.maxZoom;
+        }
+    }
```
manishPh: The function name checks for enablement while setting `.disabled` to boolean on zoomElement. Also, may be to check for enablement somewhere, do we have to do `if(!disabled)`? Double negative can be confusing. Thoughts?

src/lib/Controls.scss
```diff
@@ -153,3 +153,8 @@
         display: block;
     }
 }
+
+.bp-zoom-current-scale {
```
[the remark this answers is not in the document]
ConradJChan: So far we haven't adopted any SUIT conventions in the preview sdk, but that could be a discussion we can have

src/lib/icons/zoom_in_16px.svg
```diff
@@ -0,0 +1,6 @@
+<svg xmlns="http://www.w3.org/2000/svg" width="10" height="10" viewBox="0 0 10 10">
```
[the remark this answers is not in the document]
ConradJChan: I'll rename the files :)

src/lib/viewers/image/ImageBaseViewer.js
```diff
@@ -77,8 +82,8 @@ class ImageBaseViewer extends BaseViewer {
 
         const loadOriginalDimensions = this.setOriginalImageSize(this.imageEl);
         loadOriginalDimensions.then(() => {
-            this.loadUI();
             this.zoom();
+            this.loadUI();
```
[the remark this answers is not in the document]
ConradJChan: Shouldn't but I think I might need to put some guards in the downstream code because to make sure the controls are initialized

src/lib/Controls.js
```diff
@@ -183,40 +183,46 @@ class Controls {
...
-        button.setAttribute('type', 'button');
-        button.setAttribute('title', text);
-        button.className = `${CONTROLS_BUTTON_CLASS} ${classList}`;
-        button.addEventListener('click', handler);
+        const element = document.createElement(tag);
+        element.setAttribute('aria-label', text);
+        element.setAttribute('title', text);
+
+        if (tag === 'button') {
+            element.setAttribute('type', 'button');
```
manishPh: Create const for 'button' since being used multiple times?

src/lib/ZoomControls.js
```diff
@@ -0,0 +1,140 @@
...
+     * @param {number} currentScale - Initial scale value, assumes range of 0-1
+     * @param {number} [options.maxZoom] - Maximum zoom, range 0-1
+     * @param {number} [options.minZoom] - Minimum zoom, range 0-1
+     * @param {String} [options.zoomInClassName] - Class name for zoom in button
+     * @param {String} [options.zoomOutClassName] - Class name for zoom out button
+     * @return {void}
+     */
+    add(
+        currentScale,
+        { zoomOutClassName = '', zoomInClassName = '', minZoom = 0, maxZoom = Number.POSITIVE_INFINITY } = {},
```
manishPh: `maxZoom = Number.POSITIVE_INFINITY` what is the range of maxZoom? Is it 0-1 as the JSDoc says?

src/lib/Controls.js
ConradJChan: They're the same string, but they represent different things, tag name vs button type -- I fee like a constant here would be convenient but possibly misleading. Thoughts?

src/lib/ZoomControls.js
ConradJChan: `maxZoom` range is up to the consuming code, the scale is 0-1 though the maxZoom could be a value upwards of 1 -- I'll try to reflect this in the doc

src/lib/ZoomControls.js
ConradJChan: Thanks for catching that, I'll use `round` to be consistent

src/lib/Controls.js
```diff
@@ -183,40 +183,46 @@ class Controls {
...
+            element.className = `${CONTROLS_BUTTON_CLASS} ${classList}`;
+            element.addEventListener('click', handler);
+
+            // Maintain a reference for cleanup
+            this.buttonRefs.push({
```
jstoffan: Seems like we'd want to do this if `handler` is defined, regardless of the tag type.

src/lib/icons/zoom_in.svg
```diff
@@ -0,0 +1,6 @@
+<svg xmlns="http://www.w3.org/2000/svg" width="10" height="10" viewBox="0 0 10 10">
+    <g fill="none" fill-rule="evenodd">
```
jstoffan: Is this wrapping tag needed?

src/lib/viewers/doc/DocBaseViewer.js
```diff
@@ -167,6 +172,11 @@ class DocBaseViewer extends BaseViewer {
             this.pageControls.removeListener('pagechange', this.setPage);
         }
 
+        if (this.zoomControls) {
+            this.zoomControls.removeListener(ZOOM_CONTROLS_EVENTS.zoomin, this.zoomIn);
```
jstoffan: It seems like logic should live in `destroy` method on the ZoomControls class, which should then be called from the `this.controls.destroy()` method.

src/lib/viewers/doc/DocBaseViewer.js
```diff
@@ -536,6 +546,7 @@ class DocBaseViewer extends BaseViewer {
             });
         }
         this.pdfViewer.currentScaleValue = newScale;
+        this.zoomControls.setCurrentScale(newScale);
```
jstoffan: Can we move these calls to a shared zoom event listener within this class?

src/lib/viewers/doc/DocBaseViewer.js
```diff
@@ -523,7 +533,7 @@ class DocBaseViewer extends BaseViewer {
         let numTicks = ticks;
         let newScale = this.pdfViewer.currentScale;
         do {
-            newScale = (newScale * DEFAULT_SCALE_DELTA).toFixed(3);
+            newScale += DEFAULT_SCALE_DELTA;
```
jstoffan: It looks like we may end up losing some rounding here, since we're no longer using `toFixed(3)`. What was the reason behind this change?

src/lib/viewers/doc/DocBaseViewer.js
```diff
@@ -1006,7 +1018,10 @@ class DocBaseViewer extends BaseViewer {
     loadUI() {
         this.controls = new Controls(this.containerEl);
         this.pageControls = new PageControls(this.controls, this.docEl);
+        this.zoomControls = new ZoomControls(this.controls);
         this.pageControls.addListener('pagechange', this.setPage);
+        this.zoomControls.addListener(ZOOM_CONTROLS_EVENTS.zoomin, this.zoomIn);
```
jstoffan: Do these need to be event-based or can we pass in `this.zoomIn` as a callback option to the ZoomControls constructor? Is it the former for consistency with PageControls? Do we want that to be the standard?

src/lib/viewers/doc/DocBaseViewer.js
```diff
@@ -1065,8 +1080,12 @@ class DocBaseViewer extends BaseViewer {
             );
         }
 
-        this.controls.add(__('zoom_out'), this.zoomOut, 'bp-doc-zoom-out-icon', ICON_ZOOM_OUT);
-        this.controls.add(__('zoom_in'), this.zoomIn, 'bp-doc-zoom-in-icon', ICON_ZOOM_IN);
+        this.zoomControls.add(this.pdfViewer.currentScale, {
```
jstoffan: I think the transparency of this API could be better. It's not clear to me what the result is here without going in and looking at what `add` is doing.

src/lib/viewers/image/MultiImageViewer.js
```diff
@@ -4,6 +4,8 @@ import './MultiImage.scss';
 import { ICON_FULLSCREEN_IN, ICON_FULLSCREEN_OUT } from '../../icons/icons';
 import { CLASS_INVISIBLE, CLASS_MULTI_IMAGE_PAGE, CLASS_IS_SCROLLABLE } from '../../constants';
 import { pageNumberFromScroll } from '../../util';
+import ZoomControls from '../../ZoomControls';
```
jstoffan: Should this live up under `import PageControls...`?

src/lib/viewers/text/TextBaseViewer.js
```diff
@@ -55,17 +62,23 @@ class TextBaseViewer extends BaseViewer {
...
         if (inOrOut === 'in') {
-            newFontSize = `${size + 10}%`;
+            newFontSize = size + 10;
         } else if (inOrOut === 'out') {
-            newFontSize = `${size - 10}%`;
+            newFontSize = size - 10;
         }
 
-        el.style.fontSize = newFontSize;
+        el.style.fontSize = `${newFontSize}%`;
```
jstoffan: Can this lead to a value of `0%` due to the default value of `newFontSize`? I would probably revert this change, since the duplication of the template isn't super impactful.

src/lib/viewers/text/TextBaseViewer.js
```diff
@@ -55,17 +62,23 @@ class TextBaseViewer extends BaseViewer {
...
             zoom: newFontSize,
             canZoomIn: true,
             canZoomOut: true,
         });
+        this.zoomControls.setCurrentScale(newFontSize / 100);
+    }
+
+    getFontSize() {
+        const el = this.containerEl.querySelector('.bp-text');
+        return parseInt(el.style.fontSize, 10) || 100;
```
jstoffan: Should we be using a class property rather than the DOM to track the current state?

src/lib/viewers/text/__tests__/TextBaseViewer-test.js
```diff
@@ -4,6 +4,7 @@ import TextBaseViewer from '../TextBaseViewer';
 import BaseViewer from '../../BaseViewer';
 import * as file from '../../../file';
 import { PERMISSION_DOWNLOAD } from '../../../constants';
+import ZoomControls from '../../../ZoomControls';
```
jstoffan: Alphabetize imports?

src/lib/Controls.js
jstoffan: I agree with @ConradJChan. Further, we should only advocate for constants where they add clarity rather than as an ironclad requirement.

src/lib/icons/zoom_in.svg
ConradJChan: Yeah looks like a little black box appears without it

src/lib/viewers/text/TextBaseViewer.js
ConradJChan: Yeah `0%` is possible, but that is no different from the behavior it was before. This change was only to keep `newFontSize` as a number which could be passed into the `zoomControls.setCurrentScale` call on L76

src/lib/viewers/text/TextBaseViewer.js
ConradJChan: I agree that the font size/scale could be better represented than the DOM computedStyles. I'll address that in an follow up PR. This was just a convenience method based on what was already happening in L53-43, which I forgot to replace with this 😛

src/lib/viewers/doc/DocBaseViewer.js
ConradJChan: I'll move them to be callback options which should help with the cleanup of the listeners as well

src/lib/viewers/image/MultiImageViewer.js
```diff
@@ -255,6 +260,10 @@ class MultiImageViewer extends ImageBaseViewer {
      */
     loadUI() {
         super.loadUI();
+
+        this.zoomControls = new ZoomControls(this.controls);
```
jstoffan: Looks like this is already handled by the base class. Should the `init` call move to the base class, as well?

PR #1096, opened 2019-11-07
refactor(controls): Adjusting the size of the grouped controls

ConradJChan:
Also introduces a controls group div to group the zoom controls and page controls together

Before:
Screen Shot 2019-11-06 at 1 35 46 PM

After:
Screen Shot 2019-11-07 at 10 17 39 AM


boxcla:

src/lib/Controls.js
```diff
@@ -227,6 +230,22 @@ class Controls {
...
+    /**
+     * Add div for a group of controls
+     *
+     * @public
+     * @param {string} classNames - optional class names
+     * @return {HTMLElement} The created HTMLElement for a group of controls
+     */
+    addGroup(classNames = '') {
+        const group = document.createElement('div');
+        group.className = `bp-controls-group ${classNames}`;
```
mickr: Minor: Use `classNames` util so that in case classNames aren't provided there won't be an empty space in the DOM

ConradJChan: We don't currently depend on that in preview, but we could add that and address it elsewhere in a follow up PR

mickr: It is super minor and if we don't currently have it, no big deal 👍

src/lib/Controls.js
```diff
@@ -214,7 +217,7 @@ class Controls {
         }
 
         cell.appendChild(element);
-        this.controlsEl.appendChild(cell);
+        parentElement.appendChild(cell);
```
jstoffan: Can we repurpose `add` to just create the element and then defer the `appendChild` part to the calling method? It looks like we're already returning `element` as our result.

src/lib/Controls.scss
```diff
@@ -10,8 +10,7 @@
 .bp-controls {
     position: relative;
     left: -50%;
-    display: table;
-    table-layout: fixed;
+    display: flex;
```
jstoffan: Definitely want to double-check the new layout in IE/Edge/Safari.

src/lib/constants.js
```diff
@@ -2,6 +2,7 @@ export const CLASS_ACTIVE = 'bp-is-active';
 export const CLASS_NAVIGATION_VISIBILITY = 'bp-is-navigation-visible';
 export const CLASS_HIDDEN = 'bp-is-hidden';
 export const CLASS_PREVIEW_LOADED = 'bp-loaded';
+export const CLASS_BOX_CONTROLS_GROUP_BUTTON = 'bp-controls-group-btn';
```
jstoffan: Should this be encapsulated in the Controls class, instead?

src/lib/Controls.js
ConradJChan: It would probably end up looking something like the following in each `add` call:
 ```
const control = controls.add(...);
controls.controlsEl.appendChild(control);
```

It requires the calling code to know where to put it, but really only the `Controls` instance knows that. Thoughts?

jstoffan: Yeah, we probably should avoid optimizing it too much considering we may want to rewrite these components eventually. Go ahead and disregard my initial comment.

PR #1097, opened 2019-11-08
fix(doc zoom): Move zoom control update to pagerender handler

ConradJChan:
When PDFjs updates the page scaling, it emits the `pagerendered` event, so this PR moves the logic to update the zoom control with the current scale value based off of that event. This covers the issue where toggling fullscreen wouldn't update the current scale value in the zoom control.

### DIFF
--- a/src/lib/viewers/doc/DocBaseViewer.js
+++ b/src/lib/viewers/doc/DocBaseViewer.js
@@ -564,7 +564,6 @@ class DocBaseViewer extends BaseViewer {
             });
         }
         this.pdfViewer.currentScaleValue = newScale;
-        this.zoomControls.setCurrentScale(newScale);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


boxcla:

src/lib/viewers/doc/DocBaseViewer.js
```diff
@@ -1176,6 +1175,8 @@ class DocBaseViewer extends BaseViewer {
             return;
         }
 
+        this.zoomControls.setCurrentScale(this.pdfViewer.currentScale);
```
jstoffan: Should this go under the `if (!this.somePageRendered)` condition below?

ConradJChan: I don't think so because this isn't an initial render thing, it's everytime the pdf viewer re-renders which can result from initial render, zoom in/out, toggle fullscreen, etc...

PR #1099, opened 2019-11-15
chore(controls): add e2e tests for toggle findbar and zoom

ConradJChan:


boxcla:

src/lib/ZoomControls.js
```diff
@@ -94,7 +94,7 @@ class ZoomControls {
             groupElement,
         );
 
-        this.currentScaleElement = this.controlsElement.querySelector(`.${CLASS_ZOOM_CURRENT_SCALE}`);
+        this.currentScaleElement = this.controlsElement.querySelector(`.${CLASS_ZOOM_CURRENT_SCALE} > span`);
```
jstoffan: Is this meant to address a bug? Can we select on a class, instead?

ConradJChan: Yeah there is a bug where the `curentScaleElement` was pointing at the containing div, rather than the span that was initially created as part of the html template for the button. I'll update this to be a class

PR #1101, opened 2019-11-19
fix(viewer): Find controls in PPT files update current page

jstoffan:
It seems that PresentationViewer was written before the PDFSinglePageViewer was available. The original implementation included several hacks to override core PDF.js functionality around scroll, which broke with our upgrade in August. This change should resolve that bug and simplifies the viewer's logic.

### DIFF
--- a/src/lib/viewers/doc/PresentationViewer.js
+++ b/src/lib/viewers/doc/PresentationViewer.js
@@ -1,7 +1,6 @@
 import throttle from 'lodash/throttle';
 import DocBaseViewer from './DocBaseViewer';
 import PresentationPreloader from './PresentationPreloader';
-import { CLASS_INVISIBLE } from '../../constants';
 import './Presentation.scss';
 
 const WHEEL_THROTTLE = 200;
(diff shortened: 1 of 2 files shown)


boxcla:

src/lib/viewers/doc/PresentationViewer.js
```diff
@@ -129,16 +100,14 @@ class PresentationViewer extends DocBaseViewer {
...
-     * @return {void}
      * @protected
+     * @override
+     * @return {pdfjsViewer.PDFViewer} PDF viewer type
      */
-    initViewer(pdfUrl) {
-        super.initViewer(pdfUrl);
-        this.overwritePdfViewerBehavior();
+    initPdfViewer() {
+        return this.initPdfViewerClass(this.pdfjsViewer.PDFSinglePageViewer);
```
mxiao6: Then what's the difference between Presentation Viewer and Single Page Viewer?

jstoffan: There's some additional logic in PresentationViewer around scrolling. The code is available for comparison.

PR #1104, opened 2019-11-20
fix(find): Find bar container should only be created when needed

jstoffan:


boxcla:

PR #1105, opened 2019-11-21
fix(viewer): Set default size on SVG images if parsing fails in IE11

jstoffan:
The `parseFromString` method can throw errors in IE11 if the source SVG XML is malformed. The resulting image has no `originalSize` attribute, allowing it to expand to the size of the viewport and interfering with the new zoom behavior.


boxcla:

PR #1107, opened 2019-11-23
fix(zoom): Add zoom min/max limits for image and text files

jstoffan:


boxcla:

src/lib/viewers/text/TextBaseViewer.js
```diff
@@ -47,36 +53,31 @@ class TextBaseViewer extends BaseViewer {
     /**
      * Zooms by increasing or decreasing font size
      * @public
-     * @param {string} inOrOut - in or out
+     * @param {string} type - in, out, reset
```
mickr: Is reset currently a used type?

jstoffan: I don't believe so, but it's supported implicitly in the existing code. This change just brings the function docs more in line with the other viewers.

PR #1108, opened 2019-11-23
fix(find): Allow find input to receive focus upon find icon click

jstoffan:


boxcla:

PR #1110, opened 2019-11-26
feat(viewer): Add loading icons for raw image file extensions

jstoffan:
I noticed that we're still missing some file-specific loading icons for certain extensions.


boxcla:

PR #1112, opened 2019-12-05
feat(archive-viewer): add archive loader, viewer, and archive explorer

mxiao6:
Current demo video: https://cloud.box.com/s/93t1mxinwx3apngl2ahx16g1pfwfjpbq


boxcla:

sarahsab:
Ses quoi sa ??

Le mer. 4 déc. 2019 8 h 25 p.m., Box Bot <notifications@github.com> a
écrit :

package.json
```diff
@@ -28,6 +28,7 @@
         "babel-plugin-transform-require-ignore": "^0.1.1",
         "box-annotations": "^2.3.0",
         "box-react-ui": "^22.7.0",
+        "box-ui-elements": "^10.1.0",
```
jstoffan: Should we use 11.0.2?

package.json
```diff
@@ -28,6 +28,7 @@
         "babel-plugin-transform-require-ignore": "^0.1.1",
         "box-annotations": "^2.3.0",
         "box-react-ui": "^22.7.0",
```
jstoffan: Can we replace this now that box-ui-elements is a dependency? I recall we are only importing a single utility function.

src/lib/viewers/archive/Archive.scss
```diff
@@ -0,0 +1,6 @@
+.bp-archive {
+    display: flex;
+    width: 100%;
+    height: 100%;
+    background: white;
```
jstoffan: I think we usually want to use hex values for colors.

src/lib/viewers/archive/ArchiveExplorer.js
```diff
@@ -0,0 +1,148 @@
...
+    dateCellRenderer,
+    sizeCellRenderer,
+    fileNameCellRenderer,
+} from 'box-ui-elements/es/features/virtualized-table-renderers';
+import VirtualizedTable from 'box-ui-elements/es/features/virtualized-table';
+import { Column } from 'react-virtualized/dist/es/Table/index';
+
+class ArchiveExplorer extends React.Component {
+    static propTypes = {
+        data: PropTypes.arrayOf(
```
jstoffan: Can we name this top-level collection in a more meaningful way?

src/lib/viewers/archive/ArchiveExplorer.js
```diff
@@ -0,0 +1,148 @@
...
+            PropTypes.shape({
+                type: PropTypes.string,
```
jstoffan: Are all of these properties optional? If not, should some be marked with `isRequired`?

src/lib/viewers/archive/ArchiveExplorer.js
```diff
@@ -0,0 +1,148 @@
...
+    };
+
+    constructor(props) {
+        super(props);
+        const { cache, root } = this.buildCache(props.data);
+        this.cache = cache;
+        this.root = root;
+
+        this.state = {
+            currentData: this.getCurrentData(cache, root),
```
jstoffan: This naming is kind of duplicative, since `this.state` is always the "current" data for the component. What does this data represent? A collection of items?

src/lib/viewers/archive/ArchiveExplorer.js
```diff
@@ -0,0 +1,148 @@
...
+
+    buildCache = data => {
+        const cache = {};
+        let root;
+
+        data.forEach(info => {
+            if (!info.parent) {
+                root = info.absolute_path;
+            }
+            cache[info.absolute_path] = info;
```
jstoffan: This is derived state and is effectively duplicating the contents of `this.state.currentData` into an instance property. Can we calculate these values in render, instead?

src/lib/viewers/archive/ArchiveExplorer.js
```diff
@@ -0,0 +1,148 @@
...
+        const rowData = {
+            ...currentRow,
+            name: {
+                id: undefined,
+                name,
+                isExternal: false,
+                type,
+                onClick: type === 'folder' ? () => this.updateData(absolutePath) : undefined,
+            },
+            date: `20${modifiedAt}`,
```
jstoffan: This doesn't seem ideal. Are we expecting full date objects in the final data structure?

src/lib/viewers/archive/ArchiveExplorer.js
```diff
@@ -0,0 +1,148 @@
...
+
+    getRowData = ({ index }) => {
+        const { currentData } = this.state;
+        const currentRow = currentData[index];
+        const { name, type, modified_at: modifiedAt, absolute_path: absolutePath } = currentRow;
+
+        const rowData = {
+            ...currentRow,
+            name: {
+                id: undefined,
```
jstoffan: The data structure here seems very odd. We have `name.id` and `name.name` in the row data. Is that correct?

src/lib/viewers/archive/ArchiveExplorer.js
```diff
@@ -0,0 +1,148 @@
...
+
+    updateData = absolutePath => {
+        this.setState({
+            currentData: this.getCurrentData(this.cache, absolutePath),
+        });
+    };
+
+    getRowData = ({ index }) => {
+        const { currentData } = this.state;
+        const currentRow = currentData[index];
```
jstoffan: We can use `...rest` on the line below to avoid this temporary variable.

src/lib/viewers/archive/ArchiveExplorer.js
```diff
@@ -0,0 +1,148 @@
...
+        const { name, type, modified_at: modifiedAt, absolute_path: absolutePath } = currentRow;
+
+        const rowData = {
+            ...currentRow,
+            name: {
+                id: undefined,
+                name,
+                isExternal: false,
```
jstoffan: Let's alphabetize all properties in both data objects and types, if possible.

src/lib/viewers/archive/ArchiveExplorer.js
```diff
@@ -0,0 +1,148 @@
...
+                            label="Last modified date"
+                            width={1}
+                        />,
+                        <Column
+                            key="Size"
+                            cellRenderer={sizeCellRenderer()}
+                            dataKey="size"
+                            disableSort
+                            flexGrow={1}
+                            label="Size"
```
jstoffan: These will need to be localized.

src/lib/viewers/archive/ArchiveLoader.js
```diff
@@ -0,0 +1,42 @@
...
+    {
+        NAME: 'Archive',
+        CONSTRUCTOR: ArchiveViewer,
+        REP: 'json',
+        EXT: ['zip'],
+    },
+    {
+        NAME: 'Archive',
+        CONSTRUCTOR: ArchiveViewer,
+        REP: ORIGINAL_REP_NAME,
```
jstoffan: Is this case possible? Is the ArchiveViewer equipped to handle it?

src/lib/viewers/archive/ArchiveLoader.js
```diff
@@ -0,0 +1,42 @@
+import AssetLoader from '../AssetLoader';
+import ArchiveViewer from './ArchiveViewer';
+import { ORIGINAL_REP_NAME } from '../../constants';
+
+// Order of the viewers matters. Prefer original before others. Go from specific to general.
```
jstoffan: What does `prefer original before others` mean in this case?

src/lib/viewers/archive/ArchiveLoader.js
```diff
@@ -0,0 +1,42 @@
...
+     */
+    constructor() {
+        super();
+        this.viewers = VIEWERS;
+    }
+
+    /**
+     * @inheritdoc
+     */
+    determineViewer(file, disabledViewers = []) {
```
jstoffan: Is this override needed?

src/lib/viewers/archive/ArchiveExplorer.js
```diff
@@ -0,0 +1,148 @@
...
+            cache[info.absolute_path] = info;
+        });
+
+        return { cache, root };
+    };
+
+    getCurrentData = (cache, currentFolder) => {
+        const currentFolderInfo = cache[currentFolder];
+
+        const currentData = currentFolderInfo.item_collection.entries.map(item => cache[item.absolute_path]);
```
jstoffan: Please avoid temporary variables that are immediately returned.

src/lib/viewers/archive/ArchiveViewer.js
```diff
@@ -0,0 +1,348 @@
...
+     */
+    postLoad = () => {
+        const { representation } = this.options;
+        const template = representation.content.url_template;
+        const contentUrl = this.createContentUrlWithAuthParams(template);
+        this.startLoadTimer();
+
+        this.api
+            .get(contentUrl)
+            .then(data => {
```
jstoffan: This is a nested promise chain and can be flattened into the original chain. The second `.catch` isn't necessary.

src/lib/viewers/archive/ArchiveViewer.js
```diff
@@ -0,0 +1,348 @@
...
+    };
+
+    /**
+     * Finishes loading the archive data
+     *
+     * @private
+     * @return {void}
+     */
+    finishLoading() {
+        /* global BoxArchive */
```
jstoffan: It's not really clear why this needs to be a global. Can you add a comment with an explanation?

src/lib/viewers/archive/ArchiveViewer.js
```diff
@@ -0,0 +1,348 @@
...
+        this.archiveComponent = new BoxArchive(this.archiveEl, this.data);
+        this.archiveComponent.renderArchive();
```
jstoffan: Is this separate call needed or can we just use the constructor?

src/lib/viewers/archive/Archive.scss
```diff
@@ -0,0 +1,6 @@
+.bp-archive {
```
jstoffan: It seems like overkill to make a separate HTTP call for this small amount of CSS. I wonder if we can just inline it in the React component?

src/lib/viewers/archive/ArchiveExplorer.js
mxiao6: This is only for now, because it's not standard date format and JS cannot parse it correctly. I talked to conversion team, and they said they would fix it.

src/lib/viewers/archive/Archive.scss
mxiao6: > It seems like overkill to make a separate HTTP call for this small amount of CSS. I wonder if we can just inline it in the React component?

The final css file also contains all the css code from the renderers.

package.json
mxiao6: yeah, we can

src/lib/viewers/archive/ArchiveLoader.js
mxiao6: This is only for testing purpose, because currently the file api only gives us ORIGINAL_REP_NAME. Will remove it after the real api is in prod.

mxiao6: > Is this case possible? Is the ArchiveViewer equipped to handle it?

This is only for testing purpose, because currently the file api only gives us ORIGINAL_REP_NAME. Will remove it after the real api is in prod.

src/lib/viewers/archive/ArchiveExplorer.js
mxiao6: Changed the cache-build process. No duplication now.

mxiao6: > This is derived state and is effectively duplicating the contents of `this.state.currentData` into an instance property. Can we calculate these values in render, instead?

Changed the cache-build process. No duplication now.

src/lib/viewers/archive/ArchiveViewer.js
```diff
@@ -0,0 +1,310 @@
...
+     * @private
+     * @return {Promise} promise to get archive content
+     */
+    postLoad = () => {
+        const { representation } = this.options;
+        const template = representation.content.url_template;
+        const contentUrl = this.createContentUrlWithAuthParams(template);
+        this.startLoadTimer();
+
+        this.api.get(contentUrl).then(data => {
```
jstoffan: This promise isn't being returned, which means errors within it won't be caught by the parent `.catch`. This promise chain also still needs to be flattened. Let's move the `this.api.get` call and the logic in its anonymous `.then` clause each to separate methods and call them via two separate `.then` clauses in the main Promise.all chain.

src/lib/viewers/archive/ArchiveViewer.js
```diff
@@ -0,0 +1,310 @@
...
+            if (this.isDestroyed()) {
```
jstoffan: It seems like all of this logic should just live in `this.finishLoading`.

src/lib/viewers/archive/ArchiveExplorer.js
```diff
@@ -0,0 +1,161 @@
...
+        });
+    };
+
+    render() {
+        const { itemCollection } = this.props;
+        const { fullPath } = this.state;
+        const { cache, root } = this.buildCache(itemCollection);
+
+        if (!fullPath) {
+            this.setState({
```
jstoffan: We should always avoid calling `this.setState` within render to avoid infinite render loops. The default value should be set in the constructor.

src/lib/viewers/archive/ArchiveExplorer.js
```diff
@@ -0,0 +1,161 @@
...
+                            absolute_path: PropTypes.string.isRequired,
+                            name: PropTypes.string,
+                        }),
+                    ).isRequired,
+                }).isRequired,
+            }),
+        ).isRequired,
+    };
+
+    buildCache = memoize(itemList => {
```
jstoffan: Pass in `props` directly so this can be an instance method. There is also naming disparity between `itemList` as a parameter and `itemCollection` in render. Let's keep our naming consistent.

src/lib/viewers/archive/ArchiveExplorer.js
```diff
@@ -0,0 +1,161 @@
...
+        const nextFullPath = `${fullPath}${name}/`;
+        this.setState({
+            fullPath: nextFullPath,
+        });
+    };
+
+    render() {
+        const { itemCollection } = this.props;
+        const { fullPath } = this.state;
+        const { cache, root } = this.buildCache(itemCollection);
```
jstoffan: Do we need to build the entire cache or can we find the correct `absolute_path` and return the related item list immediately? We should also separate the logic for finding the current item list and the root path into separate methods.

src/lib/viewers/archive/ArchiveExplorer.js
```diff
@@ -0,0 +1,161 @@
...
+                            key="Last modified date"
+                            cellRenderer={readableTimeCellRenderer}
+                            dataKey={KEY_MODIFIED_AT}
+                            disableSort
+                            flexGrow={2}
+                            label={__('last_modified_date')}
+                            width={1}
+                        />,
+                        <Column
+                            key="Size"
```
jstoffan: Should these keys be lowercase/hyphenated? Should they use the column key strings to avoid duplication?

src/lib/viewers/archive/ArchiveExplorer.js
```diff
@@ -0,0 +1,161 @@
...
+                            cellRenderer={sizeCellRenderer()}
+                            dataKey="size"
```
jstoffan: Should this also be a constant?

src/lib/viewers/archive/ArchiveExplorer.js
```diff
@@ -0,0 +1,161 @@
...
+import {
+    readableTimeCellRenderer,
+    sizeCellRenderer,
+    itemNameCellRenderer,
+} from 'box-ui-elements/es/features/virtualized-table-renderers';
+import VirtualizedTable from 'box-ui-elements/es/features/virtualized-table';
+import { Column } from 'react-virtualized/dist/es/Table/index';
+import { TABLE_COLUMNS } from './constants';
+
+const { KEY_NAME, KEY_MODIFIED_AT } = TABLE_COLUMNS;
```
jstoffan: I don't think destructuring is necessary here.

src/lib/viewers/archive/ArchiveExplorer.js
```diff
@@ -0,0 +1,161 @@
...
+        });
+
+        return { cache, root };
+    });
+
+    constructor(props) {
+        super(props);
+
+        this.state = {
+            fullPath: undefined,
```
mickr: Should the initial state of full path be undefined?

src/lib/viewers/archive/ArchiveExplorer.js
```diff
@@ -0,0 +1,161 @@
...
+    getRowData = itemList => ({ index }) => {
+        const { name, type, modified_at: modifiedAt, ...rest } = itemList[index];
+
+        const rowData = {
+            [KEY_NAME]: {
+                isExternal: false,
+                name,
+                type,
+            },
+            [KEY_MODIFIED_AT]: `20${modifiedAt}`,
```
mickr: What does the 20 signify in this statement?

src/lib/viewers/archive/ArchiveViewer.js
```diff
@@ -0,0 +1,310 @@
+/* eslint-disable no-unused-vars */
+import './Archive.scss';
+import BaseViewer from '../BaseViewer';
+import { VIEWER_EVENT } from '../../events';
+
+const JS = ['archive.js'];
+const CSS = ['archive.css'];
+
+const testData = [
```
mickr: Make sure this doesn't get commited

mxiao6: I put it there in case you guys want to see how it looks like. Will remove it later.

src/lib/viewers/archive/ArchiveExplorer.js
mxiao6: It's just a temporary way to fix JS date parsing confusion (make '19' to '2019'). Conversion will fix it later.

src/lib/viewers/archive/ArchiveExplorer.js
mxiao6: why? isn't `TABLE_COLUMNS. KEY_NAME` too long?

mxiao6: > I don't think destructuring is necessary here.

why? isn't `TABLE_COLUMNS. KEY_NAME` too long?

src/lib/viewers/archive/ArchiveExplorer.js
```diff
@@ -0,0 +1,169 @@
...
+
+    /**
+     * Handle click event, update fullPath state
+     *
+     * @param {Object} cellValue - the cell being clicked
+     */
+    handleClick = cellValue => {
+        const { name } = cellValue;
+        const { fullPath } = this.state;
+        const nextFullPath = `${fullPath}${name}/`;
```
jstoffan: Try to avoid variables that are created and then used immediately only once.

src/lib/viewers/archive/ArchiveExplorer.js
```diff
@@ -0,0 +1,169 @@
...
+        return rowData;
+    };
+
+    /**
+     * Handle click event, update fullPath state
+     *
+     * @param {Object} cellValue - the cell being clicked
+     */
+    handleClick = cellValue => {
+        const { name } = cellValue;
```
jstoffan: Minor, but this can be moved to the function signature.

src/lib/viewers/archive/ArchiveExplorer.js
```diff
@@ -0,0 +1,169 @@
...
+     * Prepare data to render
+     * Will be passed as row getter into VirtaulizedTable
+     *
+     * @param {Array<Object>} itemList - list of data object
+     * @param { Number } index - row index of the data to selected
+     */
+    getRowData = itemList => ({ index }) => {
+        const { modified_at: modifiedAt, name, size, type, ...rest } = itemList[index];
+
+        const rowData = {
```
jstoffan: This can be returned immediately rather than declared as a variable.

build/webpack.karma.config.js
```diff
@@ -1,17 +1,13 @@
...
-});
-
 const { IgnorePlugin } = require('webpack');
 const commonConfig = require('./webpack.common.config');
 
 const baseConfig = commonConfig('en-US');
 
 const config = {
     ...baseConfig,
+    devtool: 'inline-source-map',
```
jstoffan: Is this needed as part of this change? Does it affect the production bundle?

mxiao6: > Is this needed as part of this change? Does it affect the production bundle?

The next line specifies the mode to be 'development', so it does not affect the production bundle. This only makes the testing slower but it solves enzyme preprocess problem.

package.json
```diff
@@ -89,6 +91,8 @@
         "raw-loader": "^3.1.0",
         "react": "^16.7.0",
         "react-dom": "^16.7.0",
+        "react-intl": "^2.9.0",
+        "react-tether": "^1.0.0",
```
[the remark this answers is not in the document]
mxiao6: > Where is react-tether used in the virtualized table?
If I don't add it, the building process gives me error about dropdown-menu. I guess it's because one of the renderers uses breadcrumb and breadcrumb imports dropdown-menu.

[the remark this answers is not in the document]
mxiao6: > Where is react-tether used in the virtualized table?

If I don't add it, the building process gives me error about dropdown-menu. I guess it's because one of the renderers uses breadcrumb and breadcrumb imports dropdown-menu.

src/lib/ThumbnailsSidebar.js
```diff
@@ -311,7 +311,7 @@ class ThumbnailsSidebar {
      * @return {Promise} - promise reolves with the image HTMLElement or null if generation is in progress
      */
     createThumbnailImage(itemIndex) {
-        const cacheEntry = this.thumbnailImageCache.get(itemIndex);
+        const cacheEntry = this.thumbnailImageCache && this.thumbnailImageCache.get(itemIndex);
```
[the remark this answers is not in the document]
mxiao6: > This change seems unrelated to the archive viewer. Is it meant to address a separate bug?

This change is to fix a sanity test. The `thumbnailImageCache` is sometimes not initialized in time because the travis testing machine is too slow. Should I open another pr for this?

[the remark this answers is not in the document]
jstoffan: Should we revert this change since we have the other PR open?

PR #1114, opened 2019-12-06
fix(viewer): Annotations render correctly for presentation files

jstoffan:
This change reverts #1101 due to issues with box-annotations, which relies on every page being rendered in the DOM, even if it's not yet visible. Find controls still work properly with this change. I renamed PresentationViewer's `pagechangeHandler` to `pagechangingHandler` so the base method is overridden properly, which was the original bug that #1101 was meant to address.


boxcla:

PR #1116, opened 2019-12-13
chore(build): Add dry run option

mickr:
Adding dry_run option to release script for easier testing.

### DIFF
--- a/build/release.sh
+++ b/build/release.sh
@@ -6,6 +6,19 @@ export NODE_PATH=$NODE_PATH:./node_modules
 OLD_VERSION="XXX"
 VERSION="XXX"
 
+dry_run=false
+
+# Wraps mutating commands with echo if dry_run is enabled.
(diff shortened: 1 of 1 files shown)


boxcla:

build/release.sh
```diff
@@ -6,6 +6,19 @@ export NODE_PATH=$NODE_PATH:./node_modules
 OLD_VERSION="XXX"
 VERSION="XXX"
 
+dry_run=false
```
jstoffan: It looks like our current variables follow SNAKE_CAPS convention.

jstoffan:
I think the scope should probably be `chore` rather than `feat` for this commit, since the change doesn't affect the end user.

build/release.sh
mickr: Most of the variables do, our switches do not. Since this was a command line switch I made it look like the others

PR #1117, opened 2019-12-13
fix(test): Fix possible sanity test bug

mxiao6:
This change is to fix a sanity test. The function `createThumbnailImage` is used as a callback. Maybe the object has been destroyed before the callback happens. So we need to check whether `this.thumbnailImageCache` still exists.

### DIFF
--- a/src/lib/ThumbnailsSidebar.js
+++ b/src/lib/ThumbnailsSidebar.js
@@ -39,6 +39,9 @@ class ThumbnailsSidebar {
     /** @property {Object} - Cache for the thumbnail image elements */
     thumbnailImageCache;
 
+    /** @property {Array<number>} - The ID values returned by the call to window.requestAnimationFrame() */
+    animationFrameRequestIds;
+
     /**
      * [constructor]
      *
@@ -46,6 +49,7 @@ class ThumbnailsSidebar {
      * @param {PDFViewer} pdfViewer - the PDFJS viewer
      */
     constructor(element, pdfViewer) {
+        this.animationFrameRequestIds = [];
         this.anchorEl = element;
         this.currentThumbnails = [];
         this.pdfViewer = pdfViewer;
@@ -132,6 +136,10 @@ class ThumbnailsSidebar {
      * @return {void}
      */
     destroy() {
+        if (this.animationFrameRequestIds.length > 0) {
+            this.animationFrameRequestIds.forEach(id => cancelAnimationFrame(id));
+        }
+
         if (this.virtualScroller) {
             this.virtualScroller.destroy();
             this.virtualScroller = null;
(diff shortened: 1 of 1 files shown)


boxcla:

src/lib/ThumbnailsSidebar.js
```diff
@@ -311,7 +311,7 @@ class ThumbnailsSidebar {
      * @return {Promise} - promise reolves with the image HTMLElement or null if generation is in progress
      */
     createThumbnailImage(itemIndex) {
-        const cacheEntry = this.thumbnailImageCache.get(itemIndex);
+        const cacheEntry = this.thumbnailImageCache && this.thumbnailImageCache.get(itemIndex);
```
mickr: I do wonder why this class isn't getting instantiated properly to kick off the constructor. There is a bounded cache on `L52`

mxiao6: > I do wonder why this class isn't getting instantiated properly to kick off the constructor. There is a bounded cache on `L52`

The function is used as a callback. Maybe the object has been destroyed before the callback happens.

src/lib/ThumbnailsSidebar.js
```diff
@@ -311,6 +311,10 @@ class ThumbnailsSidebar {
      * @return {Promise} - promise reolves with the image HTMLElement or null if generation is in progress
      */
     createThumbnailImage(itemIndex) {
```
jstoffan: It looks like this method is called as part of a `requestAnimationFrame` callback. Should we track the RAF request and cancel it in `destroy`? That's the only place the cache is destroyed, looks like.

src/lib/ThumbnailsSidebar.js
```diff
@@ -311,6 +311,10 @@ class ThumbnailsSidebar {
...
+        if (!this.thumbnailImageCache) {
+            return Promise.resolve(null);
```
mickr: Should you create the cache or call to create the cache if it hasn't been created at this point?

mxiao6: If it's destroyed already, we shouldn't create the cache again, right

mxiao6: > Should you create the cache or call to create the cache if it hasn't been created at this point?

If it's destroyed already, we shouldn't create the cache again, right

mickr: Why are we creating a thumbnail image AFTER the cache has been destroyed?

src/lib/ThumbnailsSidebar.js
```diff
@@ -289,19 +297,21 @@ class ThumbnailsSidebar {
...
-                    thumbnailEl.classList.add(CLASS_BOX_PREVIEW_THUMBNAIL_IMAGE_LOADED);
-                }
-
-                // After generating the thumbnail image, render the next one
-                this.renderNextThumbnailImage();
-            });
-        });
+        this.animationFrameRequestIds.push(
+            requestAnimationFrame(() => {
+                this.createThumbnailImage(itemIndex).then(imageEl => {
```
jstoffan: We'll need to remove the requestId from `animationFrameRequestIds` before this line as well to avoid an ever-growing array of already-resolved ids.

PR #1121, opened 2019-12-17
feat(archive-viewer): Add search

mxiao6:
Demo video: https://cloud.box.com/s/t3bs3hpzim1cjpp1hlpqwg4qr6fi5ou4

### DIFF
--- a/src/lib/viewers/archive/ArchiveExplorer.js
+++ b/src/lib/viewers/archive/ArchiveExplorer.js
@@ -1,9 +1,11 @@
 import * as React from 'react';
 import PropTypes from 'prop-types';
 import getProp from 'lodash/get';
+import fuzzysearch from 'fuzzysearch';
 import elementsMessages from 'box-elements-messages'; // eslint-disable-line
 import intlLocaleData from 'react-intl-locale-data'; // eslint-disable-line
 import Internationalize from 'box-ui-elements/es/elements/common/Internationalize';
+import SearchBar from 'box-ui-elements/es/elements/common/header';
 import {
     readableTimeCellRenderer,
     sizeCellRenderer,
@@ -18,6 +20,7 @@ import './ArchiveExplorer.scss';
 
 const language = __LANGUAGE__; // eslint-disable-line
 const { KEY_NAME, KEY_MODIFIED_AT, KEY_SIZE } = TABLE_COLUMNS;
+const { VIEW_FOLDER, VIEW_SEARCH } = VIEWS;
 
 class ArchiveExplorer extends React.Component {
     static propTypes = {
@@ -65,7 +68,8 @@ class ArchiveExplorer extends React.Component {
 
         this.state = {
             fullPath: props.itemCollection.find(info => !info.parent).absolute_path,
-            view: VIEWS.VIEW_FOLDER,
+            searchQuery: '',
+            view: VIEW_FOLDER,
         };
     }
 
@@ -96,10 +100,11 @@ class ArchiveExplorer extends React.Component {
      * @return {Object} formatted data
      */
     getRowData = itemList => ({ index }) => {
-        const { modified_at: modifiedAt, name, size, type, ...rest } = itemList[index];
+        const { absolute_path: fullPath, modified_at: modifiedAt, name, size, type, ...rest } = itemList[index];
 
         return {
             [KEY_NAME]: {
+                fullPath,
                 isExternal: false,
                 name,
                 type,
@@ -112,17 +117,12 @@ class ArchiveExplorer extends React.Component {
     };
 
     /**
-     * Handle click event, update fullPath state
+     * Handle click event, update fullPath state, reset search and view
      *
      * @param {Object} cellValue - the cell being clicked
      * @return {void}
      */
-    handleClick = ({ name }) => {
-        const { fullPath } = this.state;
-        this.setState({
-            fullPath: `${fullPath}${name}/`,
-        });
-    };
+    handleClick = ({ fullPath }) => this.setState({ view: VIEW_FOLDER, fullPath, searchQuery: '' });
 
     /**
      * Handle click event, update fullPath state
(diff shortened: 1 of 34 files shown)


boxcla:

src/lib/viewers/archive/ArchiveExplorer.js
```diff
@@ -1,9 +1,11 @@
 import * as React from 'react';
 import PropTypes from 'prop-types';
 import getProp from 'lodash/get';
+import fuzzysearch from 'fuzzysearch';
```
jstoffan: Can we use our own fuzzy search [utility](https://github.com/box/box-ui-elements/blob/master/src/utils/fuzzySearch.js) rather than adding a third-party dependency?

src/lib/viewers/archive/ArchiveExplorer.js
```diff
@@ -132,19 +132,65 @@ class ArchiveExplorer extends React.Component {
...
+     */
+    search = query => {
+        const trimmedQuery = query.trim();
+
+        if (!query) {
+            this.setState({
+                searchQuery: query,
+                view: VIEW_FOLDER,
+            });
+            return;
```
jstoffan: We should avoid multiple returns as a method of flow control. We should build up the state object and call `this.setState` once at the bottom of the method.

src/lib/viewers/archive/ArchiveExplorer.js
```diff
@@ -132,19 +132,65 @@ class ArchiveExplorer extends React.Component {
...
+
+    /**
+     * Filter item collection for search query
+     *
+     * @param {Array<Object>} itemCollection - raw data
+     * @param {string} searchQuery - user input
+     * @return {Array<Object>} filtered items for search query
+     */
+    getSearchResult = (itemCollection, searchQuery) => {
+        const trimmedQuery = searchQuery.trim();
```
jstoffan: It doesn't seem like this variable is needed, since `.trim()` can be called inline.

src/lib/viewers/archive/ArchiveExplorer.js
```diff
@@ -132,19 +132,65 @@ class ArchiveExplorer extends React.Component {
...
+        const { fullPath, searchQuery, view } = this.state;
+        const itemList =
+            view === VIEW_SEARCH
+                ? this.getSearchResult(itemCollection, searchQuery)
+                : this.getItemList(itemCollection, fullPath);
 
         return (
             <Internationalize language={language} messages={elementsMessages}>
                 <div className="bp-ArchiveExplorer">
+                    <SearchBar isSmall={false} onSearch={this.search} searchQuery={searchQuery} view={view} />
```
jstoffan: We should keep our event handler naming consistent with `this.handleSearch`.

src/lib/viewers/archive/ArchiveExplorer.scss
```diff
@@ -3,4 +3,8 @@
     flex-direction: column;
     width: 100%;
     height: 100%;
+
+    .be-logo {
```
jstoffan: Should this be scoped only to the search bar?

src/lib/viewers/archive/ArchiveExplorer.js
```diff
@@ -1,9 +1,11 @@
 import * as React from 'react';
 import PropTypes from 'prop-types';
 import getProp from 'lodash/get';
+import fuzzysearch from 'fuzzysearch';
 import elementsMessages from 'box-elements-messages'; // eslint-disable-line
 import intlLocaleData from 'react-intl-locale-data'; // eslint-disable-line
 import Internationalize from 'box-ui-elements/es/elements/common/Internationalize';
+import SearchBar from 'box-ui-elements/es/elements/common/header';
```
jstoffan: It looks like the `Header` component is doing a lot of stuff we don't need (isSmall, logo, disablingbased on view prop). I think we should be wary about importing individual pieces from the `/es/elements` directory. Should we create our own simple input, instead?

src/lib/viewers/archive/ArchiveExplorer.js
mxiao6: > It doesn't seem like this variable is needed, since `.trim()` can be called inline.

But the next line is `filter` function, it's like a for loop and the `.trim()` will be called many times.

jstoffan: True. I'd consider that a micro-optimization given the amount of logic being executed within the `fuzzysearch` function, but it's fine as-is.

src/lib/viewers/archive/ArchiveExplorer.js
```diff
@@ -132,19 +132,56 @@ class ArchiveExplorer extends React.Component {
...
+     * @param {string} query - raw query string in the search bar
+     * @return {void}
+     */
+    handleSearch = query => {
+        const trimmedQuery = query.trim();
+        const newState = {
+            searchQuery: query,
+        };
+
+        if (!query) {
```
jstoffan: It seems like this can be simplified to what's below:

```
this.setState({
    searchQuery: query,
    view: query.trim() ? VIEW_SEARCH : VIEW_FOLDER
});
```

src/lib/viewers/archive/ArchiveExplorer.js
```diff
@@ -132,19 +132,56 @@ class ArchiveExplorer extends React.Component {
      */
     handleClickFullPath = fullPath => this.setState({ fullPath });
```
jstoffan: Can we rename this to `handleBreadcrumbClick`?

src/lib/viewers/archive/ArchiveExplorer.js
```diff
@@ -112,17 +117,12 @@ class ArchiveExplorer extends React.Component {
...
      * @param {Object} cellValue - the cell being clicked
      * @return {void}
      */
-    handleClick = ({ name }) => {
-        const { fullPath } = this.state;
-        this.setState({
-            fullPath: `${fullPath}${name}/`,
-        });
-    };
+    handleClick = ({ fullPath }) => this.setState({ view: VIEW_FOLDER, fullPath, searchQuery: '' });
```
jstoffan: Can we rename this to `handleItemClick`?

src/lib/viewers/archive/SearchBar.js
```diff
@@ -0,0 +1,28 @@
...
+
+class SearchBar extends React.PureComponent {
+    static propTypes = {
+        onSearch: PropTypes.func.isRequired,
+        searchQuery: PropTypes.string.isRequired,
+    };
+
+    render() {
+        const { onSearch, searchQuery } = this.props;
+        const search = ({ currentTarget }) => onSearch(currentTarget.value);
```
jstoffan: This can be declared inline with the JSX rather than as a variable.

src/lib/viewers/archive/SearchBar.js
```diff
@@ -0,0 +1,28 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import './SearchBar.scss';
+
+class SearchBar extends React.PureComponent {
```
jstoffan: Does this need to be a class?

src/lib/viewers/archive/SearchBar.js
```diff
@@ -0,0 +1,28 @@
...
+        searchQuery: PropTypes.string.isRequired,
+    };
+
+    render() {
+        const { onSearch, searchQuery } = this.props;
+        const search = ({ currentTarget }) => onSearch(currentTarget.value);
+        return (
+            <div className="bp-SearchBar">
+                <input
+                    aria-label="search"
```
jstoffan: Does this need to be internationalized?

src/lib/viewers/archive/SearchBar.scss
```diff
@@ -0,0 +1,15 @@
...
+
+.bp-SearchBar {
+    display: flex;
+    flex: 0 0 70px;
+    align-items: center;
+    padding: 0 20px;
+    background: $almost-white;
+    border-bottom: 1px solid $bdl-gray-10;
+
+    input[type='search'] {
```
jstoffan: Let's apply a class to the input so we can style it directly and avoid nesting our styles.

src/lib/viewers/archive/ArchiveExplorer.js
mxiao6: > It seems like this can be simplified to what's below:
> 
> ```
> this.setState({
>     searchQuery: query,
>     view: query.trim() ? VIEW_SEARCH : VIEW_FOLDER
> });
> ```

It's not the same as the original logic, because if `query && !trimmedQuery`, we don't change view state in the original code. But your advice makes sense and can make the logic simpler. Will do.

src/lib/viewers/archive/SearchBar.scss
mxiao6: That doesn't work cause we have to override `.be input[type='search']`

src/lib/viewers/archive/ArchiveExplorer.js
jstoffan: Cool. Setting primitive values on state only really matters when the values change. Setting a primitive value to its current value should basically be a no-op.

src/lib/viewers/archive/SearchBar.scss
jstoffan: Should we add a comment to that effect?

PR #1126, opened 2019-12-19
feat(archive): Add sortable columns to ArchiveExplorer

mickr:


boxcla:

src/lib/viewers/archive/ArchiveExplorer.js
```diff
@@ -156,50 +160,84 @@ class ArchiveExplorer extends React.Component {
...
 
+    /**
+     * Handle sort click
+     *
+     * @param {object} sort
+     * @param {string} sort.sortBy used to sort
+     * @param {string} sort.sortDirection set direction of sort either ASC | DESC
+     * @return {void}
+     */
+    handleSort = ({ sortBy, sortDirection }) => this.setState({ sortBy, sortDirection });
```
jstoffan: Let's keep all the `handle*` methods grouped together above the helper methods?

src/lib/viewers/archive/ArchiveExplorer.js
```diff
@@ -1,18 +1,20 @@
 import * as React from 'react';
 import PropTypes from 'prop-types';
 import getProp from 'lodash/get';
+import sortCollection from 'lodash/sortBy';
```
jstoffan: Can we call this `sortBy` to make it clear that it's a generic utility method?

mickr:
@jstoffan tests are coming

src/lib/viewers/archive/ArchiveExplorer.js
mickr: Sure, the reason I named it that is that it causes conflicts when using the keys provided from virtualized-table and de-structuring, but I can rename it in state.

src/lib/viewers/archive/ArchiveExplorer.js
```diff
@@ -156,50 +160,84 @@ class ArchiveExplorer extends React.Component {
         return itemCollection.filter(item => fuzzySearch(trimmedQuery, item.name, 0));
     };
 
+    /**
+     * Handle sort click
+     *
+     * @param {object} sort
```
mxiao6: Do we still need this parameter if we do the destruction? It's not used anywhere in the function body.

src/lib/viewers/archive/ArchiveExplorer.js
```diff
@@ -156,50 +160,84 @@ class ArchiveExplorer extends React.Component {
...
+     * @param {string} sort.sortBy used to sort
+     * @param {string} sort.sortDirection set direction of sort either ASC | DESC
```
mxiao6: should add "-" in between `sortBy - used to sort`

src/lib/viewers/archive/constants.js
```diff
@@ -1,7 +1,7 @@
 const TABLE_COLUMNS = {
-    KEY_MODIFIED_AT: 'key_modified_at',
-    KEY_NAME: 'key_name',
-    KEY_SIZE: 'key_size',
+    KEY_MODIFIED_AT: 'modified_at',
+    KEY_NAME: 'name',
+    KEY_SIZE: 'size',
```
mxiao6: What's the purpose of renaming here?

src/lib/viewers/archive/ArchiveExplorer.js
mickr: Yes, this will help with type hinting for browsers that support it.

src/lib/viewers/archive/constants.js
mickr: Turns out I don't have `sortBy` will still sort properly, probably just a microsecond slower so I changed it back.

mxiao6: No it won't be slower because they're exactly the same thing. I rewrite the row data in `getRowData`, so no matter what the key is, you're sorting the same value. One more thing, `[KEY_NAME]` is an object, can the lodash sortBy algorithm sort it well? 
```
[KEY_NAME]: {
    fullPath,
    isExternal: false,
    name,
    type,
}
```

mickr: We talked off-line about this

PR #1129, opened 2019-12-20
fix(preloader): Fix preload experience for presentation viewer

jstoffan:
These overrides reset the `previewUI` value that's passed into the super call back to undefined.


boxcla:

PR #1132, opened 2019-12-21
chore(types): Add support for typescript and add types to api module

jstoffan:
I converted the API module mostly as an example and to test the build/lint steps.


boxcla:

tsconfig.json
```diff
@@ -0,0 +1,5 @@
+{
+    "extends": "@box/frontend/ts/tsconfig.json",
+    "include": ["./src/**/*"],
+    "exclude": ["./src/**/*.js", "node_modules", "dist"]
```
priyajeet: What happens if you remove JS from the exclude section? I added it in the other project because it had flow types and TS complained about "types can only be in TS files" or something like that. Unsure what happens when the file is type free.

jstoffan: Seems to work fine without it. Removed.

src/lib/api.ts
```diff
@@ -0,0 +1,208 @@
...
+};
+
+/**
+ * Pass through transformer if the response type is text
+ * @param {Object} data
+ * @return {Object}
+ */
+const transformTextResponse = <D>(data: D): D => data;
+
+export default class Api {
```
priyajeet: nit: normally we try to keep exports all at the bottom. Though I guess with types we have them wherever they are defined.

PR #1133, opened 2019-12-21
fix(video): Change event type for setting up video element

mickr:
Fixes an issue with Safari not showing videos on MacOS Catalina


boxcla:

PR #1137, opened 2019-12-31
feat(archive): Add archive loader to prepare for conversion feature flip

mxiao6:
If conversion feature flip is not enabled for a user, the user will see the error message "We're sorry, your account is unable to preview this file type."

### DIFF
--- a/src/lib/constants.js
+++ b/src/lib/constants.js
@@ -93,8 +93,7 @@ export const STATUS_SUCCESS = 'success';
 export const STATUS_VIEWABLE = 'viewable';
 
 // X-Rep-Hints for Representations API
-export const X_REP_HINT_ARCHIVE = '[json]';
-export const X_REP_HINT_BASE = '[3d][pdf][text][mp3]';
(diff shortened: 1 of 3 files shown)


boxcla:

src/lib/constants.js
```diff
@@ -93,8 +93,7 @@ export const STATUS_SUCCESS = 'success';
 export const STATUS_VIEWABLE = 'viewable';
 
 // X-Rep-Hints for Representations API
-export const X_REP_HINT_ARCHIVE = '[json]';
-export const X_REP_HINT_BASE = '[3d][pdf][text][mp3]';
+export const X_REP_HINT_BASE = '[3d][pdf][text][mp3][json]';
```
jstoffan: This rep will now be hinted for every file. Did we check with the conversion team regarding the performance impact of this change?

mxiao6: > This rep will now be hinted for every file. Did we check with the conversion team regarding the performance impact of this change?

Yeah, they said it's ok.

PR #1138, opened 2020-01-03
feat(perf): Add new events to report modern performance metrics

jstoffan:


boxcla:

src/lib/PreviewPerf.ts
```diff
@@ -0,0 +1,62 @@
...
+    fcp?: number;
+    lcp?: number;
+}
+
+export default class PreviewPerf {
+    private fcpObserver: PerformanceObserver;
+
+    private lcpObserver: PerformanceObserver;
+
+    private performance: Performance = performance;
```
mxiao6: Should this be `window.performance`?

src/lib/Preview.js
```diff
@@ -560,13 +560,10 @@ class Preview extends EventEmitter {
                 });
         }
 
-        const downloadAttemptEvent = {
+        this.emit(PREVIEW_METRIC, {
```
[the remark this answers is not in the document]
jstoffan: Nice catch, thanks.

src/lib/PreviewPerf.ts
jstoffan: Karma chokes when I use it that way, unfortunately. It doesn't really make a difference.

PR #1140, opened 2020-01-04
test(archive): Add e2e test

mxiao6:
The test is failing because conversion feature flip is not on for our e2e test account. The account is live, so we have to wait until the next week. The test works on my dev vm.

### DIFF
--- a/test/integration/archive/ArchiveViewer.e2e.test.js
+++ b/test/integration/archive/ArchiveViewer.e2e.test.js
@@ -1,0 +1,52 @@
+// <reference types="Cypress" />
+describe('Archive Viewer', () => {
+    const token = Cypress.env('ACCESS_TOKEN');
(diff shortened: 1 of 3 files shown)


boxcla:

test/integration/archive/ArchiveViewer.e2e.test.js
```diff
@@ -0,0 +1,52 @@
...
+            .first()
+            .contains('Audio.mp3');
+    });
+
+    it('Should show matched item list based on search query', () => {
+        cy.getByTitle('Preview SDK Sample Archive').within(() => {
+            cy.get('button').click();
+        });
+
+        cy.get('.bp-SearchBar input').type('level');
```
jstoffan: We should be using `data-testid` for selectors rather than CSS classes.

PR #1143, opened 2020-01-06
fix(typescript): override rootDir from remote tsconfig.json

mickr:
This fixes an error from tsc that imports are not residing in the correct root dir.


boxcla:

mickr:
@jstoffan has a fix for this in another PR that fixes it in a different manner, so closing this.

PR #1145, opened 2020-01-08
feat(archive): Change data parsing for new data format

mxiao6:


boxcla:

src/lib/viewers/archive/ArchiveExplorer.js
```diff
@@ -85,13 +64,12 @@ class ArchiveExplorer extends React.Component {
      */
     getItemList = (itemCollection, fullPath) => {
         const folderInfo = itemCollection.find(item => item.absolute_path === fullPath);
-        const subItems = getProp(folderInfo, 'item_collection.entries');
+        const subItems = folderInfo.item_collection;
```
jstoffan: Should this be property be destructured? Is `folderInfo` always defined?

src/lib/viewers/archive/ArchiveExplorer.js
```diff
@@ -68,7 +47,7 @@ class ArchiveExplorer extends React.Component {
         addLocaleData(intlLocaleData);
 
         this.state = {
-            fullPath: props.itemCollection.find(info => !info.parent).absolute_path,
+            fullPath: props.itemCollection.find(info => info.name === info.absolute_path.slice(0, -1)).absolute_path,
```
jstoffan: Can this be simplified to compare `absolute_path` rather than `name`?

src/lib/viewers/archive/ArchiveExplorer.js
```diff
@@ -85,13 +64,12 @@ class ArchiveExplorer extends React.Component {
      */
     getItemList = (itemCollection, fullPath) => {
         const folderInfo = itemCollection.find(item => item.absolute_path === fullPath);
-        const subItems = getProp(folderInfo, 'item_collection.entries');
+        const subItems = folderInfo.item_collection;
         if (!subItems) {
```
jstoffan: It seems like this can be simplified a bit with a default value:

```
const { item_collection: folderItems = [] } = itemCollection.find(item => item.absolute_path === fullPath);
return itemCollection.filter(item => folderItems.includes(item.absolute_path));
```

PR #1146, opened 2020-01-08
feat(archive): Fix virtualized table cannot scroll

mxiao6:


boxcla:

src/lib/viewers/archive/ArchiveExplorer.js
```diff
@@ -189,43 +190,48 @@ class ArchiveExplorer extends React.Component {
...
-                                headerRenderer={sortableColumnHeaderRenderer}
-                                label={__('size')}
-                                width={1}
-                            />,
-                        ]}
-                    </VirtualizedTable>
+                    <AutoSizer disableWidth>
+                        {({ height }) => (
+                            <VirtualizedTable
+                                height={height - 120} // Subtruct SearchBar and Breadcrumbs heights
```
jstoffan: The need for this is counterintuitive, since the search bar and breadcrumbs are declared outside of the AutoSizer. Does the AutoSizer calculate its height based on the viewport? Does it still work if there are multiple previews on the same page?

mxiao6: > The need for this is counterintuitive, since the search bar and breadcrumbs are declared outside of the AutoSizer. Does the AutoSizer calculate its height based on the viewport? Does it still work if there are multiple previews on the same page?

No it's not based on the viewport. AutoSizer expands to fill its parent but does not work well flex box. I added a block container and problem solved!

src/lib/viewers/archive/ArchiveExplorer.js
```diff
@@ -4,15 +4,16 @@ import elementsMessages from 'box-elements-messages'; // eslint-disable-line
 import intlLocaleData from 'react-intl-locale-data'; // eslint-disable-line
 import Internationalize from 'box-ui-elements/es/elements/common/Internationalize';
 import fuzzySearch from 'box-ui-elements/es/utils/fuzzySearch';
+import VirtualizedTable from 'box-ui-elements/es/features/virtualized-table';
+import { AutoSizer } from 'react-virtualized';
+import { Column, SortDirection } from 'react-virtualized/dist/es/Table/index';
```
jstoffan: You should be able to omit `/index` here.

src/lib/viewers/archive/ArchiveExplorer.js
```diff
@@ -4,15 +4,16 @@ import elementsMessages from 'box-elements-messages'; // eslint-disable-line
 import intlLocaleData from 'react-intl-locale-data'; // eslint-disable-line
 import Internationalize from 'box-ui-elements/es/elements/common/Internationalize';
 import fuzzySearch from 'box-ui-elements/es/utils/fuzzySearch';
+import VirtualizedTable from 'box-ui-elements/es/features/virtualized-table';
+import { AutoSizer } from 'react-virtualized';
```
jstoffan: Does this work?

```
import { AutoSizer, Column, SortDirection } from 'react-virtualized';
```

src/lib/viewers/archive/ArchiveExplorer.js
```diff
@@ -189,43 +190,50 @@ class ArchiveExplorer extends React.Component {
...
-                                cellRenderer={sizeCellRenderer()}
-                                dataKey={KEY_SIZE}
-                                flexGrow={1}
-                                headerRenderer={sortableColumnHeaderRenderer}
-                                label={__('size')}
-                                width={1}
-                            />,
-                        ]}
-                    </VirtualizedTable>
+                    <div className="bp-AutoSizerContainer">
```
jstoffan: This isn't a new component, so it should probably be called `bp-ArchiveExplorer-table` or something.

PR #1149, opened 2020-01-11
feat(archive): Add root folder

mxiao6:


boxcla:

src/i18n/en-US.properties
```diff
@@ -84,6 +84,8 @@ error_flash_not_enabled=We’re sorry, the preview didn’t load because Flash i
 filename=Filename
 # Label for last modified date column name
 last_modified_date=Last modified date
+# Label for the root folder
+root_folder=Root Folder
```
mickr: Should make sure this is the appropriate designation for the folder

src/lib/viewers/archive/Breadcrumbs.js
```diff
@@ -19,14 +19,21 @@ class Breadcrumbs extends React.PureComponent {
      * @return {Array<Object>} path items including name and path string
      */
     getPathItems = fullPath => {
+        if (fullPath === ROOT_FOLDER) {
+            return [{ name: __('root_folder'), path: ROOT_FOLDER }];
+        }
+
```
mickr: Is there a new test the encompasses this behavior?

mxiao6: > Is there a new test the encompasses this behavior?

Yeah, I modified the old test to encompass this behavior.

mxiao6: > Is there a new test the encompasses this behavior?

done

mickr:
Possible to do another patch release today so that this will go into the GA release?

src/lib/viewers/archive/ArchiveExplorer.js
```diff
@@ -66,6 +63,16 @@ class ArchiveExplorer extends React.Component {
      * @return {Array<Object>} filtered itemlist for target folder
      */
     getItemList = (itemCollection, fullPath) => {
+        if (fullPath === ROOT_FOLDER) {
+            // Trying to find the root items
+            // The only way to tell what the root items are
+            // is by comparing the name and absolute path,
+            // which are the same for files and differ by '/' for folders
+            return itemCollection.filter(
+                info => info.name === info.absolute_path || info.name === info.absolute_path.slice(0, -1),
```
jstoffan: Should these properties be destructured to reduce duplication?

src/lib/viewers/archive/Breadcrumbs.js
```diff
@@ -19,14 +19,21 @@ class Breadcrumbs extends React.PureComponent {
...
         const pathNames = fullPath.split('/').slice(0, -1);
         // join path names from root to current index to get absolute path
         const getAbsolutePath = index => pathNames.slice(0, index + 1).join('/');
 
-        return pathNames.map((name, index) => ({
-            name,
-            path: `${getAbsolutePath(index)}/`,
-        }));
+        return [
+            { name: __('root_folder'), path: ROOT_FOLDER },
```
jstoffan: Is this needed if the zip file already has a root folder? If so, can we avoid duplicating this object declaration by pushing the other pathnames onto a `results` array?

src/lib/viewers/archive/__tests__/Breadcrumbs-test-react.js
```diff
@@ -22,27 +24,40 @@ describe('lib/viewers/archive/Breadcrumbs', () => {
...
+            let pathItems = component.instance().getPathItems(ROOT_FOLDER);
 
             expect(pathItems).to.eql([
+                {
+                    name: __('root_folder'),
+                    path: ROOT_FOLDER,
+                },
+            ]);
+
+            pathItems = component.instance().getPathItems(fullPath);
```
jstoffan: Should this be a separate test?

src/lib/viewers/archive/constants.js
```diff
@@ -9,4 +9,6 @@ const VIEWS = {
     VIEW_SEARCH: 'search',
 };
 
-export { TABLE_COLUMNS, VIEWS };
+const ROOT_FOLDER = 'root/';
```
jstoffan: Can this be defined at the top of the file? Is this folder injected into the manifest by conversion? If so, can you add a comment stating such?

mxiao6: > Can this be defined at the top of the file? Is this folder injected into the manifest by conversion? If so, can you add a comment stating such?

This folder is not injected by conversion. So it can be named 'root/' or 'rootFolder/' or whatever we want.

src/lib/viewers/archive/Breadcrumbs.js
mxiao6: > Is this needed if the zip file already has a root folder? If so, can we avoid duplicating this object declaration by pushing the other pathnames onto a `results` array?

I researched G Drive and observed they add root folder for all zip files. So I think this is needed if the zip file already has a root folder.

We cannot prepend it the the results array because the filepath is 'root/' not empty string, so it will add another 'root' object.

src/lib/viewers/archive/ArchiveViewer.js
```diff
@@ -71,7 +71,7 @@ class ArchiveViewer extends BaseViewer {
 
         try {
             /* global BoxArchive loaded from archive.js */
-            this.archiveComponent = new BoxArchive(this.archiveEl, data);
+            this.archiveComponent = new BoxArchive(this.archiveEl, this.options.file.name, data);
```
jstoffan: Just to confirm: Are `this.options.file` and `this.options.file.name` always defined?

mxiao6: > Just to confirm: Are `this.options.file` and `this.options.file.name` always defined?

If file is not defined, the representations are not too, so the expected behavior is to throw the error.

PR #1153, opened 2020-01-16
fix(archive): Fix blank screen when sorting

mickr:
* Add try/catch handler for exceptions
* Handle case when sorting data is falsy

### DIFF
--- a/src/lib/viewers/archive/ArchiveExplorer.js
+++ b/src/lib/viewers/archive/ArchiveExplorer.js
@@ -162,18 +162,27 @@ class ArchiveExplorer extends React.Component {
      */
     sortItemList(itemList) {
         const { sortBy, sortDirection } = this.state;
+        let sortedItems;
 
         if (!sortBy.length) {
             return itemList;
         }
 
-        const sortedItems = itemList.sort((a, b) => {
-            if (typeof a[sortBy] === 'number' && typeof b[sortBy] === 'number') {
-                return a[sortBy] - b[sortBy];
-            }
+        try {
+            sortedItems = itemList.sort((a, b) => {
+                if (!a[sortBy] || !b[sortBy]) {
+                    return -1;
+                }
 
-            return a[sortBy].localeCompare(b[sortBy]);
-        });
+                if (typeof a[sortBy] === 'number' && typeof b[sortBy] === 'number') {
+                    return a[sortBy] - b[sortBy];
+                }
+
+                return a[sortBy].localeCompare(b[sortBy]);
+            });
+        } catch {
+            sortedItems = itemList;
+        }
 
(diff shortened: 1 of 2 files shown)


boxcla:

src/lib/viewers/archive/ArchiveExplorer.js
```diff
@@ -162,18 +162,27 @@ class ArchiveExplorer extends React.Component {
...
             return itemList;
         }
 
-        const sortedItems = itemList.sort((a, b) => {
-            if (typeof a[sortBy] === 'number' && typeof b[sortBy] === 'number') {
-                return a[sortBy] - b[sortBy];
-            }
+        try {
+            sortedItems = itemList.sort((a, b) => {
+                if (!a[sortBy] || !b[sortBy]) {
```
jstoffan: Should we also check `!a || !b` for null/undefined values?

src/lib/viewers/archive/ArchiveExplorer.js
```diff
@@ -162,18 +162,27 @@ class ArchiveExplorer extends React.Component {
...
 
         if (!sortBy.length) {
             return itemList;
         }
 
-        const sortedItems = itemList.sort((a, b) => {
-            if (typeof a[sortBy] === 'number' && typeof b[sortBy] === 'number') {
-                return a[sortBy] - b[sortBy];
-            }
+        try {
```
jstoffan: It seems like we should be able to guard against pretty much any issues without resorting to try/catch here. I think we should add a `componentDidCatch` to the parent component, though.

src/lib/viewers/archive/ArchiveExplorer.js
```diff
@@ -162,18 +162,27 @@ class ArchiveExplorer extends React.Component {
      */
     sortItemList(itemList) {
         const { sortBy, sortDirection } = this.state;
+        let sortedItems;
```
jstoffan: Should this be declared after the guard clause?

src/lib/viewers/archive/ArchiveExplorer.js
mickr: I think the keys are always there, but the values maybe be `null | value`, but I can check without much effort 👍

src/lib/viewers/archive/ArchiveExplorer.js
mickr: I think we are safe as well, but I through it in for good measure to ensure we always get an itemList returned.

jstoffan: Ideally, try/catch should be reserved for exception handling rather than used for flow control. I don't think it hurts in this case, but it can get overused.

mickr:
I will follow this up with a `componentDidCatch` in the React tree to catch exceptions in the future.

PR #1157, opened 2020-01-17
fix(archive): Fix IE11 does not respect line-height

mxiao6:


boxcla:

PR #1160, opened 2020-01-27
fix(archive): Fix column misaligned in expiring embed

mxiao6:


boxcla:

src/lib/viewers/archive/ArchiveExplorer.scss
```diff
@@ -7,4 +7,5 @@
 
 .bp-ArchiveExplorer-table {
     height: 100%;
+    text-align: start; // override embed widget styles
```
jstoffan: It looks like these values aren't [supported](https://developer.mozilla.org/en-US/docs/Web/CSS/text-align#Browser_compatibility) in IE11/Edge. Should we use something else?

mxiao6: > It looks like these values aren't [supported](https://developer.mozilla.org/en-US/docs/Web/CSS/text-align#Browser_compatibility) in IE11/Edge. Should we use something else?

Sorry! Changed to `left` which is the default value of `text-align`.

PR #1161, opened 2020-01-27
fix(archive): Fix inconsistent search cancel button in Expiring Embed

mxiao6:


boxcla:

src/lib/viewers/archive/SearchBar.scss
```diff
@@ -13,5 +13,10 @@
         width: 100%;
         height: 36px; // fix IE 11 doesn't respect line-height
         font: inherit;
+
+        &::-webkit-search-cancel-button {
+            display: inline-block; // override BUIE display none
```
jstoffan: @mxiao6, it looks like this button is hidden in the main Box webapp. Is the desire to show it here or to hide it for Expiring Embed links? It seems like it should be the latter for consistency?

ConradJChan:
@Mergifyio refresh

PR #1162, opened 2020-01-28
fix(archive): Add default sort order

mickr:
* add default sort order
* fix sorting of file sizes with a size of 0

### DIFF
--- a/src/lib/viewers/archive/ArchiveExplorer.js
+++ b/src/lib/viewers/archive/ArchiveExplorer.js
@@ -50,8 +50,8 @@ class ArchiveExplorer extends React.Component {
         this.state = {
             fullPath: ROOT_FOLDER,
             searchQuery: '',
-            sortBy: '',
-            sortDirection: SortDirection.ASC,
+            sortBy: 'name',
+            sortDirection: SortDirection.DESC,
             view: VIEW_FOLDER,
         };
     }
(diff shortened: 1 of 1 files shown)


boxcla:

src/lib/viewers/archive/ArchiveExplorer.js
```diff
@@ -168,8 +168,8 @@ class ArchiveExplorer extends React.Component {
         }
 
         const sortedItems = itemList.sort((a = {}, b = {}) => {
-            if (!a[sortBy] || !b[sortBy]) {
-                return -1;
+            if (a[sortBy] === null || b[sortBy] === null) {
```
jstoffan: Are any other non-string values possible? It seems like this change could cause us to throw an error on line 179.

mxiao6: Maybe something like
```
            if (!a[sortBy] || !b[sortBy]) {
                return !a[sortBy] ? -1 : 1;
            }
```

src/lib/viewers/archive/ArchiveExplorer.js
```diff
@@ -168,18 +168,29 @@ class ArchiveExplorer extends React.Component {
         }
 
         const sortedItems = itemList.sort((a = {}, b = {}) => {
-            if (!a[sortBy] || !b[sortBy]) {
+            const aItem = a[sortBy];
+            const bItem = b[sortBy];
+
+            if (aItem === null) {
```
jstoffan: Do we want to do the same for `undefined`?

jstoffan:
Seems like we should we have some input/output unit tests that show the sort is working as we want, as well.

PR #1163, opened 2020-01-28
chore(deps): Upgrade node and yarn version requirements

jstoffan:
This change also upgrades node-sass to avoid issues with Node 12.


boxcla:

PR #1164, opened 2020-01-29
fix(build): fix cmd to return correct codes

mickr:


boxcla:

build/release.sh
```diff
@@ -15,9 +15,7 @@ cmd() {
         return 0
     fi
 
-    "$@"
-
-    return 0
+    return "$@"
```
mxiao6: Will the `$@` return anything? Should it be simply `"$@"` without any return statement?

PR #1165, opened 2020-01-29
fix(archive): Fix inconsistent search bar style on mobile devices

mxiao6:
Screenshot(iOS)-COXP-7274
Fixed:
WX20200129-105954@2x


boxcla:

PR #1166, opened 2020-01-29
fix(archive): Fix truncated breadcrumb when file name is too long

mxiao6:
long-breadcrumb
->
correct-breadcrumb


boxcla:

PR #1167, opened 2020-01-30
fix(doc): Remove unnecessary http header for iOS

mxiao6:
The webkit bug was fixed on Jan 2018. I removed the header and tried to preview a pdf file on iOS 12.0 and iOS 13.3 (We only support two latest released versions), and they both work well. So I'm gonna remove the unnecessary http header code for iOS.

### DIFF
--- a/src/lib/viewers/doc/DocBaseViewer.js
+++ b/src/lib/viewers/doc/DocBaseViewer.js
@@ -647,7 +647,6 @@ class DocBaseViewer extends BaseViewer {
         const { file, location } = this.options;
         const { size, watermark_info: watermarkInfo } = file;
         const assetUrlCreator = createAssetUrlCreator(location);
-        const httpHeaders = {};
         const queryParams = {};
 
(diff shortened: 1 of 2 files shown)


boxcla:

src/lib/viewers/doc/DocBaseViewer.js
```diff
@@ -669,12 +669,6 @@ class DocBaseViewer extends BaseViewer {
         const rangeChunkSizeDefault = location.locale === 'en-US' ? RANGE_CHUNK_SIZE_US : RANGE_CHUNK_SIZE_NON_US;
         const rangeChunkSize = this.getViewerOption('rangeChunkSize') || rangeChunkSizeDefault;
 
-        // Fix incorrectly cached range requests on older versions of iOS webkit browsers,
-        // see: https://bugs.webkit.org/show_bug.cgi?id=82672
-        if (Browser.isIOS()) {
-            httpHeaders['If-None-Match'] = 'webkit-no-cache';
```
jstoffan: Is `httpsHeaders` still used at all? If not, let's remove it.

PR #1168, opened 2020-02-03
fix(test): Fix archive e2e test due to default sort by Name

mxiao6:


boxcla:

PR #1169, opened 2020-02-04
fix(build): guard for removing readme file

mickr:
* Check to ensure a back up of readme exists before deleting it.


boxcla:

PR #1171, opened 2020-02-05
fix(archive): Folders come before files when sorting

mxiao6:
image

### DIFF
--- a/src/lib/viewers/archive/ArchiveExplorer.js
+++ b/src/lib/viewers/archive/ArchiveExplorer.js
@@ -51,7 +51,7 @@ class ArchiveExplorer extends React.Component {
             fullPath: ROOT_FOLDER,
             searchQuery: '',
             sortBy: 'name',
-            sortDirection: SortDirection.DESC,
+            sortDirection: SortDirection.ASC,
             view: VIEW_FOLDER,
         };
     }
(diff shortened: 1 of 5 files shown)


boxcla:

src/lib/viewers/archive/ArchiveExplorer.js
```diff
@@ -190,7 +190,19 @@ class ArchiveExplorer extends React.Component {
             return 0;
         });
 
-        return sortDirection === SortDirection.DESC ? sortedItems : sortedItems.reverse();
+        if (sortDirection === SortDirection.DESC) {
```
mickr: I know this is to keep consistent with Webapp, but I've always found this sorting weird because when I sort by name I want the files/folders sorted by name, not by type and then name. Code looks good but just calling out if this is the UX we really want and design wants.

src/lib/viewers/archive/ArchiveExplorer.js
```diff
@@ -190,7 +190,19 @@ class ArchiveExplorer extends React.Component {
...
+            sortedItems.reverse();
+        }
+
+        // folders always come before files
+        sortedItems.sort((a, b) => {
```
jstoffan: It seems like we should be able to collapse all of this into a single `.sort` call, with `sortDirection` controlling the comparator logic. Can we also check to see how webapp does this sorting? I don't see an equivalent complex sort function there.

src/lib/viewers/archive/ArchiveExplorer.js
```diff
@@ -180,17 +185,15 @@ class ArchiveExplorer extends React.Component {
             }
 
             if (typeof aItem === 'number' && typeof bItem === 'number') {
-                return aItem - bItem;
+                return sortDirection === SortDirection.ASC ? aItem - bItem : bItem - aItem;
```
jstoffan: How about flipping them at assignment to avoid the repeated logic?

```
const aItem = sortDirection === SortDirection.ASC ? a : b;
const bItem = sortdirection === SortDirection.ASC ? b : a;
```

mxiao6: > How about flipping them at assignment to avoid the repeated logic?
> 
> ```
> const aItem = sortDirection === SortDirection.ASC ? a : b;
> const bItem = sortdirection === SortDirection.ASC ? b : a;
> ```

Cannot flip them before `null` check, right?

PR #1172, opened 2020-02-05
fix(archive): Reset vertical scroll when navigating folders

mxiao6:


boxcla:

test/integration/archive/ArchiveViewer.e2e.test.js
```diff
@@ -53,4 +56,30 @@ describe('Archive Viewer', () => {
...
+        cy.getByTitle('Collection').within(() => {
+            cy.get('button').click();
+        });
+
+        cy.get('.ReactVirtualized__Table__Grid').then($tables => {
+            const table = $tables[0];
+            table.scrollTo(0, 50);
+
+            const tableTop = table.getBoundingClientRect().top;
+            cy.get('.ReactVirtualized__Table__row').then($rows => {
```
jstoffan: Any way we can avoid using CSS classes here?

test/support/constants.js
```diff
@@ -2,6 +2,7 @@ Cypress.env({
     // NOTE: The token used is a readonly token accessing public data in a demo enterprise.
     ACCESS_TOKEN: 'S8wjvjOL9GEK5VtXsQNVMOwSrx1g55oC',
     FILE_ID_ARCHIVE: '589625563707',
+    FILE_ID_ARCHIVE_LONG: '611800378713',
```
jstoffan: It looks like we have `FILE_ID_DOC_LARGE`. Can we rename to `FILE_ID_ARCHIVE_LARGE`, as well?

test/integration/archive/ArchiveViewer.e2e.test.js
mxiao6: Is using filename (like `cy.contains('File001.txt')`) better?

jstoffan: I think so, yeah, since it's based on user-visible information. We do it in other tests, too.

PR #1177, opened 2020-02-28
feat(print): add check to see if printing is available

mickr:
* Adds helper method to check if printing is available on the preview instance

### DIFF
--- a/src/lib/Preview.js
+++ b/src/lib/Preview.js
@@ -1305,13 +1305,12 @@ class Preview extends EventEmitter {
         // Log now that loading is finished
         this.emitLoadMetrics(data);
 
-        // Show download and print buttons if user can download
         if (canDownload(this.file, this.options)) {
             this.ui.showDownloadButton(this.download);
+        }
 
-            if (this.canPrint()) {
-                this.ui.showPrintButton(this.print);
-            }
+        if (this.canPrint()) {
+            this.ui.showPrintButton(this.print);
         }
 
         const { error } = data;
--- a/src/lib/Preview.js
+++ b/src/lib/Preview.js
@@ -503,13 +503,23 @@ class Preview extends EventEmitter {
     }
 
     /**
+     * Checks if the file being previewed supports printing.
+     *
+     * @public
+     * @return {boolean}
+     */
+    canPrint() {
+        return !!canDownload(this.file, this.options) && checkFeature(this.viewer, 'print');
+    }
+
+    /**
      * Prints the file being previewed if the viewer supports printing.
      *
      * @public
      * @return {void}
      */
     print() {
-        if (canDownload(this.file, this.options) && checkFeature(this.viewer, 'print')) {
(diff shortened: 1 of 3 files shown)


mickr:
This will be used by the contentPreview to check whether a file type will support printing, there is a PR coming for that.

src/lib/Preview.js
```diff
@@ -1299,7 +1309,7 @@ class Preview extends EventEmitter {
         if (canDownload(this.file, this.options)) {
             this.ui.showDownloadButton(this.download);
 
-            if (checkFeature(this.viewer, 'print')) {
+            if (this.canPrint()) {
```
jstoffan: Probably move this outside the canDownload check to flatten the conditions?

src/lib/__tests__/Preview-test.js
```diff
@@ -795,6 +795,34 @@ describe('lib/Preview', () => {
...
 
+    describe('canPrint()', () => {
+        beforeEach(() => {
+            stubs.canDownload = sandbox.stub(file, 'canDownload');
+            stubs.checkFeature = sandbox.stub(file, 'checkFeature');
+        });
+
+        it('should return true is file is downloadable and has printing feature', () => {
+            stubs.canDownload.returns(true);
+            stubs.checkFeature.withArgs(sinon.match.any, 'print').returns(true);
```
jstoffan: Try to avoid using withArgs if possible. It’s difficult to migrate to other test frameworks like Jest.

src/lib/Preview.js
```diff
@@ -502,14 +502,24 @@ class Preview extends EventEmitter {
...
     }
 
+    /**
+     * Checks if the file being previewed supports printing.
+     *
+     * @public
+     * @return {boolean}
+     */
+    canPrint() {
+        return !!canDownload(this.file, this.options) && checkFeature(this.viewer, 'print');
```
ConradJChan: Is the `!!` necessary here?

mickr: Both return booleans, but figured to ensure it is a boolean in case those interfaces change, I don't believe it hurts anything.

src/lib/Preview.js
mickr: Sure

src/lib/__tests__/Preview-test.js
```diff
@@ -795,6 +795,34 @@ describe('lib/Preview', () => {
...
+
+            expect(preview.canPrint()).to.equal(true);
+        });
+
+        it('should return false is file is not downloadable and has printing feature', () => {
+            stubs.canDownload.returns(false);
+            stubs.checkFeature.withArgs(sinon.match.any, 'print').returns(true);
```
mxiao6: Forgot to remove the other two `withArgs`?

src/lib/Preview.js
jstoffan: Seems like the `!!` should be omitted if it's not needed.

PR #1180, opened 2020-03-12
fix(log): Fix log not getting serialized properly

mxiao6:


src/lib/Preview.js
```diff
@@ -204,10 +204,8 @@ class Preview extends EventEmitter {
 
                 this.emitLogEvent(PREVIEW_METRIC, {
                     event_name: PREVIEW_END_EVENT,
-                    value: {
-                        duration: previewDuration,
-                        viewer_status: this.viewer.getLoadStatus(),
-                    },
+                    value: previewDuration,
+                    viewer_status: this.viewer.getLoadStatus(),
```
ConradJChan: I don't think this will work because that field doesn't exist in the metrics table. You'll need to emit a separate event entirely

mxiao6:
> fix(log) maybe?

But this is not user facing?

ConradJChan:
imo, it changes the functionality so I would change it to fix. Also chore won't get included in a release by itself I think

PR #1183, opened 2020-03-24
feat(annotations): Add region comment button in toolbar

mxiao6:
region-comment
region-comment-active

- [x] Tests
- [x] TypeScript
- [x] SUIT styles


src/lib/AnnotationControls.js
```diff
@@ -0,0 +1,63 @@
+import noop from 'lodash/noop';
+import { ICON_REGION_COMMENT } from './icons/icons';
+import Controls, { CLASS_BOX_CONTROLS_GROUP_BUTTON } from './Controls';
+
+const CLASS_ANNOTATIONS_GROUP = 'bp-annotations-group';
+const CLASS_REGION_COMMENT_BUTTON = 'bp-region-comment-btn';
+const CLASS_BUTTON_ACTIVE = 'active';
+
+class AnnotationControls {
```
jstoffan: Since this is a new component, can it be written in TypeScript?

src/lib/AnnotationControls.js
```diff
@@ -0,0 +1,63 @@
...
+    /** @property {Controls} - Controls object */
+    controls;
+
+    /** @property {Boolean} - Region comment mode active state */
+    isRegionCommentActive = false;
```
jstoffan: We've been referring to these as "area highlights" or just area annotations. May want to check with PM on official naming.

src/lib/AnnotationControls.js
```diff
@@ -0,0 +1,63 @@
...
+     */
+    constructor(controls) {
+        if (!controls || !(controls instanceof Controls)) {
+            throw Error('controls must be an instance of Controls');
+        }
+
+        this.controls = controls;
+    }
+
+    regionCommentClickHandler = onRegionCommentClick => event => {
```
jstoffan: We should use `handle` for event handler methods, so `handleAreaButtonClick`.

src/lib/Controls.scss
```diff
@@ -180,3 +180,26 @@
         margin-left: 4px;
     }
 }
+
+.bp-annotations-group {
```
jstoffan: Let's move these styles to their own stylesheet file and write them in the SUIT convention.

src/lib/Preview.js
```diff
@@ -982,6 +982,9 @@ class Preview extends EventEmitter {
         // Add the response interceptor to the preview instance
         this.options.responseInterceptor = options.responseInterceptor;
 
+        // Option to enable use of annotations v4
+        this.options.enableAnnotations = options.enableAnnotations || false;
```
jstoffan: What do you think about naming this `showAnnotationsControls` and grouping it with `showAnnotations`? It can also be simplified to `this.options.showAnnotationsControls = !!options.showAnnotationsControls`;

src/lib/viewers/doc/DocBaseViewer.js
```diff
@@ -1072,6 +1077,8 @@ class DocBaseViewer extends BaseViewer {
             this.controls.add(__('toggle_findbar'), () => this.findBar.toggle(), 'bp-toggle-findbar-icon', ICON_SEARCH);
         }
 
+        this.pageControls.add(this.pdfViewer.currentPageNumber, this.pdfViewer.pagesCount);
```
jstoffan: What as the reason for this moving?

src/lib/viewers/image/ImageBaseViewer.js
```diff
@@ -201,8 +203,19 @@ class ImageBaseViewer extends BaseViewer {
         this.controls = new Controls(this.containerEl);
         this.zoomControls = new ZoomControls(this.controls);
         this.zoomControls.init(this.scale, { onZoomIn: this.zoomIn, onZoomOut: this.zoomOut });
+        if (this.options.enableAnnotations) {
```
jstoffan: Images aren't yet supported.

src/lib/AnnotationControls.js
mxiao6: > We've been referring to these as "area highlights" or just area annotations. May want to check with PM on official naming.

I'm confusing too, because the design is using "region comment" and tooltip hint is "Comment on Region".

src/lib/viewers/doc/DocBaseViewer.js
mxiao6: > What as the reason for this moving?

It's in the new design and I ask Matt, he says he does it on purpose.

src/lib/AnnotationControls.js
mxiao6: > Since this is a new component, can it be written in TypeScript?

Will do

src/lib/Controls.scss
mxiao6: > Let's move these styles to their own stylesheet file and write them in the SUIT convention.

Will do

src/lib/Preview.scss
```diff
@@ -1,6 +1,8 @@
 @import 'common';
 @import 'loading';
 @import 'navigation';
+@import './AnnotationControls';
```
jstoffan: Should these be imported by src/lib/Controls.scss, instead?

src/lib/icons/icons.js
```diff
@@ -6,6 +6,7 @@ import FULLSCREEN_OUT from './full_screen_out_24px.svg';
 import ROTATE_LEFT from './rotate_left_24px.svg';
 import ZOOM_IN from './zoom_in.svg';
 import ZOOM_OUT from './zoom_out.svg';
+import REGION_COMMENT from './annotation_region_comment.svg';
```
jstoffan: Should the file name be `region_comment.svg` to match the constant?

src/lib/viewers/doc/DocBaseViewer.js
```diff
@@ -101,6 +102,7 @@ class DocBaseViewer extends BaseViewer {
         this.toggleThumbnails = this.toggleThumbnails.bind(this);
         this.zoomIn = this.zoomIn.bind(this);
         this.zoomOut = this.zoomOut.bind(this);
+        this.regionClickHandler = this.regionClickHandler.bind(this);
```
jstoffan: Alphabetize?

src/lib/AnnotationControls.scss
```diff
@@ -0,0 +1,22 @@
+.bp-AnnotationControls-group {
+    padding: 0 4px 0 8px;
+    border-left: 1px solid $twos;
+
+    & .bp-AnnotationControls-regionBtn {
```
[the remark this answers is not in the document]
mxiao6: > We should avoid nesting where possible. This class can be declared at the top level in this file.

I tried it at first, but the style is overridden by another class. It has to be specific like this.

src/lib/AnnotationControls.scss
```diff
@@ -0,0 +1,22 @@
...
+        width: 32px;
```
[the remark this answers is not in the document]
mxiao6: > Are these values common between controls? Should they be shared via a variable?

yeah it's common. Will put into a variable.

src/lib/AnnotationControls.scss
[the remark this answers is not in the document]
jstoffan: Gotcha. You should be able to omit the `&`, then.

src/lib/AnnotationControls.ts
```diff
@@ -0,0 +1,73 @@
...
+     * Initialize the annotation controls with options.
+     *
+     * @param {Function} [options.onRegionClick] - Callback when region comment button is clicked
+     * @return {void}
+     */
+    public init({ onRegionClick = noop }: Options = {}): void {
+        const groupElement = this.controls.addGroup(CLASS_ANNOTATIONS_GROUP);
+        this.controls.add(
+            // eslint-disable-next-line @typescript-eslint/ban-ts-ignore
+            // @ts-ignore
```
jstoffan: What are we ignoring here?

src/lib/AnnotationControls.ts
```diff
@@ -0,0 +1,73 @@
+import noop from 'lodash/noop';
+import { ICON_REGION_COMMENT } from './icons/icons';
+import Controls, { CLASS_BOX_CONTROLS_GROUP_BUTTON } from './Controls';
+
+export const CLASS_ANNOTATIONS_GROUP = 'bp-AnnotationControls-group';
+export const CLASS_REGION_BUTTON = 'bp-AnnotationControls-regionBtn';
+export const CLASS_BUTTON_ACTIVE = 'is-active';
+
+export type Options = {
+    onRegionClick?: Function;
```
jstoffan: We should be able to define a more accurate type here and in the handler below: 

```onRegionClick?: ({ isRegionActive: boolean, event: MouseEvent }) => void```

File: src/lib/AnnotationControls.ts [hunk 15]
```diff
@@ -0,0 +1,74 @@
+import noop from 'lodash/noop';
+import { ICON_REGION_COMMENT } from './icons/icons';
+import Controls, { CLASS_BOX_CONTROLS_GROUP_BUTTON } from './Controls';
+
+export const CLASS_ANNOTATIONS_GROUP = 'bp-AnnotationControls-group';
+export const CLASS_REGION_BUTTON = 'bp-AnnotationControls-regionBtn';
+export const CLASS_BUTTON_ACTIVE = 'is-active';
+
+export type RegionHandlerType = ({ isRegionActive, event }: { isRegionActive: boolean; event: MouseEvent }) => void;
```
jstoffan: Doesn't need `Type` at the end.

PR #1185, opened 2020-03-26
fix(controls): Remove controls wrapper

mxiao6:


src/lib/Controls.js
```diff
@@ -35,13 +35,9 @@ class Controls {
     constructor(container) {
         this.containerEl = container;
 
-        const controlsWrapperEl = document.createElement('div');
-        controlsWrapperEl.className = 'bp-controls-wrapper';
-        controlsWrapperEl.setAttribute('data-testid', 'controls-wrapper');
```
jstoffan: Make sure this isn't used by any end-to-end tests in other repos (elements, webapp, etc.)

src/lib/Controls.scss
```diff
@@ -1,15 +1,6 @@
-.bp-controls-wrapper {
+.bp-controls {
     position: absolute;
     bottom: 25px;
-    left: 50%;
```
jstoffan: It doesn't seem like the control bar will be centered on the screen. Do we need to do something like `left: 50%; transform: translateX(-50%);`?

src/lib/_common.scss
```diff
@@ -160,6 +160,7 @@ $header-height: 48px;
         display: flex;
         flex: 1 1 auto;
         align-items: center;
+        justify-content: center;
```
jstoffan: What is the reason for this change? It seems pretty risky.

mxiao6: > What is the reason for this change? It seems pretty risky.

so we can avoid using `left: 50%; transform: translateX(-50%);`

src/lib/Controls.scss
mxiao6: > It doesn't seem like the control bar will be centered on the screen. Do we need to do something like `left: 50%; transform: translateX(-50%);`?

I did it in the parent container, which you think it's risky below

src/lib/Controls.js
mxiao6: > Make sure this isn't used by any end-to-end tests in other repos (elements, webapp, etc.)

Checked (elements, webapp, annotations) and there's no usage of it.

PR #1187, opened 2020-03-30
fix(a11y): fix keyboard navigation in video viewer

jeeyi:
- Made the play button in the middle focusable
- Disabled hiding of the control when user starts to navigate via keyboard (tab or shift+tab) so that buttons in the controls can be focused

### DIFF
--- a/src/lib/viewers/media/MediaControls.js
+++ b/src/lib/viewers/media/MediaControls.js
@@ -672,6 +672,16 @@ class MediaControls extends EventEmitter {
     }
 
     /**
+     * Prevents hiding of the controls and show media controls.
+     *
+     * @return {void}
+     */
+    disableHideAndShow() {
+        this.preventHiding = true;
+        this.show();
+    }
+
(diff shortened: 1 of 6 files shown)


src/lib/viewers/media/MediaControls.js
```diff
@@ -671,6 +671,16 @@ class MediaControls extends EventEmitter {
         }
     }
 
+    /**
+     * Prevents hiding of the controls and show media controls.
+     *
+     * @return {void}
+     */
+    disableHideAndShow() {
+        this.preventHiding = true;
```
mxiao6: Do you need another function to enable hiding?

jeeyi: When user starts to tab around, this user is considered as an accessibility user and hiding is disabled. Enable hiding is not necessary.

jstoffan:
We'll reassess this change at a later date. Closing for now.

PR #1188, opened 2020-03-30
feat(annotations): Hide annotation controls when fullscreen is active

mxiao6:


src/lib/viewers/BaseViewer.js
```diff
@@ -536,6 +536,9 @@ class BaseViewer extends EventEmitter {
      */
     toggleFullscreen() {
         fullscreen.toggle(this.containerEl);
+        if (this.options.showAnnotationsControls) {
```
jstoffan: There are several different ways to enter fullscreen mode, so we'll probably want to handle this from within the component. Take a look at MediaControls for an example.

src/lib/AnnotationControls.ts
```diff
@@ -32,8 +35,47 @@ export default class AnnotationControls {
...
+    }
+
+    /**
+     * [destructor]
+     *
+     * @return {void}
+     */
+    public destroy(): void {
+        if (fullscreen) {
+            fullscreen.removeAllListeners();
```
jstoffan: This will destroy all of the listeners on the fullscreen module, right? I don't think we'll want to do that when these controls are destroyed.

src/lib/AnnotationControls.ts
```diff
@@ -32,8 +35,47 @@ export default class AnnotationControls {
...
+        }
     }
 
+    /**
+     * Attaches event handlers
+     *
+     * @return {void}
+     */
+    private attachEventHandlers = (): void => {
```
jstoffan: Does this need to be a property or can it be a regular method?

src/lib/AnnotationControls.ts
mxiao6: > This will destroy all of the listeners on the fullscreen module, right? I don't think we'll want to do that when these controls are destroyed.

We destroy controls when we destroy viewer, right? I think when we construct another viewer, we will create another controls and attach the fullscreen module again. So I think it's ok to remove all the listeners when controls are destroyed.

jstoffan: I agree that's how it works today in practice, so it has the desired effect right now. What if, in six months, we need add logic to remove and destroy just these controls? Fullscreen would stop working and likely cause a production issue.

Encapsulation is important to reduce side effects in our code. Let's have the component remove its own handlers only.

src/lib/AnnotationControls.ts
```diff
@@ -32,8 +38,63 @@ export default class AnnotationControls {
...
+    /**
+     * Handle fullscreen change
+     *
+     * @param {boolean} isFullscreen - true if full screen will be active
+     * @return {void}
+     */
+    private handleFullscreenChange = (isFullscreen: boolean): void => {
+        const groupElement = this.controlsElement.querySelector(`.${CLASS_ANNOTATIONS_GROUP}`);
+
+        if (groupElement) {
```
ConradJChan: Maybe negate the if condition to reduce nesting?

src/lib/AnnotationControls.ts
```diff
@@ -32,8 +38,63 @@ export default class AnnotationControls {
...
+            if (isFullscreen) {
+                groupElement.classList.add(CLASS_GROUP_HIDE);
+            } else {
+                groupElement.classList.remove(CLASS_GROUP_HIDE);
+            }
+        }
+    };
+
+    /**
+     * Enter fullscreen handler
```
ConradJChan: Do we still need to do jsdoc with typescript?

src/lib/AnnotationControls.ts
```diff
@@ -42,13 +103,15 @@ export default class AnnotationControls {
...
-            regionButtonElement.classList.add(CLASS_BUTTON_ACTIVE);
-        } else {
-            regionButtonElement.classList.remove(CLASS_BUTTON_ACTIVE);
+        if (regionButtonElement) {
+            this.isRegionActive = !this.isRegionActive;
+            if (this.isRegionActive) {
+                regionButtonElement.classList.add(CLASS_BUTTON_ACTIVE);
+            } else {
+                regionButtonElement.classList.remove(CLASS_BUTTON_ACTIVE);
+            }
```
ConradJChan: Once the toggleAnnotationMode api is setup with the corresponding event, we may want to consider refactoring this to respond to the event instead so we can remove the local state maintained by AnnotationControls

PR #1190, opened 2020-03-31
fix(controls): Fix controls class overriden by annotations

mxiao6:


src/lib/Controls.scss
```diff
@@ -84,6 +84,21 @@ $controls-button-width: 32px;
         align-items: center;
         justify-content: center;
     }
+
+    .bp-controls-group {
```
jstoffan, line 88: Do any of the other classes in here need increased specificity? What is being overridden?

PR #1192, opened 2020-04-01
feat(annotations): Toggle region annotation mode

mxiao6:


src/lib/AnnotationControls.ts
```diff
@@ -9,31 +12,32 @@ export const CLASS_REGION_BUTTON = 'bp-AnnotationControls-regionBtn';
...
     /**
      * [constructor]
      */
-    constructor(controls: Controls) {
+    constructor(controls: Controls, annotator: EventEmitter) {
         if (!controls || !(controls instanceof Controls)) {
             throw Error('controls must be an instance of Controls');
         }
 
+        this.annotator = annotator;
```
jstoffan, line 40: We should avoid passing around and storing the annotator instance when possible. If it gets destroyed and recreated independently, this will probably break. Can the BaseViewer manager the event handlers and invoke methods on the annotation controls?

src/lib/AnnotationControls.ts
```diff
@@ -46,6 +50,7 @@ export default class AnnotationControls {
     public destroy(): void {
         fullscreen.removeListener('enter', this.handleFullscreenEnter);
         fullscreen.removeListener('exit', this.handleFullscreenExit);
+        this.annotator.removeListener('annotationcreate', this.deactivateCurrentControl);
```
jstoffan, line 53: We're going to need to do several things when this event fires, so having it in a more central place is probably preferable.

src/lib/AnnotationControls.ts
mxiao6: > We should avoid passing around and storing the annotator instance when possible. If it gets destroyed and recreated independently, this will probably break. Can the BaseViewer manage the event handlers and invoke methods on the annotation controls?

Sounds good

src/lib/AnnotationControls.ts
```diff
@@ -84,21 +84,53 @@ export default class AnnotationControls {
...
+
+    /**
+     * Region comment button click handler
+     */
+    private handleRegionClick = (onRegionClick: RegionHandler) => (event: MouseEvent): void => {
+        if (this.currentActiveControl === ANNOTATION_MODE.region) {
+            this.deactivateCurrentControl();
+        } else {
+            if (this.currentActiveControl) {
+                this.deactivateCurrentControl();
```
jstoffan, line 127: It looks like this happens no matter what? Can we simplify and move it to the top of the method, outside of the conditions?

src/lib/AnnotationControls.ts
```diff
@@ -84,21 +84,53 @@ export default class AnnotationControls {
...
+     * Deactive current control button
      */
-    private handleRegionClick = (onRegionClick: RegionHandler) => (event: MouseEvent): void => {
+    public deactivateCurrentControl = (): void => {
+        if (!this.currentActiveControl) {
+            return;
+        }
+
+        switch (this.currentActiveControl) {
+            case ANNOTATION_MODE.region:
```
jstoffan, line 95: It seems like we're planning ahead for more buttons, which is great. Maybe we could create a map for mode type -> method, like so?

```
this.controlsMap = {
    [ANNOTATION_MODE.region]: this.updateRegionButton
}
```

src/lib/AnnotationControls.ts
```diff
@@ -84,21 +84,53 @@ export default class AnnotationControls {
...
+                this.currentActiveControl = null;
```
jstoffan, line 96: It looks like this happens no matter what. Should it be moved outside of the switch?

src/lib/AnnotationControls.ts
```diff
@@ -84,21 +84,53 @@ export default class AnnotationControls {
...
-     * Region comment button click handler
+     * Deactive current control button
      */
-    private handleRegionClick = (onRegionClick: RegionHandler) => (event: MouseEvent): void => {
+    public deactivateCurrentControl = (): void => {
+        if (!this.currentActiveControl) {
+            return;
+        }
+
+        switch (this.currentActiveControl) {
```
ConradJChan, line 94: I wonder if the switch is needed at all. Should it just blindly run through all the `update*Button` methods each time? For instance if the click region and then click highlight, we would need region and highlight to update respectively

src/lib/viewers/doc/DocBaseViewer.js
```diff
@@ -1113,7 +1113,9 @@ class DocBaseViewer extends BaseViewer {
      * @private
      * @return {void}
      */
-    regionClickHandler() {}
+    regionClickHandler({ activeControl }) {
```
ConradJChan, line 1116: Is this needed? Wouldn't this always call with 'region'?

mxiao6: > Is this needed? Wouldn't this always call with 'region'?

It might be `null`. I thought this would simplify togglemode method in box-annotations

src/lib/AnnotationControls.ts
mxiao6: > I wonder if the switch is needed at all. Should it just blindly run through all the `update*Button` methods each time? For instance if the click region and then click highlight, we would need region and highlight to update respectively

This is `deactivateCurrentControl()`, which only deactivate the current control. Update next control will happen in `handle*Click`.

src/lib/AnnotationControls.ts
mxiao6: > It looks like this happens no matter what. Should it be moved outside of the switch?

Yeah, but I need to create another variable to store the original value of it, because the switch is using the its value.

src/lib/AnnotationControls.ts
```diff
@@ -84,21 +93,48 @@ export default class AnnotationControls {
...
                 regionButtonElement.classList.remove(CLASS_BUTTON_ACTIVE);
             }
         }
+    };
+
+    /**
+     * Region comment button click handler
+     */
+    private handleRegionClick = (onRegionClick: RegionHandler) => (event: MouseEvent): void => {
+        const activeControl = this.currentActiveControl;
```
ConradJChan, line 128: The call to this.deactivateCurrentControl() seems unnecessary. It seems like this method's responsibility should be (and all `handle([Region|Highlight|Draw])Click` for that matter):

- update `currentActiveControl`
- update all buttons states accordingly
- callback to Viewer with the type toggled

so maybe something like this
```
this.currentActiveControl = this.currentActiveControl === 'region' ? 'none' : 'region';
this.controlsMap.keys().forEach(updateButton => updateButton());
onModeClick('region');
```
Thoughts?

src/lib/AnnotationControls.ts
```diff
@@ -84,21 +93,48 @@ export default class AnnotationControls {
     private handleFullscreenExit = (): void => this.handleFullscreenChange(false);
 
     /**
-     * Region comment button click handler
+     * Deactive current control button
```
jstoffan, line 96: Typo. Should be `Deactivate`.

src/lib/AnnotationControls.ts
```diff
@@ -84,21 +93,48 @@ export default class AnnotationControls {
...
+        updateButton();
+    };
+
+    /**
+     * Update region button UI
+     */
+    private updateRegionButton = (): void => {
         const regionButtonElement = this.controlsElement.querySelector(`.${CLASS_REGION_BUTTON}`);
 
         if (regionButtonElement) {
```
jstoffan, line 115: Convert to a guard clause to reduce nesting?

PR #1194, opened 2020-04-08
chore(typescript): Fix wrong rootDir in tsconfig

mxiao6:


tsconfig.json
```diff
@@ -1,5 +1,9 @@
 {
     "extends": "@box/frontend/ts/tsconfig.json",
-    "include": ["./src/**/*.ts", "./src/**/*.tsx"],
-    "exclude": ["node_modules", "dist"]
+    "compilerOptions": {
```
jstoffan, line 3: Can you try upgrading `@box/frontend`, first? I think this was fixed in the latest version.

PR #1195, opened 2020-04-08
feat(annotations): Hide annotations when fullscreen is active

mxiao6:


src/lib/viewers/BaseViewer.js
```diff
@@ -545,6 +545,7 @@ class BaseViewer extends EventEmitter {
      */
     handleFullscreenEnter() {
         this.resize();
+        this.emit(ANNOTATOR_EVENT.toggleVisibility);
```
jstoffan, line 548: Should this be `this.annotator.emit`, instead? If not, how does it get into the annotator?

mxiao6: > Should this be `this.annotator.emit`, instead? If not, how does it get into the annotator?

😱

src/lib/viewers/BaseViewer.js
```diff
@@ -545,6 +545,9 @@ class BaseViewer extends EventEmitter {
      */
     handleFullscreenEnter() {
         this.resize();
+        if (this.annotator) {
+            this.annotator.emit(ANNOTATOR_EVENT.toggleVisibility);
```
jstoffan, line 549: This may be a silly question, but if the visibility is toggled off for some other reason, this will turn them back on. Should we consider passing in an explicit boolean value, instead?

mxiao6: > This may be a silly question, but if the visibility is toggled off for some other reason, this will turn them back on. Should we consider passing in an explicit boolean value, instead?

Yeah, toggleMode also has this problem. Should I change them together?

jstoffan: I think the toggleMode method is already present in the 2.x and 3.x version of annotations, which is why it's still a direct method invocation.

ConradJChan:
Update unit tests?

PR #1196, opened 2020-04-13
feat(annotations): Exit annotations mode when enter fullscreen

mxiao6:


src/lib/viewers/BaseViewer.js
```diff
@@ -33,6 +33,7 @@ import {
 } from '../constants';
 import { getIconFromExtension, getIconFromName } from '../icons/icons';
 import { VIEWER_EVENT, ERROR_CODE, LOAD_METRIC, DOWNLOAD_REACHABILITY_METRICS } from '../events';
+import { AnnotationMode } from '../AnnotationControls';
```
jstoffan, line 36: Should we expose an `exit` or `reset` method to avoid exposing the internals of the controls here?

src/lib/viewers/BaseViewer.js
```diff
@@ -545,8 +546,12 @@ class BaseViewer extends EventEmitter {
      */
     handleFullscreenEnter() {
         this.resize();
-        if (this.annotator) {
+
+        if (this.annotator && this.options.showAnontationsControls && this.annotationControls) {
             this.annotator.emit(ANNOTATOR_EVENT.setVisibility, false);
+
+            this.annotator.toggleAnnotationMode(AnnotationMode.NONE);
```
jstoffan, line 553: Does passing nothing here work?

src/lib/viewers/BaseViewer.js
```diff
@@ -545,8 +546,12 @@ class BaseViewer extends EventEmitter {
      */
     handleFullscreenEnter() {
         this.resize();
-        if (this.annotator) {
+
+        if (this.annotator && this.options.showAnontationsControls && this.annotationControls) {
```
jstoffan, line 550: Do we need these checks in `handleFullscreenExit`?

src/lib/viewers/BaseViewer.js
mxiao6: > Does passing nothing here work?

No, it won't work because it sets state to `undefined` instead of `'none'`;

PR #1197, opened 2020-04-14
feat(images): support multi-page jpg reps for tif files

ikerr:


src/lib/viewers/image/ImageLoader.js
```diff
@@ -65,6 +72,31 @@ class ImageLoader extends AssetLoader {
             return viewer.REP === entry.representation;
         });
     }
+
+    /**
+     * @inheritdoc
+     */
+    determineViewer(file, disabledViewers = []) {
+        return this.viewers.find(viewer => {
```
jstoffan, line 80: Can we avoid duplicating the logic from AssetLoader? Can we call `super.determineViewer` and then further filter the list as needed?

src/lib/viewers/image/ImageLoader.js
```diff
@@ -65,6 +72,31 @@ class ImageLoader extends AssetLoader {
...
+            const multiPageViewer = (viewer.ASSET || '').indexOf('{page}') !== -1;
+
+            return file.representations.entries.some(entry => {
+                if (viewer.REP !== entry.representation) {
+                    return false;
+                }
+
+                // If the viewer uses paged assets (e.g., image_2048_jpg), then
+                // ensure that there is a compatible paged representation.
+                const pagedRep = (entry.properties || {}).paged === 'true';
```
jstoffan, line 94: I'm finding the logic here hard to follow. Can we have non-paged TIFF files? What case is this guarding against?

ikerr: @jstoffan I was seeing the following issue... Suppose we have the following reps available:

```
representations: {                                              
  entries: [                                                  
    {                                                       
      representation: 'jpg',                              
      properties: {                                       
        dimensions: '1024x1024',                        
        paged: 'false',                                 
      },                                                  
    },                                                      
    {                                                       
      representation: 'png',                              
      properties: {                                       
        dimensions: '2048x2048',                        
        paged: 'true',
      },                                                  
    },                                                      
  ],                                                          
}
```

When `AssetLoader.determineViewer(file)` is called for a TIFF file, it will see that a "jpg" representation is available (the first entry), and select the following viewer:

```
{                                                                           
  NAME: 'MultiImage',                                                     
  CONSTRUCTOR: MultiImageViewer,                                          
  REP: 'jpg',                                                             
  EXT: ['tif', 'tiff'],                                                   
  ASSET: '{page}.jpg',                                                    
}
```

However, the 1024x1024 jpg rep that is available is not a paged rep. So, when we then call `ImageLoader.determineRepresentation()`, we get back `null/undefined` because it specifically filters out these reps:

```
determineRepresentation(file, viewer) {                                     
  return file.representations.entries.find(entry => {                     
    // Do not use the dimensions=1024x1024&paged=false rep that is for document preloading
    if (entry.properties && entry.properties.paged === 'false') {       
      return false;                                                   
    }                                                                   
                                                                                
    return viewer.REP === entry.representation;                         
   });                                                                     
}
```

This most recent change aims to only use the `MultiImage` viewer (which is currently only used for TIFF files) when a paged representation exists.

src/lib/viewers/image/ImageLoader.js
ikerr: We could probably do something like that if `determineViewer()` returned a set of viewers rather than a single viewer. Then we could call `super.determineViewer()` and then further filter the results. Happy to make that change if you'd like, but it'll increase the scope of this change.

ikerr: Alternatively, I could probably create a new method, `AssetLoader.determineViewerSet()` that returns a set of viewers and then do this:

```
// AssetLoader
determineViewer(file, disabledViewers = []) {
  const viewers = determineViewerSet(file, disabledViewers);
  if (viewers.length > 0) {
    return viewers[0];
  }

  return undefined;
}
```

and then in `ImageLoader`:

```
// ImageLoader
determineViewer(file, disabledViewers = []) {
  const viewers = super.determineViewerSet(file, disabledViewers);
  // perform further filtering...
}
```

PR #1199, opened 2020-04-16
feat(annotation): handle create event

ConradJChan:
Updated the create event to `annotations_create` and added logic to `toggleAnnotationMode` and set the active annotation id in the annotator

Todo:
 - [x] unit tests

### DIFF
--- a/src/lib/AnnotationControls.ts
+++ b/src/lib/AnnotationControls.ts
@@ -95,6 +95,14 @@ export default class AnnotationControls {
     private handleFullscreenExit = (): void => this.handleFullscreenChange(false);
 
     /**
+     * Returns the current active annotation mode
+     * @return {string} Current active annotation mode
+     */
+    public getActiveMode = (): AnnotationMode => {
+        return this.currentActiveControl;
+    };
+
+    /**
      * Deactivate current control button
      */
     public resetControls = (): void => {
--- a/src/lib/AnnotationControls.ts
+++ b/src/lib/AnnotationControls.ts
@@ -94,10 +94,6 @@ export default class AnnotationControls {
      */
     private handleFullscreenExit = (): void => this.handleFullscreenChange(false);
 
-    /**
-     * Returns the current active annotation mode
-     * @return {string} Current active annotation mode
-     */
     public getActiveMode = (): AnnotationMode => {
         return this.currentActiveControl;
     };
--- a/src/lib/viewers/BaseViewer.js
+++ b/src/lib/viewers/BaseViewer.js
@@ -149,6 +149,7 @@ class BaseViewer extends EventEmitter {
         this.mobileZoomChangeHandler = this.mobileZoomChangeHandler.bind(this);
         this.mobileZoomEndHandler = this.mobileZoomEndHandler.bind(this);
         this.handleAnnotatorEvents = this.handleAnnotatorEvents.bind(this);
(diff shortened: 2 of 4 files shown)


src/lib/AnnotationControls.ts
```diff
@@ -94,6 +94,14 @@ export default class AnnotationControls {
      */
     private handleFullscreenExit = (): void => this.handleFullscreenChange(false);
 
+    /**
+     * Returns the current active annotation mode
+     * @return {string} Current active annotation mode
```
mxiao6, line 99: The type is `AnnotationMode` right? And I don't think js doc is needed if we have `AnnotationMode` as return type below.

src/lib/viewers/BaseViewer.js
```diff
@@ -1100,6 +1101,18 @@ class BaseViewer extends EventEmitter {
         this.emit('annotatorevent', data);
     }
 
+    handleAnnotationCreateEvent({ annotation, meta: { status } = {} }) {
+        if (status !== 'pending') {
+            const activeMode = this.annotationControls.getActiveMode();
+            this.annotator.toggleAnnotationMode(activeMode);
```
mxiao6, line 1107: Should we do this inside box-annotations?

src/lib/viewers/BaseViewer.js
```diff
@@ -1100,6 +1101,18 @@ class BaseViewer extends EventEmitter {
...
+            this.annotationControls.resetControls();
+        }
+
+        if (status === 'success') {
+            this.annotator.emit('annotations_active_set', annotation.id);
```
mxiao6, line 1112: You can get `id` directly in the function signature;

jstoffan, line 1112: Is it possible that the user may click on another annotation during the save XHR? Do we want to try to respect that or is it an edge case we're not worried about right now?

src/lib/viewers/BaseViewer.js
```diff
@@ -1100,6 +1101,18 @@ class BaseViewer extends EventEmitter {
         this.emit('annotatorevent', data);
     }
 
+    handleAnnotationCreateEvent({ annotation, meta: { status } = {} }) {
+        if (status !== 'pending') {
```
jstoffan, line 1105: What would the experience be on a failure? Would the unsaved annotation be lost?

src/lib/AnnotationControls.ts
ConradJChan: Sounds good, I'll remove it then

src/lib/viewers/BaseViewer.js
ConradJChan: That's a good question, according to the designs, it looks like the failed annotation remains -- I'll follow up with design

src/lib/viewers/BaseViewer.js
ConradJChan: Given how we're trying to balance box-annotations being a full blown app vs sdk, I think I would lean towards this logic to be determined by the users of the SDK

src/lib/viewers/BaseViewer.js
ConradJChan: I guess it's possible? I think for now we can leave it, I'll follow up with design, my guess is we may want to disable interactivity during the `isPending` state?

jstoffan: Probably just leave it as-is for now after validating with design.

PR #1201, opened 2020-04-29
feat(annotations): Support multiple annotations at a time

mxiao6:


src/lib/viewers/BaseViewer.js
```diff
@@ -1105,9 +1105,6 @@ class BaseViewer extends EventEmitter {
         // Only on success do we exit create annotation mode. If error occurs,
         // we remain in create mode
         if (status === 'success') {
-            const activeMode = this.annotationControls.getActiveMode();
-            this.annotator.toggleAnnotationMode(activeMode);
-            this.annotationControls.resetControls();
             this.annotator.emit('annotations_active_set', id);
```
jstoffan, line 1108: Do we want to remove this, as well? @ConradJChan @mickr

ConradJChan: yeah I think there's a separate ticket, might as well remove this whole handler while we're at it

mickr: If we remove the emit this will affect the setActive state when an annotation is created so we will need to follow up on it.

PR #1202, opened 2020-04-30
feat(annotations): Disable annotations for excel and iWork formats

mxiao6:


src/lib/extensions.js
```diff
@@ -48,8 +48,14 @@ export const CODE_EXTENSIONS = [
     'yaml',
 ];
 
+export const EXCEL_EXTENSIONS = ['xls', 'xlsm', 'xlsx'];
```
jstoffan, line 51: Do we want to include `xlsb`? Should we update OfficeViewer to use parts of these constants?

src/lib/viewers/BaseViewer.js
```diff
@@ -173,6 +174,10 @@ class BaseViewer extends EventEmitter {
             const fileExt = this.options.file.extension;
             this.fileLoadingIcon = getIconFromExtension(fileExt);
             this.startAt = getProp(this.options, `fileOptions.${this.options.file.id}.${FILE_OPTION_START}`, {});
+
+            if (EXCEL_EXTENSIONS.includes(fileExt) || IWORK_EXTENSIONS.includes(fileExt)) {
```
jstoffan, line 178: Seems like this logic should live in the `areAnnotationsEnabled` method.

src/lib/extensions.js
mxiao6: > Do we want to include `xlsb`? Should we update OfficeViewer to use parts of these constants?

`xlsb` conversion to pdf is not supported. But OfficeViewer supports `xlsb` but not `xls`...

src/lib/viewers/BaseViewer.js
```diff
@@ -1030,6 +1031,11 @@ class BaseViewer extends EventEmitter {
             return this.viewerConfig.enabled;
         }
 
+        // Disable new annotations for Excel and iWork formats
+        if (EXCEL_EXTENSIONS.includes(extension) || IWORK_EXTENSIONS.includes(extension)) {
+            this.options.showAnnotationsControls = false;
```
jstoffan, line 1036: Do we want to mutate the options object in here? Seems like we should just `return false`. Do we need a new method for the new experience?

src/lib/extensions.js
ConradJChan: Should we add a comment so this nuance isn't lost?

src/lib/viewers/__tests__/BaseViewer-test.js
```diff
@@ -1254,6 +1254,27 @@ describe('lib/viewers/BaseViewer', () => {
         });
     });
 
+    describe('areNewAnnotationsEnabled()', () => {
+        it('should return false if the file is excel or iWork formats', () => {
```
ConradJChan, line 1258: Should we have a parameterized test that runs over all the excluded extensions? Might make it easier to assert any changes in the future

src/lib/viewers/doc/DocBaseViewer.js
```diff
@@ -1015,7 +1015,7 @@ class DocBaseViewer extends BaseViewer {
         this.controls = new Controls(this.containerEl);
         this.pageControls = new PageControls(this.controls, this.docEl);
         this.zoomControls = new ZoomControls(this.controls);
-        if (this.options.showAnnotationsControls) {
+        if (this.areNewAnnotationsEnabled()) {
             this.annotationControls = new AnnotationControls(this.controls);
```
jstoffan, line 1019: This disables the controls, but doesn't disable the Annotations SDK, so annotations that great created in a different way (platform, API, etc.) will still be shown. Are we concerned about that scenario?

mxiao6: > This disables the controls, but doesn't disable the Annotations SDK, so annotations that great created in a different way (platform, API, etc.) will still be shown. Are we concerned about that scenario?

Yeah, Rohit said we only want to disable new annotations for these extensions.

jstoffan: Does that mean "newly-created" or "new experience"? I suspect that he meant the former. Can you check?

mxiao6: > Does that mean "newly-created" or "new experience"? I suspect that he meant the latter. Can you check?

Oh I realized this difference just now...

PR #1203, opened 2020-05-05
chore(tests): Add data-testid attribute to region toggle button

jstoffan:


src/lib/AnnotationControls.ts
```diff
@@ -150,13 +150,15 @@ export default class AnnotationControls {
...
+        const regionButton = this.controls.add(
             __('region_comment'),
             this.handleClick(onRegionClick, AnnotationMode.REGION),
             `${CLASS_BOX_CONTROLS_GROUP_BUTTON} ${CLASS_REGION_BUTTON}`,
             ICON_REGION_COMMENT,
             'button',
             groupElement,
         );
+
+        regionButton.setAttribute('data-testid', 'bp-AnnotationsControls-regionBtn');
```
mxiao6, line 162: Can use `CLASS_REGION_BUTTON` here?

jstoffan: I intentionally separated them to avoid changes in the class name affecting the test selector.

PR #1206, opened 2020-05-08
feat(annotations): add support for passed annotations options

mickr:
* Add support for annotation options
* Use messages and locale if preset in options and forward them to the annotator to populate messages and locale

### DIFF
--- a/src/lib/i18n.js
+++ b/src/lib/i18n.js
@@ -7,6 +7,10 @@ try {
 
 const language = __LANGUAGE__ || 'en-US'; // eslint-disable-line
 
+const getLocale = lang => {
+    return lang.substr(0, lang.indexOf('-'));
+};
+
 /**
  * Creates Intl object used by annotations
  *
@@ -15,8 +19,9 @@ const language = __LANGUAGE__ || 'en-US'; // eslint-disable-line
  */
 const createAnnotatorIntl = () => {
     return {
-        messages: annotationMessages,
         language,
+        locale: getLocale(language),
+        messages: annotationMessages,
     };
 };
 
--- a/src/lib/viewers/BaseViewer.js
+++ b/src/lib/viewers/BaseViewer.js
@@ -939,11 +939,10 @@ class BaseViewer extends EventEmitter {
         }
 
         const options = boxAnnotations.getOptions && boxAnnotations.getOptions();
-        const { intl } = options || intlUtil.createAnnotatorIntl();
 
         const annotatorOptions = this.createAnnotatorOptions({
             annotator: this.annotatorConf,
-            intl,
+            intl: (options && options.intl) || intlUtil.createAnnotatorIntl(),
             modeButtons: ANNOTATION_BUTTONS,
         });
 
--- a/src/lib/viewers/BaseViewer.js
+++ b/src/lib/viewers/BaseViewer.js
@@ -931,29 +931,22 @@ class BaseViewer extends EventEmitter {
             return;
         }
 
-        const boxAnnotations =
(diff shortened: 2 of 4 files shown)


src/lib/viewers/BaseViewer.js
```diff
@@ -931,17 +931,29 @@ class BaseViewer extends EventEmitter {
...
+        const boxAnnotations =
+            this.options.boxAnnotations || new global.BoxAnnotations(viewerOptions, intl.createAnnotatorIntl());
         this.annotatorConf = boxAnnotations.determineAnnotator(this.options, this.viewerConfig);
 
         if (!this.annotatorConf) {
             return;
         }
 
-        const annotatorOptions = this.createAnnotatorOptions({
+        const options = {
```
jstoffan, line 942:
```
const { intl: intlProvider } = boxAnnotations.getOptions ? boxAnnotations.getOptions() : {};
const annotatorOptions = this.createAnnotatorOptions({
    annotator: this.annotatorConf,
    intl: intlProvider || intl.createAnnotatorIntl(),
    modeButtons: ANNOTATION_BUTTONS
});
```

src/lib/viewers/BaseViewer.js
```diff
@@ -931,17 +931,29 @@ class BaseViewer extends EventEmitter {
...
             annotator: this.annotatorConf,
             modeButtons: ANNOTATION_BUTTONS,
-        });
+        };
+
+        if (this.areNewAnnotationsEnabled() && boxAnnotations.getOptions) {
+            const annotationsOptions = boxAnnotations.getOptions();
+
+            if (annotationsOptions) {
+                const { locale, messages } = annotationsOptions;
```
jstoffan, line 951: This seems a little confusing. Is `language` omitted intentionally? Can we simplify the options logic?

```
const { intl } = boxAnnotations.getOptions ? boxAnnotations.getOptions() : {};
const annotatorOptions = this.createAnnotatorOptions({
    annotator: this.annotatorConf,
    intl: intl || intlUtil.createAnnotatorIntl(),
    modeButtons: ANNOTATION_BUTTONS
});
```

src/lib/viewers/__tests__/BaseViewer-test.js
```diff
@@ -1448,8 +1499,6 @@ describe('lib/viewers/BaseViewer', () => {
             expect(combinedOptions.location).to.deep.equal({ locale: 'en-US' });
             expect(combinedOptions.randomOption).to.equal('derp');
             expect(combinedOptions.localizedStrings).to.not.be.undefined;
-            expect(combinedOptions.intl.language).to.equal('en-US');
```
jstoffan, line 1451: I think we'll want more tests for the final output of the options object.

src/lib/viewers/BaseViewer.js
mickr: Sure, I passed it to the boxAnnotations instantiation on purpose to be explicit how it should be used.  This should be fine as well. Thanks for the input

src/lib/viewers/__tests__/BaseViewer-test.js
mickr: I do check in the `createAnnotator()` test  that this gets called with the correct properties. I have added back the check for language as well.

src/lib/i18n.js
```diff
@@ -7,6 +7,10 @@ try {
 
 const language = __LANGUAGE__ || 'en-US'; // eslint-disable-line
 
+const getLocale = lang => {
+    return lang.substr(0, lang.indexOf('-'));
```
ConradJChan, line 11: Do we need to handle the case where the lang doesn't have a dash?

mickr: I believe all of our supported langs have a dash.

PR #1209, opened 2020-05-12
fix(annotations): Fix annotations control shows if no create permission

mxiao6:


src/lib/viewers/doc/__tests__/DocBaseViewer-test.js
```diff
@@ -2285,6 +2302,9 @@ describe('src/lib/viewers/doc/DocBaseViewer', () => {
                     'bp-exit-fullscreen-icon',
                     ICON_FULLSCREEN_OUT,
                 );
+                expect(docBase.annotationControls.init).to.be.calledWith({
```
jstoffan, line 2305: Do we have a negative test case?

PR #1210, opened 2020-05-14
fix(controls): stop event propagation from controls

mickr:
* Stop event propagation from preview controls, this is interfering with annotations.
* Add new tests for Controls.js constructor

### DIFF
--- a/src/lib/Controls.js
+++ b/src/lib/Controls.js
@@ -44,10 +44,10 @@ class Controls {
         this.controlsEl.addEventListener('mouseleave', this.mouseleaveHandler);
         this.controlsEl.addEventListener('focusin', this.focusinHandler);
         this.controlsEl.addEventListener('focusout', this.focusoutHandler);
+        this.controlsEl.addEventListener('click', this.clickHandler);
 
         if (this.hasTouch) {
             this.containerEl.addEventListener('touchstart', this.mousemoveHandler);
-            this.controlsEl.addEventListener('click', this.clickHandler);
         }
     }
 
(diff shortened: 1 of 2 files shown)


src/lib/Controls.js
```diff
@@ -175,9 +175,15 @@ class Controls {
      * @param {Event} event - A DOM-normalized event object.
      * @return {void}
      */
-    clickHandler = () => {
+    clickHandler = event => {
+        if (event) {
+            event.stopPropagation();
+        }
+
+        if (this.hasTouch) {
```
jstoffan, line 183: Do we want to hide the controls only for touch-enabled devices?

src/lib/Controls.js
```diff
@@ -175,9 +175,15 @@ class Controls {
...
+            this.shouldHide = true;
+        }
         // If we are not focused in on the page num input, allow hiding after timeout
```
jstoffan, line 186: Comment seems to have been orphaned.

src/lib/Controls.js
mickr: Yes, I believe that was the intention of the original code as it didn’t register the click handler unless it was a touch device. Now that we are using it, probably should keep the same intention, while only hiding on mobile devices.

src/lib/Controls.js
```diff
@@ -44,10 +44,10 @@ class Controls {
         this.controlsEl.addEventListener('mouseleave', this.mouseleaveHandler);
         this.controlsEl.addEventListener('focusin', this.focusinHandler);
         this.controlsEl.addEventListener('focusout', this.focusoutHandler);
+        this.controlsEl.addEventListener('click', this.clickHandler);
```
jstoffan, line 47: Should we move the associated `removeEventListener` call out of its conditional check as well?

jstoffan: Also, this component is used in all of the viewers across Preview, so this is a wide-ranging change. Does it have any side effects? Have we tested all the various viewers?

PR #1212, opened 2020-05-18
fix(controls): update region annotation icon

mickr:
Updating to the newest region annotation icon. I increased the size a little bit because the new icon looked a bit small on the toolbar.


jstoffan:
Do we have a separate change for normalizing the padding of the toggle button with the thumbnail sidebar toggle button?

src/lib/AnnotationControls.scss
```diff
@@ -1,18 +1,18 @@
...
-        width: $controls-button-width;
-        height: $controls-button-width;
-        border-radius: 4px;
-
         svg {
             fill: $white;
         }
 
         &.is-active {
+            width: 34px;
```
jstoffan, line 11: Seems like these measurements should remain the same whether `is-active` is applied or not.

src/lib/AnnotationControls.scss
```diff
@@ -1,18 +1,18 @@
...
+            height: 32px;
+            margin: 7px;
```
jstoffan, line 13: Seems kind of magic number-y.

mickr:
Default State
Screen Shot 2020-05-18 at 4 58 42 PM

Active State
Screen Shot 2020-05-18 at 4 58 51 PM

Focused State
Screen Shot 2020-05-18 at 4 59 01 PM

src/lib/AnnotationControls.scss
```diff
@@ -1,23 +1,32 @@
...
             svg {
                 fill: $black;
             }
         }
+
+        &.is-active::before {
+            position: absolute;
+            z-index: -1;
+            width: 34px;
+            height: 32px;
```
mxiao6, line 25: Can we use `$controls-button-width`?

src/lib/AnnotationControls.scss
```diff
@@ -1,23 +1,32 @@
...
+            background-color: $white;
+            border-radius: 4px;
```
mxiao6, line 27: `bdl-border-radius-size`?

src/lib/AnnotationControls.ts
```diff
@@ -153,7 +153,7 @@ export default class AnnotationControls {
         const regionButton = this.controls.add(
             __('region_comment'),
             this.handleClick(onRegionClick, AnnotationMode.REGION),
-            `${CLASS_BOX_CONTROLS_GROUP_BUTTON} ${CLASS_REGION_BUTTON}`,
+            `${CONTROLS_BUTTON_CLASS} ${CLASS_REGION_BUTTON}`,
```
mxiao6, line 156: `CONTROLS_BUTTON_CLASS` is added in `Controls.js`, right? I think in the future, there will be more annotations buttons. Do we want to remove the group class?

src/lib/AnnotationControls.scss
mickr: The design calls for these dimensions... not 32x32

src/lib/AnnotationControls.ts
mickr: I don't think the group class is what we need with the individual buttons.

mxiao6: Is `CONTROLS_BUTTON_CLASS` needed? I believe `this.controls` already has it.

PR #1214, opened 2020-05-28
fix(annotations): handle scrollToAnnotation in Preview

mickr:
* Add handler to handle the scrolltoannotation event.
* Add override to PresentationViewer to setPage since we only show 1 page at a time.

### DIFF
--- a/src/lib/viewers/BaseViewer.js
+++ b/src/lib/viewers/BaseViewer.js
@@ -1025,11 +1025,16 @@ class BaseViewer extends EventEmitter {
 
     /**
      * Handles the 'scrolltoannotation' event and calls the annotator scroll method
-     * @param {Object} data
-     * @param {string} data.id Annotation id
+     * @param {Object} data - Annotator event data
+     * @param {string} [data.id] - Annotation Id
+     * @param {Object} [data.target] - Annotation target
+     * @param {Object} [data.target.location] - Target location of the annotation
+     * @param {number} [data.target.location.value] - location value
+     *
+     * @return {void}
      */
-    handleScrollToAnnotation({ id }) {
-        this.annotator.scrollToAnnotation(id);
+    handleScrollToAnnotation(data) {
+        this.annotator.scrollToAnnotation(data);
     }
 
     /**
--- a/src/lib/viewers/BaseViewer.js
+++ b/src/lib/viewers/BaseViewer.js
@@ -980,7 +980,7 @@ class BaseViewer extends EventEmitter {
         });
 
         // Add a custom listener to scroll to the specified annotation
-        this.addListener('scrolltoannotation', data => this.annotator.scrollToAnnotation(data));
+        this.addListener('scrolltoannotation', this.handleScrollToAnnotation);
 
         // Add a custom listener for events emmited by the annotator
         this.annotator.addListener('annotatorevent', this.handleAnnotatorEvents);
@@ -1021,6 +1021,15 @@ class BaseViewer extends EventEmitter {
             return false;
         }
         return permissions.can_annotate || permissions.can_create_annotations;
+    }
+
+    /**
+     * Handles the 'scrolltoannotation' event and calls the annotator scroll method
+     * @param {Object} data
+     * @param {string} data.id Annotation id
+     */
+    handleScrollToAnnotation({ id }) {
+        this.annotator.scrollToAnnotation(id);
     }
 
     /**
--- a/src/lib/viewers/BaseViewer.js
+++ b/src/lib/viewers/BaseViewer.js
@@ -1026,7 +1026,7 @@ class BaseViewer extends EventEmitter {
     /**
      * Handles the 'scrolltoannotation' event and calls the annotator scroll method
      * @param {string | Object} event - Annotation Event
-     * @param {Object} event.id - Annotation Id
+     * @param {string} event.id - Annotation Id
      * @return {void}
      */
     handleScrollToAnnotation(event) {
--- a/src/lib/viewers/BaseViewer.js
+++ b/src/lib/viewers/BaseViewer.js
@@ -1025,15 +1025,13 @@ class BaseViewer extends EventEmitter {
 
     /**
      * Handles the 'scrolltoannotation' event and calls the annotator scroll method
-     * @param {Object} data - Annotator event data
-     * @param {string} [data.id] - Annotation Id
-     * @param {Object} [data.target] - Annotation target
-     * @param {Object} [data.target.location] - Target location of the annotation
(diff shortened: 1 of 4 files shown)


src/lib/viewers/doc/PresentationViewer.js
```diff
@@ -313,6 +313,19 @@ class PresentationViewer extends DocBaseViewer {
...
     }
+
+    /**
+     * @override
+     * @param {Object} data
+     * @param {string}  data.id - location value of the annotation
+     * @param {Object} data.location - location value of the annotation
+     * @param {number} data.location.value - location value of the annotation
+     */
+    handleScrollToAnnotation({ id, location: { value = 1 } }) {
```
jstoffan, line 324: Will this affect the legacy annotations as well? If so, does it work properly?

src/lib/viewers/doc/PresentationViewer.js
```diff
@@ -313,6 +313,19 @@ class PresentationViewer extends DocBaseViewer {
...
+        this.setPage(value);
+
+        this.annotator.scrollToAnnotation(id);
```
jstoffan, line 327: Can/should this call `super.handleScrollToAnnotation`?

src/lib/viewers/doc/PresentationViewer.js
```diff
@@ -313,6 +313,19 @@ class PresentationViewer extends DocBaseViewer {
             };
         };
     }
+
+    /**
+     * @override
```
jstoffan, line 318: It looks like this is under the `// Private` section of the class, though it's protected/public.

src/lib/viewers/BaseViewer.js
```diff
@@ -1019,6 +1019,15 @@ class BaseViewer extends EventEmitter {
         return permissions.can_annotate || permissions.can_create_annotations;
     }
 
+    /**
+     * Handles the 'scrolltoannotation' event and calls the annotator scroll method
+     * @param {Object} data
+     * @param {string} data.id Annotation id
+     */
+    handleScrollToAnnotation({ id }) {
+        this.annotator.scrollToAnnotation(id);
```
jstoffan, line 1028: Is this the value that's passed to legacy annotations? It looks like the whole `data` argument is passed down.

src/lib/viewers/BaseViewer.js
```diff
@@ -976,7 +976,7 @@ class BaseViewer extends EventEmitter {
         });
 
         // Add a custom listener to scroll to the specified annotation
-        this.addListener('scrolltoannotation', data => this.annotator.scrollToAnnotation(data));
+        this.addListener('scrolltoannotation', this.handleScrollToAnnotation);
```
jstoffan, line 979: Will we update BUIE to emit this event?

src/lib/viewers/BaseViewer.js
mickr, line 1028: Also thinking that if we emit the location information to the scrollToAnnotation we could use that information in box-annotations instead of looking it up from the store.

ConradJChan, line 1028: We've been keeping the new annotations loosely coupled via events instead of explicit function calls, should we do the same here?

src/lib/viewers/BaseViewer.js
ConradJChan, line 979: Who calls this?

src/lib/viewers/doc/PresentationViewer.js
```diff
@@ -313,6 +313,19 @@ class PresentationViewer extends DocBaseViewer {
...
+     * @param {Object} data
+     * @param {string}  data.id - location value of the annotation
```
ConradJChan, line 320: extra space? Also is the description correct?

src/lib/viewers/doc/PresentationViewer.js
ConradJChan, line 327: should this call `super.handleScrollToAnnotation()`?

src/lib/viewers/BaseViewer.js
mickr: Yeah, I am getting the other PR's together.

ConradJChan:
Sorry, some duplicate comments

src/lib/viewers/BaseViewer.js
mickr: Ah, yeah it should be passed down for legacy... I'll update that.

src/lib/viewers/doc/PresentationViewer.js
```diff
@@ -124,6 +124,19 @@ class PresentationViewer extends DocBaseViewer {
...
     }
 
+    /**
+     * @override
+     */
+    handleScrollToAnnotation(data) {
+        const { target } = data;
+
+        if (target && target.location) {
+            this.setPage(target.location.value);
```
jstoffan, line 134: Would lodash/get help here?

src/lib/viewers/BaseViewer.js
mickr: BUIE emits to this event

src/lib/viewers/BaseViewer.js
mickr: This is a legacy event so we should keep it around.

src/lib/viewers/doc/PresentationViewer.js
```diff
@@ -124,6 +125,17 @@ class PresentationViewer extends DocBaseViewer {
         return hasXOverflow || hasYOverflow;
     }
 
+    /**
+     * @override
+     */
+    handleScrollToAnnotation(data) {
+        const location = getProp(data, 'target.location.value', 0);
```
jstoffan, line 132: Should the default value be `1` or `undefined`?

src/lib/viewers/doc/PresentationViewer.js
```diff
@@ -124,6 +125,17 @@ class PresentationViewer extends DocBaseViewer {
...
+
+        this.setPage(location);
```
jstoffan, line 134: Seems like the value of `location` could be inlined.

src/lib/viewers/BaseViewer.js
```diff
@@ -1023,6 +1023,18 @@ class BaseViewer extends EventEmitter {
         return permissions.can_annotate || permissions.can_create_annotations;
     }
 
+    /**
+     * Handles the 'scrolltoannotation' event and calls the annotator scroll method
+     * @param {string | Object} event - Annotation Event
+     * @param {Object} event.id - Annotation Id
```
jstoffan, line 1029: Duplicate parameter doc?

PR #1219, opened 2020-06-03
feat(annotations): Emit annotator before init

mxiao6:


src/lib/viewers/BaseViewer.js
```diff
@@ -952,6 +952,10 @@ class BaseViewer extends EventEmitter {
 
         this.annotator = new this.annotatorConf.CONSTRUCTOR(annotatorOptions);
 
+        // Once the annotator instance has been created, emit it so that clients can attach their events.
+        // Annotator object will still be sent along with the viewer in the load event also.
+        this.emit('annotator', this.annotator);
```
jstoffan, line 957: This seems a little risky, since this event is used by legacy annotations. Did you test the older versions against this change?

mickr: This will also affect the work of loading an annotation on hard load, but I can also create a new event that gets emitted AFTER the annotations and event listeners have been added.

mxiao6: I checked out v2.3.0 and didn't find any usage of this event. The only occurrence of this event is in a markdown file.

mxiao6: > This will also affect the work of loading an annotation on hard load, but I can also create a new event that gets emitted AFTER the annotations and event listeners have been added.

Sorry for the inconvenience. The api call happens in the constructor which is before the init. So the `annotator` event has to be emitted before init so that the sidebar can listen to the api call.

jstoffan: Can we create a new "created" event instead of relocating an existing event?

mxiao6: > Can we create a new "created" event instead of relocating an existing event?

Sounds good

PR #1220, opened 2020-06-04
fix(annotations): scroll to annotation on load

mickr:
* Added `setpage` event that will allow setting the page in PresentationViewer through an event.

* Added `annotatations_loaded` event that will be emitted at the end of the `initAnnotations` method execution. This will enable listeners to know when it is safe to emit the scrolltoannotation event on a hard load.

- [ ] tests

### DIFF
--- a/src/lib/viewers/BaseViewer.js
+++ b/src/lib/viewers/BaseViewer.js
@@ -988,6 +988,9 @@ class BaseViewer extends EventEmitter {
         if (this.areNewAnnotationsEnabled() && this.annotationControls) {
             this.annotator.addListener('annotations_create', this.handleAnnotationCreateEvent);
         }
+
+        // Signal that the annotator and listeners have been loaded.
+        this.emit('annotations_loaded');
     }
 
     /**
--- a/src/lib/viewers/doc/PresentationViewer.js
+++ b/src/lib/viewers/doc/PresentationViewer.js
@@ -2,7 +2,6 @@ import throttle from 'lodash/throttle';
 import DocBaseViewer from './DocBaseViewer';
 import PresentationPreloader from './PresentationPreloader';
 import { CLASS_INVISIBLE } from '../../constants';
-import { getProp } from '../../util';
 import './Presentation.scss';
 
 const WHEEL_THROTTLE = 200;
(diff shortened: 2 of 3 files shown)


src/lib/viewers/BaseViewer.js
```diff
@@ -988,6 +988,9 @@ class BaseViewer extends EventEmitter {
         if (this.areNewAnnotationsEnabled() && this.annotationControls) {
             this.annotator.addListener('annotations_create', this.handleAnnotationCreateEvent);
         }
+
+        // Signal that the annotator and listeners have been loaded.
+        this.emit('annotations_loaded');
```
ConradJChan, line 993: Is this an event that box-annotations itself should emit?

jstoffan, line 993: Should this be `annotator_loaded` to follow the existing event's convention somewhat. Would `annotator_ready` work, since the Annotations SDK is loaded much earlier?

src/lib/viewers/doc/PresentationViewer.js
```diff
@@ -125,15 +129,6 @@ class PresentationViewer extends DocBaseViewer {
         return hasXOverflow || hasYOverflow;
     }
 
-    /**
-     * @override
-     */
-    handleScrollToAnnotation(data) {
```
jstoffan, line 131: This is a breaking change and require another joint release with BUIE. We'll need to start considering the effect of cross-library changes like this as we move into release.

src/lib/viewers/doc/PresentationViewer.js
```diff
@@ -42,6 +41,9 @@ class PresentationViewer extends DocBaseViewer {
         // Set up preloader
         this.preloader = new PresentationPreloader(this.previewUI, { api: this.api });
         this.preloader.addListener('preload', this.onPreload.bind(this));
+
+        // listen for set page events
+        this.addListener('setpage', this.setPage);
```
jstoffan, line 46: Can we just call this method on the viewer itself? Or proxy it in Preview.js?

src/lib/viewers/BaseViewer.js
mickr: Yeah, that is much better!

src/lib/viewers/doc/PresentationViewer.js
mickr: In the current implementation, I am emitting this from `box-annotations` with the page from the annotation location.

PR #1223, opened 2020-06-10
fix(annotations): Scroll to presentation page only when necessary

jstoffan:
This change prevents the scroll position from being reset if the user has zoomed in and selects multiple annotations on the same page.

### DIFF
--- a/src/lib/viewers/doc/PresentationViewer.js
+++ b/src/lib/viewers/doc/PresentationViewer.js
@@ -129,7 +129,11 @@ class PresentationViewer extends DocBaseViewer {
      * @override
      */
     handleScrollToAnnotation(data) {
-        this.setPage(getProp(data, 'target.location.value', 1));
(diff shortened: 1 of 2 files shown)


src/lib/viewers/doc/PresentationViewer.js
```diff
@@ -129,7 +129,11 @@ class PresentationViewer extends DocBaseViewer {
      * @override
      */
     handleScrollToAnnotation(data) {
-        this.setPage(getProp(data, 'target.location.value', 1));
+        const targetPage = getProp(data, 'target.location.value', 1);
+
+        if (this.pdfViewer.currentPageNumber !== targetPage) {
```
ConradJChan, line 134: Do we need to check if pdfViewer exists?

jstoffan: We don't really do that anywhere else, including `setPage`. Seems like overkill?

PR #1224, opened 2020-06-11
feat(annotations): scroll to annotation on load

mickr:
* Add handler that will look for a set annotations activeId in the file options and if present, scroll to annotation.

### DIFF
--- a/src/lib/viewers/BaseViewer.js
+++ b/src/lib/viewers/BaseViewer.js
@@ -989,6 +989,7 @@ class BaseViewer extends EventEmitter {
 
         if (this.areNewAnnotationsEnabled() && this.annotationControls) {
             this.annotator.addListener('annotations_create', this.handleAnnotationCreateEvent);
+            this.handleAnnotationsOnLoad();
         }
     }
 
@@ -1036,6 +1037,19 @@ class BaseViewer extends EventEmitter {
 
         this.annotator.scrollToAnnotation(data);
     }
+
+    handleAnnotationsOnLoad = () => {
+        const {
+            file: { id },
+        } = this.options;
+
+        const loadedAnnotationId = getProp(this.options, `fileOptions.${id}.annotations.activeId`, null);
+        const annotation = loadedAnnotationId ? this.annotator.getAnnotationById(loadedAnnotationId) : null;
+
+        if (annotation) {
+            this.handleScrollToAnnotation(annotation);
+        }
+    };
 
     /**
      * Returns whether or not annotations are enabled for this viewer.
--- a/src/lib/viewers/__tests__/BaseViewer-test.js
+++ b/src/lib/viewers/__tests__/BaseViewer-test.js
@@ -1296,6 +1296,49 @@ describe('lib/viewers/BaseViewer', () => {
         });
     });
 
+    describe('handleAnnotationsOnLoad()', () => {
+        let getAnnotationStub;
+        let scrollToAnnotationStub;
+
+        beforeEach(() => {
+            getAnnotationStub = sandbox.stub().returns({ id: 'ABC' });
+            scrollToAnnotationStub = sandbox.stub();
+
+            base.annotator = {
+                init: sandbox.stub(),
+                scrollToAnnotation: scrollToAnnotationStub,
+                getAnnotationById: getAnnotationStub,
+            };
+        });
+
+        it('should not call handleScrollToAnnotation if there is not an active annotation', () => {
+            base.options.fileOptions = {
+                '0': {
+                    annotations: {},
+                },
+            };
+
+            base.handleAnnotationsOnLoad();
+
+            expect(scrollToAnnotationStub).not.to.be.called;
+            expect(getAnnotationStub).not.to.be.called;
+        });
+        it('should call scroll to annotation if active annotation is set', () => {
+            base.options.fileOptions = {
+                '0': {
+                    annotations: {
+                        activeId: 'ABC',
+                    },
+                },
+            };
+
+            base.handleAnnotationsOnLoad();
+
+            expect(scrollToAnnotationStub).to.be.calledWith('ABC');
+            expect(getAnnotationStub).to.be.called;
+        });
+    });
+
     describe('areAnnotationsEnabled()', () => {
         beforeEach(() => {
             stubs.getViewerOption = sandbox


src/lib/viewers/BaseViewer.js
```diff
@@ -989,6 +989,7 @@ class BaseViewer extends EventEmitter {
 
         if (this.areNewAnnotationsEnabled() && this.annotationControls) {
             this.annotator.addListener('annotations_create', this.handleAnnotationCreateEvent);
+            this.handleAnnotationsOnLoad();
```
ConradJChan, line 992: Can we guarantee that the annotations will have loaded by this time? Should we do something event based like an `annotations_fetched` event?

src/lib/viewers/BaseViewer.js
```diff
@@ -1037,6 +1038,19 @@ class BaseViewer extends EventEmitter {
         this.annotator.scrollToAnnotation(data);
     }
 
+    handleAnnotationsOnLoad = () => {
+        const {
+            file: { id },
+        } = this.options;
+
+        const loadedAnnotationId = getProp(this.options, `fileOptions.${id}.annotations.activeId`, null);
```
ConradJChan, line 1046: Should we just return if `loadedAnnotationId` is falsy?

src/lib/viewers/BaseViewer.js
jstoffan, line 992: This seems likely to lead to a race condition, since we are assuming the fetch call from box-annotations has completed by this point. Should we emit a `annotations_fetch_success` event from box-annotations with a payload of the fetched annotations? Then we can pull the annotation we want out of that payload and scroll to it?

src/lib/viewers/BaseViewer.js
```diff
@@ -1037,6 +1038,19 @@ class BaseViewer extends EventEmitter {
...
+        const annotation = loadedAnnotationId ? this.annotator.getAnnotationById(loadedAnnotationId) : null;
```
jstoffan, line 1047: We should avoid calling methods on the annotator when possible, especially when they involve the internal state of the library.

src/lib/viewers/BaseViewer.js
mickr: Conrad eluded to this as well, I think that is a good approach as well.

src/lib/viewers/BaseViewer.js
```diff
@@ -966,6 +966,9 @@ class BaseViewer extends EventEmitter {
      * @return {void}
      */
     initAnnotations() {
+        // Annotations must be loaded and ready before scrolling can occur on a deep-linked annotation.
+        this.annotator.addListener('annotations_ready', this.handleAnnotationsOnLoad);
```
ConradJChan, line 970: I think @mxiao6 is adding a `annotations_fetch_error` event, would `annotations_fetch_success` make more sense?

src/lib/viewers/BaseViewer.js
```diff
@@ -1038,17 +1040,23 @@ class BaseViewer extends EventEmitter {
...
         const loadedAnnotationId = getProp(this.options, `fileOptions.${id}.annotations.activeId`, null);
-        const annotation = loadedAnnotationId ? this.annotator.getAnnotationById(loadedAnnotationId) : null;
 
-        if (annotation) {
-            this.handleScrollToAnnotation(annotation);
+        if (!loadedAnnotationId) {
+            return;
         }
+
+        this.handleScrollToAnnotation(data.find(entry => entry.id === loadedAnnotationId));
```
ConradJChan, line 1059: Should we make sure an annotation is found before calling into `this.handleScrollToAnnotation`? Otherwise it looks like we call into the `this.annotator.scrollToAnnotation` with `undefined`

src/lib/viewers/BaseViewer.js
jstoffan: Can we keep our terminology consistent? We're using `init`, `ready`, and `load` interchangeably.

src/lib/viewers/BaseViewer.js
```diff
@@ -1037,6 +1040,25 @@ class BaseViewer extends EventEmitter {
...
+     * @return {void}
+     */
+    handleAnnotationsOnLoad = (data = []) => {
+        const {
+            file: { id },
+        } = this.options;
+
+        const loadedAnnotationId = getProp(this.options, `fileOptions.${id}.annotations.activeId`, null);
+
+        if (!loadedAnnotationId) {
```
jstoffan, line 1055: What does `loaded` mean in this case? Should it be `activeAnnotationId`?

mickr: Loaded meaning it was the activeAnnotationId on load... but I can re-name it.

src/lib/viewers/BaseViewer.js
```diff
@@ -1037,6 +1040,32 @@ class BaseViewer extends EventEmitter {
         this.annotator.scrollToAnnotation(data);
     }
 
+    /**
+     * Handles retrieving the active annotation id from a deep-linked annotation and scrolls to the annotation.
+     * @param {Object} event - annotations array from emitted from box-annotations.
+     * @param {Array} event.annotations - annotations array from emitted from box-annotations.
+     * @return {void}
+     */
+    handleAnnotationsOnLoad = ({ annotations = [] }) => {
```
ConradJChan, line 1049: Rename to `handleAnnotationsInitialized`?

src/lib/viewers/BaseViewer.js
```diff
@@ -1037,6 +1040,32 @@ class BaseViewer extends EventEmitter {
...
+        const {
```
ConradJChan, line 1050: Maybe use getProp to extract the fileId too?

src/lib/viewers/__tests__/BaseViewer-test.js
```diff
@@ -1296,6 +1300,44 @@ describe('lib/viewers/BaseViewer', () => {
         });
     });
 
+    describe('handleAnnotationsOnLoad()', () => {
```
ConradJChan, line 1303: Add test for if specified active annotation id is not found in the list of annotations?

src/lib/viewers/BaseViewer.js
```diff
@@ -1037,6 +1040,32 @@ class BaseViewer extends EventEmitter {
...
+
+        const activeAnnotationId = getProp(this.options, `fileOptions.${id}.annotations.activeId`, null);
+
+        if (!activeAnnotationId) {
+            return;
+        }
+
+        const annotation = annotations.find(entry => entry.id === activeAnnotationId);
+
+        if (!annotation) {
```
jstoffan, line 1062: Do these need to be separate guard clauses? If `activeAnnotationId` is undefined, it won't match any results in the `.find` call above.

src/lib/viewers/BaseViewer.js
```diff
@@ -1037,6 +1040,32 @@ class BaseViewer extends EventEmitter {
...
+     * @param {Object} event - annotations array from emitted from box-annotations.
+     * @param {Array} event.annotations - annotations array from emitted from box-annotations.
+     * @return {void}
+     */
+    handleAnnotationsOnLoad = ({ annotations = [] }) => {
+        const {
+            file: { id },
+        } = this.options;
+
+        const activeAnnotationId = getProp(this.options, `fileOptions.${id}.annotations.activeId`, null);
```
jstoffan, line 1054: Is the `null` value meaningful here or does `undefined` serve the same purpose?

src/lib/viewers/BaseViewer.js
mickr: Agreed, but the feedback from Conrad was to add the guard which I didn't really see a problem with.

jstoffan: Sure, keep the guard, just replace the `annotationId` check with it.

src/lib/viewers/BaseViewer.js
```diff
@@ -1037,6 +1040,27 @@ class BaseViewer extends EventEmitter {
         this.annotator.scrollToAnnotation(data);
     }
 
+    /**
+     * Handles retrieving the active annotation id from a deep-linked annotation and scrolls to the annotation.
+     * @param {Object} event - annotations array from emitted from box-annotations.
+     * @param {Array} event.annotations - annotations array from emitted from box-annotations.
+     * @return {void}
+     */
+    handleAnnotationsInitialized = ({ annotations = [] }) => {
```
jstoffan, line 1049: Is `event` always defined? Should we guard against it being null or undefined?

mickr: For now it should always be defined from annotations.

PR #1226, opened 2020-06-16
feat(annotations): Press Esc to exit annotations mode

mxiao6:
- [x] Tests

### DIFF
--- a/src/lib/AnnotationControls.ts
+++ b/src/lib/AnnotationControls.ts
@@ -17,6 +17,7 @@ export enum AnnotationMode {
 export type ClickHandler = ({ activeControl, event }: { activeControl: AnnotationMode; event: MouseEvent }) => void;
 export type Options = {
     onRegionClick?: ClickHandler;
+    onReset?: () => void;
 };
 
 declare const __: (key: string) => string;
@@ -33,6 +34,10 @@ export default class AnnotationControls {
     private controlsMap: ControlsMap;
 
     private currentActiveControl: AnnotationMode = AnnotationMode.NONE;
+
+    private hasInit = false;
+
+    private onReset: () => void = noop;
 
     /**
      * [constructor]
@@ -55,8 +60,12 @@ export default class AnnotationControls {
      * [destructor]
      */
     public destroy(): void {
+        if (!this.hasInit) {
+            return;
+        }
         fullscreen.removeListener('enter', this.handleFullscreenEnter);
         fullscreen.removeListener('exit', this.handleFullscreenExit);
+        document.removeEventListener('keydown', this.handleKeyDown);
     }
 
     /**
@@ -110,6 +119,8 @@ export default class AnnotationControls {
 
         this.currentActiveControl = AnnotationMode.NONE;
         updateButton();
+
+        this.onReset();
     };
 
     /**
@@ -146,9 +157,27 @@ export default class AnnotationControls {
     };
 
     /**
+     * Escape key handler, reset all control buttons,
+     * and stop propagation to prevent preview modal from exiting
+     */
+    private handleKeyDown = (event: KeyboardEvent): void => {
+        if (event.key !== 'Escape' || this.currentActiveControl === AnnotationMode.NONE) {
+            return;
+        }
+
+        this.resetControls();
+
+        event.preventDefault();
+        event.stopPropagation();
+    };
+
+    /**
      * Initialize the annotation controls with options.
      */
-    public init({ onRegionClick = noop }: Options = {}): void {
+    public init({ onRegionClick = noop, onReset = noop }: Options = {}): void {
+        if (this.hasInit) {
+            return;
+        }
         const groupElement = this.controls.addGroup(CLASS_ANNOTATIONS_GROUP);
         const regionButton = this.controls.add(
             __('region_comment'),
@@ -160,5 +189,10 @@ export default class AnnotationControls {
         );
 
         regionButton.setAttribute('data-testid', 'bp-AnnotationsControls-regionBtn');
+
+        this.onReset = onReset;
+        document.addEventListener('keydown', this.handleKeyDown);
+
+        this.hasInit = true;
     }
 }
--- a/src/lib/AnnotationControls.ts
+++ b/src/lib/AnnotationControls.ts
@@ -66,6 +66,8 @@ export default class AnnotationControls {
         fullscreen.removeListener('enter', this.handleFullscreenEnter);
         fullscreen.removeListener('exit', this.handleFullscreenExit);
         document.removeEventListener('keydown', this.handleKeyDown);
+
+        this.hasInit = false;
     }
 
     /**
--- a/src/lib/viewers/BaseViewer.js
+++ b/src/lib/viewers/BaseViewer.js
@@ -151,6 +151,7 @@ class BaseViewer extends EventEmitter {
         this.mobileZoomEndHandler = this.mobileZoomEndHandler.bind(this);
         this.handleAnnotatorEvents = this.handleAnnotatorEvents.bind(this);
         this.handleAnnotationCreateEvent = this.handleAnnotationCreateEvent.bind(this);
+        this.handleAnnotationControlsReset = this.handleAnnotationControlsReset.bind(this);
         this.handleFullscreenEnter = this.handleFullscreenEnter.bind(this);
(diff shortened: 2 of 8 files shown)


src/lib/AnnotationControls.ts
```diff
@@ -59,6 +59,10 @@ export default class AnnotationControls {
         fullscreen.removeListener('exit', this.handleFullscreenExit);
     }
 
+    public focus(): void {
+        this.controlsElement.querySelector(`.${CLASS_REGION_BUTTON}`).focus();
```
jstoffan, line 63: What happens if we have more than one button? Or if this button isn't available for a given viewer?

src/lib/viewers/BaseViewer.js
```diff
@@ -1176,6 +1178,12 @@ class BaseViewer extends EventEmitter {
         }
     }
 
+    handleAnnotationStagedEvent(staged) {
+        if (!staged && this.annotationControls) {
```
jstoffan, line 1182: Let's check for the presence of `this.annotationControls` first.

src/lib/viewers/BaseViewer.js
```diff
@@ -1256,6 +1264,18 @@ class BaseViewer extends EventEmitter {
...
+
+    /**
+     * Method to exit annotation mode
+     *
+     * @return {void}
+     */
+    exitAnnotationMode = () => {
+        if (this.areAnnotationsEnabled() && this.annotator && this.annotationControls) {
+            this.annotator.toggleAnnotationMode(AnnotationMode.NONE);
+            this.annotationControls.resetControls();
```
jstoffan, line 1276: I'm not sure I understand the logical flow. Can we just call `this.annotatonControls.focus()` as the last line here?

src/lib/viewers/doc/DocBaseViewer.js
```diff
@@ -593,6 +593,9 @@ class DocBaseViewer extends BaseViewer {
             case ']':
                 this.nextPage();
                 break;
+            case 'Escape':
+                this.exitAnnotationMode();
```
jstoffan, line 597: Should this handler be managed by the AnnotationControls component, instead? Otherwise we'll have to reimplement it in each viewer.

src/lib/viewers/BaseViewer.js
```diff
@@ -1176,6 +1178,12 @@ class BaseViewer extends EventEmitter {
         }
     }
 
+    handleAnnotationStagedEvent({ staged }) {
+        if (this.annotationControls && !staged) {
+            this.annotationControls.focus();
```
jstoffan, line 1183: What is the desired user experience here?

src/lib/AnnotationControls.ts
```diff
@@ -160,5 +172,20 @@ export default class AnnotationControls {
         );
 
         regionButton.setAttribute('data-testid', 'bp-AnnotationsControls-regionBtn');
+        regionButton.onkeydown = (event: KeyboardEvent): void => {
```
jstoffan, line 175: We should be using `addEventListener` rather than raw attributes. Should we be listening at the document level?

src/lib/AnnotationControls.ts
```diff
@@ -160,5 +172,20 @@ export default class AnnotationControls {
...
+            if (!annotator) {
+                return;
+            }
+
+            if (event.key !== 'Escape') {
```
jstoffan, line 180: Combine guards when possible.

src/lib/AnnotationControls.ts
```diff
@@ -160,5 +172,20 @@ export default class AnnotationControls {
...
+                return;
+            }
+
+            this.resetControls();
+            annotator.toggleAnnotationMode(AnnotationMode.NONE);
```
jstoffan, line 185: Can `annotator` be undefined here? Should the controls instead provide a callback to the parent, like `onReset`, which the parent can use to reset the mode?

src/lib/AnnotationControls.ts
```diff
@@ -59,6 +63,14 @@ export default class AnnotationControls {
         fullscreen.removeListener('exit', this.handleFullscreenExit);
     }
 
+    public focus(): void {
+        const buttonClass = buttonClassMap[this.currentActiveControl];
```
jstoffan, line 67: We already have `this.controlsMap` for handlers. Can we keep this consistent?

src/lib/AnnotationControls.ts
```diff
@@ -25,6 +25,10 @@ interface ControlsMap {
     [key: string]: () => void;
 }
 
+const buttonClassMap = {
+    region: CLASS_REGION_BUTTON,
```
jstoffan, line 29: Should this use the mode enum?

src/lib/viewers/BaseViewer.js
mxiao6: > What is the desired user experience here?

I was listening `keydown` event on `regionButton`, so I have to focus that button after `cancel` etc. However, listening at the document level solves this problem. I'm going to remove all the focus code.

src/lib/AnnotationControls.ts
```diff
@@ -145,10 +149,22 @@ export default class AnnotationControls {
...
+    private handleKeyDown = (onReset: () => void) => (event: KeyboardEvent): void => {
+        if (event.key !== 'Escape' || this.currentActiveControl === AnnotationMode.NONE) {
+            return;
+        }
+
+        this.resetControls();
+        onReset();
+
+        event.preventDefault();
+        event.stopPropagation();
```
jstoffan, line 161: This seems kind of risky on a document-level handler where other non-Preview listeners might live. What is the reasoning for it?

mxiao6: Stop exiting preview modal.

src/lib/AnnotationControls.ts
```diff
@@ -145,10 +149,22 @@ export default class AnnotationControls {
         onClick({ activeControl: this.currentActiveControl, event });
     };
 
+    private handleKeyDown = (onReset: () => void) => (event: KeyboardEvent): void => {
+        if (event.key !== 'Escape' || this.currentActiveControl === AnnotationMode.NONE) {
+            return;
+        }
+
+        this.resetControls();
+        onReset();
```
jstoffan, line 158: Should this occur within `this.resetControls`?

src/lib/viewers/doc/DocBaseViewer.js
```diff
@@ -1098,7 +1098,10 @@ class DocBaseViewer extends BaseViewer {
         this.controls.add(__('exit_fullscreen'), this.toggleFullscreen, 'bp-exit-fullscreen-icon', ICON_FULLSCREEN_OUT);
 
         if (this.areNewAnnotationsEnabled() && this.hasAnnotationCreatePermission()) {
-            this.annotationControls.init({ onRegionClick: this.handleRegionClick });
+            this.annotationControls.init({
+                onRegionClick: this.handleRegionClick,
+                onReset: this.handleAnnotationControlsReset,
```
jstoffan, line 1103: Do we need to add this to ImageViewer, as well?

src/lib/viewers/image/ImageViewer.js
```diff
@@ -288,7 +288,10 @@ class ImageViewer extends ImageBaseViewer {
 
         if (this.areNewAnnotationsEnabled() && this.hasAnnotationCreatePermission()) {
             this.annotationControls = new AnnotationControls(this.controls);
-            this.annotationControls.init({ onRegionClick: this.handleRegionClick });
+            this.annotationControls.init({
+                onRegionClick: this.handleRegionClick,
+                onReset: this.handleAnnotationControlsReset,
```
jstoffan, line 293: Test?

src/lib/AnnotationControls.ts
```diff
@@ -55,8 +60,12 @@ export default class AnnotationControls {
      * [destructor]
      */
     public destroy(): void {
+        if (!this.hasInit) {
+            return;
+        }
         fullscreen.removeListener('enter', this.handleFullscreenEnter);
         fullscreen.removeListener('exit', this.handleFullscreenExit);
+        document.removeEventListener('keydown', this.handleKeyDown);
```
jstoffan, line 68: Should we set `this.hasInit` back to false?

jstoffan:
Seems worthwhile to add lifecycle tests for multiple init/destroy. Can be done in a follow-up, though.

PR #1230, opened 2020-06-29
chore(resin): Add annotations resin tags

mxiao6:
image

### DIFF
--- a/src/lib/AnnotationControls.ts
+++ b/src/lib/AnnotationControls.ts
@@ -16,7 +16,7 @@ export enum AnnotationMode {
 }
 export type ClickHandler = ({ event }: { event: MouseEvent }) => void;
 export type Options = {
-    fileId?: string;
+    fileId: string;
     onEscape?: () => void;
     onRegionClick?: ClickHandler;
 };
@@ -172,7 +172,7 @@ export default class AnnotationControls {
     /**
      * Initialize the annotation controls with options.
      */
-    public init({ onEscape = noop, onRegionClick = noop, fileId = '' }: Options = {}): void {
+    public init({ fileId, onEscape = noop, onRegionClick = noop }: Options): void {
         if (this.hasInit) {
             return;
         }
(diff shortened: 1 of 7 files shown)


src/lib/viewers/doc/DocBaseViewer.js
```diff
@@ -1099,6 +1099,7 @@ class DocBaseViewer extends BaseViewer {
 
         if (this.areNewAnnotationsEnabled() && this.hasAnnotationCreatePermission()) {
             this.annotationControls.init({
+                fileId: this.options.file.id,
```
jstoffan, line 1102: Is it possible for this value to change after `init` has been called?

src/lib/AnnotationControls.ts
```diff
@@ -171,7 +172,7 @@ export default class AnnotationControls {
     /**
      * Initialize the annotation controls with options.
      */
-    public init({ onEscape = noop, onRegionClick = noop }: Options = {}): void {
+    public init({ onEscape = noop, onRegionClick = noop, fileId = '' }: Options = {}): void {
```
jstoffan, line 175: Is an empty string an appropriate default value?

src/lib/AnnotationControls.ts
```diff
@@ -185,6 +186,9 @@ export default class AnnotationControls {
             groupElement,
         );
 
+        groupElement.setAttribute('data-resin-feature', 'annotations');
+        regionButton.setAttribute('data-resin-target', 'highlightRegion');
+        regionButton.setAttribute('data-resin-file_id', fileId);
```
jstoffan, line 191: Should this be `fileId` or `fileID`?

src/lib/viewers/doc/DocBaseViewer.js
mxiao6: If file id changes, the preview will be destroyed and re-init everything, right?

jstoffan: Is that right?

src/lib/AnnotationControls.ts
```diff
@@ -185,6 +186,12 @@ export default class AnnotationControls {
             groupElement,
         );
 
+        groupElement.setAttribute('data-resin-feature', 'annotations');
+        regionButton.setAttribute('data-resin-target', 'highlightRegion');
+        if (fileId) {
```
jstoffan, line 191: Why is this an optional parameter? What scenario would lead this value to be undefined/null? Is that scenario handled?

PR #1232, opened 2020-07-01
fix(image): Disallow creation of annotations on rotated images

jstoffan:


src/lib/viewers/image/ImageViewer.js
```diff
@@ -128,6 +128,13 @@ class ImageViewer extends ImageBaseViewer {
         this.imageEl.style.transform = `rotate(${this.currentRotationAngle}deg)`;
         this.emit('rotate');
 
+        // Disallow creating annotations on rotated images
+        if (this.currentRotationAngle === 0) {
```
ConradJChan, line 132: TIL: -0 === 0 is true

PR #1234, opened 2020-07-10
chore(log): Add log for image zoom click

mxiao6:


src/lib/viewers/image/ImageBaseViewer.js
```diff
@@ -378,6 +378,8 @@ class ImageBaseViewer extends BaseViewer {
                 // click mouse up. In that case reset the image size, mimicking single-click-unzoom.
                 this.zoom('reset');
             }
+
+            this.emitMetric(VIEWER_EVENT.imageZoomClick, this.options.file.id);
```
ConradJChan, line 382: Would it be more useful to send the `in` or `reset` value since file id can be extracted in other ways?

mxiao6:
> I'm not sure how useful of a viewer event this would be from an SDK perspective. Is this way preferred over using resin tags to track activity?

Discussed offline. Summary: I first tried adding resin tag to the `imageEl` which has `mouseup` event listener. However resin does’t work in that way. I looked into the resin source code and found the element with resin target tag must be `button` `li` etc. Therefore I decided to use event to log.

src/lib/viewers/image/ImageBaseViewer.js
```diff
@@ -373,10 +373,12 @@ class ImageBaseViewer extends BaseViewer {
             if (!this.isPannable && this.isZoomable) {
                 // If the mouse up was not due to panning, and the image is zoomable, then zoom in.
                 this.zoom('in');
+                this.emitMetric(VIEWER_EVENT.imageZoomClick, 'in');
```
ConradJChan, line 376: It seems like we already have a `zoom` event emitted from the viewers -- could we enhance that to include an `interactionType` or something?

src/lib/events.js
```diff
@@ -3,6 +3,7 @@ export const VIEWER_EVENT = {
     default: 'viewerevent', // The default viewer event.
     download: 'download', // Begin downloading the file.
     error: 'error', // When an error occurs.
+    imageZoomClick: 'imagezoomclick', // When image is clicked directly to zoom/reset
```
jstoffan, line 6: Baking the file type into the event name doesn't seem very scalable. It should be available as a separate property on the event.

src/lib/viewers/image/ImageBaseViewer.js
mxiao6: `zoom`  event emitted from viewers doesn't get logged. Adding listener to viewer to specifically log `zoom` event in `Preview.js` adds a lot of complications to this PR. Instead, I just changed the event name to `zoom` for scalability and consistence.

src/lib/viewers/image/ImageBaseViewer.js
```diff
@@ -373,10 +373,12 @@ class ImageBaseViewer extends BaseViewer {
             if (!this.isPannable && this.isZoomable) {
                 // If the mouse up was not due to panning, and the image is zoomable, then zoom in.
                 this.zoom('in');
+                this.emitMetric('zoom', 'imageZoomInClick');
             } else if (!this.didPan) {
                 // If the mouse up was not due to ending of panning, then assume it was a regular
                 // click mouse up. In that case reset the image size, mimicking single-click-unzoom.
                 this.zoom('reset');
+                this.emitMetric('zoom', 'imageZoomResetClick');
```
jstoffan, line 381: Again, we should avoid baking the viewer name into the event. The file and/or viewer type should already be available as a separate property.

src/lib/viewers/image/ImageBaseViewer.js
jstoffan: In what way does it add complexity?

mxiao6: > In what way does it add complexity?

So now it's a two-lines change. If we use `emit('zoom')`, we need to add `zoom` event handler in `attachViewerListeners` in `Preview.js`, which sounds more complex than two lines.

src/lib/viewers/image/ImageBaseViewer.js
```diff
@@ -373,10 +373,12 @@ class ImageBaseViewer extends BaseViewer {
             if (!this.isPannable && this.isZoomable) {
                 // If the mouse up was not due to panning, and the image is zoomable, then zoom in.
                 this.zoom('in');
+                this.emitMetric('zoom', 'inClick');
```
jstoffan, line 376: It looks like `emitMetric` takes a string for the first parameter and a `data` object for the second. What event payload does this actually create? Can we discuss the approach in more detail offline?

jstoffan: Looks like `data` can be anything, including a string.

mxiao6: Yeah, it'll be `JSON.stringify`

PR #1236, opened 2020-07-24
feat(annotations): Add Highlight Text button in toolbar

mxiao6:
- [x] Add tests

image


src/i18n/en-US.properties
```diff
@@ -12,6 +12,8 @@ enter_fullscreen=Enter fullscreen
 exit_fullscreen=Exit fullscreen
 # Button tooltip for annotation region comment
 region_comment=Comment on Region
+# Button tooltip for annotation highlight text
+highlight_text=Text Highlighter
```
jstoffan, line 16: It seems like the copy here should be written as a directive rather than a description.

Region: Comment on Region
Highlight: Highlight text

src/lib/AnnotationControls.scss
```diff
@@ -2,7 +2,8 @@
     padding-left: 4px;
     border-left: 1px solid $twos;
 
-    .bp-AnnotationControls-regionBtn {
+    .bp-AnnotationControls-regionBtn,
```
jstoffan, line 5: Should we add a shared class to both buttons?

src/lib/AnnotationControls.ts
```diff
@@ -147,27 +147,60 @@ export default class AnnotationControls {
...
     };
 
+    private addButton = (mode: AnnotationMode, handler: ClickHandler, parent: HTMLElement, fileId: string): void => {
+        let icon;
+        let resinTarget;
+        let testid;
+        let text;
+
+        if (mode === AnnotationMode.REGION) {
+            icon = ICON_REGION_COMMENT;
```
jstoffan, line 157: Maybe add a map with an object for each mode?

PR #1241, opened 2020-08-18
feat(annotations): Support annotations features option

ConradJChan:


src/lib/Preview.js
```diff
@@ -987,6 +986,12 @@ class Preview extends EventEmitter {
         // Add the response interceptor to the preview instance
         this.options.responseInterceptor = options.responseInterceptor;
 
+        this.options.annotationsOptions = merge(options.annotationsOptions, {
```
mxiao6, line 989: Should we combine `this.options. showAnnotationsControls` into the new `this.options.annotationsOptions`?

jstoffan, line 989: Can we update the host app to pass in the `features` option, instead of `showAnnotationsHighlightText`?

src/lib/viewers/BaseViewer.js
```diff
@@ -1294,7 +1295,8 @@ class BaseViewer extends EventEmitter {
             whoHighlighted: __('annotation_who_highlighted'),
         };
         return cloneDeep({
-            ...this.options,
+            ...restOptions,
+            features,
```
jstoffan, line 1299: This seems a little awkward. What is the reason for having it between the two spread operations?

src/lib/Preview.js
jstoffan: I think `showAnnotationsControls` is preview-specific and shouldn't be passed on to the annotations instance. It would also increase the risk of the change.

src/lib/viewers/BaseViewer.js
ConradJChan: The reason is I wanted to make the options object flatter -- without the destructuring of `features` the annotator would get an options object such as:
```
{
  annotationsOptions: {
    features: {
       ...
    },
    ...
}
```

src/lib/Preview.js
ConradJChan: Yeah, I figured that seemed specific to the preview controls and not really an option for the annotations sdk

ConradJChan: Sure, I was trying to be backwards compatible, but we can just make a clean transition over... across 3 codebases 😄

jstoffan: Okay. We have to make the change in all three anyway, don't we?

PR #1242, opened 2020-08-19
feat(pdf): Upgrade pdf.js to v2.5.207 and closure compiler to v20200719

jstoffan:


jstoffan:
Unfortunately, we won't be able to upgrade to this version of pdf.js due to a known issue that prevents errors from being serialized properly in the minified library. The maintainers merged a commit a few of days ago to address the [issue](https://github.com/mozilla/pdf.js/issues/12209), but it hasn’t been released yet. We can revisit this once it is.

PR #1245, opened 2020-08-28
feat(doc): Updated text selection color

ConradJChan:
Updated text selection background-color to box blue with 10% opacity

Before:
Screen Shot 2020-08-28 at 10 37 14 AM

After:
Screen Shot 2020-08-28 at 11 24 14 AM

### DIFF
--- a/src/lib/viewers/doc/DocFindBar.scss
+++ b/src/lib/viewers/doc/DocFindBar.scss
@@ -1,6 +1,3 @@
-$pdfjs-highlight: #b400aa !default;
-$pdfjs-highlight-selected: #006400 !default;
-
 .bp-find-bar {
     $bdlGray10: #e8e8e8;
 
@@ -118,18 +115,5 @@ $pdfjs-highlight-selected: #006400 !default;
 
     .bp-doc-find-controls {
         display: flex;
-    }
-}
-
-.bp-doc {
-    .textLayer {
-        .highlight {
-            display: inline-block;
-            background-color: fade-out($pdfjs-highlight, .8);
-
-            &.selected {
-                background-color: fade-out($pdfjs-highlight-selected, .8);
-            }
-        }
     }
 }
--- a/src/lib/viewers/doc/DocFindBar.scss
+++ b/src/lib/viewers/doc/DocFindBar.scss
@@ -1,6 +1,5 @@
-$medium-aquamarine: #5aeda8 !default;
-$moon-yellow: #ffba1a !default;
-$blue: #00f !default;
+$pdfjs-highlight: #b400aa !default;
+$pdfjs-highlight-selected: #006400 !default;
 
 .bp-find-bar {
     $bdlGray10: #e8e8e8;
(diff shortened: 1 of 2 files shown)


src/lib/viewers/doc/DocFindBar.scss
```diff
@@ -122,19 +121,15 @@ $blue: #00f !default;
     }
 }
 
-.textLayer {
-    opacity: 1;
+.bp-doc {
+    .textLayer {
+        .highlight {
```
jstoffan, line 126: Does this affect embedded PDF annotations, as well?

src/lib/viewers/doc/_docBase.scss
```diff
@@ -252,14 +252,15 @@ $thumbnail-sidebar-width: 226px;
...
         bottom: auto;
+        opacity: 1;
 
         // Only allow text divs to be selected
         > div {
             user-select: text;
         }
 
-        .highlight {
-            display: inline-block;
```
jstoffan, line 262: This seems like kind of a risky change. Do we understand all the use cases for this class?

src/lib/viewers/doc/_docBase.scss
```diff
@@ -252,14 +252,15 @@ $thumbnail-sidebar-width: 226px;
     .textLayer {
         top: 15px; // Match 15px padding top on page
         bottom: auto;
+        opacity: 1;
```
jstoffan, line 255: Why is this needed?

src/lib/viewers/doc/_docBase.scss
ConradJChan: I just moved it from _docBase.scss to L127 of DocFindBar.scss

src/lib/viewers/doc/_docBase.scss
ConradJChan: This is needed because pdfjs styles it with `opacity: .2` so all colors get .2 transparency. But the colors we'll be applying need more control than that

src/lib/viewers/doc/_docBase.scss
jstoffan: It looks like we reduced its specificity in the process, though. Given the high usage of these components, can we limit the changes as much as possible to reduce risk?

src/lib/viewers/doc/_docBase.scss
jstoffan: I suspect this value is set by pdfjs as a default for a reason, though. Does it have any consequences on the overall user experience? It looks like the test document is mostly text with few images.

src/lib/viewers/doc/DocFindBar.scss
ConradJChan: Those seem to live on their own layer `annotationLayer`

src/lib/viewers/doc/_docBase.scss
ConradJChan: Looking at https://github.com/mozilla/pdfjs-dist/blob/c17c20fb126fcb284811616662156022bbde1947/web/pdf_viewer.css, the only elements of the .textLayer that is affected by this `opacity` was the find bar's highlights which is what I tried to account for below

src/lib/viewers/doc/DocFindBar.scss
ConradJChan: I played around with a few documents with annotations (drawings & links) and things still seemed to be working well

src/lib/viewers/doc/_docBase.scss
jstoffan: Gotcha. So long as the text layer itself doesn't have a background color or something, it should be okay.

PR #1246, opened 2020-08-31
feat(annotations): Handle promote event

mxiao6:


mxiao6:
Will use staged change event instead

PR #1247, opened 2020-09-03
feat(annotations): Handle annotations staged change event

mxiao6:
- [x] Annotations PR: https://github.com/box/box-annotations/pull/570
- [x] Unit tests

image

### DIFF
--- a/src/lib/AnnotationControls.ts
+++ b/src/lib/AnnotationControls.ts
@@ -91,8 +91,6 @@ export default class AnnotationControls {
         this.hasInit = false;
     }
 
-    public getMode = (): AnnotationMode => this.currentMode;
-
     /**
      * Deactivate current control button
      */
@@ -157,21 +155,6 @@ export default class AnnotationControls {
     }
 
     /**
-     * Annotation control button click handler
-     */
-    private handleClick = (onClick: ClickHandler, mode: AnnotationMode) => (event: MouseEvent): void => {
-        const prevMode = this.currentMode;
-        this.resetControls();
-
-        if (prevMode !== mode) {
-            this.currentMode = mode as AnnotationMode;
-            this.updateButton(mode);
-        }
-
-        onClick({ event, mode: this.currentMode });
-    };
-
-    /**
      * Escape key handler, reset all control buttons,
      * and stop propagation to prevent preview modal from exiting
      */
@@ -187,7 +170,7 @@ export default class AnnotationControls {
         event.stopPropagation();
     };
 
-    private addButton = (mode: AnnotationMode, handler: ClickHandler, parent: HTMLElement, fileId: string): void => {
+    private addButton = (mode: AnnotationMode, onClick: ClickHandler, parent: HTMLElement, fileId: string): void => {
         const buttonProps = buttonPropsMap[mode];
 
         if (!buttonProps) {
@@ -196,7 +179,7 @@ export default class AnnotationControls {
 
         const buttonElement = this.controls.add(
             buttonProps.text,
-            this.handleClick(handler, mode),
+            (event: MouseEvent) => onClick({ event, mode }),
             buttonProps.classname,
             buttonProps.icon,
             'button',
--- a/src/lib/AnnotationControls.ts
+++ b/src/lib/AnnotationControls.ts
@@ -16,12 +16,11 @@ export enum AnnotationMode {
     NONE = 'none',
     REGION = 'region',
 }
-export type ClickHandler = ({ event }: { event: MouseEvent }) => void;
+export type ClickHandler = ({ event, mode }: { event: MouseEvent; mode: AnnotationMode }) => void;
 export type Options = {
     fileId: string;
+    onClick?: ClickHandler;
     onEscape?: () => void;
-    onHighlightClick?: ClickHandler;
-    onRegionClick?: ClickHandler;
     showHighlightText: boolean;
 };
 
@@ -91,6 +90,8 @@ export default class AnnotationControls {
 
         this.hasInit = false;
     }
+
+    public getMode = (): AnnotationMode => this.currentMode;
 
     /**
      * Deactivate current control button
@@ -167,7 +168,7 @@ export default class AnnotationControls {
             this.updateButton(mode);
         }
 
-        onClick({ event });
+        onClick({ event, mode: this.currentMode });
     };
 
     /**
@@ -210,22 +211,16 @@ export default class AnnotationControls {
     /**
      * Initialize the annotation controls with options.
      */
-    public init({
-        fileId,
-        onEscape = noop,
-        onRegionClick = noop,
-        onHighlightClick = noop,
-        showHighlightText = false,
-    }: Options): void {
+    public init({ fileId, onEscape = noop, onClick = noop, showHighlightText = false }: Options): void {
         if (this.hasInit) {
             return;
         }
         const groupElement = this.controls.addGroup(CLASS_ANNOTATIONS_GROUP);
         groupElement.setAttribute('data-resin-feature', 'annotations');
 
-        this.addButton(AnnotationMode.REGION, onRegionClick, groupElement, fileId);
+        this.addButton(AnnotationMode.REGION, onClick, groupElement, fileId);
         if (showHighlightText) {
-            this.addButton(AnnotationMode.HIGHLIGHT, onHighlightClick, groupElement, fileId);
+            this.addButton(AnnotationMode.HIGHLIGHT, onClick, groupElement, fileId);
         }
 
         this.onEscape = onEscape;
--- a/src/lib/AnnotationControlsFSM.ts
+++ b/src/lib/AnnotationControlsFSM.ts
@@ -30,13 +30,13 @@ export const stateModeMap = {
     [AnnotationState.REGION_TEMP]: AnnotationMode.REGION,
 };
 
-export default class AnnotationFSM {
+export default class AnnotationControlsFSM {
     private currentState = AnnotationState.NONE;
 
-    public send = (input: AnnotationInput, mode: AnnotationMode): AnnotationMode => {
+    public transition = (input: AnnotationInput, mode: AnnotationMode): AnnotationMode => {
         if (input === AnnotationInput.CLICK) {
-            this.currentState = modeStateMap[mode];
-            return mode;
+            this.currentState = mode === stateModeMap[this.currentState] ? AnnotationState.NONE : modeStateMap[mode];
+            return stateModeMap[this.currentState];
         }
 
         switch (this.currentState) {
@@ -46,7 +46,6 @@ export default class AnnotationFSM {
                         mode === AnnotationMode.HIGHLIGHT
                             ? AnnotationState.HIGHLIGHT_TEMP
                             : AnnotationState.REGION_TEMP;
-                    return mode;
                 }
                 break;
             case AnnotationState.HIGHLIGHT_TEMP:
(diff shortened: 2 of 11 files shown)


src/lib/AnnotationControls.ts
```diff
@@ -153,6 +154,16 @@ export default class AnnotationControls {
         this.resetControls();
         this.currentMode = mode;
         this.updateButton(mode);
+
+        this.isModeTemp = mode !== AnnotationMode.NONE;
```
jstoffan, line 158: It seems like this should use a separate setter that accepts a boolean and is not based on `mode` at all.

src/lib/AnnotationControls.ts
```diff
@@ -65,6 +64,8 @@ export default class AnnotationControls {
 
     private hasInit = false;
 
+    private isModeTemp = false;
```
jstoffan, line 67: Should this be used to control whether the button is marked as "active' or not? What does this actually do?

src/lib/viewers/BaseViewer.js
```diff
@@ -1259,12 +1249,21 @@ class BaseViewer extends EventEmitter {
...
 
-    handleAnnotationModeChangeEvent({ mode }) {
-        if (this.annotationControls) {
-            this.annotationControls.setMode(mode);
+    handleAnnotationStagedChangeEvent({ status, type }) {
+        if (!this.annotationControls) {
+            return;
+        }
+        if (!status || status === 'cancel') {
+            this.annotationControls.resetTempMode();
```
jstoffan, line 1264: Should the controls component no longer be managing the "current mode" state at all? Only which (if any) control is active?

src/lib/AnnotationControls.ts
mxiao6: I changed the name to `isTemporary` and it controls whether the current active control is temporary or not. If it's temporary, when we receive "annotations creation success" or "staged cancel", we reset the active control to none.

src/lib/AnnotationControls.ts
```diff
@@ -167,7 +168,7 @@ export default class AnnotationControls {
             this.updateButton(mode);
         }
 
-        onClick({ event });
+        onClick({ event, mode: this.currentMode });
```
ConradJChan, line 171: Could we make `AnnotationControls` even dumber by removing `handleClick` and just calling `onClick` whenever one of the annotation buttons are clicked? Then give the responsibility to `BaseViewer` to get the new mode from the `AnnotationFSM` and call back in `this.annotationControls.setMode(...)`?

It would make the flow of data more consistent for `AnnotationControls`

src/lib/AnnotationFSM.ts
```diff
@@ -0,0 +1,63 @@
...
+
+export const stateModeMap = {
+    [AnnotationState.HIGHLIGHT]: AnnotationMode.HIGHLIGHT,
+    [AnnotationState.HIGHLIGHT_TEMP]: AnnotationMode.HIGHLIGHT,
+    [AnnotationState.NONE]: AnnotationMode.NONE,
+    [AnnotationState.REGION]: AnnotationMode.REGION,
+    [AnnotationState.REGION_TEMP]: AnnotationMode.REGION,
+};
+
+export default class AnnotationFSM {
```
ConradJChan, line 33: Should this be `AnnotationControlsFSM`?

src/lib/AnnotationFSM.ts
```diff
@@ -0,0 +1,63 @@
...
+    private currentState = AnnotationState.NONE;
+
+    public send = (input: AnnotationInput, mode: AnnotationMode): AnnotationMode => {
```
ConradJChan, line 36: Should this be `transition`?

src/lib/AnnotationFSM.ts
```diff
@@ -0,0 +1,63 @@
...
+        if (input === AnnotationInput.CLICK) {
+            this.currentState = modeStateMap[mode];
+            return mode;
```
ConradJChan, line 39: Should this method always return `stateModeMap[this.currentState]`?

src/lib/AnnotationFSM.ts
```diff
@@ -0,0 +1,63 @@
...
+    [AnnotationState.NONE]: AnnotationMode.NONE,
+    [AnnotationState.REGION]: AnnotationMode.REGION,
+    [AnnotationState.REGION_TEMP]: AnnotationMode.REGION,
+};
+
+export default class AnnotationFSM {
+    private currentState = AnnotationState.NONE;
+
+    public send = (input: AnnotationInput, mode: AnnotationMode): AnnotationMode => {
+        if (input === AnnotationInput.CLICK) {
```
ConradJChan, line 37: Will this transition from `highlight|region_temp` back to `none` if `CLICK` is the input? Also, from 'highlight|region' back to `none` if `CLICK` is the input?

mxiao6: Yes, if `CLICK` is the input, and `none` is the mode.

mxiao6: But if we make `AnnotationControls` dumber as you described in your above comment, mode will never be  `none` and the current FSM implementation won't work.

src/lib/AnnotationControlsFSM.ts
```diff
@@ -0,0 +1,62 @@
...
+        if (input === AnnotationInput.CLICK) {
+            this.currentState = mode === stateModeMap[this.currentState] ? AnnotationState.NONE : modeStateMap[mode];
+            return stateModeMap[this.currentState];
+        }
+
+        switch (this.currentState) {
+            case AnnotationState.NONE:
+                if (input === AnnotationInput.CREATE) {
+                    this.currentState =
+                        mode === AnnotationMode.HIGHLIGHT
```
ConradJChan, line 46: We may want to consider that at some point drawing staged events may be emitted in which case this would break

src/lib/viewers/BaseViewer.js
```diff
@@ -1257,7 +1254,21 @@ class BaseViewer extends EventEmitter {
         // we remain in create mode
         if (status === 'success') {
             this.annotator.emit('annotations_active_set', id);
+
+            if (this.annotationControls) {
+                this.annotationControls.setMode(
+                    this.annotationControlsFSM.transition(AnnotationInput.SUCCESS, AnnotationMode.NONE),
```
ConradJChan, line 1260: Is the second argument of `transition` required for any of the staged change event transitions?

ConradJChan:
Might be worth writing the unit tests for the fsm so that we can make sure all the transitions are covered

src/lib/AnnotationControlsFSM.ts
```diff
@@ -33,7 +39,7 @@ export const stateModeMap = {
 export default class AnnotationControlsFSM {
     private currentState = AnnotationState.NONE;
 
-    public transition = (input: AnnotationInput, mode: AnnotationMode): AnnotationMode => {
+    public transition = (input: AnnotationInput, mode: AnnotationMode = AnnotationMode.NONE): AnnotationMode => {
```
ConradJChan, line 42: Having mode default to `none` seems ripe for bugs to be introduced. Would it work if it was just optionally undefined?

ConradJChan (commented):
Does the `type` still need to be passed into the `transition` call here: https://github.com/box/box-content-preview/pull/1247/commits/61187930c57039e9f84dcc621b541d4d209f88d2#diff-c7f4ce178012f15dc209092ae0c317f9R1269?

mxiao6:
> Does the `type` still need to be passed into the `transition` call here: [6118793#diff-c7f4ce178012f15dc209092ae0c317f9R1269](https://github.com/box/box-content-preview/commit/61187930c57039e9f84dcc621b541d4d209f88d2#diff-c7f4ce178012f15dc209092ae0c317f9R1269)?

`type` is needed to determine next state is `HIGHLIGHT_TEMP` or `REGION_TEMP`.

src/lib/AnnotationControlsFSM.ts
mxiao6: The state in FSM cannot be `undefined`, so if mode is `undefined` and state transition depends on mode, the state has to be `none` which is the same as default mode to `none`.

src/lib/AnnotationControlsFSM.ts
```diff
@@ -39,6 +39,12 @@ export const stateModeMap = {
 export default class AnnotationControlsFSM {
     private currentState = AnnotationState.NONE;
 
+    public getState = (): AnnotationState => this.currentState;
+
+    public setState = (state: AnnotationState): void => {
```
ConradJChan, line 44: Maybe instead of exposing a setState, could we have an `initialState` in the constructor?

src/lib/__tests__/AnnotationControlsFSM-test.js
```diff
@@ -0,0 +1,90 @@
...
+let annotationControlsFSM;
+
+describe('lib/AnnotationControlsFSM', () => {
+    beforeEach(() => {
+        annotationControlsFSM = new AnnotationControlsFSM();
+    });
+
+    [
+        {
+            state: AnnotationState.NONE,
```
ConradJChan, line 13: Is this all the possible state changes? Maybe consider a test for each state and test the transition (if any) based on all possible inputs?

src/lib/__tests__/AnnotationControlsFSM-test.js
```diff
@@ -0,0 +1,90 @@
...
+            input: AnnotationInput.CLICK,
+            mode: AnnotationMode.HIGHLIGHT,
+            nextState: AnnotationState.HIGHLIGHT,
```
ConradJChan, line 16: Does `nextState` and `output` ever differ?

mxiao6: `nextState` could be `highlight_temp` and output is `highlight`

src/lib/__tests__/AnnotationControlsFSM-test.js
mxiao6: The FSM graph is symmetry, so I think it's enough to only test upper/lower half of the states and transitions.

mxiao6: It makes sense to group tests by state. Will do.

src/lib/__tests__/AnnotationControlsFSM-test.js
```diff
@@ -1,86 +1,134 @@
...
+                    });
+                },
+            );
+
+            // Click input
+            [AnnotationMode.HIGHLIGHT, AnnotationMode.REGION].forEach(mode => {
+                it(`should go to state none/${mode} if input is click and mode is ${mode}`, () => {
+                    const annotationControlsFSM = new AnnotationControlsFSM(state);
+
+                    const output = state === mode ? AnnotationMode.NONE : mode;
```
ConradJChan, line 69: We should avoid conditional logic in the tests, could we include it in the input parameters for the test?

src/lib/__tests__/AnnotationControlsFSM-test.js
```diff
@@ -1,86 +1,134 @@
...
+
+                    expect(annotationControlsFSM.transition(AnnotationInput.CLICK, mode)).to.equal(output);
+                    expect(annotationControlsFSM.getState()).to.equal(output);
+                });
+            });
+        });
+    });
+
+    describe('AnnotationState.HIGHLIGHT_TEMP/REGION_TEMP', () => {
```
ConradJChan, line 78: Seems like for readability, it might make sense to just keep these combined tests separate, also L50

ConradJChan (approved):

PR #1249, opened 2020-09-08
feat(perf): Add support for document viewer performance options

jstoffan:
Todo
- [x] Unit tests

### DIFF
--- a/src/lib/viewers/doc/DocBaseViewer.js
+++ b/src/lib/viewers/doc/DocBaseViewer.js
@@ -665,16 +665,16 @@ class DocBaseViewer extends BaseViewer {
         const disableFontFace = Browser.hasFontIssue() || this.getViewerOption('disableFontFace');
 
         // Disable range requests for files smaller than minimum range request size
-        const isRangeSupported = size >= this.getViewerOption('rangeMinSize', RANGE_REQUEST_MINIMUM_SIZE);
+        const isRangeSupported = size >= (this.getViewerOption('rangeMinSize') || RANGE_REQUEST_MINIMUM_SIZE);
         const isWatermarked = watermarkInfo && watermarkInfo.is_watermarked;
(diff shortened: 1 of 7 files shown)


src/lib/viewers/doc/DocBaseViewer.js
```diff
@@ -1202,6 +1209,18 @@ class DocBaseViewer extends BaseViewer {
                 this.resize();
             }
         }
+
+        // Fire rendered metric to indicate that the specific page of content the user requested has been shown
+        if (!this.startPageRendered && (this.startPageNum === pageNumber || this.getCachedPage() === pageNumber)) {
+            const pageRenderTag = Timer.createTag(this.options.file.id, RENDER_METRIC);
+            const pageRenderTimer = Timer.stop(pageRenderTag) || {};
```
ConradJChan, line 1216: Should we still emit the event even if `pageRenderTimer` is an empty object?

src/lib/viewers/doc/DocBaseViewer.js
```diff
@@ -1202,6 +1209,18 @@ class DocBaseViewer extends BaseViewer {
...
+
+            this.emitMetric({
+                name: RENDER_EVENT,
+                data: pageRenderTimer.elapsed,
+            });
+            this.startPageRendered = true;
```
ConradJChan, line 1222: Should this be declared at the top as a property of the class?

src/lib/viewers/doc/DocBaseViewer.js
jstoffan: Fair enough, I can add a guard clause against it.

src/lib/viewers/doc/DocBaseViewer.js
jstoffan: We don't do that for any of the other instance properties in this class, save one.

ConradJChan (approved):

PR #1250, opened 2020-09-08
feat(annotations): Remove mode concept from AnnotationControls

mxiao6:


src/lib/viewers/BaseViewer.js
```diff
@@ -906,7 +906,7 @@ class BaseViewer extends EventEmitter {
 
     disableAnnotationControls() {
         if (this.annotator && this.annotationControls && this.areNewAnnotationsEnabled()) {
-            this.annotator.toggleAnnotationMode(AnnotationMode.NONE);
+            this.annotator.toggleAnnotationMode(AnnotationType.NONE);
```
jstoffan, line 909: Seems like this change kind of breaks down once it moves outside of the controls, since Mode is more appropriate for the Annotations SDK.

src/lib/AnnotationControls.ts
```diff
@@ -11,7 +11,7 @@ export const CLASS_REGION_BUTTON = 'bp-AnnotationControls-regionBtn';
 export const CLASS_BUTTON_ACTIVE = 'is-active';
 export const CLASS_GROUP_HIDE = 'is-hidden';
 
-export enum AnnotationMode {
+export enum AnnotationType {
     HIGHLIGHT = 'highlight',
     NONE = 'none',
```
jstoffan, line 16: If this is truly unrelated to Mode, this enum value should no longer be needed. The active control would just be undefined or null.

PR #1253, opened 2020-09-11
feat(annotations): Don't show highlight button if no download permission

ConradJChan:


src/lib/viewers/doc/DocBaseViewer.js
```diff
@@ -1105,11 +1105,13 @@ class DocBaseViewer extends BaseViewer {
         this.controls.add(__('exit_fullscreen'), this.toggleFullscreen, 'bp-exit-fullscreen-icon', ICON_FULLSCREEN_OUT);
 
         if (this.areNewAnnotationsEnabled() && this.hasAnnotationCreatePermission()) {
+            const canDownload = checkPermission(this.options.file, PERMISSION_DOWNLOAD);
```
jstoffan, line 1108: Should we use the `canDownload` util from lib/file.js?

src/lib/viewers/doc/__tests__/DocBaseViewer-test.js
```diff
@@ -2380,6 +2381,18 @@ describe('src/lib/viewers/doc/DocBaseViewer', () => {
                 }),
             );
 
+            it('should not showHighlightText if file has no download permission', () => {
+                stubs.checkPermission.withArgs(docBase.options.file, PERMISSION_DOWNLOAD).returns(false);
```
jstoffan, line 2385: Any way we can avoid using `withArgs` here? Will make converting to Jest (someday) easier.

src/lib/viewers/doc/DocBaseViewer.js
ConradJChan: Seems like `canDownload` from `lib/file.js` checks a lot more than just the download permission related to actually being able to download the file. Might be too restrictive?

src/lib/viewers/doc/__tests__/DocBaseViewer-test.js
ConradJChan: I looked into this but it turns out `stubs.checkPermission` is already being used in the other tests

src/lib/viewers/doc/DocBaseViewer.js
jstoffan: Yeah, fair point. It also looks like we only check the permission for `textLayerMode`. Should be good to go here.

jstoffan (approved):

PR #1255, opened 2020-09-14
feat(annotations): Set region as default mode if discover FF is on

mxiao6:
- [x] Annotations Constructor change: https://github.com/box/box-annotations/pull/591
- [x] Unit test

### DIFF
--- a/src/lib/viewers/BaseViewer.js
+++ b/src/lib/viewers/BaseViewer.js
@@ -972,6 +972,7 @@ class BaseViewer extends EventEmitter {
         const annotatorOptions = this.createAnnotatorOptions({
             annotator: this.annotatorConf,
             features: options && options.features,
+            initialMode: this.options.enableAnnotationsDiscoverability ? AnnotationMode.REGION : AnnotationMode.NONE,
             intl: (options && options.intl) || intlUtil.createAnnotatorIntl(),
             modeButtons: ANNOTATION_BUTTONS,
         });
@@ -1082,7 +1083,11 @@ class BaseViewer extends EventEmitter {
      */
     handleAnnotationControlsClick({ mode }) {
         const nextMode = this.annotationControlsFSM.transition(AnnotationInput.CLICK, mode);
-        this.annotator.toggleAnnotationMode(nextMode);
+        this.annotator.toggleAnnotationMode(
+            this.options.enableAnnotationsDiscoverability && nextMode === AnnotationMode.NONE
+                ? AnnotationMode.REGION
+                : nextMode,
+        );
         this.annotationControls.setMode(nextMode);
     }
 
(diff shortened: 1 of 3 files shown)


src/lib/viewers/BaseViewer.js
```diff
@@ -1081,7 +1085,11 @@ class BaseViewer extends EventEmitter {
      */
     handleAnnotationControlsClick({ mode }) {
         const nextMode = this.annotationControlsFSM.transition(AnnotationInput.CLICK, mode);
-        this.annotator.toggleAnnotationMode(nextMode);
+        this.annotator.toggleAnnotationMode(
```
ConradJChan, line 1088: Would it make more sense to somehow have this discoverability FF live in the FSM?

src/lib/viewers/BaseViewer.js
```diff
@@ -997,6 +997,10 @@ class BaseViewer extends EventEmitter {
 
         this.annotator.init(this.scale);
 
+        if (this.options.enableAnnotationsDiscoverability) {
+            this.annotator.toggleAnnotationMode(AnnotationMode.REGION);
+        }
```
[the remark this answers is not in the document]
jstoffan: The `annotator.init` method gets called every time the `scale` event fires. We'd probably want to pass in such an option via the constructor, instead.

src/lib/viewers/BaseViewer.js
mxiao6: I first want to put it in FSM too. But later I find we cannot do that because of the line below. The annotations control still needs to set mode to `none` instead of `region`.

src/lib/viewers/__tests__/BaseViewer-test.js
```diff
@@ -1113,6 +1113,7 @@ describe('lib/viewers/BaseViewer', () => {
     describe('createAnnotator()', () => {
         const annotatorMock = {};
         const annotationsOptions = {
+            initialMode: 'none',
```
ConradJChan, line 1116: Should these modes use the enum?

ConradJChan (approved):

src/lib/viewers/__tests__/BaseViewer-test.js
mxiao6: Yeah I realize this when addressing comments in another PR lol

src/lib/viewers/__tests__/BaseViewer-test.js
```diff
@@ -1203,6 +1204,20 @@ describe('lib/viewers/BaseViewer', () => {
...
+            sandbox.stub(base, 'createAnnotatorOptions');
+
+            base.options.boxAnnotations = {
+                determineAnnotator: sandbox.stub().returns(conf),
+            };
+            base.options.enableAnnotationsDiscoverability = true;
+
+            base.createAnnotator();
+
+            expect(base.createAnnotatorOptions).to.be.calledWith(sinon.match({ initialMode: 'region' }));
```
ChenCodes, line 1218: Can we also use the constant for `'region'`?

ConradJChan (approved):

PR #1256, opened 2020-09-15
feat(discoverability): add class to bp-content if FF is enabled

ChenCodes:
- [x] add `bp-annotatons-discoverable` to `bp-content` if FF is enabled in `setup` method
- [x] add `bp-annotations-create--region` to `bp-content` if FF is enabled and mode was explicitly changed to `REGION`
- [x] remove `bp-annotations-create--region` from `bp-content` if FF is enabled and mode was changed to anything other than `REGION`
- [x] update tests

### DIFF
--- a/src/lib/constants.js
+++ b/src/lib/constants.js
@@ -53,6 +53,8 @@ export const CLASS_IS_SELECTABLE = 'bp-is-selectable';
 export const CLASS_IS_BUFFERING = 'bp-is-buffering';
 export const CLASS_DARK = 'bp-dark';
 export const CLASS_CRAWLER = 'bp-crawler';
+export const CLASS_ANNOTATIONS_DISCOVERABLE = 'bp-annotations-discoverable';
+export const CLASS_ANNOTATIONS_CREATE_REGION = 'bp-annotations-create--region';
 
 export const SELECTOR_BOX_PREVIEW_CONTAINER = `.${CLASS_BOX_PREVIEW_CONTAINER}`;
 export const SELECTOR_BOX_PREVIEW = `.${CLASS_BOX_PREVIEW}`;
--- a/src/lib/constants.js
+++ b/src/lib/constants.js
@@ -1,4 +1,6 @@
 export const CLASS_ACTIVE = 'bp-is-active';
+export const CLASS_ANNOTATIONS_CREATE_REGION = 'bp-annotations-create--region';
+export const CLASS_ANNOTATIONS_DISCOVERABLE = 'bp-annotations-discoverable';
 export const CLASS_NAVIGATION_VISIBILITY = 'bp-is-navigation-visible';
 export const CLASS_HIDDEN = 'bp-is-hidden';
 export const CLASS_PREVIEW_LOADED = 'bp-loaded';
@@ -53,8 +55,6 @@ export const CLASS_IS_SELECTABLE = 'bp-is-selectable';
 export const CLASS_IS_BUFFERING = 'bp-is-buffering';
 export const CLASS_DARK = 'bp-dark';
 export const CLASS_CRAWLER = 'bp-crawler';
-export const CLASS_ANNOTATIONS_CREATE_REGION = 'bp-annotations-create--region';
-export const CLASS_ANNOTATIONS_DISCOVERABLE = 'bp-annotations-discoverable';
 
 export const SELECTOR_BOX_PREVIEW_CONTAINER = `.${CLASS_BOX_PREVIEW_CONTAINER}`;
 export const SELECTOR_BOX_PREVIEW = `.${CLASS_BOX_PREVIEW}`;
--- a/src/lib/constants.js
+++ b/src/lib/constants.js
@@ -53,8 +53,8 @@ export const CLASS_IS_SELECTABLE = 'bp-is-selectable';
 export const CLASS_IS_BUFFERING = 'bp-is-buffering';
 export const CLASS_DARK = 'bp-dark';
 export const CLASS_CRAWLER = 'bp-crawler';
-export const CLASS_ANNOTATIONS_DISCOVERABLE = 'bp-annotations-discoverable';
 export const CLASS_ANNOTATIONS_CREATE_REGION = 'bp-annotations-create--region';
+export const CLASS_ANNOTATIONS_DISCOVERABLE = 'bp-annotations-discoverable';
 
 export const SELECTOR_BOX_PREVIEW_CONTAINER = `.${CLASS_BOX_PREVIEW_CONTAINER}`;
 export const SELECTOR_BOX_PREVIEW = `.${CLASS_BOX_PREVIEW}`;
--- a/src/lib/viewers/BaseViewer.js
+++ b/src/lib/viewers/BaseViewer.js
@@ -19,6 +19,8 @@ import {
 } from '../util';
 import {
     ANNOTATOR_EVENT,
+    CLASS_ANNOTATIONS_CREATE_REGION,
+    CLASS_ANNOTATIONS_DISCOVERABLE,
     CLASS_BOX_PREVIEW_MOBILE,
     CLASS_HIDDEN,
     FILE_OPTION_START,
@@ -44,8 +46,6 @@ const VIEWER_STATUSES = {
     loaded: 'loaded',
     loading: 'loading',
 };
-
-export const CSS_ANNOTATIONS_DISCOVERABLE_CLASS = 'bp-annotations-discoverable';
 
 const ANNOTATIONS_JS = 'annotations.js';
 const ANNOTATIONS_CSS = 'annotations.css';
@@ -217,7 +217,7 @@ class BaseViewer extends EventEmitter {
         }
 
         if (this.options.enableAnnotationsDiscoverability) {
-            this.containerEl.classList.add(CSS_ANNOTATIONS_DISCOVERABLE_CLASS);
+            this.containerEl.classList.add(CLASS_ANNOTATIONS_DISCOVERABLE);
         }
 
         this.isSetup = true;
@@ -266,7 +266,7 @@ class BaseViewer extends EventEmitter {
         if (this.containerEl) {
             this.containerEl.removeEventListener('contextmenu', this.preventDefault);
             this.containerEl.innerHTML = '';
-            this.containerEl.classList.remove(CSS_ANNOTATIONS_DISCOVERABLE_CLASS);
+            this.containerEl.classList.remove(CLASS_ANNOTATIONS_DISCOVERABLE);
         }
 
         // Destroy the annotator
@@ -1100,10 +1100,10 @@ class BaseViewer extends EventEmitter {
         if (this.options.enableAnnotationsDiscoverability) {
             switch (nextMode) {
                 case AnnotationMode.REGION:
-                    this.containerEl.classList.add('bp-annotations-create--region');
+                    this.containerEl.classList.add(CLASS_ANNOTATIONS_CREATE_REGION);
                     break;
                 default:
-                    this.containerEl.classList.remove('bp-annotations-create--region');
+                    this.containerEl.classList.remove(CLASS_ANNOTATIONS_CREATE_REGION);
                     break;
             }
         }
--- a/src/lib/viewers/BaseViewer.js
+++ b/src/lib/viewers/BaseViewer.js
@@ -214,6 +214,10 @@ class BaseViewer extends EventEmitter {
             });
         }
 
+        if (this.options.enableAnnotationsDiscoverability) {
+            this.rootEl.classList.add('bp-annotations-discoverable');
+        }
+
         this.isSetup = true;
     }
 
@@ -1089,6 +1093,17 @@ class BaseViewer extends EventEmitter {
                 : nextMode,
         );
         this.annotationControls.setMode(nextMode);
+
+        if (this.options.enableAnnotationsDiscoverability) {
+            switch (mode) {
+                case AnnotationMode.REGION:
+                    this.rootEl.classList.add('bp-annotations-create--region');
+                    break;
+                default:
+                    this.rootEl.classList.remove('bp-annotations-create--region');
+                    break;
+            }
+        }
     }
 
     /**
--- a/src/lib/viewers/BaseViewer.js
+++ b/src/lib/viewers/BaseViewer.js
@@ -215,7 +215,7 @@ class BaseViewer extends EventEmitter {
         }
 
         if (this.options.enableAnnotationsDiscoverability) {
-            this.rootEl.classList.add('bp-annotations-discoverable');
+            this.containerEl.classList.add('bp-annotations-discoverable');
         }
 
         this.isSetup = true;
@@ -264,6 +264,10 @@ class BaseViewer extends EventEmitter {
         if (this.containerEl) {
             this.containerEl.removeEventListener('contextmenu', this.preventDefault);
             this.containerEl.innerHTML = '';
+
+            if (this.options.enableAnnotationsDiscoverability) {
+                this.containerEl.classList.remove('bp-annotations-discoverable');
+            }
         }
 
         // Destroy the annotator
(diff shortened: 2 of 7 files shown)


src/lib/viewers/doc/DocBaseViewer.js
```diff
@@ -126,6 +126,10 @@ class DocBaseViewer extends BaseViewer {
         this.docEl.classList.add('bp-doc');
         this.docEl.tabIndex = '0';
 
+        if (this.options.enableAnnotationsDiscoverability) {
+            this.docEl.classList.add('ba-discoverability-enabled');
```
mxiao6, line 130: Do we need to clean it up in `destroy`?

ConradJChan, line 130: I wonder if this class should be named something else, like `bp-annotations-discoverable`? We don't reference any `.ba-*` classes in preview anywhere. Thoughts?

ConradJChan: I'm not sure that we do, since `destroy()` will do `this.containerEl.innerHTML = '';`

ChenCodes: Yes I think this makes sense.

src/lib/viewers/BaseViewer.js
```diff
@@ -214,6 +214,10 @@ class BaseViewer extends EventEmitter {
             });
         }
 
+        if (this.options.enableAnnotationsDiscoverability) {
+            this.rootEl.classList.add('bp-annotations-discoverable');
```
ChenCodes, line 218: Moved to BaseViewer because we want to add `bp-annotations-discoverable` to `bp` rather than `bp-doc`.

src/lib/viewers/BaseViewer.js
```diff
@@ -1089,6 +1093,17 @@ class BaseViewer extends EventEmitter {
                 : nextMode,
         );
         this.annotationControls.setMode(nextMode);
+
+        if (this.options.enableAnnotationsDiscoverability) {
+            switch (mode) {
+                case AnnotationMode.REGION:
+                    this.rootEl.classList.add('bp-annotations-create--region');
```
ChenCodes, line 1100: Add class on explicit mode change to Region.

src/lib/viewers/__tests__/BaseViewer-test.js
```diff
@@ -1858,24 +1861,25 @@ describe('lib/viewers/BaseViewer', () => {
...
 
+        it('should call toggleAnnotationMode and setMode', () => {
             base.handleAnnotationControlsClick({ mode: AnnotationMode.REGION });
 
             expect(base.annotator.toggleAnnotationMode).to.be.calledWith(AnnotationMode.REGION);
             expect(base.annotationControls.setMode).to.be.calledWith(AnnotationMode.REGION);
         });
 
         it('should call toggleAnnotationMode with appropriate mode if discoverability is enabled', () => {
-            base.annotator = {
```
ChenCodes, line 1875: Moved up to the `beforeEach`.

src/lib/viewers/doc/_docBase.scss
```diff
@@ -13,6 +13,24 @@ $thumbnail-sidebar-width: 226px;
 }
 
 .bp {
+    &.bp-annotations-discoverable {
+        .textLayer {
```
ChenCodes, line 17: Will need a separate ticket to address removing similar styling code from annotations repo.

src/lib/viewers/BaseViewer.js
ConradJChan: Maybe `this.containerEl` might make more sense than the `rootEl` of preview

ConradJChan: Also may need to handle cleanup of the the class in the `destroy`

src/lib/viewers/BaseViewer.js
```diff
@@ -1089,6 +1093,17 @@ class BaseViewer extends EventEmitter {
                 : nextMode,
         );
         this.annotationControls.setMode(nextMode);
+
+        if (this.options.enableAnnotationsDiscoverability) {
+            switch (mode) {
```
ConradJChan, line 1098: Should this be `mode` or `nextMode`?

ChenCodes: I think it should be `nextMode`.

src/lib/viewers/BaseViewer.js
```diff
@@ -264,6 +264,10 @@ class BaseViewer extends EventEmitter {
         if (this.containerEl) {
             this.containerEl.removeEventListener('contextmenu', this.preventDefault);
             this.containerEl.innerHTML = '';
+
+            if (this.options.enableAnnotationsDiscoverability) {
```
mxiao6, line 268: Is this guard clause needed?

src/lib/viewers/BaseViewer.js
```diff
@@ -215,7 +215,7 @@ class BaseViewer extends EventEmitter {
         }
 
         if (this.options.enableAnnotationsDiscoverability) {
-            this.rootEl.classList.add('bp-annotations-discoverable');
+            this.containerEl.classList.add('bp-annotations-discoverable');
```
mxiao6, line 218: Should we also modify the stylesheet?

src/lib/viewers/BaseViewer.js
```diff
@@ -1095,12 +1099,13 @@ class BaseViewer extends EventEmitter {
         this.annotationControls.setMode(nextMode);
 
         if (this.options.enableAnnotationsDiscoverability) {
-            switch (mode) {
+            switch (nextMode) {
+                case AnnotationMode.NONE:
```
mxiao6, line 1103: Should we add `NONE` case? We still want to select texts in `NONE` (default) mode, right?

src/lib/viewers/BaseViewer.js
ChenCodes: I think it wouldn't hurt to attempt to remove `'bp-annotations-discoverable'` even if the FF was off so this guard clause isn't needed. Would be open to removing it

src/lib/viewers/BaseViewer.js
ChenCodes: Yep, fixed

src/lib/viewers/BaseViewer.js
ChenCodes: Yep I've pushed a fix to nest the styles under `bp-content`.

src/lib/viewers/doc/_docBase.scss
```diff
@@ -354,3 +336,23 @@ $thumbnail-sidebar-width: 226px;
 .bp-content.bp-is-fullscreen .bp-toggle-thumbnails-icon {
     display: none;
 }
+
+.bp-content {
+    &.bp-annotations-discoverable {
```
mxiao6, line 341: Can these two line be combined into one line just like the style above? `
`.bp-content.bp-annotations-discoverable`

src/lib/viewers/BaseViewer.js
ConradJChan: I don't think we want to apply this class for the `NONE` case because this class turns off pointer-events for the textLayer span's which is what we want only when in explicit region mode

src/lib/viewers/doc/_docBase.scss
```diff
@@ -336,3 +336,23 @@ $thumbnail-sidebar-width: 226px;
 .bp-content.bp-is-fullscreen .bp-toggle-thumbnails-icon {
     display: none;
 }
+
+.bp-content {
```
ConradJChan, line 340: Maybe we can roll the styles in L336 into this one as well?

src/lib/viewers/BaseViewer.js
mxiao6: I actually mean "should we remove `NONE` case?"...

src/lib/viewers/doc/_docBase.scss
mxiao6: Yeah, instead of combining L340 and L341, you can move L336 too

src/lib/viewers/doc/_docBase.scss
ChenCodes: So if we were to go with Conrad's approach below, I don't think I can combine them into one line.

This is what I'm thinking:

```
.bp-content {
    &.bp-annotations-discoverable {
        .textLayer {
            pointer-events: none;

            span {
                pointer-events: auto;
            }
        }

        &.bp-annotations-create--region {
            .textLayer {
                span {
                    pointer-events: none;
                }
            }
        }
    }

    // Hide the toggle thumbnails button when in fullscreen
    &.bp-is-fullscreen .bp-toggle-thumbnails-icon {
        display: none;
    }
}
```

mxiao6: Yeah, this looks good to me.

src/lib/viewers/BaseViewer.js
ChenCodes: Discussed offline, will make code simpler so will remove it.

src/lib/viewers/doc/_docBase.scss
ChenCodes: Done

src/lib/viewers/BaseViewer.js
```diff
@@ -214,6 +214,10 @@ class BaseViewer extends EventEmitter {
             });
         }
 
+        if (this.options.enableAnnotationsDiscoverability) {
+            this.containerEl.classList.add('bp-annotations-discoverable');
```
ConradJChan, line 218: Does it make sense to move these classes to constants since they're being used in a few places?

src/lib/viewers/__tests__/BaseViewer-test.js
```diff
@@ -1858,24 +1863,26 @@ describe('lib/viewers/BaseViewer', () => {
                 destroy: sandbox.stub(),
                 setMode: sandbox.stub(),
             };
-        });
-
-        it('should call toggleAnnotationMode and setMode', () => {
+            base.containerEl = {
```
ConradJChan, line 1866: Instead of stubbing the `containerEl` could we create an empty `div` and do assertions on the presence of the class on the `div` itself? That way we can treat it as more black box testing rather than rely on underlying method calls

src/lib/viewers/__tests__/BaseViewer-test.js
```diff
@@ -1858,24 +1863,26 @@ describe('lib/viewers/BaseViewer', () => {
...
+                classList: {
+                    add: sandbox.stub(),
+                    remove: sandbox.stub(),
+                },
+                removeEventListener: sandbox.stub(),
```
ConradJChan, line 1871: Is `removeEventListener` needed?

src/lib/viewers/BaseViewer.js
mxiao6: +1

src/lib/viewers/__tests__/BaseViewer-test.js
ChenCodes: Yeah it looks like it's needed:
```
 undefined is not a constructor (evaluating 'this.containerEl.removeEventListener('contextmenu', this.preventDefault)')
```

src/lib/viewers/__tests__/BaseViewer-test.js
ChenCodes: Yep that makes sense

src/lib/viewers/BaseViewer.js
```diff
@@ -45,6 +45,8 @@ const VIEWER_STATUSES = {
     loading: 'loading',
 };
 
+export const CSS_ANNOTATIONS_DISCOVERABLE_CLASS = 'bp-annotations-discoverable';
```
ConradJChan, line 48: Maybe we can put this with the other constants in the `constants.js`? And also the naming convention for those constants seems to be `CLASS_*`

src/lib/viewers/__tests__/BaseViewer-test.js
```diff
@@ -1884,5 +1890,26 @@ describe('lib/viewers/BaseViewer', () => {
...
+            base.handleAnnotationControlsClick({ mode: AnnotationMode.NONE });
+
+            expect(base.containerEl).to.not.have.class(CSS_ANNOTATIONS_CREATE_REGION);
+        });
+
+        it('should remove create region class if discoverability is enabled and mode is HIGHLIGHT', () => {
+            base.options.enableAnnotationsDiscoverability = true;
+            base.handleAnnotationControlsClick({ mode: AnnotationMode.HIGHLIGHT });
+
+            expect(base.containerEl).to.not.have.class(CSS_ANNOTATIONS_CREATE_REGION);
```
ConradJChan, line 1912: Could combine these in a parameterized test?

src/lib/viewers/doc/_docBase.scss
```diff
@@ -332,7 +332,27 @@ $thumbnail-sidebar-width: 226px;
     }
 }
 
-// Hide the toggle thumbnails button when in fullscreen
-.bp-content.bp-is-fullscreen .bp-toggle-thumbnails-icon {
-    display: none;
+.bp-content {
+    &.bp-annotations-discoverable {
+        .textLayer {
```
ConradJChan, line 337: Can you also see if we need to mention the pdfjs annotationLayer as well?

src/lib/viewers/__tests__/BaseViewer-test.js
ChenCodes: Is this how it would look?

```
const parameters = [
            { description: 'should remove create region class if discoverability is enabled and mode is NONE', mode: AnnotationMode.NONE },
            { description: 'should remove create region class if discoverability is enabled and mode is HIGHLIGHT', mode: AnnotationMode.HIGLIGHT },
        ];

        parameters.forEach((parameter) => {
            it(parameter.description, () => {
                base.options.enableAnnotationsDiscoverability = true;
                base.handleAnnotationControlsClick({ mode: parameter.mode });
    
                expect(base.containerEl).to.not.have.class(CLASS_ANNOTATIONS_CREATE_REGION);
            });
        });
```

ConradJChan: Probably could be simpler like:
```
[AnnotationMode.NONE, AnnotationMode.HIGHLIGHT].forEach(mode => {
  it(`should remove create region class if discoverability is enabled and mode is ${mode}`, () => { ... })
});
```

src/lib/viewers/doc/_docBase.scss
ChenCodes: Yep, we do need it. Added a fix.

src/lib/viewers/doc/_docBase.scss
```diff
@@ -348,6 +348,12 @@ $thumbnail-sidebar-width: 226px;
                     pointer-events: none;
                 }
             }
+
+            .annotationLayer {
+                section {
+                    pointer-events: none;
+                }
+            }
```
mxiao6, line 356: Is `section` needed? Can we ignore the whole layer?
```
.annotationLayer {
    pointer-event: none;
}
```

ChenCodes: We can ignore the whole layer.

src/lib/constants.js
```diff
@@ -53,6 +53,8 @@ export const CLASS_IS_SELECTABLE = 'bp-is-selectable';
 export const CLASS_IS_BUFFERING = 'bp-is-buffering';
 export const CLASS_DARK = 'bp-dark';
 export const CLASS_CRAWLER = 'bp-crawler';
+export const CLASS_ANNOTATIONS_DISCOVERABLE = 'bp-annotations-discoverable';
+export const CLASS_ANNOTATIONS_CREATE_REGION = 'bp-annotations-create--region';
```
ConradJChan, line 57: 🔤

ConradJChan (approved):

mxiao6 (approved):

mxiao6 (approved):

ConradJChan (approved):

PR #1258, opened 2020-09-17
feat(annotations): Elevate region targets

ConradJChan:
When discoverability is turned on, the region annotation targets need to be elevated above the region creation layer


mxiao6:
Does page flicker when region annotation is created on IE11?

mxiao6:
With the change, I don't think we can select texts within region target in explicit highlight mode. Is this expected behavior?

PR #1260, opened 2020-09-18
fix(discoverability): set mode to REGION instead of NONE

ChenCodes:
- [x] update toggleAnnotationMode to be called with REGION instead of NONE if FF is on
- [x] update tests

**Demo:**
ezgif com-video-to-gif (5)

### DIFF
--- a/src/lib/viewers/BaseViewer.js
+++ b/src/lib/viewers/BaseViewer.js
@@ -587,6 +587,9 @@ class BaseViewer extends EventEmitter {
 
         if (this.annotator && this.areNewAnnotationsEnabled()) {
             this.annotator.emit(ANNOTATOR_EVENT.setVisibility, true);
+            this.annotator.toggleAnnotationMode(
+                this.options.enableAnnotationsDiscoverability ? AnnotationMode.REGION : AnnotationMode.NONE,
+            );
             this.enableAnnotationControls();
         }
     }
--- a/src/lib/viewers/BaseViewer.js
+++ b/src/lib/viewers/BaseViewer.js
@@ -587,9 +587,9 @@ class BaseViewer extends EventEmitter {
 
         if (this.annotator && this.areNewAnnotationsEnabled()) {
             this.annotator.emit(ANNOTATOR_EVENT.setVisibility, true);
-            this.annotator.toggleAnnotationMode(
-                this.options.enableAnnotationsDiscoverability ? AnnotationMode.REGION : AnnotationMode.NONE,
-            );
+            if (this.options.enableAnnotationsDiscoverability) {
+                this.annotator.toggleAnnotationMode(AnnotationMode.REGION);
+            }
             this.enableAnnotationControls();
         }
     }
--- a/src/lib/viewers/BaseViewer.js
+++ b/src/lib/viewers/BaseViewer.js
@@ -572,7 +572,9 @@ class BaseViewer extends EventEmitter {
 
         if (this.annotator && this.areNewAnnotationsEnabled()) {
             this.annotator.emit(ANNOTATOR_EVENT.setVisibility, false);
-            this.annotator.toggleAnnotationMode(AnnotationMode.NONE);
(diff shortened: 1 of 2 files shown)


src/lib/viewers/BaseViewer.js
```diff
@@ -915,7 +917,9 @@ class BaseViewer extends EventEmitter {
 
     disableAnnotationControls() {
         if (this.annotator && this.annotationControls && this.areNewAnnotationsEnabled()) {
-            this.annotator.toggleAnnotationMode(AnnotationMode.NONE);
+            this.annotator.toggleAnnotationMode(
```
ChenCodes, line 920: Something that I noticed is that we have two toggleAnnotationMode method called being made in the 
`handleFullscreenEnter` -> `disableAnnotationControls` flow. My thinking is that either the first or second is redundant?

mxiao6: Good point. I think the first is redundant.

mxiao6:
I notice there another `this.annotator.toggleAnnotationMode` call in `handleAnnotationControlsEscape()` and I believe it is also causing bug. Basically, I think we should wrap `this.annotator.toggleAnnotationMode` into another function with discovery FF logic and change all of the current calls to the new function. Does that sound reasonable?

src/lib/viewers/BaseViewer.js
jstoffan: I'm not sure. It seems like these changes somewhat defeat the purpose of the original code, which was to forcibly disallow annotation creation. What effect does this have on creating annotations on rotated images, for example?

ChenCodes:
> I notice there is another `this.annotator.toggleAnnotationMode` call in `handleAnnotationControlsEscape()` and I believe it is also causing bug. Basically, I think we should wrap `this.annotator.toggleAnnotationMode` into another function with discovery FF logic and change all of the current `this.annotator.toggleAnnotationMode` calls to the new function. Does that sound reasonable?

I'm not sure if we want to move all of the `this.annotator.toggleAnnotation` to the new function. The reason is that in some locations, we always want `this.annotation.toggleAnnotationMode` to be called with `NONE` because we are explicitly defining that the user should not be able to create regions. In other locations, we do want to set the default to be `REGION` (for example, in handleFullScreenExit, we do want to call it  `this.annotation.toggleAnnotationMode` with REGION here if FF is on because we want the default mode to be REGION.

src/lib/viewers/BaseViewer.js
ChenCodes: My change was not in the right location, so I have moved it over to handleFullScreenExit instead.

src/lib/viewers/__tests__/BaseViewer-test.js
```diff
@@ -568,20 +568,42 @@ describe('lib/viewers/BaseViewer', () => {
             expect(base.resize).to.be.called;
         });
 
-        it('should show annotations', () => {
+        it('should show annotations and toggle annotations mode to NONE if discoverability FF is not enabled', () => {
```
ConradJChan, line 571: Seems like we could have a parameterized test for this as well based on `enableAnnotationsDiscoverability` and the called annotation `mode`

src/lib/viewers/BaseViewer.js
```diff
@@ -587,6 +587,9 @@ class BaseViewer extends EventEmitter {
 
         if (this.annotator && this.areNewAnnotationsEnabled()) {
             this.annotator.emit(ANNOTATOR_EVENT.setVisibility, true);
+            this.annotator.toggleAnnotationMode(
```
ConradJChan, line 590: Should this instead be
```
}
```
?

ChenCodes: Yeah that makes sense - we probably don't need to call with NONE again if the FF is off since it's already NONE.

ConradJChan (approved):

PR #1261, opened 2020-09-18
fix(annotations): Fix cursor when drawing region in default mode

mxiao6:
Pay attention to the cursor
Before:
text-cursor-prev

After:
text-cursor-after

### DIFF
--- a/src/lib/viewers/BaseViewer.js
+++ b/src/lib/viewers/BaseViewer.js
@@ -1089,9 +1089,11 @@ class BaseViewer extends EventEmitter {
      * @param {AnnotationMode} mode Next annotation mode
      */
     processAnnotationModeChange = mode => {
-        if (this.annotationControls) {
-            this.annotationControls.setMode(mode);
+        if (!this.annotationControls) {
+            return;
         }
+
+        this.annotationControls.setMode(mode);
 
         if (this.options.enableAnnotationsDiscoverability && this.containerEl) {
             switch (mode) {
--- a/src/lib/viewers/BaseViewer.js
+++ b/src/lib/viewers/BaseViewer.js
@@ -1082,6 +1082,30 @@ class BaseViewer extends EventEmitter {
     }
 
     /**
+     * Handler for annotation mode change
+     * 1. Set annotationsControls mode
+     * 2. If discoverability FF is on, add mode classes to container
+     *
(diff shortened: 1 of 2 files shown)


src/lib/viewers/BaseViewer.js
```diff
@@ -1081,6 +1081,30 @@ class BaseViewer extends EventEmitter {
...
     }
 
+    /**
+     * Handler for annotation mode change
+     * 1. Set annotationsControls mode
+     * 2. If discoverability FF is on, add mode classes to container
+     *
+     * @param {AnnotationMode} mode Next annotation mode
+     */
+    handleAnnotationModeChange = mode => {
```
ConradJChan, line 1091: Maybe instead of `handle*` since by convention that tends to be an event handler, `processAnnotationModeChange`?

src/lib/viewers/BaseViewer.js
```diff
@@ -1081,6 +1081,30 @@ class BaseViewer extends EventEmitter {
...
+        if (this.annotationControls) {
+            this.annotationControls.setMode(mode);
+        }
+
+        if (this.options.enableAnnotationsDiscoverability) {
```
ConradJChan, line 1096: This is unrelated to this change, but should we check to see if `this.containerEl` exists?

src/lib/viewers/BaseViewer.js
```diff
@@ -1081,6 +1081,30 @@ class BaseViewer extends EventEmitter {
...
 
+    /**
+     * Handler for annotation mode change
+     * 1. Set annotationsControls mode
+     * 2. If discoverability FF is on, add mode classes to container
+     *
+     * @param {AnnotationMode} mode Next annotation mode
+     */
+    processAnnotationModeChange = mode => {
+        if (this.annotationControls) {
```
ConradJChan, line 1092: It seemed like previously, if `this.annotationControls` didn't exist, no further processing would occur, but this now decouples the presence of `this.annotationControls` from managing the classes on `this.containerEl`. Is that desired?

mxiao6: No, we shouldn't decouple these two. If `annotationsControls` doesn't exist, there is no meaning managing the classes on `this.container`. Will add a negative guard clause at the top of this function.

ConradJChan (approved):

PR #1263, opened 2020-09-22
fix(discoverability): set mode to REGION on escape key press

ChenCodes:
ezgif com-video-to-gif (8)

### DIFF
--- a/src/lib/__tests__/AnnotationControlsFSM-test.js
+++ b/src/lib/__tests__/AnnotationControlsFSM-test.js
@@ -203,5 +203,15 @@ describe('lib/AnnotationControlsFSM', () => {
                 });
             });
         });
+
+        // reset the currentState to be AnnotationState.NONE
+        describe('reset()', () => {
+            it('should set this.currentState to AnnotationState.NONE', () => {
+                const annotationControlsFSM = new AnnotationControlsFSM(AnnotationState.REGION_TEMP);
+
+                annotationControlsFSM.reset();
+                expect(annotationControlsFSM.getState()).to.equal(AnnotationState.NONE);
+            });
+        });
     });
 });
--- a/src/lib/__tests__/AnnotationControlsFSM-test.js
+++ b/src/lib/__tests__/AnnotationControlsFSM-test.js
@@ -88,18 +88,30 @@ describe('lib/AnnotationControlsFSM', () => {
         describe('AnnotationState.HIGHLIGHT', () => {
             [
                 {
+                    input: AnnotationInput.CLICK,
                     mode: AnnotationMode.HIGHLIGHT,
                     output: AnnotationMode.NONE,
                 },
                 {
+                    input: AnnotationInput.CLICK,
                     mode: AnnotationMode.REGION,
                     output: AnnotationMode.REGION,
                 },
-            ].forEach(({ mode, output }) => {
-                it(`should output ${output} if input is click and mode is ${mode}`, () => {
+                {
+                    input: AnnotationInput.RESET,
+                    mode: AnnotationMode.HIGHLIGHT,
+                    output: AnnotationMode.NONE,
+                },
+                {
+                    input: AnnotationInput.RESET,
+                    mode: AnnotationMode.REGION,
+                    output: AnnotationMode.NONE,
+                },
+            ].forEach(({ input, mode, output }) => {
+                it(`should output ${output} if input is ${input} and mode is ${mode}`, () => {
                     const annotationControlsFSM = new AnnotationControlsFSM(AnnotationState.HIGHLIGHT);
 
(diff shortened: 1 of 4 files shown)


src/lib/viewers/BaseViewer.js
```diff
@@ -595,6 +595,8 @@ class BaseViewer extends EventEmitter {
             this.annotator.emit(ANNOTATOR_EVENT.setVisibility, true);
             if (this.options.enableAnnotationsDiscoverability) {
                 this.annotator.toggleAnnotationMode(AnnotationMode.REGION);
+                this.annotationControlsFSM.reset();
+                this.processAnnotationModeChange(AnnotationMode.NONE);
```
ChenCodes, line 599: If needed, I can move this change out to a separate PR. The bug that @mxiao6 described affects both `handleFullScreenExit` as well as `handleAnnotationControlsEscape` so I thought I would lump this change in with the `handleAnnotationControlsEscape` change.

src/lib/viewers/BaseViewer.js
```diff
@@ -1087,7 +1089,13 @@ class BaseViewer extends EventEmitter {
      * @return {void}
      */
     handleAnnotationControlsEscape() {
-        this.annotator.toggleAnnotationMode(AnnotationMode.NONE);
+        if (this.options.enableAnnotationsDiscoverability) {
+            this.annotator.toggleAnnotationMode(AnnotationMode.REGION);
+            this.annotationControlsFSM.reset();
+            this.processAnnotationModeChange(AnnotationMode.NONE);
```
ChenCodes, line 1095: The idea behind this change is that we not only want to call `toggleAnnotationMode`, we also want to reset `annotationControlsFSM` currentState because that is the source of truth. Without the above changes, we would end up seeing buggy behavior.

ChenCodes: Also, since L1093 -> L1095 is the same as L597 -> L599, I could move those out to a shared method - would be open to naming ideas for that method.

src/lib/viewers/BaseViewer.js
ConradJChan: Instead of changing the usage pattern, can we instead introduce a new `AnnotationInput.RESET` that can be passed as an argument into `this.processAnnotationModeChange(this.annotationControlsFSM.transition(AnnotationInput.RESET))`?

ChenCodes: This would be cleaner - will implement.

src/lib/__tests__/AnnotationControlsFSM-test.js
```diff
@@ -59,13 +59,21 @@ describe('lib/AnnotationControlsFSM', () => {
...
+            expect(annotationControlsFSM.getState()).to.equal(AnnotationState.NONE);
+        });
     });
 
     describe('AnnotationState.HIGHLIGHT/REGION', () => {
         // Stay in the same state
         [AnnotationState.HIGHLIGHT, AnnotationState.REGION].forEach(state => {
             Object.values(AnnotationInput)
-                .filter(input => input !== AnnotationInput.CLICK)
+                .filter(input => input !== AnnotationInput.CLICK && input !== AnnotationInput.RESET)
```
ChenCodes, line 76: We want to filter out the AnnotationInput.RESET case as well.

src/lib/__tests__/AnnotationControlsFSM-test.js
```diff
@@ -59,13 +59,21 @@ describe('lib/AnnotationControlsFSM', () => {
                 expect(annotationControlsFSM.getState()).to.equal(AnnotationState.NONE);
             });
         });
+
+        // Should reset state
+        it('should reset state if input is AnnotationInput.RESET', () => {
```
mxiao6, line 64: The tests are grouped with starting state. We have three large groups (`describe('AnnotationState.NONE'`, `describe('AnnotationState.HIGHLIGHT/REGION'` and `describe('AnnotationState.HIGHLIGHT_TEMP/REGION_TEMP'`)
You introduced a new input, so you should add one test under each of above three groups.

mxiao6:
Looks good to me besides tests.

src/lib/__tests__/AnnotationControlsFSM-test.js
```diff
@@ -59,13 +59,21 @@ describe('lib/AnnotationControlsFSM', () => {
...
+            const annotationControlsFSM = new AnnotationControlsFSM(AnnotationMode.REGION);
```
mxiao6, line 65: This test is under `NONE` group. The initial state should be `NONE`. The path `NONE -> NONE` still needs to be tested, right?

ChenCodes: Yes, that's correct

mxiao6 (approved):

PR #1266, opened 2020-09-25
fix(discoverability): add AnnotationInput.INIT in AnnotationControlsFSM

ChenCodes:
This change is needed due to a change in box-annotations where we always call `handleStop` which would trigger 
```
handleAnnotationCreatorChangeEvent({ status, type }) {
    this.processAnnotationModeChange(this.annotationControlsFSM.transition(status, type));
}
```

Status in this case would be AnnotationInput.INIT which is `cancel`.

When the input is `cancel`, this means that we want to reset AnnotationControlsFSM state to be `AnnotationState.NONE` only if our current state is `HIGHLIGHT_TEMP` or `REGION_TEMP`.

### DIFF
--- a/src/lib/AnnotationControlsFSM.ts
+++ b/src/lib/AnnotationControlsFSM.ts
@@ -12,6 +12,7 @@ export enum AnnotationInput {
     CANCEL = 'cancel',
     CLICK = 'click',
     CREATE = 'create',
+    INIT = 'init',
     RESET = 'reset',
     STARTED = 'started',
     SUCCESS = 'success',
@@ -53,7 +54,7 @@ export default class AnnotationControlsFSM {
             return stateModeMap[this.currentState];
         }
 
-        if (input === AnnotationInput.RESET) {
+        if (input === AnnotationInput.RESET || input === AnnotationInput.INIT) {
             this.currentState = AnnotationState.NONE;
             return stateModeMap[this.currentState];
         }
--- a/src/lib/AnnotationControlsFSM.ts
+++ b/src/lib/AnnotationControlsFSM.ts
@@ -54,7 +54,7 @@ export default class AnnotationControlsFSM {
             return stateModeMap[this.currentState];
         }
 
-        if (input === AnnotationInput.RESET || input === AnnotationInput.INIT) {
+        if (input === AnnotationInput.RESET) {
             this.currentState = AnnotationState.NONE;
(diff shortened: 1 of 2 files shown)


src/lib/AnnotationControlsFSM.ts
```diff
@@ -53,7 +54,7 @@ export default class AnnotationControlsFSM {
             return stateModeMap[this.currentState];
         }
 
-        if (input === AnnotationInput.RESET) {
+        if (input === AnnotationInput.RESET || input === AnnotationInput.INIT) {
```
mxiao6, line 57: Should we go to NONE state if we're in explicit region/highlight mode and the input is `INIT`?

ChenCodes: I'm not sure - my guess is we still want to stay in explicit region / highlight mode.

ChenCodes: We probably only want to go to NONE state only when we are not in explicit region / highlight mode.

mxiao6: I agree, should be under
```
case AnnotationState.HIGHLIGHT_TEMP:
case AnnotationState. REGION_TEMP:
```

ChenCodes: So something like this:
```
if (input === AnnotationInput.INIT && this.currentState === AnnotationState.NONE) {
    this.currentState = AnnotationState.NONE;
    return stateModeMap[this.currentState];
}
```

mxiao6:
Is this PR still needed if we use `cancel` which is already there?

ChenCodes:
I think this PR is a nice to have - the only change in it is the addition of some tests for the AnnotationInput.CANCEL case which wasn't previously there. There is no net-new functionality change though.

mxiao6:
> I think this PR is a nice to have - the only change in it is the addition of some tests for the AnnotationInput.CANCEL case which were not previously included. There is no net-new functionality change though.

`CANCEL` is covered at L73

ChenCodes:
> > I think this PR is a nice to have - the only change in it is the addition of some tests for the AnnotationInput.CANCEL case which were not previously included. There is no net-new functionality change though.
> 
> `CANCEL` is covered at L73 and L175

Gotcha, yeah I can close this one.

PR #1268, opened 2020-09-28
fix(discoverability): add cancel input case to FSM

ChenCodes:
This change accompanies the change in [box-annotations](https://github.com/box/box-annotations/pull/602)

### DIFF
--- a/src/lib/AnnotationControlsFSM.ts
+++ b/src/lib/AnnotationControlsFSM.ts
@@ -63,6 +63,9 @@ export default class AnnotationControlsFSM {
                 if (input === AnnotationInput.CREATE || input === AnnotationInput.STARTED) {
                     this.currentState = modeTempStateMap[mode] || AnnotationState.NONE;
                 }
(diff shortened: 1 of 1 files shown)


src/lib/AnnotationControlsFSM.ts
```diff
@@ -63,6 +63,9 @@ export default class AnnotationControlsFSM {
                 if (input === AnnotationInput.CREATE || input === AnnotationInput.STARTED) {
                     this.currentState = modeTempStateMap[mode] || AnnotationState.NONE;
                 }
+                if (input === AnnotationInput.CANCEL) {
+                    this.currentState = AnnotationState.NONE;
```
mxiao6, line 67: Is this change needed? When we're in `NONE` state, if we don't do anything, we remain in `NONE` state, right?

ChenCodes: Yes that's correct, this change isn't needed.

PR #1270, opened 2020-10-05
feat(discoverability): add image cursor experience

ChenCodes:
Changes in this PR:
- [x] store FF in options in Preview.js
- [x] add FF into init method for AnnotationControls
- [x] add logic for turning button mode to be off when the the image overflows the viewport after zoom
- [x] add unit tests

**Demo:**
ezgif com-gif-maker (1)

### DIFF
--- a/src/lib/AnnotationControls.ts
+++ b/src/lib/AnnotationControls.ts
@@ -18,8 +18,8 @@ export enum AnnotationMode {
 }
 export type ClickHandler = ({ event, mode }: { event: MouseEvent; mode: AnnotationMode }) => void;
 export type Options = {
-    enableAnnotationsImageDiscoverability: boolean;
     fileId: string;
+    initialMode?: AnnotationMode;
     onClick?: ClickHandler;
     onEscape?: () => void;
     showHighlightText: boolean;
@@ -196,8 +196,8 @@ export default class AnnotationControls {
      * Initialize the annotation controls with options.
      */
     public init({
-        enableAnnotationsImageDiscoverability = false,
         fileId,
+        initialMode = AnnotationMode.NONE,
         onEscape = noop,
         onClick = noop,
         showHighlightText = false,
@@ -213,13 +213,11 @@ export default class AnnotationControls {
             this.addButton(AnnotationMode.HIGHLIGHT, onClick, groupElement, fileId);
         }
 
-        if (enableAnnotationsImageDiscoverability) {
-            this.setMode(AnnotationMode.REGION);
-        }
-
         this.onEscape = onEscape;
         document.addEventListener('keydown', this.handleKeyDown);
 
         this.hasInit = true;
+
+        this.setMode(initialMode);
     }
 }
--- a/src/lib/AnnotationControls.ts
+++ b/src/lib/AnnotationControls.ts
@@ -18,6 +18,7 @@ export enum AnnotationMode {
 }
 export type ClickHandler = ({ event, mode }: { event: MouseEvent; mode: AnnotationMode }) => void;
 export type Options = {
+    enableAnnotationsImageDiscoverability: boolean;
     fileId: string;
     onClick?: ClickHandler;
     onEscape?: () => void;
@@ -194,7 +195,13 @@ export default class AnnotationControls {
     /**
      * Initialize the annotation controls with options.
      */
-    public init({ fileId, onEscape = noop, onClick = noop, showHighlightText = false }: Options): void {
+    public init({
+        enableAnnotationsImageDiscoverability = false,
+        fileId,
+        onEscape = noop,
+        onClick = noop,
+        showHighlightText = false,
+    }: Options): void {
         if (this.hasInit) {
             return;
         }
@@ -204,6 +211,10 @@ export default class AnnotationControls {
         this.addButton(AnnotationMode.REGION, onClick, groupElement, fileId);
         if (showHighlightText) {
             this.addButton(AnnotationMode.HIGHLIGHT, onClick, groupElement, fileId);
+        }
+
+        if (enableAnnotationsImageDiscoverability) {
+            this.setMode(AnnotationMode.REGION);
         }
 
         this.onEscape = onEscape;
--- a/src/lib/__tests__/AnnotationControls-test.js
+++ b/src/lib/__tests__/AnnotationControls-test.js
@@ -125,8 +125,8 @@ describe('lib/AnnotationControls', () => {
             expect(document.addEventListener).not.toBeCalled();
         });
 
-        test('should call setMode with mode REGION if enableAnnotationsImageDiscoverability is true', () => {
-            annotationControls.init({ enableAnnotationsImageDiscoverability: true });
+        test('should call setMode with the initialMode that is passed in', () => {
+            annotationControls.init({ initialMode: AnnotationMode.REGION });
 
             expect(annotationControls.setMode).toBeCalledWith(AnnotationMode.REGION);
         });
--- a/src/lib/__tests__/AnnotationControls-test.js
+++ b/src/lib/__tests__/AnnotationControls-test.js
@@ -128,7 +128,7 @@ describe('lib/AnnotationControls', () => {
         test('should call setMode with the initialMode that is passed in', () => {
             annotationControls.init({ initialMode: AnnotationMode.REGION });
 
-            expect(annotationControls.setMode).toBeCalledWith(AnnotationMode.REGION);
+            expect(annotationControls.currentMode).toBe(AnnotationMode.REGION);
         });
     });
 
--- a/src/lib/__tests__/AnnotationControls-test.js
+++ b/src/lib/__tests__/AnnotationControls-test.js
@@ -72,6 +72,7 @@ describe('lib/AnnotationControls', () => {
     describe('init()', () => {
         beforeEach(() => {
             jest.spyOn(annotationControls, 'addButton');
+            jest.spyOn(annotationControls, 'setMode');
         });
 
         test('should only add region button', () => {
@@ -122,6 +123,12 @@ describe('lib/AnnotationControls', () => {
 
             expect(annotationControls.addButton).not.toBeCalled();
             expect(document.addEventListener).not.toBeCalled();
+        });
+
+        test('should call setMode with mode REGION if enableAnnotationsImageDiscoverability is true', () => {
+            annotationControls.init({ enableAnnotationsImageDiscoverability: true });
+
+            expect(annotationControls.setMode).toBeCalledWith(AnnotationMode.REGION);
         });
     });
 
--- a/src/lib/__tests__/AnnotationControls-test.js
+++ b/src/lib/__tests__/AnnotationControls-test.js
@@ -72,7 +72,6 @@ describe('lib/AnnotationControls', () => {
     describe('init()', () => {
         beforeEach(() => {
             jest.spyOn(annotationControls, 'addButton');
-            jest.spyOn(annotationControls, 'setMode');
         });
 
         test('should only add region button', () => {
@@ -125,7 +124,7 @@ describe('lib/AnnotationControls', () => {
             expect(document.addEventListener).not.toBeCalled();
         });
 
-        test('should call setMode with the initialMode that is passed in', () => {
+        test('should set annotationControls currentMode to be REGION', () => {
             annotationControls.init({ initialMode: AnnotationMode.REGION });
 
             expect(annotationControls.currentMode).toBe(AnnotationMode.REGION);
--- a/src/lib/viewers/image/ImageBaseViewer.js
+++ b/src/lib/viewers/image/ImageBaseViewer.js
@@ -1,5 +1,4 @@
 import BaseViewer from '../BaseViewer';
-import { AnnotationInput } from '../../AnnotationControlsFSM';
 import Browser from '../../Browser';
 import Controls from '../../Controls';
 import PreviewError from '../../PreviewError';
@@ -8,9 +7,6 @@ import ZoomControls from '../../ZoomControls';
 import { BROWSERS, CLASS_INVISIBLE } from '../../constants';
 import { ERROR_CODE, VIEWER_EVENT } from '../../events';
 import { openContentInsideIframe } from '../../util';
-
-export const IMAGE_PADDING = 15;
-export const IMAGE_ZOOM_SCALE = 1.2;
 
 const CSS_CLASS_PANNING = 'panning';
 const CSS_CLASS_ZOOMABLE = 'zoomable';
@@ -30,9 +26,6 @@ class ImageBaseViewer extends BaseViewer {
         this.stopPanning = this.stopPanning.bind(this);
         this.zoomIn = this.zoomIn.bind(this);
         this.zoomOut = this.zoomOut.bind(this);
-
-        this.getTransformWidthAndHeight = this.getTransformWidthAndHeight.bind(this);
-        this.isRotated = this.isRotated.bind(this);
 
         this.handleMouseDown = this.handleMouseDown.bind(this);
         this.handleMouseUp = this.handleMouseUp.bind(this);
@@ -99,29 +92,6 @@ class ImageBaseViewer extends BaseViewer {
      * @return {void}
      */
     zoomIn() {
-        if (this.options.enableAnnotationsImageDiscoverability) {
-            const isRotated = this.isRotated();
-            const { width, height } = this.getTransformWidthAndHeight(
-                this.imageEl.offsetWidth,
-                this.imageEl.offsetHeight,
-                isRotated,
-            );
-
-            const viewport = {
-                width: this.wrapperEl.clientWidth - 2 * IMAGE_PADDING,
-                height: this.wrapperEl.clientHeight - 2 * IMAGE_PADDING,
-            };
-            // We want to calculate the zoomedWidth / zoomedHeight to determine if we should toggle annotation mode to None
-            const ratio = Math.min(viewport.width / width, viewport.height / height);
-            const zoomedWidth = width * IMAGE_ZOOM_SCALE * ratio;
-            const zoomedHeight = height * IMAGE_ZOOM_SCALE * ratio;
-
-            // If the image is overflowing the viewport, we should set annotation mode to be NONE so that the user can pan
-            if (zoomedWidth > viewport.width || zoomedHeight > viewport.height) {
-                this.processAnnotationModeChange(this.annotationControlsFSM.transition(AnnotationInput.CANCEL));
-            }
-        }
-
         this.zoom('in');
     }
 
@@ -132,40 +102,6 @@ class ImageBaseViewer extends BaseViewer {
      */
     zoomOut() {
         this.zoom('out');
-    }
-
-    /**
-     * Determines if Image file has been rotated 90 or 270 degrees to the left
-     *
-     * @return {boolean} Whether image has been rotated -90 or -270 degrees
-     */
-    isRotated() {
-        return Math.abs(this.currentRotationAngle) % 180 === 90;
-    }
-
-    /**
-     * Gets the width & height after the transforms are applied.
-     * When an image is rotated, the width & height of an image, e.g. offsetWidth & offsetHeight,
-     * are the values before transforms are applied, so if the image is rotated we need to swap the
-     * width & height
-     *
-     * @private
-     * @param {number} [width] - The width in px
-     * @param {number} [height] - The height in px
-     * @param {boolean} [isRotated] - if the image has a transform rotate applied to it
-     * @return {Object} the width & height of the image after tranformations
-     */
-    getTransformWidthAndHeight(width, height, isRotated) {
-        if (isRotated) {
-            return {
-                width: height,
-                height: width,
-            };
-        }
-        return {
-            width,
-            height,
-        };
     }
 
     /**
--- a/src/lib/viewers/image/ImageBaseViewer.js
+++ b/src/lib/viewers/image/ImageBaseViewer.js
@@ -1,3 +1,4 @@
+import { AnnotationInput } from '../../AnnotationControlsFSM';
 import BaseViewer from '../BaseViewer';
 import Browser from '../../Browser';
 import Controls from '../../Controls';
@@ -8,6 +9,7 @@ import { BROWSERS, CLASS_INVISIBLE } from '../../constants';
 import { ERROR_CODE, VIEWER_EVENT } from '../../events';
 import { openContentInsideIframe } from '../../util';
 
+export const IMAGE_PADDING = 15;
 const CSS_CLASS_PANNING = 'panning';
 const CSS_CLASS_ZOOMABLE = 'zoomable';
 const CSS_CLASS_PANNABLE = 'pannable';
@@ -31,6 +33,8 @@ class ImageBaseViewer extends BaseViewer {
         this.handleMouseUp = this.handleMouseUp.bind(this);
         this.cancelDragEvent = this.cancelDragEvent.bind(this);
         this.finishLoading = this.finishLoading.bind(this);
+        this.handleZoomEvent = this.handleZoomEvent.bind(this);
+        this.getViewportDimensions = this.getViewportDimensions.bind(this);
 
         if (this.isMobile) {
             if (Browser.isIOS()) {
@@ -62,6 +66,10 @@ class ImageBaseViewer extends BaseViewer {
             this.imageEl.removeEventListener('mouseup', this.handleMouseUp);
         }
 
+        if (this.options.enableAnnotationsImageDiscoverability) {
+            this.removeListener('zoom', this.handleZoomEvent);
+        }
+
         super.destroy();
     }
 
@@ -84,6 +92,36 @@ class ImageBaseViewer extends BaseViewer {
             this.loaded = true;
             this.emit(VIEWER_EVENT.load);
         });
+    }
+
+    /**
+     * Gets the viewport dimensions.
+     *
+     * @return {Object} the width & height of the viewport
+     */
+
+    getViewportDimensions() {
+        return {
+            width: this.wrapperEl.clientWidth - 2 * IMAGE_PADDING,
+            height: this.wrapperEl.clientHeight - 2 * IMAGE_PADDING,
+        };
+    }
+
+    /**
+     * Sets mode to be AnnotationMode.NONE if the zoomed image overflows the viewport.
+     *
+     * @return {void}
+     */
+    handleZoomEvent({ newScale }) {
+        const width = newScale[0];
+        const height = newScale[1];
+
+        const viewport = this.getViewportDimensions();
+
+        // If the image is overflowing the viewport, we should set annotation mode to be NONE so that the user can pan
+        if (width > viewport.width || height > viewport.height) {
+            this.processAnnotationModeChange(this.annotationControlsFSM.transition(AnnotationInput.RESET));
+        }
     }
 
     /**
--- a/src/lib/viewers/image/ImageBaseViewer.js
+++ b/src/lib/viewers/image/ImageBaseViewer.js
@@ -1,10 +1,10 @@
-import { AnnotationInput } from '../../AnnotationControlsFSM';
 import BaseViewer from '../BaseViewer';
 import Browser from '../../Browser';
 import Controls from '../../Controls';
 import PreviewError from '../../PreviewError';
 import ZoomControls from '../../ZoomControls';
 
+import { AnnotationInput } from '../../AnnotationControlsFSM';
 import { BROWSERS, CLASS_INVISIBLE } from '../../constants';
 import { ERROR_CODE, VIEWER_EVENT } from '../../events';
 import { openContentInsideIframe } from '../../util';
--- a/src/lib/viewers/image/ImageBaseViewer.js
+++ b/src/lib/viewers/image/ImageBaseViewer.js
@@ -1,4 +1,5 @@
 import BaseViewer from '../BaseViewer';
+import { AnnotationInput } from '../../AnnotationControlsFSM';
 import Browser from '../../Browser';
 import Controls from '../../Controls';
 import PreviewError from '../../PreviewError';
@@ -7,6 +8,9 @@ import ZoomControls from '../../ZoomControls';
 import { BROWSERS, CLASS_INVISIBLE } from '../../constants';
 import { ERROR_CODE, VIEWER_EVENT } from '../../events';
 import { openContentInsideIframe } from '../../util';
+
+export const IMAGE_PADDING = 15;
+export const IMAGE_ZOOM_SCALE = 1.2;
 
 const CSS_CLASS_PANNING = 'panning';
 const CSS_CLASS_ZOOMABLE = 'zoomable';
@@ -26,6 +30,9 @@ class ImageBaseViewer extends BaseViewer {
         this.stopPanning = this.stopPanning.bind(this);
         this.zoomIn = this.zoomIn.bind(this);
         this.zoomOut = this.zoomOut.bind(this);
+
+        this.getTransformWidthAndHeight = this.getTransformWidthAndHeight.bind(this);
+        this.isRotated = this.isRotated.bind(this);
 
         this.handleMouseDown = this.handleMouseDown.bind(this);
         this.handleMouseUp = this.handleMouseUp.bind(this);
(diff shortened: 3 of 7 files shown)


src/lib/viewers/image/ImageViewer.js
```diff
@@ -178,6 +179,10 @@ class ImageViewer extends ImageBaseViewer {
         let height;
         let width;
         const isRotated = this.isRotated();
+        const viewport = {
```
ChenCodes, line 182: Moved this const up to the top because it's being shared by both the if and else blocks.

src/lib/AnnotationControls.ts
```diff
@@ -194,7 +195,13 @@ export default class AnnotationControls {
     /**
      * Initialize the annotation controls with options.
      */
-    public init({ fileId, onEscape = noop, onClick = noop, showHighlightText = false }: Options): void {
+    public init({
+        enableAnnotationsImageDiscoverability = false,
```
ConradJChan, line 199: It doesn't seem like this is what we should do, we try to keep `AnnotationControls` as a dumb component that just renders the buttons based on state. If we want to initialize the controls in a certain state, then we should just provide a `initialMode` or something more generic

src/lib/viewers/image/ImageViewer.js
```diff
@@ -192,6 +197,20 @@ class ImageViewer extends ImageBaseViewer {
...
+            const zoomedHeight =
+                type === 'in' ? height * IMAGE_ZOOM_SCALE * ratio : (height / IMAGE_ZOOM_SCALE) * ratio;
+
+            // If the image is overflowing the viewport, we should set annotation mode to be NONE so that the user can pan
+            if (zoomedWidth > viewport.width || zoomedHeight > viewport.height) {
+                if (this.options.enableAnnotationsImageDiscoverability) {
+                    this.processAnnotationModeChange(this.annotationControlsFSM.transition(AnnotationInput.CANCEL));
+                    this.annotator.toggleAnnotationMode(AnnotationMode.NONE);
+                }
+            }
```
ConradJChan, line 213: I'd be inclined to leave this `zoom` method alone and perhaps react to it via event or perhaps here https://github.com/box/box-content-preview/blob/c29dddc59d652cd597fd518a8af5bb78bdd5f16c/src/lib/viewers/image/ImageBaseViewer.js#L95

src/lib/viewers/image/ImageViewer.js
```diff
@@ -192,6 +197,20 @@ class ImageViewer extends ImageBaseViewer {
...
+            const ratio = Math.min(viewport.width / width, viewport.height / height);
+            const zoomedWidth = newWidth * ratio;
+            const zoomedHeight =
+                type === 'in' ? height * IMAGE_ZOOM_SCALE * ratio : (height / IMAGE_ZOOM_SCALE) * ratio;
+
+            // If the image is overflowing the viewport, we should set annotation mode to be NONE so that the user can pan
+            if (zoomedWidth > viewport.width || zoomedHeight > viewport.height) {
+                if (this.options.enableAnnotationsImageDiscoverability) {
+                    this.processAnnotationModeChange(this.annotationControlsFSM.transition(AnnotationInput.CANCEL));
+                    this.annotator.toggleAnnotationMode(AnnotationMode.NONE);
```
ConradJChan, line 211: Is this necessary? Seems like `processAnnotationModeChange` should already handle it

ChenCodes: Nope this isn't needed, will remove.

src/lib/AnnotationControls.ts
ChenCodes: I see. So maybe we can instead pass in `initialMode` in as a new parameter in the init method?

src/lib/viewers/image/ImageViewer.js
ChenCodes: Sounds good, I've moved this logic up to the ImageBaseViewer class. I also had to move a few helper methods as well because it seems to make more sense for the child to call the parent helpers rather than the parent calling the child methods.

src/lib/viewers/image/ImageBaseViewer.js
```diff
@@ -104,6 +134,40 @@ class ImageBaseViewer extends BaseViewer {
         this.zoom('out');
     }
 
+    /**
```
ChenCodes, line 137: Moved these helper methods out from ImageViewer -> ImageBaseViewer because they are now being used by both classes.

src/lib/viewers/image/ImageViewer.js
```diff
@@ -1,5 +1,6 @@
-import AnnotationControls from '../../AnnotationControls';
 import ImageBaseViewer from './ImageBaseViewer';
+import AnnotationControls, { AnnotationMode } from '../../AnnotationControls';
```
ConradJChan, line 2: for 🔤 , this should still be on L1

src/lib/viewers/image/ImageViewer.js
```diff
@@ -389,6 +411,15 @@ class ImageViewer extends ImageBaseViewer {
         if (this.isMobile) {
             this.imageEl.addEventListener('orientationchange', this.handleOrientationChange);
         }
+
+        if (this.options.enableAnnotationsImageDiscoverability) {
```
ConradJChan, line 415: This isn't really a DOM listener, maybe it would make more sense in `setup`?

src/lib/viewers/image/ImageViewer.js
```diff
@@ -389,6 +411,15 @@ class ImageViewer extends ImageBaseViewer {
...
+            this.addListener('zoom', ({ newScale }) => {
```
ConradJChan, line 416: Instead of an anonymous function, should we just have `handleZoom` as a named function on the `ImageViewer` so that it can also be cleaned up in `destroy`?

src/lib/viewers/image/ImageViewer.js
```diff
@@ -297,6 +316,9 @@ class ImageViewer extends ImageBaseViewer {
             this.annotationControls = new AnnotationControls(this.controls);
             this.annotationControls.init({
                 fileId: this.options.file.id,
+                initialMode: this.options.enableAnnotationsImageDiscoverability
```
ConradJChan, line 319: We also need to init the `AnnotationControlsFSM` with `AnnotationState.REGION` in `ImageViewer` to keep the source of truth up to date

src/lib/viewers/image/ImageViewer.js
```diff
@@ -78,6 +80,23 @@ class ImageViewer extends ImageBaseViewer {
         super.handleAssetAndRepLoad();
     }
 
+    /**
+     * Sets mode to be AnnotationMode.NONE if the zoomed image overflows the viewport.
+     *
+     * @return {void}
+     */
+    handleZoomEvent(height, width) {
+        const viewport = {
```
ConradJChan, line 89: Maybe it's worth creating a helper function like `getViewportDimensions` to be shared with `zoom`?

src/lib/viewers/image/ImageViewer.js
```diff
@@ -78,6 +80,23 @@ class ImageViewer extends ImageBaseViewer {
...
+            width: this.wrapperEl.clientWidth - 2 * IMAGE_PADDING,
+            height: this.wrapperEl.clientHeight - 2 * IMAGE_PADDING,
+        };
+
+        // If the image is overflowing the viewport, we should set annotation mode to be NONE so that the user can pan
+        if (width > viewport.width || height > viewport.height) {
+            this.processAnnotationModeChange(this.annotationControlsFSM.transition(AnnotationInput.CANCEL));
```
ConradJChan, line 96: Should this be `AnnotationInput.RESET`? I only see a transition for `AnnotationInput.CANCEL` from `AnnotationState.REGION_TEMP`

src/lib/viewers/image/__tests__/ImageViewer-test.js
```diff
@@ -351,19 +351,28 @@ describe('lib/viewers/image/ImageViewer', () => {
...
-                onClick: image.handleAnnotationControlsClick,
-                onEscape: image.handleAnnotationControlsEscape,
-            });
-        });
+        test.each`
+            enableAnnotationsImageDiscoverability | initialMode              | should
+            ${false}                              | ${AnnotationMode.NONE}   | ${'AnnotationMode.NONE if enableAnnotationsImageDiscoverability is false'}
+            ${true}                               | ${AnnotationMode.REGION} | ${'AnnotationMode.REGION if enableAnnotationsImageDiscoverability is true '}
+        `(
+            'should call annotation controls init with callbacks and with initialMode set to $should ',
```
ConradJChan, line 359: is the `should` param needed? Could it just be `should call annotation controls init with $initialMode when enableAnnotationsImageDiscoverability is $enableAnnotationsImageDiscoverability`?

ChenCodes: Not needed, will remove.

src/lib/viewers/image/ImageViewer.js
ChenCodes: Yep it should be RESET instead.

src/lib/viewers/image/ImageViewer.js
ChenCodes: Yep that's a good idea.

src/lib/viewers/image/ImageViewer.js
ChenCodes: I can have it be a named function instead. I'm not seeing a `destroy` method within ImageViewer though.

src/lib/__tests__/AnnotationControls-test.js
```diff
@@ -123,6 +124,12 @@ describe('lib/AnnotationControls', () => {
             expect(annotationControls.addButton).not.toBeCalled();
             expect(document.addEventListener).not.toBeCalled();
         });
+
+        test('should call setMode with the initialMode that is passed in', () => {
+            annotationControls.init({ initialMode: AnnotationMode.REGION });
+
+            expect(annotationControls.setMode).toBeCalledWith(AnnotationMode.REGION);
```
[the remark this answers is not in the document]
ChenCodes: Yeah that would be better to make sure that the functionality is completely correct.

src/lib/__tests__/AnnotationControls-test.js
```diff
@@ -72,6 +72,7 @@ describe('lib/AnnotationControls', () => {
     describe('init()', () => {
         beforeEach(() => {
             jest.spyOn(annotationControls, 'addButton');
+            jest.spyOn(annotationControls, 'setMode');
```
ConradJChan, line 75: Can this spy be removed then?

src/lib/__tests__/AnnotationControls-test.js
```diff
@@ -123,6 +124,12 @@ describe('lib/AnnotationControls', () => {
             expect(annotationControls.addButton).not.toBeCalled();
             expect(document.addEventListener).not.toBeCalled();
         });
+
+        test('should call setMode with the initialMode that is passed in', () => {
```
ConradJChan, line 128: Probably needs to update the description

src/lib/viewers/image/ImageBaseViewer.js
```diff
@@ -1,4 +1,5 @@
 import BaseViewer from '../BaseViewer';
+import { AnnotationInput } from '../../AnnotationControlsFSM';
```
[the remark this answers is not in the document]
ConradJChan: Comment still applies :)

src/lib/viewers/image/ImageBaseViewer.js
```diff
@@ -86,6 +94,36 @@ class ImageBaseViewer extends BaseViewer {
...
+        const width = newScale[0];
+        const height = newScale[1];
+
+        const viewport = this.getViewportDimensions();
+
+        // If the image is overflowing the viewport, we should set annotation mode to be NONE so that the user can pan
+        if (width > viewport.width || height > viewport.height) {
+            this.processAnnotationModeChange(this.annotationControlsFSM.transition(AnnotationInput.RESET));
+        }
+    }
```
ConradJChan, line 125: These seem to be `ImageViewer` specific, i.e. `IMAGE_PADDING` may not apply to the `MultiImageViewer` and similarly, the `handleZoomEvent` should only apply to the `ImageViewer`

src/lib/viewers/image/ImageViewer.js
```diff
@@ -44,6 +48,10 @@ class ImageViewer extends ImageBaseViewer {
         this.imageEl.classList.add(CLASS_INVISIBLE);
 
         this.currentRotationAngle = 0;
+
+        if (this.options.enableAnnotationsImageDiscoverability) {
```
ConradJChan, line 52: Seems like this should be cleaned up in `ImageViewer.destroy` instead of in `ImageBaseViewer`

src/lib/viewers/image/ImageViewer.js
```diff
@@ -21,6 +21,10 @@ class ImageViewer extends ImageBaseViewer {
         if (this.isMobile) {
             this.handleOrientationChange = this.handleOrientationChange.bind(this);
         }
+
+        if (this.options.enableAnnotationsImageDiscoverability) {
```
ConradJChan, line 25: To make it clear that ImageViewer has it's own `annotationControlFSM`, perhaps we should always instantiate it's own instance and move the `this.options.enableAnnotationsImageDiscoverability` as a ternary for the initial state

src/lib/__tests__/AnnotationControls-test.js
ChenCodes: Yep good catch.

src/lib/viewers/image/ImageViewer.js
ChenCodes: Yep good idea.

src/lib/viewers/image/ImageViewer.js
```diff
@@ -17,12 +18,30 @@ class ImageViewer extends ImageBaseViewer {
         this.updatePannability = this.updatePannability.bind(this);
         this.handleImageDownloadError = this.handleImageDownloadError.bind(this);
         this.handleAssetAndRepLoad = this.handleAssetAndRepLoad.bind(this);
+        this.handleZoomEvent = this.handleZoomEvent.bind(this);
+        this.getViewportDimensions = this.getViewportDimensions.bind(this);
```
ConradJChan, line 22: Is this binding needed if the 'handleZoomEvent' is already bound?

src/lib/viewers/image/ImageViewer.js
```diff
@@ -44,6 +63,10 @@ class ImageViewer extends ImageBaseViewer {
         this.imageEl.classList.add(CLASS_INVISIBLE);
 
         this.currentRotationAngle = 0;
+
+        if (this.options.enableAnnotationsImageDiscoverability) {
+            this.addListener('zoom', this.handleZoomEvent);
```
ConradJChan, line 68: Is there an existing constant for this?

src/lib/viewers/image/ImageViewer.js
```diff
@@ -165,6 +188,36 @@ class ImageViewer extends ImageBaseViewer {
         };
     }
 
+    /**
+     * Gets the viewport dimensions.
+     *
+     * @return {Object} the width & height of the viewport
+     */
+
```
ConradJChan, line 196: Extra line?

src/lib/viewers/image/ImageViewer.js
```diff
@@ -165,6 +188,36 @@ class ImageViewer extends ImageBaseViewer {
...
+        };
+    }
+
+    /**
+     * Sets mode to be AnnotationMode.NONE if the zoomed image overflows the viewport.
+     *
+     * @return {void}
+     */
+    handleZoomEvent({ newScale }) {
+        const width = newScale[0];
```
ConradJChan, line 210: Could these be restructured in one line?

src/lib/viewers/image/ImageViewer.js
```diff
@@ -165,6 +188,36 @@ class ImageViewer extends ImageBaseViewer {
...
+        const height = newScale[1];
+
+        const viewport = this.getViewportDimensions();
+
+        // If the image is overflowing the viewport, we should set annotation mode to be NONE so that the user can pan
+        if (width > viewport.width || height > viewport.height) {
```
ConradJChan, line 216: How about the case where the user has explicitly entered region mode? Seems like this would kick them out of it

src/lib/viewers/image/ImageViewer.js
ChenCodes: Not needed

src/lib/viewers/image/ImageViewer.js
ChenCodes: I couldn't find an existing constant for 'zoom'.

src/lib/viewers/image/ImageViewer.js
ChenCodes: Yep will do.

src/lib/viewers/image/ImageViewer.js
```diff
@@ -17,12 +18,30 @@ class ImageViewer extends ImageBaseViewer {
         this.updatePannability = this.updatePannability.bind(this);
         this.handleImageDownloadError = this.handleImageDownloadError.bind(this);
         this.handleAssetAndRepLoad = this.handleAssetAndRepLoad.bind(this);
+        this.handleZoomEvent = this.handleZoomEvent.bind(this);
+        this.getViewportDimensions = this.getViewportDimensions.bind(this);
+        this.annotationControlsFSM = new AnnotationControlsFSM(
+            this.options.enableAnnotationsImageDiscoverability ? AnnotationState.REGION : AnnotationState.NONE,
```
mxiao6, line 24: This makes users in explicit region mode by default (region control button is on). I thought we want the image to have the same experience as document (annotations in region mode, but region control button is off). Can you confirm with product that this is the desired behavior?

src/lib/viewers/image/ImageViewer.js
ChenCodes: Good catch, discussed offline that we will try a REGION_TEMP approach instead.

src/lib/viewers/image/ImageViewer.js
ChenCodes: Discussed offline, the desired behavior is that the button is on.

src/lib/viewers/image/ImageViewer.js
```diff
@@ -165,6 +187,38 @@ class ImageViewer extends ImageBaseViewer {
...
+            height: this.wrapperEl.clientHeight - 2 * IMAGE_PADDING,
+        };
+    }
+
+    /**
+     * Sets mode to be AnnotationMode.NONE if the zoomed image overflows the viewport.
+     *
+     * @return {void}
+     */
+
```
ConradJChan, line 208: nit: extra lines here & L195

src/lib/viewers/image/ImageViewer.js
```diff
@@ -165,6 +187,38 @@ class ImageViewer extends ImageBaseViewer {
...
+    handleZoomEvent({ newScale }) {
+        const [width, height] = newScale;
+
+        const viewport = this.getViewportDimensions();
+
+        // We only set AnnotationMode to be NONE if the image overflows the viewport and the state is not explicit region creation
+        const { currentState } = this.annotationControlsFSM;
```
ConradJChan, line 215: Can we use the `getState` method instead of accessing `currentState` directly?

src/lib/viewers/image/ImageViewer.js
```diff
@@ -165,6 +187,38 @@ class ImageViewer extends ImageBaseViewer {
...
+        if (currentState !== AnnotationState.REGION && (width > viewport.width || height > viewport.height)) {
```
ConradJChan, line 216: Would looking for `AnnotationState.REGION_TEMP` make more sense since we only want to execute this logic in that case?

src/lib/viewers/image/ImageViewer.js
```diff
@@ -165,6 +187,38 @@ class ImageViewer extends ImageBaseViewer {
...
+            this.processAnnotationModeChange(this.annotationControlsFSM.transition(AnnotationInput.CANCEL));
+            this.annotator.toggleAnnotationMode(AnnotationMode.NONE);
```
ConradJChan, line 218: Seems like in `BaseViewer` we typically follow a `toggleAnnotationMode` followed by a `processAnnotationModeChange` -- should we just order it the same to be consistent?

src/lib/viewers/image/__tests__/ImageViewer-test.js
```diff
@@ -576,4 +606,55 @@ describe('lib/viewers/image/ImageViewer', () => {
...
+            Object.defineProperty(image.wrapperEl, 'clientHeight', { value: 100 });
+            jest.spyOn(image, 'processAnnotationModeChange');
+        });
+
+        test.each`
+            currentState                   | height | width  | should
+            ${AnnotationState.REGION}      | ${60}  | ${60}  | ${'image does not overflow the viewport'}
+            ${AnnotationState.REGION}      | ${110} | ${110} | ${'image does overflow the viewport'}
+            ${AnnotationState.REGION_TEMP} | ${110} | ${110} | ${'image does overflow the viewport'}
+        `('should not call processAnnotationImage if $should and currentState is $currentState', () => {
```
ConradJChan, line 638: `processAnnotationModeChange`?

src/lib/viewers/image/__tests__/ImageViewer-test.js
```diff
@@ -576,4 +606,55 @@ describe('lib/viewers/image/ImageViewer', () => {
...
+            image.annotationControlsFSM.currentState = AnnotationState.REGION;
```
ConradJChan, line 639: Instead of setting `currentState`, we could instead instantiate a new instance with `initialState`

src/lib/viewers/image/__tests__/ImageViewer-test.js
```diff
@@ -576,4 +606,55 @@ describe('lib/viewers/image/ImageViewer', () => {
...
+            Object.defineProperty(image.wrapperEl, 'clientWidth', { value: 100 });
+            Object.defineProperty(image.wrapperEl, 'clientHeight', { value: 100 });
+            jest.spyOn(image, 'processAnnotationModeChange');
+        });
+
+        test.each`
+            currentState                   | height | width  | should
+            ${AnnotationState.REGION}      | ${60}  | ${60}  | ${'image does not overflow the viewport'}
+            ${AnnotationState.REGION}      | ${110} | ${110} | ${'image does overflow the viewport'}
+            ${AnnotationState.REGION_TEMP} | ${110} | ${110} | ${'image does overflow the viewport'}
```
ConradJChan, line 637: Shouldn't this case cause `processAnnotationModeChange` to be called?

ConradJChan: Ah it looks like the `currentState` param is never used in the test

src/lib/viewers/image/ImageViewer.js
ChenCodes: Yep that makes more sense.

src/lib/viewers/image/__tests__/ImageViewer-test.js
ChenCodes: Yeah I forgot to push a change to address that line

src/lib/viewers/image/ImageViewer.js
```diff
@@ -17,12 +18,29 @@ class ImageViewer extends ImageBaseViewer {
         this.updatePannability = this.updatePannability.bind(this);
         this.handleImageDownloadError = this.handleImageDownloadError.bind(this);
         this.handleAssetAndRepLoad = this.handleAssetAndRepLoad.bind(this);
+        this.handleZoomEvent = this.handleZoomEvent.bind(this);
+        this.annotationControlsFSM = new AnnotationControlsFSM(
```
mxiao6, line 22: `AnnotationControlsFSM` has already been created in `BaseViewer` constructor, right? Should we move the logic to `BaseViewer`? Or call something like `this.annotationControlsFSM.setState` here?

src/lib/viewers/image/ImageViewer.js
```diff
@@ -165,6 +187,36 @@ class ImageViewer extends ImageBaseViewer {
...
+
+    /**
+     * Sets mode to be AnnotationMode.NONE if the zoomed image overflows the viewport.
+     *
+     * @return {void}
+     */
+    handleZoomEvent({ newScale }) {
+        const [width, height] = newScale;
+
+        const viewport = this.getViewportDimensions();
```
mxiao6, line 210: Destructure?

ChenCodes: I think I can keep it as is because if I were to destructure it there would be a naming collision between the existing width / height defined on L208.

I could do this though:

`const { height: viewportHeight, width: viewportWidth } = this.getViewportDimensions();`

mxiao6: Yeah, `viewport.height/width` is only used once so it's probably fine.

src/lib/viewers/image/ImageViewer.js
ChenCodes: I think @ConradJChan and I discussed that we wanted ImageViewer to instantiate its own instance of AnnotationControlsFSM.

src/lib/viewers/image/__tests__/ImageViewer-test.js
```diff
@@ -576,4 +606,59 @@ describe('lib/viewers/image/ImageViewer', () => {
...
+
+    describe('handleZoomEvent', () => {
+        beforeEach(() => {
+            image.wrapperEl = document.createElement('img');
+            Object.defineProperty(image.wrapperEl, 'clientWidth', { value: 100 });
+            Object.defineProperty(image.wrapperEl, 'clientHeight', { value: 100 });
+            jest.spyOn(image, 'processAnnotationModeChange');
+        });
+
+        test.each`
```
ConradJChan, line 630: Seems like this test may be combining too many things. Maybe split it out to a test based on `type` to make sure it doesn't get processed and the another test for the combinations of `currentState` and whether the image overflows

src/lib/viewers/image/ImageViewer.js
```diff
@@ -165,6 +187,58 @@ class ImageViewer extends ImageBaseViewer {
...
+
+    /**
+     * Sets mode to be AnnotationMode.NONE if the zoomed image overflows the viewport.
+     *
+     * @return {void}
+     */
+    handleZoomEvent({ newScale, type }) {
+        const [width, height] = newScale;
+        // This method is fired on first render - we don't want toggleAnnotationMode to be called on the first render
+        const isFirstRender = type === undefined;
```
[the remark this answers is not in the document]
ChenCodes: Yeah that sounds like a better name.

src/lib/viewers/image/ImageViewer.js
```diff
@@ -165,6 +187,56 @@ class ImageViewer extends ImageBaseViewer {
...
+    }
+
+    /**
+     * Sets mode to be AnnotationMode.NONE if the zoomed image overflows the viewport.
+     *
+     * @return {void}
+     */
+    handleZoomEvent({ newScale, type }) {
+        const [width, height] = newScale;
+        const isUserInitiated = type !== undefined;
```
mxiao6, line 225: Why the existence of `type` is connected with `isUserInitiated`? So if we accidentally change `type` parameter in `zoom` event in the future, this logic might be broken then?

src/lib/viewers/image/ImageViewer.js
```diff
@@ -165,6 +187,56 @@ class ImageViewer extends ImageBaseViewer {
...
+        if (!isUserInitiated) {
+            return;
+        }
+
+        const viewport = this.getViewportDimensions();
+
+        // We only set AnnotationMode to be NONE if the image overflows the viewport and the state is not explicit region creation
+        const currentState = this.annotationControlsFSM.getState();
+        if (currentState === AnnotationState.REGION_TEMP && (width > viewport.width || height > viewport.height)) {
```
mxiao6, line 234: So if user explicitly turns off region mode, the state in FSM will become `NONE` and the mode in annotations will become `REGION`, and then user zoom in, but this will not work anymore, because the `currentState` is `NONE`. Then user cannot "pan" anymore.

src/lib/viewers/image/ImageViewer.js
ChenCodes: Right now there are two possibilities of `handleZoomEvent` being called.

1. On initial render, `handleZoomEvent` is called (even though the user has not zoomed).
2. The user performs a zoom action (type may be `in` or `out`.

If we change the type parameter to some other string, since that is still a user action, this logic would not break because we are checking against whether or not a user has performed an action.

mxiao6: How about
```
// explanation
if (!type) {
    return;
}
```

src/lib/viewers/image/ImageViewer.js
ChenCodes: Yes I see what you mean. Maybe I can do this:

```
if ((currentState === AnnotationState.REGION_TEMP || currentState === AnnotationState.NONE) && (width > viewport.width || height > viewport.height)) {
    this.annotator.toggleAnnotationMode(AnnotationMode.NONE);
    this.processAnnotationModeChange(this.annotationControlsFSM.transition(AnnotationInput.CANCEL));
}
```

src/lib/viewers/image/ImageViewer.js
ConradJChan: Let's also make sure that if `enableAnnotationsImageDiscoverability === false`, that we get the old `ImageViewer` experience, i.e. not the `enableAnnotationsDiscoverability` experience

src/lib/viewers/image/ImageViewer.js
ConradJChan: Good callout @mxiao6, we probably need to account for the button click -- probably need to override `handleAnnotationControlsClick` for how the `ImageViewer` experience should be

ChenCodes: Since @ConradJChan merged his refactor change, I think we no longer need to override `handleAnnotationControlsClick`.

src/lib/viewers/image/ImageViewer.js
ChenCodes: Could you elaborate a little on this? Do you mean that we should have a check like this:

`
if (enableAnnotationsImageDiscoverability) {
     this.annotationControlsFSM = new AnnotationControlsFSM(AnnotationState.REGION_TEMP);
}
`

src/lib/viewers/image/ImageViewer.js
ConradJChan: Seems like this comment is still outstanding

src/lib/viewers/image/__tests__/ImageViewer-test.js
```diff
@@ -52,11 +53,31 @@ describe('lib/viewers/image/ImageViewer', () => {
         stubs = {};
     });
 
+    describe('destroy()', () => {
+        test('should remove the zoom event listener', () => {
+            image.options.enableAnnotationsImageDiscoverability = true;
+            image.removeListener = jest.fn();
```
ConradJChan, line 59: This seems like it could be a spy rather than an outright stub to preserve the implementation

src/lib/viewers/image/__tests__/ImageViewer-test.js
```diff
@@ -52,11 +53,31 @@ describe('lib/viewers/image/ImageViewer', () => {
...
+        });
+    });
+
     describe('setup()', () => {
         test('should set up layout', () => {
             expect(image.wrapperEl).toHaveClass('bp-image');
             expect(image.imageEl).toHaveClass('bp-is-invisible');
         });
+
+        test('should bind zoom listener if enableAnnotationsImageDiscoverability is true', () => {
```
ConradJChan, line 73: Should there be a similar test for asserting that no zoom listener is added if the feature is false?

src/lib/viewers/image/__tests__/ImageViewer-test.js
ConradJChan: The `should` param is probably unnecessary. Maybe just update the test description to include the `height` and `width`, otherwise consider breaking out a separate test to test the `currentState` vs `overflow` conditions

src/lib/viewers/image/__tests__/ImageViewer-test.js
```diff
@@ -54,30 +54,52 @@ describe('lib/viewers/image/ImageViewer', () => {
...
+        test('should remove the zoom event listener if enableAnnotationsImageDiscoverability is true', () => {
             image.options.enableAnnotationsImageDiscoverability = true;
-            image.removeListener = jest.fn();
 
             image.destroy();
 
             expect(image.removeListener).toBeCalledWith('zoom', expect.any(Function));
         });
+
+        test('should not remove the zoom event listener if enableAnnotationsImageDiscoverability is false', () => {
```
ConradJChan, line 69: Could combine these two into 1 with a `test.each` and assert with `toHaveBeenCalledTimes`

ConradJChan (approved):

ConradJChan (approved):

PR #1271, opened 2020-10-09
fix(annotations): move discoverability changes into DocBaseViewer

ConradJChan:
TODO:
- [x] unit tests
- [x] regression testing of region, highlight & discoverability for docs
- [x] regression testing of region for images

### DIFF
--- a/src/lib/viewers/doc/DocBaseViewer.js
+++ b/src/lib/viewers/doc/DocBaseViewer.js
@@ -1568,7 +1568,7 @@ class DocBaseViewer extends BaseViewer {
     }
 
     // Annotation overrides
-    getInitialAnnotionMode() {
+    getInitialAnnotationMode() {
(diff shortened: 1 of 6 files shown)


src/lib/viewers/doc/DocBaseViewer.js
```diff
@@ -1555,6 +1566,54 @@ class DocBaseViewer extends BaseViewer {
         // For documents of only 1 page, default thumbnails as closed
         return toggledState && numPages > 1;
     }
+
+    // Annotation overrides
+    getInitialAnnotationMode() {
+        return this.options.enableAnnotationsDiscoverability ? AnnotationMode.REGION : AnnotationMode.NONE;
+    }
+
+    initAnnotations() {
```
jstoffan, line 1575: Would it be worthwhile to offload this to a separate helper "component" and that accepts the `annotator` and `options` objects?

jstoffan (approved):

PR #1274, opened 2020-10-23
fix(pdf): Revert pdf.js to v2.2.228 due to font rendering issue

jstoffan:
We found a font rendering regression when previewing large documents in Microsoft Edge via VM. There appears to be an emerging rendering issue with the latest version of pdf.js in resource-constrained environments. We'll need to revert to the prior working version until it's resolved.


ConradJChan (approved):

ChenCodes (approved):
LGTM

PR #1275, opened 2020-10-23
feat(discoverability): Add instrumentation for annotation creation

ConradJChan:


src/lib/viewers/doc/DocBaseViewer.js
```diff
@@ -1590,6 +1590,8 @@ class DocBaseViewer extends BaseViewer {
                 : nextMode,
         );
         this.processAnnotationModeChange(nextMode);
+
+        this.updateDiscoverabilityResinTag();
```
mxiao6, line 1594: Should we call this function for every FSM transition? There're also transitions in `handleAnnotationCreateEvent`, should we call this there? Is it adding a state change callback in FSM a good idea?

src/lib/viewers/BaseViewer.js
```diff
@@ -53,6 +53,12 @@ const ANNOTATION_CLASSES = {
     [AnnotationMode.REGION]: CLASS_ANNOTATIONS_CREATE_REGION,
 };
 
+export const ANNOTATION_DISCOVERABILITY_STATES = [
```
jstoffan, line 56: Can we move this into the FSM and expose a `isDiscoverable` method on it?

src/lib/viewers/BaseViewer.js
```diff
@@ -223,9 +229,23 @@ class BaseViewer extends EventEmitter {
...
+    updateDiscoverabilityResinTag() {
+        if (!this.containerEl) {
+            return;
+        }
+
+        const isDiscoverabilityState = ANNOTATION_DISCOVERABILITY_STATES.includes(
+            this.annotationControlsFSM.getState(),
+        );
+        const isUsingDiscoverability = this.options.enableAnnotationsDiscoverability && isDiscoverabilityState;
+        this.containerEl.setAttribute('data-resin-usingdiscoverability', isUsingDiscoverability);
```
jstoffan, line 246: Can we simplify this to `data-resin-discoverable`?

src/lib/viewers/doc/DocBaseViewer.js
jstoffan: It seems okay to me. That's basically how redux works with `subscribe`, so probably fine?

ConradJChan: The reality is, we only really care to change the resin attribute when toggling in and out of explicit modes, but moving this to some sort of subscribe callback would be more comprehensive. If we introduce a subscribe callback, we could probably move the calls to `processAnnotationModeChange` there as well. Any objections to doing this in a follow up refactor?

jstoffan (approved):

src/lib/viewers/doc/DocBaseViewer.js
ConradJChan: I had to do it anyway to account for the differences between image discoverability and doc discoverability

src/lib/AnnotationControlsFSM.ts
```diff
@@ -41,20 +43,28 @@ export const stateModeMap = {
...
 
+    private notify = (): void => this.subscriptions.forEach(callback => callback(this.currentState));
+
+    public subscribe = (callback: Subscription): void => this.subscriptions.push(callback);
+
     public transition = (input: AnnotationInput, mode: AnnotationMode = AnnotationMode.NONE): AnnotationMode => {
         if (input === AnnotationInput.CLICK) {
             this.currentState = mode === stateModeMap[this.currentState] ? AnnotationState.NONE : modeStateMap[mode];
+            this.notify();
             return stateModeMap[this.currentState];
```
jstoffan, line 62: Seems like this is a good example of why we typically avoid multiple returns. Should this be refactored?

src/lib/AnnotationControlsFSM.ts
```diff
@@ -41,20 +43,28 @@ export const stateModeMap = {
...
         }
 
         if (input === AnnotationInput.RESET) {
             this.currentState = AnnotationState.NONE;
```
jstoffan, line 66: Could we introduce a `setState` method that calls `this.notify` every time it's called?

src/lib/AnnotationControlsFSM.ts
```diff
@@ -41,20 +43,28 @@ export const stateModeMap = {
...
 
+    private subscriptions: Subscription[] = [];
+
     constructor(initialState = AnnotationState.NONE) {
         this.currentState = initialState;
     }
 
     public getState = (): AnnotationState => this.currentState;
 
+    private notify = (): void => this.subscriptions.forEach(callback => callback(this.currentState));
```
jstoffan, line 54: Methods are typically grouped by visibility, public -> private.

src/lib/viewers/doc/DocBaseViewer.js
```diff
@@ -1614,6 +1624,24 @@ class DocBaseViewer extends BaseViewer {
     handleAnnotationCreatorChangeEvent({ status, type }) {
         this.processAnnotationModeChange(this.annotationControlsFSM.transition(status, type));
     }
+
+    handleFSMChange = () => {
```
jstoffan, line 1628: We should probably use `.bind` in the constructor to stay consistent with all other handlers in this class.

src/lib/viewers/doc/DocBaseViewer.js
```diff
@@ -94,6 +100,8 @@ class DocBaseViewer extends BaseViewer {
     constructor(options) {
         super(options);
 
+        this.annotationControlsFSM.subscribe(this.handleFSMChange);
```
jstoffan, line 103: This should probably happen last to allow its handler to be bound.

src/lib/viewers/image/ImageViewer.js
```diff
@@ -507,6 +511,23 @@ class ImageViewer extends ImageBaseViewer {
...
+        if (!this.containerEl) {
+            return;
+        }
+
+        const isDiscoverable = this.annotationControlsFSM.getState() === AnnotationState.REGION_TEMP;
+        const isUsingDiscoverability = this.options.enableAnnotationsImageDiscoverability && isDiscoverable;
+
+        // For tracking purposes, set property to true when the annotation controls are in a state
+        // in which the default discoverability experience is enabled
+        this.containerEl.setAttribute('data-resin-discoverability', isUsingDiscoverability);
```
jstoffan, line 529: Should this string be made into a constant since it's now used in two separate viewers?

src/lib/viewers/doc/DocBaseViewer.js
```diff
@@ -1614,6 +1624,24 @@ class DocBaseViewer extends BaseViewer {
     handleAnnotationCreatorChangeEvent({ status, type }) {
         this.processAnnotationModeChange(this.annotationControlsFSM.transition(status, type));
     }
+
+    handleFSMChange = () => {
+        this.updateDiscoverabilityResinTag();
```
jstoffan, line 1629: Does this need to be a whole separate method?

src/lib/AnnotationControlsFSM.ts
```diff
@@ -41,38 +43,52 @@ export const stateModeMap = {
 export default class AnnotationControlsFSM {
     private currentState: AnnotationState;
 
+    private subscriptions: Subscription[] = [];
+
     constructor(initialState = AnnotationState.NONE) {
         this.currentState = initialState;
```
jstoffan, line 49: Should this also call `setState`? If so, the separate calls to `updateDiscoverabilityResinTag` wouldn't be needed.

jstoffan (approved):

src/lib/AnnotationControlsFSM.ts
ConradJChan: It could, but I feel like the `subscribe` is more like a change listener while setting an initial state in the constructor is just that -- more of an initialization

jstoffan (approved):

src/lib/AnnotationControlsFSM.ts
```diff
@@ -41,38 +43,54 @@ export const stateModeMap = {
...
+                default:
+            }
         }
 
-        if (input === AnnotationInput.RESET) {
-            this.currentState = AnnotationState.NONE;
-            return stateModeMap[this.currentState];
-        }
+        this.setState(newState);
+        return stateModeMap[this.currentState];
```
mxiao6, line 83: Should this be `stateModeMap[newState];`?

mxiao6: Never mind, the currentState has already changed.

mxiao6 (approved):

PR #1276, opened 2020-10-26
fix(annotations): Select newly created annotation

ConradJChan:


src/lib/viewers/image/ImageViewer.js
```diff
@@ -507,6 +508,22 @@ class ImageViewer extends ImageBaseViewer {
         this.annotator.toggleAnnotationMode(nextMode);
         this.processAnnotationModeChange(nextMode);
     }
+
+    handleAnnotationCreateEvent({ annotation: { id } = {}, meta: { status } = {} }) {
+        // Only on success do we exit create annotation mode. If error occurs,
+        // we remain in create mode
+        if (status === 'success') {
```
jstoffan, line 515: Should this be a guard clause?

src/lib/viewers/image/ImageViewer.js
```diff
@@ -507,6 +508,22 @@ class ImageViewer extends ImageBaseViewer {
...
+            this.annotator.emit('annotations_active_set', id);
+        }
+    }
+
+    initAnnotations() {
+        super.initAnnotations();
+
+        if (this.areNewAnnotationsEnabled() && this.annotationControls) {
```
jstoffan, line 523: Should we be checking `this.annotator`, instead or as well?

src/lib/viewers/image/ImageViewer.js
```diff
@@ -507,6 +508,22 @@ class ImageViewer extends ImageBaseViewer {
...
+            this.annotator.addListener('annotations_create', this.handleAnnotationCreateEvent);
```
jstoffan, line 524: Does this listener need to be removed at some point?

src/lib/viewers/image/ImageViewer.js
ConradJChan: `BaseViewer.initAnnotations` already assumes `this.annotator` exists, so it seems safe enough?

src/lib/viewers/image/ImageViewer.js
ConradJChan: I don't think so, `BaseViewer.destroy` takes care of it by calling `this.annotator.removeAllListeners()`

src/lib/viewers/image/ImageViewer.js
jstoffan: I'm confused why the `this.annotationControls` check is needed, then. It's not referenced in the subsequent code or in the handler.

ConradJChan: Will remove that check since it seems unnecessary

jstoffan (approved):

PR #1281, opened 2020-11-03
feat(annotations): add ftux cursor logic

ChenCodes:
Changes in this PR:
- add logic to hide popup cursor after the first time that the user explicitly clicks the create region button
- hide cursor using styles

Todo:
- add tests

Demo:
ezgif com-gif-maker (3)


src/lib/viewers/BaseViewer.js
```diff
@@ -979,11 +982,23 @@ class BaseViewer extends EventEmitter {
 
         const options = boxAnnotations.getOptions && boxAnnotations.getOptions();
 
+        // Set disabled ftux cursor flag depending on if the local storage key is set
+        let isDocumentFtuxCursorDisabled = false;
+        let isImageFtuxCursorDisabled = false;
+        if (this.cache.get(DOCUMENT_FTUX_CURSOR_DISABLED_KEY)) {
```
ChenCodes, line 988: I could inline this into the `this.createAnnotatorOptions`, I kept it outside of that for readability purposes.

src/lib/viewers/doc/DocBaseViewer.js
```diff
@@ -1592,10 +1593,25 @@ class DocBaseViewer extends BaseViewer {
             this.annotator.addListener('creator_staged_change', this.handleAnnotationCreatorChangeEvent);
             this.annotator.addListener('creator_status_change', this.handleAnnotationCreatorChangeEvent);
         }
+
+        if (this.cache.get(DOCUMENT_FTUX_CURSOR_DISABLED_KEY)) {
+            this.annotator.emit('annotations_document_explicit_create_toggled');
```
ChenCodes, line 1598: Open to other naming suggestions.

src/lib/viewers/image/ImageViewer.js
```diff
@@ -531,6 +541,10 @@ class ImageViewer extends ImageBaseViewer {
         if (this.areNewAnnotationsEnabled()) {
             this.annotator.addListener('annotations_create', this.handleAnnotationCreateEvent);
         }
+
+        if (this.cache.get(IMAGE_FTUX_CURSOR_DISABLED_KEY)) {
+            this.annotator.emit('annotations_image_explicit_create_toggled');
```
ChenCodes, line 546: This is required because we want to make sure that the PopupCursor is hidden properly on a refresh.

src/lib/viewers/image/ImageViewer.js
```diff
@@ -5,6 +5,7 @@ import { CLASS_INVISIBLE, DISCOVERABILITY_ATTRIBUTE } from '../../constants';
 import { ICON_FULLSCREEN_IN, ICON_FULLSCREEN_OUT, ICON_ROTATE_LEFT } from '../../icons/icons';
 import './Image.scss';
 
+const IMAGE_FTUX_CURSOR_DISABLED_KEY = 'image-ftux-cursor-disabled';
```
jstoffan, line 8: We should avoid duplicating keys.

src/lib/viewers/image/ImageViewer.js
```diff
@@ -513,6 +514,15 @@ class ImageViewer extends ImageBaseViewer {
      */
     handleAnnotationControlsClick({ mode }) {
         const nextMode = this.annotationControlsFSM.transition(AnnotationInput.CLICK, mode);
+
+        if (nextMode === AnnotationMode.REGION) {
+            if (!this.cache.get(IMAGE_FTUX_CURSOR_DISABLED_KEY)) {
+                this.cache.set(IMAGE_FTUX_CURSOR_DISABLED_KEY, true, true /* useLocalStorage */);
```
jstoffan, line 520: What is the reasoning for this block comment? Does it affect webpack in some way?

src/lib/viewers/image/ImageViewer.js
```diff
@@ -513,6 +514,15 @@ class ImageViewer extends ImageBaseViewer {
      */
     handleAnnotationControlsClick({ mode }) {
         const nextMode = this.annotationControlsFSM.transition(AnnotationInput.CLICK, mode);
+
+        if (nextMode === AnnotationMode.REGION) {
```
jstoffan, line 518: We should avoid duplicating logic between files. Can this be moved to a reusable location?

src/lib/viewers/image/ImageViewer.js
```diff
@@ -513,6 +514,15 @@ class ImageViewer extends ImageBaseViewer {
...
+            if (!this.cache.get(IMAGE_FTUX_CURSOR_DISABLED_KEY)) {
+                this.cache.set(IMAGE_FTUX_CURSOR_DISABLED_KEY, true, true /* useLocalStorage */);
+            } else {
+                this.annotator.emit('annotations_image_explicit_create_toggled');
```
jstoffan, line 522: It looks like we're both passing in options to the annotator and emitting events to it. Are they related? Can this be simplified?

src/lib/viewers/BaseViewer.js
```diff
@@ -53,6 +53,9 @@ const ANNOTATION_CLASSES = {
     [AnnotationMode.REGION]: CLASS_ANNOTATIONS_CREATE_REGION,
 };
 
+const IMAGE_FTUX_CURSOR_DISABLED_KEY = 'image-ftux-cursor-disabled';
```
jstoffan, line 56: We should try to use positive conditions to reduce double-negatives. How about `bp-ftux-image-cursor-seen`?

src/lib/viewers/BaseViewer.js
```diff
@@ -979,11 +982,23 @@ class BaseViewer extends EventEmitter {
 
         const options = boxAnnotations.getOptions && boxAnnotations.getOptions();
 
+        // Set disabled ftux cursor flag depending on if the local storage key is set
+        let isDocumentFtuxCursorDisabled = false;
```
jstoffan, line 986: It seems like this could be inlined and converted to a constant.

src/lib/viewers/BaseViewer.js
```diff
@@ -979,11 +982,23 @@ class BaseViewer extends EventEmitter {
...
+        if (this.cache.get(IMAGE_FTUX_CURSOR_DISABLED_KEY)) {
+            isImageFtuxCursorDisabled = true;
+        }
+
         const annotatorOptions = this.createAnnotatorOptions({
             annotator: this.annotatorConf,
             features: options && options.features,
             initialMode: this.getInitialAnnotationMode(),
             intl: (options && options.intl) || intlUtil.createAnnotatorIntl(),
+            isDocumentFtuxCursorDisabled,
```
jstoffan, line 1000: We should avoid adding more top-level options, if possible. Should these properties be added and/or excluded from the `features` object?

jstoffan (changes requested):

src/lib/viewers/image/ImageViewer.js
ChenCodes: I don't think it affects webpack, I initially kept it there because I thought it was used to indicate to the user what the parameter was for. I'm open to removing it as I'm seeing instances of where it's used as well as not used.

src/lib/viewers/BaseViewer.js
ChenCodes: That sounds good, thanks for the suggestion.

src/lib/viewers/image/ImageViewer.js
ChenCodes: Yes I can move it to BaseViewer.js and have both children share a parent method.

src/lib/viewers/image/ImageViewer.js
ChenCodes: They are related:
- After a refresh of the page, we want to pass the ftux cursor flag via options to the annotator.
- When the user clicks on the Create Region button, we want to emit an event that would cause a rerender of the RegionCreator component.

I'm not aware of how this could be simplified but would be open to suggestions.

src/lib/viewers/BaseViewer.js
ChenCodes: Removed this line as it is no longer needed.

src/lib/viewers/BaseViewer.js
ChenCodes: Removed this line as it's no longer needed.

src/lib/viewers/image/ImageViewer.js
ChenCodes: We no longer need to pass in options to the annotator after implementing the css approach.

src/lib/viewers/BaseViewer.js
```diff
@@ -53,6 +55,9 @@ const ANNOTATION_CLASSES = {
     [AnnotationMode.REGION]: CLASS_ANNOTATIONS_CREATE_REGION,
 };
 
+export const DOCUMENT_FTUX_CURSOR_SEEN_KEY = 'bp-document-ftux-cursor-seen';
+export const IMAGE_FTUX_CURSOR_SEEN_KEY = 'bp-image-ftux-cursor-seen';
```
ConradJChan, line 59: Maybe `bp-ftux-cursor-seen-[document|image]`?

src/lib/viewers/BaseViewer.js
```diff
@@ -1000,6 +1006,30 @@ class BaseViewer extends EventEmitter {
...
     }
 
+    /**
+     * Updates localStorage with the state of the ftux cursor toggle
+     *
+     * @protected
+     * @param {string} key
+     * @return {void}
+     */
+    handleFtuxCursorToggle(key) {
```
ConradJChan, line 1016: Typically we use `handle*` naming convention for event handlers, maybe in this case it could be `applyCursorFtux` or something along those lines?

src/lib/viewers/BaseViewer.js
```diff
@@ -1000,6 +1006,30 @@ class BaseViewer extends EventEmitter {
...
+        if (!this.cache.get(key)) {
+            this.cache.set(key, true, true);
+        } else if (this.cache.get(key) && this.containerEl) {
+            switch (key) {
```
ConradJChan, line 1020: Could we have a map of `key` -> `classname` so we could avoid this switch statement?

src/lib/viewers/BaseViewer.js
```diff
@@ -1000,6 +1006,30 @@ class BaseViewer extends EventEmitter {
...
+    /**
+     * Updates localStorage with the state of the ftux cursor toggle
+     *
+     * @protected
+     * @param {string} key
+     * @return {void}
+     */
+    handleFtuxCursorToggle(key) {
+        if (!this.cache.get(key)) {
+            this.cache.set(key, true, true);
```
ConradJChan, line 1018: Seems like we might want to pull out marking the ftux status in a separate method

src/lib/viewers/doc/DocBaseViewer.js
```diff
@@ -1596,6 +1596,11 @@ class DocBaseViewer extends BaseViewer {
 
     handleAnnotationControlsClick({ mode }) {
         const nextMode = this.annotationControlsFSM.transition(AnnotationInput.CLICK, mode);
+
+        if (nextMode === AnnotationMode.REGION) {
```
ConradJChan, line 1600: `annotationControlsFSM` currently subscribes to the transition changes but calls `updateDiscoverabilityResinTag` directly, perhaps we could introduce a `handler` that also checks for this condition and applies the class there.

Also it seems like we would want to update localStorage on the transition from `REGION` to `NONE`, or even to `REGION` but only trigger the logic to apply the class on any subsequent transition to `REGION`

src/lib/viewers/doc/_docBase.scss
```diff
@@ -370,6 +370,13 @@ $thumbnail-sidebar-width: 226px;
         }
     }
 
+    &.bp-annotations-ftux-document-cursor-seen,
+    &.bp-annotations-ftux-image-cursor-seen {
```
ConradJChan, line 374: This doesn't seem applicable to `_docBase.scss`

src/lib/viewers/BaseViewer.js
ConradJChan, line 1018: Also, should we perform this ftux on a per user id basis?

src/lib/viewers/BaseViewer.js
ChenCodes: That sounds good.

src/lib/viewers/doc/DocBaseViewer.js
ChenCodes: Discussed offline that since the transition changes only happen when the user clicks, we're okay with not needing the handler.

src/lib/viewers/BaseViewer.js
ChenCodes: After doing some digging, it looks like we don't need to do it on a per user id basis because localStorage sessions are different for unique users.

src/lib/viewers/doc/DocBaseViewer.js
ConradJChan: I'm still debating whether this should be moved out to a fsm callback. This method is pretty mode agnostic so I'm wary of cluttering it with some region specific code. Let's do the callback because that'll be easier to strip out eventually too since it'll be it's own separate function. Something like:
```
this.annotationControlsFSM.subscribe(this.applyCursorFtux);
```

src/lib/viewers/image/Image.scss
```diff
@@ -48,3 +48,13 @@
     display: block;
     width: 100%;
 }
+
+.bp-content {
+    &.bp-annotations-ftux-image-cursor-seen {
+        .bp-image {
```
ConradJChan, line 54: Is the `.bp-image` a needed specificity?

src/lib/viewers/BaseViewer.js
```diff
@@ -1000,6 +1007,35 @@ class BaseViewer extends EventEmitter {
...
     }
 
+    /**
+     * Sets the cursor ftux key in localStorage
+     *
+     * @protected
+     * @param {string} key
+     * @return {void}
+     */
+    setCursorFtux(key) {
```
jstoffan, line 1017: We want to reduce the amount of code we're introducing to BaseViewer. This method seems to have very little code in it, so its logic can be moved to the subclasses.

src/lib/viewers/BaseViewer.js
```diff
@@ -1000,6 +1007,35 @@ class BaseViewer extends EventEmitter {
...
+     * @param {string} key
+     * @return {void}
+     */
+    applyCursorFtux(key) {
+        const keyToClassMap = {
+            [IMAGE_FTUX_CURSOR_SEEN_KEY]: CLASS_ANNOTATIONS_IMAGE_FTUX_CURSOR_SEEN,
+            [DOCUMENT_FTUX_CURSOR_SEEN_KEY]: CLASS_ANNOTATIONS_DOCUMENT_FTUX_CURSOR_SEEN,
+        };
+
+        if (this.containerEl) {
```
jstoffan, line 1034: Can each viewer (document, image) maintain its own version of this with its specific key and class name? There's not much logic left anymore to require a method in BaseViewer.

ChenCodes:
https://github.com/box/box-content-preview/pull/1281#discussion_r519064903

No, it's not needed.

ChenCodes:
https://github.com/box/box-content-preview/pull/1281#discussion_r519061106

The binds are not required.

src/lib/viewers/image/Image.scss
ChenCodes: Yes it is required.

src/lib/viewers/BaseViewer.js
```diff
@@ -162,13 +167,15 @@ class BaseViewer extends EventEmitter {
         this.mobileZoomEndHandler = this.mobileZoomEndHandler.bind(this);
         this.handleAnnotatorEvents = this.handleAnnotatorEvents.bind(this);
         this.handleAnnotationControlsEscape = this.handleAnnotationControlsEscape.bind(this);
+        this.applyCursorFtux = this.applyCursorFtux.bind(this);
```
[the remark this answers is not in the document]
ChenCodes: They weren't required previously, but with the addition of the `subscribe` call, they are now required.

src/lib/viewers/doc/DocBaseViewer.js
```diff
@@ -1677,6 +1681,26 @@ class DocBaseViewer extends BaseViewer {
...
+     *
+     * @protected
+     * @return {void}
+     */
+    applyCursorFtux() {
+        if (!this.containerEl) {
+            return;
+        }
+
+        if (this.annotationControlsFSM.getState() === AnnotationState.REGION) {
```
ConradJChan, line 1696: Could we move this into the guard clause above? then we can flatten out this nested if-else statement

src/lib/viewers/doc/DocBaseViewer.js
```diff
@@ -1677,6 +1681,26 @@ class DocBaseViewer extends BaseViewer {
...
+            if (this.cache.get(DOCUMENT_FTUX_CURSOR_SEEN_KEY)) {
+                this.containerEl.classList.add(CLASS_ANNOTATIONS_DOCUMENT_FTUX_CURSOR_SEEN);
```
ConradJChan, line 1698: Is there any chance this class won't get applied by the time the `PopupCursor` is already mounted in box-annotations and we might see a flash of the popup?

src/lib/viewers/image/Image.scss
ConradJChan: What about if we did `&.bp-annotations--create-region.bp-annotations-ftux-image-cursor-seen`?

src/lib/viewers/doc/DocBaseViewer.js
ChenCodes: Discussed offline, it looks like this issue shouldn't happen.

src/lib/viewers/doc/_docBase.scss
```diff
@@ -370,6 +370,12 @@ $thumbnail-sidebar-width: 226px;
         }
     }
 
+    &.bp-annotations-ftux-document-cursor-seen {
```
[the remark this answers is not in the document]
ChenCodes: Yep this works as well.

ConradJChan (approved):

PR #1282, opened 2020-11-03
feat(controls): Add react versions of core control components

jstoffan:


src/lib/viewers/controls/controls-bar/ControlsBar.tsx
```diff
@@ -0,0 +1,14 @@
+import React from 'react';
+import './ControlsBar.scss';
+
+export type Props = {
+    children: React.ReactNode;
+};
+
+export default function ControlsBar({ children, ...rest }: Props): JSX.Element {
+    return (
+        <div className="bp-ControlsBar" {...rest}>
```
ConradJChan, line 10: Would there be any need to provide a custom class name as a prop as each viewer may have their own styles or would that be done at a different component?

src/lib/viewers/controls/controls-layer/ControlsLayer.scss
```diff
@@ -0,0 +1,11 @@
+@import 'node_modules/box-ui-elements/es/styles/variables';
```
ConradJChan, line 1: Needed?

src/lib/viewers/controls/controls-layer/ControlsLayer.tsx
```diff
@@ -0,0 +1,86 @@
...
+export const HIDE_DELAY_MS = 2000;
+
+export default function ControlsLayer({ children, onMount = noop }: Props): JSX.Element {
+    const [isShown, setIsShown] = React.useState(false);
+    const hasFocusRef = React.useRef(false);
+    const hasCursorRef = React.useRef(false);
+    const hideTimeoutRef = React.useRef<number>();
+
+    // Visibility helpers
+    const helpers = React.useRef({
```
ConradJChan, line 27: `helpersRef`?

src/lib/viewers/controls/controls-layer/ControlsLayer.tsx
```diff
@@ -0,0 +1,86 @@
...
+        hide() {
+            window.clearTimeout(hideTimeoutRef.current);
+
+            hideTimeoutRef.current = window.setTimeout(() => {
```
ConradJChan, line 31: Are there any scenarios where we would want `hide` to immediately hide the control bar instead of waiting 2 seconds?

src/lib/viewers/controls/controls-layer/ControlsLayer.tsx
```diff
@@ -0,0 +1,86 @@
...
+    };
+
+    // Expose helpers to parent
+    React.useEffect(() => {
+        onMount(helpers.current);
+    }, [onMount]);
+
+    return (
+        <div
+            className={`bp-ControlsLayer ${isShown ? HIDE_CLASSNAME : ''}`}
```
ConradJChan, line 77: Should we import classnames?

src/lib/viewers/controls/controls-root/ControlsRoot.tsx
```diff
@@ -0,0 +1,71 @@
...
+    }, 100);
+
+    destroy(): void {
+        if (!this.containerEl) {
+            return;
+        }
+
+        this.containerEl.removeEventListener('mousemove', this.handleMouseMove);
+        this.containerEl.removeEventListener('touchstart', this.handleMouseMove);
+        this.containerEl.removeChild(this.controlsEl);
```
ConradJChan, line 57: Do we need to do any react unmount during cleanup?

src/lib/viewers/controls/controls-bar/ControlsBar.tsx
jstoffan: I don't see that being necessary at the moment. It would not be difficult to extend this component later if needed.

src/lib/viewers/controls/controls-layer/ControlsLayer.tsx
jstoffan: Not currently, outside of the enable/disable methods in the root component. If it comes up at a later point, we can add more code to support it.

src/lib/viewers/controls/controls-layer/ControlsLayer.tsx
jstoffan: If we run into more conditional styles/modifiers down the road, sure. I don't think it's needed for this one case.

src/lib/viewers/controls/controls-root/ControlsRoot.tsx
jstoffan: Good catch.

src/lib/viewers/controls/controls-layer/__tests__/ControlsLayer-test.tsx
```diff
@@ -0,0 +1,115 @@
...
+            ${'mouseenter'} | ${'mouseleave'}
+        `('should show $showTrigger and hide $hideTrigger', ({ hideTrigger, showTrigger }) => {
+            const wrapper = getWrapper();
+
+            act(() => {
+                getElement(wrapper).simulate(showTrigger);
+            });
+            wrapper.update();
+
+            expect(getElement(wrapper).hasClass(HIDE_CLASSNAME)).toBe(true);
```
ConradJChan, line 39: Should these checks be reversed? After show trigger shouldn't  HIDE_CLASSNAME not be present?

ConradJChan (dismissed):

src/lib/viewers/controls/controls-layer/__tests__/ControlsLayer-test.tsx
jstoffan: Oh man, good catch. The test is right, but I inverted the name of the variable somehow; it should be `SHOW_CLASSNAME`.

ConradJChan (approved):

PR #1285, opened 2020-11-05
feat(controls): Create react core controls for image viewers

jstoffan:
Todo
- [x] Unit tests

### DIFF
--- a/src/lib/viewers/image/ImageViewer.js
+++ b/src/lib/viewers/image/ImageViewer.js
@@ -1,6 +1,8 @@
-import ImageBaseViewer from './ImageBaseViewer';
+import React from 'react';
 import AnnotationControls, { AnnotationMode } from '../../AnnotationControls';
 import AnnotationControlsFSM, { AnnotationInput, AnnotationState, stateModeMap } from '../../AnnotationControlsFSM';
+import ImageBaseViewer from './ImageBaseViewer';
+import ImageControls from './ImageControls';
 import { CLASS_INVISIBLE, DISCOVERABILITY_ATTRIBUTE } from '../../constants';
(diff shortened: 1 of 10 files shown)


src/lib/viewers/image/ImageViewer.js
```diff
@@ -378,6 +379,28 @@ class ImageViewer extends ImageBaseViewer {
...
+        super.loadUIReact();
+        this.renderUI();
+    }
+
+    renderUI() {
+        if (this.zoomControls) {
+            this.zoomControls.setCurrentScale(this.scale);
+        }
+
+        if (this.controls && this.options.useReactControls) {
```
ConradJChan, line 392: Should `this.options.useReactControls` be first?

src/lib/viewers/image/ImageViewer.js
```diff
@@ -378,6 +379,28 @@ class ImageViewer extends ImageBaseViewer {
         }
     }
 
+    loadUIReact() {
+        super.loadUIReact();
+        this.renderUI();
+    }
+
+    renderUI() {
+        if (this.zoomControls) {
```
ConradJChan, line 388: Could this be an `else` condition attached to below's `if`?

src/lib/viewers/image/ImageViewer.js
jstoffan: I don't think it makes an objective difference? I just prefer the visual symmetry of the existence check being immediately on top of its invocation below.

src/lib/viewers/image/ImageViewer.js
jstoffan: I'd rather keep it as simple as possible, since much of this is temporary and will be removed once we fully convert the controls.

ConradJChan (approved):

PR #1287, opened 2020-11-05
fix(annotations): Fix AnnotationControlsFSM reset when fullscreen enter

mxiao6:


src/lib/viewers/BaseViewer.js
```diff
@@ -575,8 +575,8 @@ class BaseViewer extends EventEmitter {
 
         if (this.annotator && this.areNewAnnotationsEnabled()) {
             this.annotator.emit(ANNOTATOR_EVENT.setVisibility, false);
-            this.annotator.toggleAnnotationMode(AnnotationMode.NONE);
```
jstoffan, line 578: Is this not still needed to exit "create" mode in the annotator?

src/lib/viewers/BaseViewer.js
```diff
@@ -575,8 +575,8 @@ class BaseViewer extends EventEmitter {
...
             this.disableAnnotationControls();
+            this.processAnnotationModeChange(this.annotationControlsFSM.transition(AnnotationInput.RESET));
```
jstoffan, line 579: Should this be in `disableAnnotationControls`, instead?

mxiao6: Yeah, you're right

src/lib/viewers/BaseViewer.js
mxiao6: It's already in `disableAnnotationControls`

src/lib/viewers/BaseViewer.js
```diff
@@ -919,8 +918,8 @@ class BaseViewer extends EventEmitter {
     disableAnnotationControls() {
         if (this.annotator && this.annotationControls && this.areNewAnnotationsEnabled()) {
             this.annotator.toggleAnnotationMode(AnnotationMode.NONE);
-            this.annotationControls.resetControls();
             this.annotationControls.toggle(false);
+            this.processAnnotationModeChange(this.annotationControlsFSM.transition(AnnotationInput.RESET));
```
jstoffan, line 922: Can we move this outside of the conditional above? We'll need it for the react rewrite, which doesn't set `this.annotationControls`.

jstoffan (approved):

PR #1291, opened 2020-11-06
feat(controls): Add react version of annotations controls

jstoffan:
Todo
- [x] Unit tests


src/lib/viewers/controls/annotations/AnnotationsButton.tsx
```diff
@@ -0,0 +1,44 @@
...
+    if (!isEnabled) {
+        return null;
+    }
+
+    return (
+        <button
+            className={classNames('bp-AnnotationsButton', className, {
+                'bp-is-active': isActive,
+            })}
+            data-resin-fileid={fileId}
```
ConradJChan, line 36: Could this resin tag be applied at a higher level, like `ControlsBar` to avoid having to propagate it down to each button?

src/lib/viewers/controls/annotations/AnnotationsControls.tsx
```diff
@@ -0,0 +1,65 @@
...
+
+    return (
+        <div className="bp-AnnotationsControls">
+            <AnnotationsButton
+                data-resin-target="highlightRegion"
+                data-testid="bp-AnnotationsControls-regionBtn"
+                fileId={fileId}
+                isActive={annotationMode === AnnotationMode.REGION}
+                isEnabled={showRegion}
+                mode={AnnotationMode.REGION}
```
ConradJChan, line 45: Instead of having to pass the `mode` as a prop into each button, what if we just declare that in the callback logic in this component? We could have separate callbacks like `handleRegionClick` or wrap `handleModeClick` in a closure with the mode?

src/lib/viewers/doc/DocBaseViewer.js
```diff
@@ -1051,6 +1051,7 @@ class DocBaseViewer extends BaseViewer {
 
     loadUIReact() {
         this.controls = new ControlsRoot({ containerEl: this.containerEl });
+        this.annotationControlsFSM.subscribe(() => this.renderUI());
```
ConradJChan, line 1054: Is the anonymous function needed? Would there be any benefit to combining the fsm callbacks?

src/lib/viewers/controls/annotations/AnnotationsControls.tsx
```diff
@@ -0,0 +1,65 @@
...
+        return null;
+    }
+
+    const handleModeClick = (mode: AnnotationMode): void => {
+        onAnnotationModeClick({ mode: annotationMode === mode ? AnnotationMode.NONE : mode });
+    };
+
+    return (
+        <div className="bp-AnnotationsControls">
+            <AnnotationsButton
```
ConradJChan, line 39: Is the `isEnabled` pattern preferred over doing a conditional render like `{ showRegion && <AnnotationsButton .../>`

src/lib/viewers/doc/DocBaseViewer.js
jstoffan: It's needed, otherwise I need to manually bind `this.renderUI` in the constructor. I think part of the value of the observer pattern is its support for distributed callbacks; combining them would be the equivalent of having a single container connected to a redux store for an entire app.

src/lib/viewers/controls/annotations/AnnotationsControls.tsx
jstoffan: I figure there will be more modes and buttons coming in the future, so this cuts down on the repetitive callbacks a bit.

src/lib/viewers/controls/annotations/AnnotationsButton.tsx
jstoffan: Not sure. This is how the controls are built today, so I'm just emulating that.

src/lib/viewers/controls/annotations/AnnotationsControls.tsx
jstoffan: I don't think there is a strong signal either way. If it's just a preference, then it seems okay as-is?

src/lib/viewers/controls/annotations/AnnotationsButton.tsx
jstoffan: It works on higher levels. I'll move it up. Thanks for the feedback!

src/lib/viewers/image/ImageViewer.js
```diff
@@ -390,8 +392,18 @@ class ImageViewer extends ImageBaseViewer {
...
         if (this.controls && this.options.useReactControls) {
+            const canAnnotate =
+                this.areNewAnnotationsEnabled() &&
+                this.hasAnnotationCreatePermission() &&
+                this.currentRotationAngle === 0;
+
             this.controls.render(
                 <ImageControls
+                    annotationMode={this.annotationControlsFSM.getMode()}
+                    hasHighlight={false}
```
mxiao6, line 403: Could we remove this property and set `false` directly on `AnnotationsControls` inside `ImageControls`?

mxiao6 (dismissed):

src/lib/viewers/image/ImageViewer.js
jstoffan: We can, but I figured keeping the two settings together was preferable. The `has` props default to false, anyway.

mxiao6 (approved):

PR #1293, opened 2020-11-17
feat(annotations): Update CSS pointer-events for drawing

mxiao6:


src/lib/viewers/doc/_docBase.scss
```diff
@@ -354,17 +354,16 @@ $thumbnail-sidebar-width: 226px;
         }
     }
 
-    &.bp-annotations-create--highlight {
+    &.bp-annotations-create--drawing,
```
ConradJChan, line 357: I wonder if adding `.bp-annotations-create--drawing` here is necessary, due to the layering, drawing creator will be on the top and shouldn't have any interference from the other annotation targets

ConradJChan (approved):

PR #1294, opened 2020-11-18
chore(annotations): Remove highlight text option

ConradJChan:


jstoffan (approved):

src/lib/viewers/doc/__tests__/DocBaseViewer-test.js
```diff
@@ -1702,6 +1702,7 @@ describe('src/lib/viewers/doc/DocBaseViewer', () => {
                 expect(docBase.controls.render).toBeCalledWith(
                     <DocControls
                         annotationMode="none"
+                        hasHighlight={false}
```
jstoffan, line 1705: Is this supposed to be here?

ConradJChan: I think the test passed before because `this.options.showAnnotationsHighlight` was undefined, so as a result what was L1070 in DocBaseViewer `const canHighlight = this.options.showAnnotationsHighlightText && canAnnotate && canDownload;` evaluated to undefined which allowed this test to pass previously but now it will always evaluate to `true` or `false` and that value will propagate to `DocControls`

PR #1300, opened 2020-12-01
feat(drawing): add color picker component

ChenCodes:
**Changes in this PR:**
- [x] add color palette component
- [x] add color control component
- [x] add unit tests
- [x] cross browser testing

**Demo:**
Screen Shot 2020-12-03 at 1 35 42 PM


src/lib/viewers/controls/color-picker/ColorPickerToggle.scss
```diff
@@ -0,0 +1,59 @@
...
+
+.bp-ColorPickerToggle {
+    position: relative;
+    display: flex;
+    align-items: center;
+    justify-content: center;
+    width: 48px;
+    height: 48px;
+
+    .bp-ColorPickerToggle-button {
```
jstoffan, line 12: The SUIT naming convention allows us to avoid nesting rules and increasing specificity. Let's flatten this file out a bit.

src/lib/viewers/controls/color-picker/ColorPickerPalette.scss
```diff
@@ -0,0 +1,21 @@
+.bp-ColorPickerPalette {
+    position: fixed;
+    display: flex;
+    flex-direction: row;
+    justify-content: center;
+    width: 146px;
```
jstoffan, line 6: We should try to avoid "magic" numbers. Can we allow the content to determine the width of the parent?

src/lib/viewers/controls/color-picker/ColorPickerPalette.scss
```diff
@@ -0,0 +1,21 @@
...
+        width: 20px;
+        height: 20px;
+        margin-right: 2px;
+        margin-left: 2px;
+        border: none;
+        border-radius: 4px;
+        cursor: pointer;
+
+        &:hover {
+            box-shadow: 0 0 0 2pt white;
```
jstoffan, line 18: We should stick to pixels for absolute units.

src/lib/viewers/controls/controls-bar/ControlsBar.scss
```diff
@@ -3,6 +3,7 @@
 .bp-ControlsBar {
     display: flex;
     align-items: center;
+    margin-right: 6px;
```
jstoffan, line 6: This should be applied via a sibling selector, if it's needed.

src/lib/viewers/controls/color-picker/ColorPickerToggle.tsx
```diff
@@ -0,0 +1,51 @@
+import React, { useState } from 'react';
+import classNames from 'classnames';
+import noop from 'lodash/noop';
+
+import ColorPickerPalette from './ColorPickerPalette';
+
+import { AnnotationMode } from '../annotations/types';
```
jstoffan, line 7: We should group all imports together, including type-only imports.

src/lib/viewers/controls/color-picker/ColorPickerToggle.scss
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    // &.bp-is-active {
+    //     svg {
+    //         background-color: $white;
+    //         fill: $black;
+    //     }
+    // }
+
+    .bp-ControlsBar {
```
jstoffan, line 36: We should try to avoid re-styling the control bars. Can we rely on the default styling and absolutely position the palette above the control bar?

src/lib/viewers/controls/color-picker/ColorPickerToggle.scss
jstoffan, line 12: Should these buttons be using the same mixins as the other existing buttons?

ChenCodes: Just to clarify, for this scenario, do you mean moving `.bp-ColorPickerToggle-button` to be on the same level as `.bp-ColorPickerToggle`?

For example:
```
.bp-ColorPickerToggle { ... }

.bp-ColorPickerToggle-button { ... }
```

I'm looking at the example of FeedCard.scss in https://confluence.inside-box.net/display/ETO/CSS+Class+Naming and it looks like that's how the flattening works.

jstoffan: Yep!

ChenCodes: Yep, they should.

src/lib/viewers/controls/color-picker/ColorPickerPalette.scss
ChenCodes: Yep we can do that instead.

src/lib/viewers/controls/color-picker/ColorPickerToggle.scss
ChenCodes: Yep I'll give that a try.

src/lib/viewers/controls/color-picker/ColorPickerPalette.tsx
```diff
@@ -0,0 +1,34 @@
...
+import ControlsBar from '../controls-bar';
+
+import './ColorPickerPalette.scss';
+
+export type Props = {
+    onColorSelect: () => void; // TODO: this method wil need to take in a color string param
+};
+
+export default function ColorPickerPalette({ onColorSelect = noop }: Props): JSX.Element {
+    const colors = ['#0061d5', '#26c281', '#ed3757', '#f5b31b', '#ffd700', '#4826c2'];
```
ChenCodes, line 13: These colors will be moved to a better location (AnnotationModule) after this change has been merged.

src/lib/viewers/controls/color-picker/ColorPickerToggle.scss
```diff
@@ -0,0 +1,21 @@
...
+    justify-content: center;
+    width: 48px;
+    height: 48px;
+}
+
+.bp-ColorPickerToggle-button {
+    @include bp-ControlButton;
+
+    svg {
+        background-color: #0061d5;
```
ChenCodes, line 17: I'll remove this line once we have the AnnotationModule business logic implemented.

src/lib/viewers/controls/color-picker/ColorPickerToggle.tsx
```diff
@@ -0,0 +1,40 @@
...
+
+    if (annotationMode !== AnnotationMode.DRAWING) {
+        return null;
+    }
+
+    return (
+        <div className="bp-ColorPickerToggle">
+            {isColorPickerToggled && (
+                <ColorPickerPalette
+                    onColorSelect={(): void => {
```
ChenCodes, line 23: Later when I implement the business logic for updating the selected color, the parameter will no longer be empty, we will be passing in `color: string` instead.

src/lib/viewers/controls/color-picker/ColorPickerPalette.tsx
```diff
@@ -0,0 +1,31 @@
...
+
+export type Props = {
+    onColorSelect: () => void; // TODO: this method wil need to take in a color string param
+};
+
+export default function ColorPickerPalette({ onColorSelect }: Props): JSX.Element {
+    const colors = ['#0061d5', '#26c281', '#ed3757', '#f5b31b', '#ffd700', '#4826c2'];
+
+    return (
+        <ControlsBar>
```
jstoffan, line 13: This component is already inside a `ControlsBar`. Is this needed?

src/lib/viewers/controls/color-picker/ColorPickerPalette.tsx
```diff
@@ -0,0 +1,31 @@
...
+            <div className="bp-ColorPickerPalette">
+                {colors.map(color => {
+                    return (
+                        <button
+                            key={color}
+                            className="bp-ColorPickerPalette-button"
+                            onClick={(): void => onColorSelect()}
```
jstoffan, line 20: We'll want to pass back the `color` value here.

src/lib/viewers/controls/color-picker/ColorPickerPalette.scss
```diff
@@ -0,0 +1,46 @@
+.bp-ColorPickerPalette {
+    position: absolute;
+    left: -55px;
```
jstoffan, line 3: I'm a little unsure what we're trying to achieve with this value. If the goal is to center the palette, consider using `left: -50%; transform: translateX(50%);` or some variation.

src/lib/viewers/controls/color-picker/ColorPickerPalette.scss
```diff
@@ -0,0 +1,46 @@
...
+    z-index: 1;
+    display: flex;
+    align-items: center;
+    justify-content: center;
+    height: 35px;
+    margin-bottom: 125px;
```
jstoffan, line 9: We should try to avoid this by using a `top` value that positions the palette above its control bar.

src/lib/viewers/controls/color-picker/ColorPickerPalette.scss
```diff
@@ -0,0 +1,46 @@
+.bp-ColorPickerPalette {
+    position: absolute;
+    left: -55px;
+    z-index: 1;
+    display: flex;
+    align-items: center;
+    justify-content: center;
+    height: 35px;
```
jstoffan, line 8: This is a little confusing. Is the palette itself only 35 pixels tall?

src/lib/viewers/controls/color-picker/ColorPickerPalette.scss
```diff
@@ -0,0 +1,46 @@
...
+    margin-bottom: 125px;
+    background: inherit;
+    border-radius: 5px;
```
jstoffan, line 11: Can we keep the border radii consistent with the other controls? Otherwise it will look out of place.

src/lib/viewers/controls/color-picker/ColorPickerPalette.scss
```diff
@@ -0,0 +1,46 @@
...
+
+    &::after {
```
jstoffan, line 13: Is this an arrow? Do any of our other popups have arrows? If not, should we omit this for now?

src/lib/viewers/controls/color-picker/ColorPickerToggle.tsx
```diff
@@ -0,0 +1,40 @@
...
+            )}
+            <button
+                className={classNames('bp-ColorPickerToggle-button', {
+                    'bp-is-active': isActive,
+                })}
+                onClick={(): void => setIsColorPickerToggled(!isColorPickerToggled)}
+                type="button"
+                {...rest}
+            >
+                <svg focusable="false" height="16" viewBox="0 0 16 16" width="16" />
```
jstoffan, line 36: Is this SVG needed? It's currently empty.

src/lib/viewers/controls/color-picker/ColorPickerToggle.tsx
```diff
@@ -0,0 +1,40 @@
...
+    const [isColorPickerToggled, setIsColorPickerToggled] = useState(false);
+
+    if (annotationMode !== AnnotationMode.DRAWING) {
+        return null;
+    }
+
+    return (
+        <div className="bp-ColorPickerToggle">
+            {isColorPickerToggled && (
+                <ColorPickerPalette
```
jstoffan, line 22: The palette should live outside of its toggle button. Otherwise, we will have nested buttons, which isn't valid markup.

src/lib/viewers/controls/color-picker/ColorPickerToggle.tsx
```diff
@@ -0,0 +1,40 @@
...
+import ColorPickerPalette from './ColorPickerPalette';
+import { AnnotationMode } from '../annotations/types';
+import './ColorPickerToggle.scss';
+
+export type Props = {
+    isActive?: boolean;
+    annotationMode?: AnnotationMode;
+};
+
+export default function ColorPickerToggle({ annotationMode, isActive = false, ...rest }: Props): JSX.Element | null {
```
jstoffan, line 12: This naming is very confusing, since a "toggle" in other controls is just a button. Maybe `ColorPickerControl` or something to that effect?

src/lib/viewers/controls/controls-bar/ControlsBar.scss
```diff
@@ -6,3 +6,7 @@
     background: fade-out($black, .2);
     border-radius: 3px;
 }
+
+.bp-ControlsBar ~ .bp-ControlsBar {
```
jstoffan, line 10:
```
.bp-ControlsBar {
    display: flex;
    align-items: center;
    background: fade-out($black, .2);
    border-radius: 3px;

    & + & {
        margin-left: 6px;
    }
}

File: src/lib/viewers/controls/color-picker/ColorPickerPalette.tsx (hunk 10 above)
Comment (ChenCodes) [reply src/lib/viewers/controls/color-picker/ColorPickerPalette.tsx:13]: Yes it looks like it's needed. When I remove the ControlsBar wrapper component, the `background: inherit` style is not being applied properly.

File: src/lib/viewers/controls/color-picker/ColorPickerPalette.scss (hunk 12 above)
Comment (ChenCodes) [reply src/lib/viewers/controls/color-picker/ColorPickerPalette.scss:3]: Good callout, it looks like we can go with 
```
right: 50%;
transform: translateX(50%);
```

File: src/lib/viewers/controls/color-picker/ColorPickerPalette.scss (hunk 14 above)
Comment (ChenCodes) [reply src/lib/viewers/controls/color-picker/ColorPickerPalette.scss:8]: The portion of the palette that doesn't include the arrow on the bottom should be 35px tall.

File: src/lib/viewers/controls/color-picker/ColorPickerPalette.scss (hunk 13 above)
Comment (ChenCodes) [reply src/lib/viewers/controls/color-picker/ColorPickerPalette.scss:9]: Sounds good, I can use this instead:

```
top: -50px;
```

File: src/lib/viewers/controls/color-picker/ColorPickerPalette.scss (hunk 15 above)
Comment (ChenCodes) [reply src/lib/viewers/controls/color-picker/ColorPickerPalette.scss:11]: Yep, will fix.

File: src/lib/viewers/controls/color-picker/ColorPickerPalette.scss (hunk 16 above)
Comment (ChenCodes) [reply src/lib/viewers/controls/color-picker/ColorPickerPalette.scss:13]: Yep this is an arrow. I checked in with design and design confirmed that even though this breaks consistency, for this specific component we would like to keep the arrow there.

File: src/lib/viewers/controls/color-picker/ColorPickerToggle.tsx (hunk 17 above)
Comment (ChenCodes) [reply src/lib/viewers/controls/color-picker/ColorPickerToggle.tsx:36]: This SVG is needed - it's for displaying the swatch shown in the toggle button.

File: src/lib/viewers/controls/color-picker/ColorPickerToggle.tsx (hunk 19 above)
Comment (ChenCodes) [reply src/lib/viewers/controls/color-picker/ColorPickerToggle.tsx:12]: Yes, I think that makes sense.

File: src/lib/viewers/controls/color-picker/ColorPickerPalette.scss (hunk 14 above)
Comment (jstoffan) [reply src/lib/viewers/controls/color-picker/ColorPickerPalette.scss:8]: We should allow the content to determine the size of the container. Can you post a screenshot?

File: src/lib/viewers/controls/color-picker/ColorPickerPalette.tsx (hunk 10 above)
Comment (jstoffan) [reply src/lib/viewers/controls/color-picker/ColorPickerPalette.tsx:13]: We should avoid nesting components solely for visual styling. If we need a particular background color, let's add a style for it to the `bp-ColorPickerPalette` element.

Comment (ChenCodes) [reply src/lib/viewers/controls/color-picker/ColorPickerPalette.tsx:13]: That makes sense, will fix.

File: src/lib/viewers/controls/color-picker/ColorPickerPalette.scss (hunk 14 above)
Comment (ChenCodes) [reply src/lib/viewers/controls/color-picker/ColorPickerPalette.scss:9]: Screen Shot 2020-12-02 at 5 13 06 PM

Comment (ChenCodes) [reply src/lib/viewers/controls/color-picker/ColorPickerPalette.scss:9]: Ah I see what you mean. I can just apply some margin to the top / bottom of the swatches so we do not need the 35px defined.

File: src/lib/viewers/controls/color-picker/ColorPickerPalette.scss [hunk 21]
```diff
@@ -0,0 +1,48 @@
...
+.bp-ColorPickerPalette-button {
+    width: 20px;
+    height: 20px;
+    margin: 10px 2px 10px 2px;
+    padding: 0;
+    border: none;
+    border-radius: 4px;
+    cursor: pointer;
+
+    &:first-child {
```
jstoffan, line 37: Can we add left/right padding to the parent element, instead?

src/lib/viewers/controls/color-picker/ColorPickerPalette.scss
```diff
@@ -0,0 +1,48 @@
...
+        transform: rotate(-45deg);
+        content: '';
+        clip-path: polygon(0% 0%, 100% 100%, 0% 100%);
+    }
+}
+
+.bp-ColorPickerPalette-button {
+    width: 20px;
+    height: 20px;
+    margin: 10px 2px 10px 2px;
```
jstoffan, line 31: This can be simplified to `margin: 10px 2px;`

src/lib/viewers/controls/color-picker/ColorPickerPalette.scss
```diff
@@ -0,0 +1,48 @@
+@import '~box-ui-elements/es/styles/variables';
+
+.bp-ColorPickerPalette {
+    position: absolute;
+    top: -55px;
```
jstoffan, line 5: How is this value derived? Can we use a relative value, instead?

src/lib/viewers/controls/color-picker/ColorPickerPalette.scss
```diff
@@ -0,0 +1,48 @@
...
+    border: none;
+    border-radius: 4px;
+    cursor: pointer;
+
+    &:first-child {
+        margin-left: 9px;
+    }
+
+    &:hover {
+        box-shadow: 0 0 0 2px white;
```
jstoffan, line 42: We should try to avoid named CSS colors. Can we use `$white` or `#fff`, instead?

src/lib/viewers/controls/color-picker/__tests__/ColorPickerControl-test.tsx
```diff
@@ -0,0 +1,41 @@
...
+    const getWrapper = (props = {}): ShallowWrapper =>
+        shallow(<ColorPickerControl onAnnotationColorClick={onAnnotationColorClick} {...props} />);
+
+    const getToggleButton = (wrapper: ShallowWrapper): ShallowWrapper => wrapper.find('.bp-ColorPickerControl-button');
+
+    describe('render', () => {
+        test('should render null if annotationMode is not AnnotationMode.DRAWING', () => {
+            const wrapper = getWrapper({ annotationMode: AnnotationMode.REGION });
+
+            expect(wrapper.find('.bp-ColorPickerControl').length).toBe(0);
```
jstoffan, line 18: Let's use `.exists` rather than checking the length. We also have `.isEmptyRender` for null cases.

src/lib/viewers/controls/color-picker/ColorPickerControl.scss
```diff
@@ -0,0 +1,21 @@
+@import '~box-ui-elements/es/styles/variables';
```
jstoffan, line 1: I don't think this is needed since `../styles` is imported below.

src/lib/viewers/controls/color-picker/ColorPickerControl.scss
```diff
@@ -0,0 +1,21 @@
...
+    height: 48px;
+}
+
+.bp-ColorPickerControl-button {
+    @include bp-ControlButton;
+
+    svg {
+        background-color: #0061d5;
+        border-radius: 2px;
+        box-shadow: 0 0 0 2px white;
```
jstoffan, line 19: This will create a double shadow on hover. Is that desirable?

src/lib/viewers/controls/color-picker/ColorPickerControl.scss
```diff
@@ -0,0 +1,21 @@
...
+    justify-content: center;
+    width: 48px;
+    height: 48px;
+}
+
+.bp-ColorPickerControl-button {
+    @include bp-ControlButton;
+
+    svg {
+        background-color: #0061d5;
```
jstoffan, line 17: Should we use a variable here?

src/lib/viewers/controls/color-picker/ColorPickerToggle.tsx
jstoffan: Can the background color of the button itself be the "swatch"?

src/lib/viewers/controls/color-picker/ColorPickerPalette.scss
```diff
@@ -0,0 +1,48 @@
...
+
+.bp-ColorPickerPalette {
+    position: absolute;
+    top: -55px;
+    right: 50%;
+    z-index: 1;
+    display: flex;
+    align-items: center;
+    justify-content: center;
+    background: fade-out($black, .2);
```
jstoffan, line 11: We should probably move this to a shared variable since it's used in ControlsBar, as well.

src/lib/viewers/controls/color-picker/ColorPickerControl.scss
ChenCodes: Yep I'll use $box-blue.

src/lib/viewers/controls/color-picker/ColorPickerControl.scss
ChenCodes: I don't think there will be an issue - the box-shadow here is what gives the color swatch the white border. I'm guessing the other box-shadow that you're referring to comes from the mixin that is imported above this block. That box shadow is for the entire button element so the effect that it has shouldn't cause any conflicts with the box-shadow defined in this svg element.

src/lib/viewers/controls/color-picker/ColorPickerPalette.scss
ChenCodes: Sounds good.

src/lib/viewers/controls/color-picker/ColorPickerPalette.scss
ChenCodes: Yep, I can do a combination of a few other element values.

src/lib/viewers/controls/color-picker/ColorPickerControl.scss
jstoffan: Ah, okay. Should we use `border: 2px solid #fff;` here if the shadow is solid?

src/lib/viewers/controls/color-picker/ColorPickerPalette.scss
jstoffan: We should try to avoid using any absolute values, if possible. Since this is absolutely-positioned, we should be able to use a relative (percentage) value along with a transform, similar to the `right` calculation.

jstoffan: Otherwise, if the goal is to offset the element above its parent, can we use `bottom: 0` and allow the arrow to push the palette up as needed?

src/lib/viewers/controls/color-picker/ColorPickerToggle.tsx
ChenCodes: I think that complicates things a bit in terms of how the styling would work. For example, if I set the background color of the button to be the swatch, I would need to apply some padding so that the swatch in the middle of the button has the background color - the padding portion would have a different background color. We would also need to find a way to apply some kind of inner border that is on the inner edge of the padding and outer edge of the swatch.

I would prefer keeping the svg element here for the swatch implementation but am open to suggestions for how we would be able to have a clean implementation for setting the background color of the button to be the swatch.

src/lib/viewers/controls/color-picker/ColorPickerControl.scss
ChenCodes: Yep, that sounds good.

src/lib/viewers/controls/color-picker/ColorPickerPalette.scss
ChenCodes: Would using `bottom: 135%` suffice? It looks like that achieves the desired result without the usage of a transform.
Screen Shot 2020-12-03 at 11 18 28 AM

src/lib/viewers/controls/color-picker/ColorPickerToggle.tsx
jstoffan: Can we use a div for the inner element? What value does the SVG add?

src/lib/viewers/controls/color-picker/ColorPickerPalette.scss
```diff
@@ -0,0 +1,43 @@
...
+    align-items: center;
+    justify-content: center;
+    padding: 0 8px;
+    background: $controls-background;
+    border-radius: 4px;
+    transform: translateX(50%);
+
+    &::after {
+        position: absolute;
+        top: 75%;
```
jstoffan, line 20: Can we use the following?

```
bottom: 0;
transform: translateY(100%) rotate(-45deg);?
```

src/lib/viewers/controls/color-picker/ColorPickerPalette.scss
```diff
@@ -0,0 +1,43 @@
...
+.bp-ColorPickerPalette-button {
+    width: 20px;
+    height: 20px;
+    margin: 10px 2px;
+    padding: 0;
+    border: none;
+    border-radius: 4px;
+    cursor: pointer;
+
+    &:hover {
```
jstoffan, line 40: How about focus?

src/lib/viewers/controls/color-picker/ColorPickerPalette.scss
```diff
@@ -0,0 +1,43 @@
...
+        box-shadow: 0 0 0 2px $white;
```
jstoffan, line 41: Do we need to disable the outline, as well?

src/lib/viewers/controls/color-picker/ColorPickerControl.scss
```diff
@@ -0,0 +1,22 @@
...
+    align-items: center;
+    justify-content: center;
+    width: $color-picker-control-height;
+    height: $color-picker-control-height;
+}
+
+.bp-ColorPickerControl-button {
+    @include bp-ControlButton;
+
+    svg {
```
jstoffan, line 17: We should avoid styling via tag name, when possible.

src/lib/viewers/controls/color-picker/ColorPickerPalette.scss
```diff
@@ -0,0 +1,43 @@
...
+.bp-ColorPickerPalette {
+    position: absolute;
+    top: calc(-1 * (#{$controls-bar-bottom-offset} + #{$color-picker-control-height / 2} + 10px)); // 10px is the padding between the popup and the rest of the control bar
+    right: 50%;
+    z-index: 1;
+    display: flex;
+    align-items: center;
+    justify-content: center;
+    padding: 0 8px;
+    background: $controls-background;
```
jstoffan, line 14: SCSS variables are global and should be namedspaced to avoid collisions in case they're imported for some reason. Let's use `$bp-controls-background`?

src/lib/viewers/controls/color-picker/ColorPickerPalette.scss
```diff
@@ -0,0 +1,43 @@
+@import '~box-ui-elements/es/styles/variables';
+@import './ColorPickerControl';
+@import '../../../Controls';
+
+.bp-ColorPickerPalette {
+    position: absolute;
+    top: calc(-1 * (#{$controls-bar-bottom-offset} + #{$color-picker-control-height / 2} + 10px)); // 10px is the padding between the popup and the rest of the control bar
```
jstoffan, line 7: Let's sync up offline about this change.

src/lib/Controls.scss
```diff
@@ -1,14 +1,16 @@
+$controls-background: fade-out($black, .2);
```
jstoffan, line 1: This is the legacy controls stylesheet. We should not need to modify it.

src/lib/viewers/controls/color-picker/ColorPickerPalette.scss
```diff
@@ -0,0 +1,43 @@
+@import '~box-ui-elements/es/styles/variables';
+@import './ColorPickerControl';
+@import '../../../Controls';
```
jstoffan, line 3: We should avoid importing other component stylesheets. Can we move the common variable to `../styles`?

src/lib/viewers/controls/color-picker/ColorPickerControl.scss
ChenCodes: Actually, on second thought, we can't use `border` here. The reason is `border` takes up the space within the bounds of the element when drawing the border whereas `box-shadow` behaves more like `outline`. I initially tried to use `outline` but then had trouble applying a border radius to the `outline`, and the solution that I found was that we could use `box-shadow` with a `border-radius` to achieve the outline effect while also having the border radius effect.

src/lib/viewers/controls/color-picker/ColorPickerControl.scss
ChenCodes: I'll add a class for it.

src/lib/viewers/controls/color-picker/ColorPickerPalette.scss
ChenCodes: Yep, we should apply it on focus as well.

src/lib/viewers/controls/color-picker/ColorPickerPalette.scss
ChenCodes: Yes we do.

src/lib/viewers/controls/color-picker/ColorPickerToggle.tsx
ChenCodes: Yep, we can use the div instead of the svg, I just need an element there for the swatch.

src/lib/viewers/controls/color-picker/ColorPickerControl.scss
```diff
@@ -0,0 +1,44 @@
+@import '../styles';
+
+.bp-ColorPickerControl {
+    position: relative;
+    display: flex;
+    align-items: center;
+    justify-content: center;
+    width: 48px;
```
ChenCodes, line 8: Don't need this and the below line.

src/lib/viewers/controls/color-picker/ColorPickerControl.scss
```diff
@@ -0,0 +1,44 @@
...
+        border-top: $arrow-height solid $bp-controls-background;
+        border-right: $arrow-height solid transparent;
+        border-left: $arrow-height solid transparent;
+        transform: translateX(-50%);
+        content: '';
+    }
+}
+
+.bp-ColorPickerControl-swatch {
+    width: 16px;
```
jstoffan, line 39: Should this be 18x18 or 20x20?

jstoffan (commented):
LGTM

src/lib/viewers/controls/color-picker/ColorPickerControl.scss
ChenCodes: I believe 16x16 is correct.

Screen Shot 2020-12-03 at 1 30 25 PM

ChenCodes: Yes, it should be 18x18, the border is taking up part of the space.

src/lib/viewers/controls/color-picker/ColorPickerPalette.scss
```diff
@@ -0,0 +1,27 @@
+@import '~box-ui-elements/es/styles/variables';
```
jstoffan, line 1: Not needed, since `../styles` is imported below.

src/lib/viewers/controls/color-picker/ColorPickerControl.tsx
```diff
@@ -0,0 +1,49 @@
+import React, { useState } from 'react';
+import classNames from 'classnames';
+import ColorPickerPalette from './ColorPickerPalette';
+import { AnnotationMode } from '../annotations/types';
+import './ColorPickerControl.scss';
+
+export type Props = {
+    annotationMode?: AnnotationMode;
+    onAnnotationColorClick: (color: string) => void;
+    isActive?: boolean;
```
ConradJChan, line 10: nit: 🔤

src/lib/viewers/controls/color-picker/ColorPickerControl.tsx
```diff
@@ -0,0 +1,49 @@
...
+
+export default function ColorPickerControl({
+    annotationMode,
+    isActive = false,
+    onAnnotationColorClick,
+    ...rest
+}: Props): JSX.Element | null {
+    const [isColorPickerToggled, setIsColorPickerToggled] = useState(false);
+
+    if (annotationMode !== AnnotationMode.DRAWING) {
```
ConradJChan, line 21: Does this picker need to know about annotations at all? Could the rendering of this `ColorPickerControl` be controlled by the consuming component?

src/lib/viewers/controls/color-picker/ColorPickerControl.tsx
```diff
@@ -0,0 +1,49 @@
...
+                    />
+                </div>
+            )}
+            <button
+                className={classNames('bp-ColorPickerControl-button', {
+                    'bp-is-active': isActive,
+                })}
+                onClick={(): void => setIsColorPickerToggled(!isColorPickerToggled)}
+                type="button"
+                {...rest}
```
ConradJChan, line 43: Seems a bit odd to spread the `rest` props onto a nested elements instead of the root element. Is this needed for something?

src/lib/viewers/controls/color-picker/ColorPickerPalette.tsx
```diff
@@ -0,0 +1,28 @@
+import React from 'react';
+import './ColorPickerPalette.scss';
+
+export type Props = {
+    onColorSelect: (color: string) => void;
+};
+
+export default function ColorPickerPalette({ onColorSelect }: Props): JSX.Element {
+    const colors = ['#0061d5', '#26c281', '#ed3757', '#f5b31b', '#ffd700', '#4826c2'];
```
ConradJChan, line 9: Should these colors be moved outside of the function?

src/lib/viewers/controls/color-picker/ColorPickerControl.tsx
```diff
@@ -0,0 +1,49 @@
...
+                <div className="bp-ColorPickerControl-palette">
+                    <ColorPickerPalette
+                        onColorSelect={(color: string): void => {
+                            setIsColorPickerToggled(false);
+                            onAnnotationColorClick(color);
+                        }}
+                    />
+                </div>
+            )}
+            <button
```
ConradJChan, line 37: Does this need a resin tag and/or testid attributes? Similarly for the `ColorPickerPalette`

src/lib/viewers/doc/DocControls.tsx
```diff
@@ -35,31 +38,36 @@ export default function DocControls({
...
+                <AnnotationsControls
+                    annotationMode={annotationMode}
+                    hasDrawing={hasDrawing}
+                    hasHighlight={hasHighlight}
+                    hasRegion={hasRegion}
+                    onAnnotationModeClick={onAnnotationModeClick}
+                    onAnnotationModeEscape={onAnnotationModeEscape}
+                />
+            </ControlsBar>
+            <ControlsBar>
```
ConradJChan, line 68: Should this be added to the `ImageControls` as well?

ConradJChan (commented):
Sorry, was late to the game. Just a few comments for future revisions

PR #1305, opened 2020-12-04
Add annotation module

ChenCodes:
**Changes in this PR:**
- [x] add module to handle business logic for annotation color updates (via localStorage)
- [x] cross browser testing
- [x] unit tests

**Demo:**
ezgif com-gif-maker (8)


src/lib/AnnotationModule.ts
```diff
@@ -0,0 +1,50 @@
...
+}
+
+type Subscription = (state: AnnotationColor) => void;
+
+export default class AnnotationModule {
+    private cache: Cache;
+
+    private subscriptions: Subscription[] = [];
+
+    constructor(cache = new Cache()) {
```
ChenCodes, line 20: Not sure if I should be instantiating a new cache in this constructor - I added this here because there was a type error that was showing up. I don't think we'll ever be using the new cache as there should be a cache being passed in from BaseViewer.

src/lib/viewers/controls/annotations/AnnotationsControls.tsx
```diff
@@ -62,18 +65,20 @@ export default function AnnotationsControls({
...
                 data-testid="bp-AnnotationsControls-drawBtn"
-                isActive={annotationMode === AnnotationMode.DRAWING}
+                isActive={isDrawingActive}
                 isEnabled={showDrawing}
                 mode={AnnotationMode.DRAWING}
                 onClick={handleModeClick}
                 title={__('drawing_comment')}
             >
-                <IconDrawing24 />
+                <IconDrawing24 style={{ fill: isDrawingActive ? annotationColor : '#fff' }} />
```
ChenCodes, line 81: Maybe this can be moved to a constant? It seems a little hardcoded here.

src/lib/viewers/controls/color-picker/ColorPickerPalette.tsx
```diff
@@ -1,12 +1,20 @@
...
 import './ColorPickerPalette.scss';
 
 export type Props = {
-    onColorSelect: (color: string) => void;
+    onColorSelect: (color: AnnotationColor) => void;
 };
 
 export default function ColorPickerPalette({ onColorSelect }: Props): JSX.Element {
-    const colors = ['#0061d5', '#26c281', '#ed3757', '#f5b31b', '#ffd700', '#4826c2'];
+    const colors = [
```
ChenCodes, line 10: Should `colors` be passed in as a prop for ColorPickerPalette? That way it'll make this component a lot more reusable for other devs who have other colors that want to use, not just these 6.

src/lib/viewers/image/ImageControls.tsx
```diff
@@ -26,6 +32,7 @@ export default function ImageControls({
             <RotateControl onRotateLeft={onRotateLeft} />
             <FullscreenToggle onFullscreenToggle={onFullscreenToggle} />
             <AnnotationsControls
+                annotationColor={annotationColor}
```
ChenCodes, line 35: I haven't added the business logic yet for ImageControls (as well as some of the UI for showing the ColorPickerControl for an image file). Would it be okay for me to include that within this PR as well?

src/lib/AnnotationModule.ts
```diff
@@ -0,0 +1,50 @@
...
+        this.cache = cache;
+
+        if (!this.cache.get(ANNOTATION_COLOR_KEY)) {
+            this.cache.set(ANNOTATION_COLOR_KEY, AnnotationColor.BLUE, true);
+        }
+    }
+
+    public getColor = (): AnnotationColor => this.cache.get(ANNOTATION_COLOR_KEY) as AnnotationColor;
+
+    public transition = (color: AnnotationColor): AnnotationColor => {
```
jstoffan, line 30: This seems to perpetuate the issue we have with AnnotationsFSM, in that it isn't extensible. It also duplicates quite a bit of logic around subscriptions. Is this much complexity needed for one property? If so, I think we'll want to consider a more holistic solution. Should migrate to redux?

src/lib/viewers/controls/annotations/AnnotationsControls.tsx
jstoffan: Why does this icon have different behavior and visual treatment from the others?

src/lib/viewers/controls/color-picker/ColorPickerPalette.tsx
jstoffan, line 10: Does this work? `const colors = Object.values(AnnotationColor);`?

jstoffan: If so, does it need to be redefined per-render?

src/lib/viewers/image/ImageControls.tsx
```diff
@@ -1,13 +1,19 @@
 import React from 'react';
 import AnnotationsControls, { Props as AnnotationsControlsProps } from '../controls/annotations';
+import { Props as ColorPickerControlProps } from '../controls/color-picker';
```
jstoffan, line 3: Can this be exposed via the AnnotationsControlsProps, instead?

src/lib/viewers/controls/annotations/AnnotationsControls.tsx
ChenCodes: In the designs, it looks like whenever the drawing button is active, the fill color matches the color in the ColorPickerControl swatch.

 
Screen Shot 2020-12-04 at 4 18 03 PM

src/lib/AnnotationModule.ts
ChenCodes: I think we may not need to migrate to redux. Instead, for this case, whenever the user updates the current selected color, we can just call `annotator.emit` to propagate the latest color down to annotations repo.

ChenCodes: We no longer need the subscribe / transition logic if we use the `annotator.emit` approach.

src/lib/viewers/controls/annotations/AnnotationsControls.tsx
ChenCodes: Another reason is that with the introduction of drawing annotations, there is an added context of a color that the user has set (whereas region / highlight annotations are color agnostic).

src/lib/viewers/controls/color-picker/ColorPickerPalette.tsx
ChenCodes: Yes, the above suggestion works, and no, we can define it just once.

src/lib/viewers/image/ImageControls.tsx
ChenCodes: There are two props that are referenced in ColorPickerControlProps that are not referenced in AnnotationsControlsProps:
`isActive` and `onAnnotationColorClick`. Since AnnotationsControlsProps properties is not a superset of all of the props of ColorPickerControlProps, I don't think we can remove the above line.

src/lib/AnnotationModule.ts
ChenCodes: As for the `annotator.emit` approach, I think that will be done in a separate PR (it makes more sense to add it when we have the drawing logic already merged so that we can test that when we update the color, we do see that a new drawing is using that updated color).

src/lib/AnnotationModule.ts
```diff
@@ -0,0 +1,27 @@
+import { ANNOTATION_COLOR_KEY } from './constants';
```
ConradJChan, line 1: default imports before named imports

src/lib/AnnotationModule.ts
```diff
@@ -0,0 +1,27 @@
...
+import Cache from './Cache';
+
+export enum AnnotationColor {
+    BLUE = '#0061d5',
```
ConradJChan, line 5: Do these exist as variables in BUIE?

src/lib/AnnotationModule.ts
ConradJChan: Should this be an options object so it could be more easily extended in the future? Maybe could define the `type Options`

src/lib/AnnotationModule.ts
```diff
@@ -0,0 +1,27 @@
...
+}
+
+export default class AnnotationModule {
+    private cache: Cache;
+
+    constructor(cache = new Cache()) {
+        this.cache = cache;
+
+        if (!this.cache.get(ANNOTATION_COLOR_KEY)) {
+            this.cache.set(ANNOTATION_COLOR_KEY, AnnotationColor.BLUE, true);
```
ConradJChan, line 20: Could this call `this.setColor`?

src/lib/viewers/controls/annotations/AnnotationsControls.tsx
ConradJChan: Is it just me or is there an extra white border on the left of the active state of the drawing button above?

src/lib/viewers/controls/annotations/AnnotationsControls.tsx
```diff
@@ -5,10 +5,12 @@ import IconDrawing24 from '../icons/IconDrawing24';
 import IconHighlightText16 from '../icons/IconHighlightText16';
 import IconRegion24 from '../icons/IconRegion24';
 import useFullscreen from '../hooks/useFullscreen';
+import { AnnotationColor } from '../../../AnnotationModule';
 import { AnnotationMode } from './types';
 import './AnnotationsControls.scss';
 
 export type Props = {
+    annotationColor: AnnotationColor;
```
ConradJChan, line 13: Should this be an optional prop with a default?

src/lib/viewers/controls/color-picker/ColorPickerControl.tsx
```diff
@@ -1,16 +1,19 @@
 import React, { useState } from 'react';
 import classNames from 'classnames';
 import ColorPickerPalette from './ColorPickerPalette';
+import { AnnotationColor } from '../../../AnnotationModule';
 import { AnnotationMode } from '../annotations/types';
 import './ColorPickerControl.scss';
 
 export type Props = {
+    annotationColor: AnnotationColor;
     annotationMode?: AnnotationMode;
```
ConradJChan, line 10: Had left the comment on a previous PR, but does `annotationMode` need to live here? A color picker doesn't seem to need to know about annotations at all

src/lib/viewers/controls/color-picker/ColorPickerPalette.tsx
```diff
@@ -1,13 +1,14 @@
 import React from 'react';
+import { AnnotationColor } from '../../../AnnotationModule';
 import './ColorPickerPalette.scss';
 
+const colors = Object.values(AnnotationColor);
+
 export type Props = {
-    onColorSelect: (color: string) => void;
+    onColorSelect: (color: AnnotationColor) => void;
```
ConradJChan, line 8: In the same vein of keeping the color picker palette annotations agnostic, could the colors be passed in as a prop?

src/lib/AnnotationModule.ts
ChenCodes: Yes, I'll reference those instead.

src/lib/viewers/controls/annotations/AnnotationsControls.tsx
ChenCodes: I think that might be a bug in zeplin, in this new screenshot I no longer see it.
Screen Shot 2020-12-07 at 1 00 07 PM

src/lib/viewers/controls/color-picker/ColorPickerControl.tsx
ChenCodes: I can remove it from this component and perform the null check in DocControls instead.

src/lib/viewers/controls/color-picker/ColorPickerPalette.tsx
ChenCodes: Yes, that sounds good.

src/lib/viewers/controls/annotations/AnnotationsControls.tsx
ChenCodes: Yes I think it would be safer to make it optional + add a default just in case annotationColor is somehow undefined.

src/lib/AnnotationModule.ts
ChenCodes: Yep that sounds good.

src/lib/viewers/controls/annotations/AnnotationsControls.tsx
mxiao6: Should we move this logic to upper level so it will become `<IconDrawing24 style={{ fill: annotationColor }} />`

src/lib/viewers/controls/annotations/__tests__/AnnotationsControls-test.tsx
```diff
@@ -1,11 +1,13 @@
-import React from 'react';
+import * as React from 'react';
```
mxiao6, line 1: Is this change necessary?

src/lib/viewers/controls/color-picker/ColorPickerControl.tsx
```diff
@@ -1,33 +1,32 @@
...
-    }
-
     return (
         <div className="bp-ColorPickerControl">
             {isColorPickerToggled && (
                 <div className="bp-ColorPickerControl-palette">
                     <ColorPickerPalette
-                        onColorSelect={(color: string): void => {
+                        colors={colors}
+                        onColorSelect={(color): void => {
```
mxiao6, line 29: Is return type necessary? Should we declare this function outside instead of inline?

src/lib/viewers/controls/color-picker/ColorPickerPalette.tsx
```diff
@@ -1,13 +1,13 @@
 import React from 'react';
+import { AnnotationColor } from '../../../AnnotationModule';
 import './ColorPickerPalette.scss';
 
 export type Props = {
-    onColorSelect: (color: string) => void;
+    colors: Array<AnnotationColor>;
+    onColorSelect: (color: AnnotationColor) => void;
```
mxiao6, line 7: How about `onSelect`?

src/lib/viewers/doc/DocControls.tsx
```diff
@@ -65,9 +67,14 @@ export default function DocControls({
                     onAnnotationModeEscape={onAnnotationModeEscape}
                 />
             </ControlsBar>
-            <ControlsBar>
-                <ColorPickerControl annotationMode={annotationMode} onAnnotationColorClick={onAnnotationColorClick} />
-            </ControlsBar>
+            {annotationMode === 'drawing' && (
```
mxiao6, line 70: Should we use a constant instead of string?

src/lib/viewers/controls/annotations/__tests__/AnnotationsControls-test.tsx
ConradJChan: The changes to this file should be unnecessary now since `annotationColor` is optional and defaulted

src/lib/viewers/controls/color-picker/ColorPickerControl.tsx
```diff
@@ -1,33 +1,32 @@
...
-    annotationMode?: AnnotationMode;
-    onAnnotationColorClick: (color: string) => void;
+    annotationColor?: string;
     isActive?: boolean;
+    onAnnotationColorClick: (color: string) => void;
 };
 
 export default function ColorPickerControl({
-    annotationMode,
+    annotationColor = AnnotationColor.BOX_BLUE,
```
ConradJChan, line 16: Could this be from BUIE instead so we can avoid having a dependency on `AnnotationModule`?

src/lib/viewers/controls/color-picker/ColorPickerControl.tsx
```diff
@@ -1,33 +1,32 @@
...
 import './ColorPickerControl.scss';
 
+const colors = Object.values(AnnotationColor);
+
 export type Props = {
-    annotationMode?: AnnotationMode;
-    onAnnotationColorClick: (color: string) => void;
+    annotationColor?: string;
     isActive?: boolean;
+    onAnnotationColorClick: (color: string) => void;
```
ConradJChan, line 12: `onColorSelect` instead?

src/lib/viewers/controls/annotations/AnnotationsControls.tsx
ChenCodes: Could you clarify a bit about what you mean by upper level? Do you mean instead of the ternary inside of the style prop, I do a ternary with the icon?

`isDrawingActive ? <IconDrawing24 style={{ fill: annotationColor }} /> : <IconDrawing24 style={{ fill: '#fff' }} />`

src/lib/viewers/controls/annotations/__tests__/AnnotationsControls-test.tsx
ChenCodes: Yep, will revert.

src/lib/viewers/controls/color-picker/ColorPickerControl.tsx
ChenCodes: I don't think the return type is necessary - yes I can declare it outside as it will be a bit more readable.

src/lib/viewers/controls/color-picker/ColorPickerPalette.tsx
ChenCodes: Yeah that works, color doesn't seem as necessary because this component doesn't have anything else that needs to be selected.

src/lib/viewers/controls/color-picker/ColorPickerControl.tsx
ChenCodes: Yep that's a good point, will fix.

src/lib/viewers/controls/annotations/__tests__/AnnotationsControls-test.tsx
```diff
@@ -2,10 +2,12 @@ import React from 'react';
 import { act } from 'react-dom/test-utils';
 import { ReactWrapper, mount } from 'enzyme';
 import AnnotationsControls from '../AnnotationsControls';
+import { AnnotationColor } from '../../../../AnnotationModule';
 import { AnnotationMode } from '../types';
 
 describe('AnnotationsControls', () => {
-    const getWrapper = (props = {}): ReactWrapper => mount(<AnnotationsControls {...props} />);
+    const getWrapper = (props = {}): ReactWrapper =>
```
ConradJChan, line 9: Since `annotationColor` is optional and defaulted, is this change needed?

src/lib/AnnotationModule.ts
```diff
@@ -0,0 +1,39 @@
...
+    bdlGreenLight,
+    bdlWatermelonRed,
+    bdlYellorange,
+    bdlYellow,
+    bdlGrimace,
+} from 'box-ui-elements/es/styles/variables';
+import { ANNOTATION_COLOR_KEY } from './constants';
+import Cache from './Cache';
+
+export const AnnotationColor = {
```
ConradJChan, line 12: Should this move to a constants file? Then we could avoid having to keep importing from `AnnotationsModule` in other files that aren't necessarily related. Maybe we need to have a directory for annotations like
```
src/lib/annotations
src/lib/annotations/AnnotationModule.ts
src/lib/annotations/constants.ts
src/lib/annotations/index.ts
```
Thoughts?

src/lib/viewers/controls/color-picker/ColorPickerControl.tsx
```diff
@@ -1,37 +1,36 @@
 import React, { useState } from 'react';
 import classNames from 'classnames';
+import { bdlBoxBlue } from 'box-ui-elements/es/styles/variables';
 import ColorPickerPalette from './ColorPickerPalette';
-import { AnnotationMode } from '../annotations/types';
+import { AnnotationColor } from '../../../AnnotationModule';
 import './ColorPickerControl.scss';
 
+const colors = Object.values(AnnotationColor);
```
ConradJChan, line 8: Instead of importing the colors here and passing them in, should they pass through from `ColorPickerControl` down to `ColorPickerPalette`? So essentially it would be `(Doc|Image)Controls` that imports the colors and passes it down

src/lib/viewers/doc/DocBaseViewer.js
```diff
@@ -111,11 +111,11 @@ class DocBaseViewer extends BaseViewer {
         this.emitMetric = this.emitMetric.bind(this);
         this.handleAssetAndRepLoad = this.handleAssetAndRepLoad.bind(this);
         this.handleFindBarClose = this.handleFindBarClose.bind(this);
-        this.handleAnnotationColorClick = this.handleAnnotationColorClick.bind(this);
         this.handleAnnotationControlsClick = this.handleAnnotationControlsClick.bind(this);
         this.handleAnnotationControlsEscape = this.handleAnnotationControlsEscape.bind(this);
         this.handleAnnotationCreateEvent = this.handleAnnotationCreateEvent.bind(this);
         this.handleAnnotationCreatorChangeEvent = this.handleAnnotationCreatorChangeEvent.bind(this);
+        this.handleColorSelect = this.handleColorSelect.bind(this);
```
ConradJChan, line 118: For consistency seems like `handleAnnotationColorSelect` is more appropriate

src/lib/viewers/doc/DocControls.tsx
```diff
@@ -17,13 +18,14 @@ export type Props = AnnotationsControlsProps &
...
 export default function DocControls({
+    annotationColor,
     annotationMode,
     hasDrawing,
     hasHighlight,
     hasRegion,
     maxScale,
     minScale,
-    onAnnotationColorClick,
+    onColorSelect,
```
ConradJChan, line 28: `onAnnotationColorSelect`?

src/lib/constants.js
```diff
@@ -157,5 +157,6 @@ export const METADATA = {
 export const ERROR_CODE_403_FORBIDDEN_BY_POLICY = 'forbidden_by_policy';
 
 // LocalStorage Keys
+export const ANNOTATION_COLOR_KEY = 'annotation-color';
```
jstoffan, line 160: If this is only used in one file, can we encapsulate it there?

src/lib/viewers/BaseViewer.js
```diff
@@ -153,6 +154,10 @@ class BaseViewer extends EventEmitter {
 
         this.annotationControlsFSM = new AnnotationControlsFSM();
 
+        if (this.options.showAnnotationsDrawingCreate) {
```
jstoffan, line 157: Can we always instantiate the module to avoid existence checks every time we want to use it?

src/lib/viewers/controls/annotations/AnnotationsControls.tsx
```diff
@@ -5,10 +5,12 @@ import IconDrawing24 from '../icons/IconDrawing24';
 import IconHighlightText16 from '../icons/IconHighlightText16';
 import IconRegion24 from '../icons/IconRegion24';
 import useFullscreen from '../hooks/useFullscreen';
+import { AnnotationColor } from '../../../AnnotationModule';
 import { AnnotationMode } from './types';
 import './AnnotationsControls.scss';
 
 export type Props = {
+    annotationColor?: string;
```
jstoffan, line 13: Should this be `AnnotationColor` instead of string?

src/lib/viewers/controls/annotations/AnnotationsControls.tsx
jstoffan: Can `IconDrawing` accept `fill` as a prop instead of using inline styles?

src/lib/viewers/controls/color-picker/ColorPickerControl.tsx
```diff
@@ -1,37 +1,36 @@
...
-    annotationMode?: AnnotationMode;
-    onAnnotationColorClick: (color: string) => void;
+    annotationColor?: string;
     isActive?: boolean;
+    onColorSelect: (color: string) => void;
 };
 
 export default function ColorPickerControl({
-    annotationMode,
+    annotationColor = bdlBoxBlue,
```
jstoffan, line 17: If this component is meant to be generic, should this be renamed `activeColor`?

src/lib/viewers/controls/color-picker/ColorPickerControl.tsx
```diff
@@ -1,37 +1,36 @@
...
-    }
+    const handleSelect = (color: string): void => {
+        setIsColorPickerToggled(false);
+        onColorSelect(color);
+    };
 
     return (
         <div className="bp-ColorPickerControl">
             {isColorPickerToggled && (
                 <div className="bp-ColorPickerControl-palette">
```
jstoffan, line 32: Not related to this PR, but we'll need to address the accessibility portion of this control prior to release. There's no association between the toggle button and the palette right now, nor is there any keyboard handling for the palette buttons other than tab/shift+tab.

src/lib/viewers/controls/color-picker/__tests__/ColorPickerPalette-test.tsx
```diff
@@ -1,12 +1,14 @@
 import React from 'react';
 import { shallow, ShallowWrapper } from 'enzyme';
 import ColorPickerPalette from '../ColorPickerPalette';
+import { AnnotationColor } from '../../../../AnnotationModule';
 
 describe('ColorPickerPalette', () => {
-    const onColorSelect = jest.fn();
+    const colors = Object.values(AnnotationColor);
+    const onSelect = jest.fn();
```
jstoffan, line 8: This should be inlined as a default value so it's not accidentally reused between tests. Generally, the fewer dependencies each file has, the better.

src/lib/viewers/controls/color-picker/__tests__/ColorPickerPalette-test.tsx
```diff
@@ -1,12 +1,14 @@
 import React from 'react';
 import { shallow, ShallowWrapper } from 'enzyme';
 import ColorPickerPalette from '../ColorPickerPalette';
+import { AnnotationColor } from '../../../../AnnotationModule';
```
jstoffan, line 4: We should avoid importing annotations-specific code into a generic component or its tests, if possible.

src/lib/viewers/doc/DocBaseViewer.js
```diff
@@ -1089,15 +1089,16 @@ class DocBaseViewer extends BaseViewer {
...
                     annotationMode={this.annotationControlsFSM.getMode()}
                     hasDrawing={canDraw}
                     hasHighlight={canHighlight}
                     hasRegion={canAnnotate}
                     maxScale={MAX_SCALE}
                     minScale={MIN_SCALE}
-                    onAnnotationColorClick={this.handleAnnotationColorClick}
                     onAnnotationModeClick={this.handleAnnotationControlsClick}
                     onAnnotationModeEscape={this.handleAnnotationControlsEscape}
+                    onColorSelect={this.handleColorSelect}
```
jstoffan, line 1101: Can we be a bit more specific here, such as `onAnnotationColorChange`?

src/lib/viewers/doc/DocBaseViewer.js
```diff
@@ -1089,15 +1089,16 @@ class DocBaseViewer extends BaseViewer {
 
             this.controls.render(
                 <DocControls
+                    annotationColor={this.annotationModule ? this.annotationModule.getColor() : null}
```
jstoffan, line 1092: We should generally try to avoid `null` unless it adds functionality we need. Does `undefined` not work here?

src/lib/viewers/doc/DocControls.tsx
```diff
@@ -7,6 +7,7 @@ import FullscreenToggle, { Props as FullscreenToggleProps } from '../controls/fu
 import PageControls, { Props as PageControlsProps } from '../controls/page';
 import ThumbnailsToggle, { Props as ThumbnailsToggleProps } from '../controls/sidebar';
 import ZoomControls, { Props as ZoomControlsProps } from '../controls/zoom';
+import { AnnotationMode } from '../../AnnotationControls';
```
jstoffan, line 10: We already import AnnotationControls above.

src/lib/viewers/doc/DocControls.tsx
```diff
@@ -65,9 +68,11 @@ export default function DocControls({
                     onAnnotationModeEscape={onAnnotationModeEscape}
                 />
             </ControlsBar>
-            <ControlsBar>
-                <ColorPickerControl annotationMode={annotationMode} onAnnotationColorClick={onAnnotationColorClick} />
-            </ControlsBar>
+            {annotationMode === AnnotationMode.DRAWING && (
```
jstoffan, line 71: Should this also be behind a `hasDrawing` check for consistency with the main controls?

src/lib/viewers/image/ImageControls.tsx
jstoffan: I'm a little confused. The only prop added in this PR to this file is `annotationColor`, which is passed down to AnnotationControls and so should be exposed in its exported props object. Why is `ColorPickerControlProps` needed?

src/lib/constants.js
ChenCodes: Yes that works.

src/lib/viewers/controls/annotations/AnnotationsControls.tsx
ChenCodes: AnnotationColor is an object rather than an enum so I think we need to use string here.

```
export const AnnotationColor = {
    BOX_BLUE: bdlBoxBlue,
    GREEN_LIGHT: bdlGreenLight,
    WATERMELON_RED: bdlWatermelonRed,
    YELLORANGE: bdlYellorange,
    YELLOW: bdlYellow,
    GRIMACE: bdlGrimace,
};
```

src/lib/viewers/controls/annotations/AnnotationsControls.tsx
ChenCodes: Yep that works.

src/lib/viewers/controls/color-picker/ColorPickerControl.tsx
ChenCodes: Thanks for the callout, I've added a ticket to track the progress for that.

src/lib/viewers/controls/annotations/__tests__/AnnotationsControls-test.tsx
ChenCodes: Nope we don't need this anymore.

src/lib/viewers/controls/color-picker/ColorPickerControl.tsx
ChenCodes: Yes that sounds good.

src/lib/viewers/controls/annotations/AnnotationsControls.tsx
jstoffan: It was an enum previously, which seems superior to a plain object. Why did we change it?

src/lib/viewers/controls/color-picker/ColorPickerControl.tsx
ChenCodes: Yes I think that makes sense.

src/lib/viewers/controls/annotations/AnnotationsControls.tsx
ChenCodes: We changed it because I was running into a typing issue where TS didn't like how I was trying to create an enum where the right hand assignment was an externally imported constant.

Here's what I tried before:
```
export enum AnnotationColor = {
     BOX_BLUE = bdlBoxBlue,
     ...
}
```

ChenCodes: I couldn't find any other instances in our codebases where we were able to achieve the desired effect so I opted for using an object instead.

src/lib/viewers/doc/DocBaseViewer.js
ChenCodes: `undefined` works here.

src/lib/viewers/doc/DocControls.tsx
ChenCodes: I think we have `AnnotationsControls`, not `AnnotationControls`.

src/lib/viewers/doc/DocControls.tsx
ChenCodes: Yes that sounds good.

src/lib/viewers/image/ImageControls.tsx
ChenCodes: Yes that's correct, I don't need `ColorPickerControlProps`.

src/lib/viewers/doc/DocControls.tsx
jstoffan: Yes, that is intentional. We intend to remove the latter component soon.

src/lib/AnnotationModule.ts
ChenCodes: I think we should be okay with leaving it as is. The only main usage is in `DocControls`, since it's just this one location, I don't know if we need to have a separate directory for it. Open to other suggestions though.

src/lib/viewers/doc/DocControls.tsx
```diff
@@ -7,23 +7,28 @@ import FullscreenToggle, { Props as FullscreenToggleProps } from '../controls/fu
...
 
 export type Props = AnnotationsControlsProps &
     ColorPickerControlProps &
     FindBarToggleProps &
     FullscreenToggleProps &
     PageControlsProps &
     ThumbnailsToggleProps &
-    ZoomControlsProps;
+    ZoomControlsProps & {
+        onAnnotationColorChange: (color: string) => void;
```
ChenCodes, line 20: This isn't consistent with the above props imports because `onAnnotationColorChange` is not the name that's being used in the ColorPickerControl props (it's using `onSelect` instead).

src/lib/AnnotationModule.ts
```diff
@@ -0,0 +1,40 @@
...
+    BOX_BLUE: bdlBoxBlue,
+    GREEN_LIGHT: bdlGreenLight,
+    WATERMELON_RED: bdlWatermelonRed,
+    YELLORANGE: bdlYellorange,
+    YELLOW: bdlYellow,
+    GRIMACE: bdlGrimace,
+};
+
+type Options = {
```
jstoffan, line 20: Does this need to be a separate alias if it's only used once? If so, can we declare it above the `const` variables?

src/lib/AnnotationModule.ts
```diff
@@ -0,0 +1,40 @@
...
+    cache: Cache;
+};
+
+const ANNOTATION_COLOR_KEY = 'annotation-color';
+
+export default class AnnotationModule {
+    private cache: Cache;
+
+    constructor(options: Options) {
+        this.cache = options.cache;
```
jstoffan, line 30: We can destructure `cache` in the constructor signature.

src/lib/viewers/controls/color-picker/__tests__/ColorPickerControl-test.tsx
```diff
@@ -1,39 +1,31 @@
 import * as React from 'react';
 import { shallow, ShallowWrapper } from 'enzyme';
-import { AnnotationMode } from '../../annotations/types';
+import { bdlBoxBlue } from 'box-ui-elements/es/styles/variables';
```
jstoffan, line 3: Do we need to import this specific variable for the tests? Can it be a generic color, instead?

src/lib/viewers/controls/color-picker/__tests__/ColorPickerControl-test.tsx
```diff
@@ -1,39 +1,31 @@
...
 describe('ColorPickerControl', () => {
-    const onAnnotationColorClick = jest.fn();
-
     const getWrapper = (props = {}): ShallowWrapper =>
-        shallow(<ColorPickerControl onAnnotationColorClick={onAnnotationColorClick} {...props} />);
+        shallow(<ColorPickerControl colors={[bdlBoxBlue]} onColorSelect={jest.fn()} {...props} />);
 
     const getToggleButton = (wrapper: ShallowWrapper): ShallowWrapper => wrapper.find('.bp-ColorPickerControl-button');
 
     describe('render', () => {
```
jstoffan, line 12: Should there be a test for the `onSelect` callback?

src/lib/viewers/controls/color-picker/__tests__/ColorPickerPalette-test.tsx
```diff
@@ -3,16 +3,16 @@ import { shallow, ShallowWrapper } from 'enzyme';
...
-    const onColorSelect = jest.fn();
+    const colors = ['#fff'];
 
     const getWrapper = (props = {}): ShallowWrapper =>
-        shallow(<ColorPickerPalette onColorSelect={onColorSelect} {...props} />);
+        shallow(<ColorPickerPalette colors={colors} onSelect={jest.fn()} {...props} />);
 
     describe('render', () => {
-        test('should render the six color swatches', () => {
+        test('should render a single color swatch', () => {
```
jstoffan, line 12: The coverage here seems a little lacking. Do we want to assert anything else about this component?

src/lib/viewers/controls/icons/IconDrawing24.tsx
```diff
@@ -1,9 +1,13 @@
 import * as React from 'react';
 
 function IconDrawing24(props: React.SVGProps<SVGSVGElement>): JSX.Element {
+    const { fill } = props;
```
jstoffan, line 4: Can this be inlined with the function declaration? Do we want to spread `fill` onto both elements or should we use `...rest` for the SVG?

src/lib/viewers/doc/DocBaseViewer.js
```diff
@@ -1089,13 +1089,14 @@ class DocBaseViewer extends BaseViewer {
 
             this.controls.render(
                 <DocControls
+                    annotationColor={this.annotationModule ? this.annotationModule.getColor() : undefined}
```
jstoffan, line 1092: Is this ternary needed? What does `this.annotationModule.getColor()` return by default?

src/lib/viewers/doc/DocControls.tsx
```diff
@@ -37,6 +42,8 @@ export default function DocControls({
     pageNumber,
     scale,
 }: Props): JSX.Element {
+    const colors = Object.values(AnnotationColor);
```
jstoffan, line 45: Does this need to be redefined every render?

src/lib/viewers/controls/color-picker/__tests__/ColorPickerControl-test.tsx
ChenCodes: We can use a generic color.

src/lib/viewers/controls/color-picker/__tests__/ColorPickerControl-test.tsx
ChenCodes: Yes we should test that.

src/lib/viewers/doc/DocControls.tsx
ChenCodes: Nope, will move above this function.

src/lib/viewers/doc/DocBaseViewer.js
ChenCodes: We don't need it, it should return bdlBoxBlue by default.

src/lib/AnnotationModule.ts
ChenCodes: I think we can inline this to save on LOC.

src/lib/AnnotationModule.ts
ChenCodes: Will do.

src/lib/viewers/controls/icons/IconDrawing24.tsx
ChenCodes: Yes to inlining, yes for using ...rest as well.

src/lib/viewers/controls/color-picker/__tests__/ColorPickerPalette-test.tsx
ChenCodes: I think we can assert that `onSelect` is called properly on button click.

src/lib/__tests__/AnnotationModule-test.js
```diff
@@ -0,0 +1,16 @@
+import { bdlBoxBlue } from 'box-ui-elements/es/styles/variables';
+import AnnotationModule, { ANNOTATION_COLOR_KEY } from '../AnnotationModule';
+
+describe('lib/AnnotationModule', () => {
+    test('initial load', () => {
```
jstoffan, line 5: Should we test the getter and setter?

src/lib/viewers/controls/color-picker/__tests__/ColorPickerControl-test.tsx
```diff
@@ -1,41 +1,47 @@
...
+
+    describe('handleSelect', () => {
+        test('should close the palette and call onColorSelect if a color is selected', () => {
+            const onColorSelect = jest.fn();
+            const wrapper = getWrapper({ onColorSelect });
+
+            getToggleButton(wrapper).simulate('click');
+            wrapper.find('ColorPickerPalette').simulate('select', defaultColor);
+
+            expect(wrapper.find('ColorPickerPalette').exists()).toBe(false);
```
jstoffan, line 43: I believe this can be written as `wrapper.exists('ColorPickerPalette')`, but I'm not 100% sure on it.

src/lib/viewers/doc/__tests__/DocBaseViewer-test.js
```diff
@@ -1,5 +1,6 @@
 /* eslint-disable no-unused-expressions */
 import React from 'react';
+import { bdlBoxBlue } from 'box-ui-elements/es/styles/variables';
```
jstoffan, line 3: Do we need this specific color for the test or can we use a generic color?

src/lib/viewers/doc/DocControls.tsx
```diff
@@ -7,23 +7,30 @@ import FullscreenToggle, { Props as FullscreenToggleProps } from '../controls/fu
 import PageControls, { Props as PageControlsProps } from '../controls/page';
 import ThumbnailsToggle, { Props as ThumbnailsToggleProps } from '../controls/sidebar';
 import ZoomControls, { Props as ZoomControlsProps } from '../controls/zoom';
+import { AnnotationColor } from '../../AnnotationModule';
+import { AnnotationMode } from '../controls/annotations/types';
```
jstoffan, line 11: We should centralize these types in an annotations-specific types file in a future PR.

ChenCodes:
The comment I deleted earlier was related to us not needing to import ColorPickControlProps, I think there was a duplicate comment in there.

```
import { Props as ColorPickerControlProps } from '../controls/color-picker';
```

src/lib/viewers/controls/color-picker/ColorPickerControl.tsx
```diff
@@ -1,37 +1,35 @@
...
+    colors: Array<string>;
     isActive?: boolean;
+    onColorSelect: (color: string) => void;
 };
 
 export default function ColorPickerControl({
-    annotationMode,
+    activeColor = bdlBoxBlue,
+    colors,
     isActive = false,
```
jstoffan, line 17: Is this used? What does it do?

src/lib/viewers/controls/annotations/__tests__/AnnotationsControls-test.tsx
```diff
@@ -1,6 +1,7 @@
 import React from 'react';
 import { act } from 'react-dom/test-utils';
 import { ReactWrapper, mount } from 'enzyme';
+import { bdlBoxBlue } from 'box-ui-elements/es/styles/variables';
```
ConradJChan, line 4: nit: 🔤

src/lib/__tests__/AnnotationModule-test.js
ChenCodes: Yes we should.

src/lib/viewers/doc/__tests__/DocBaseViewer-test.js
ChenCodes: Nope, we can use a generic color.

src/lib/viewers/controls/color-picker/__tests__/ColorPickerPalette-test.tsx
```diff
@@ -3,16 +3,31 @@ import { shallow, ShallowWrapper } from 'enzyme';
...
+        });
+    });
+
+    describe('onSelect', () => {
+        test('should call onSelect with a button is clicked', () => {
+            const onSelect = jest.fn();
+            const wrapper = getWrapper({ onSelect });
+
+            wrapper
+                .find('button')
```
ConradJChan, line 26: Should we consider adding `testid` instead of using tag names? Should help when we do e2e tests too

src/lib/viewers/controls/color-picker/__tests__/ColorPickerControl-test.tsx
ChenCodes: Yes that works.

src/lib/viewers/doc/DocControls.tsx
ChenCodes: Will keep in mind.

src/lib/viewers/controls/color-picker/ColorPickerControl.tsx
ChenCodes: This isn't being used, will remove it.

src/lib/viewers/controls/color-picker/__tests__/ColorPickerPalette-test.tsx
ChenCodes: That sounds good, will add.

src/lib/AnnotationModule.ts
```diff
@@ -0,0 +1,36 @@
...
+    GRIMACE: bdlGrimace,
+};
+
+export default class AnnotationModule {
+    private cache: Cache;
+
+    constructor({ cache }: { cache: Cache }) {
+        this.cache = cache;
+
+        if (this.cache && !this.cache.get(ANNOTATION_COLOR_KEY)) {
```
jstoffan, line 28: Can `this.cache` be undefined? We don't have this check elsewhere. Can this call `this.getColor()`, instead?

ChenCodes: We can use this.getColor and I've removed the existence check / updated tests for all classes that inherit from BaseViewer.

src/lib/AnnotationModule.ts
```diff
@@ -0,0 +1,36 @@
...
+
+    constructor({ cache }: { cache: Cache }) {
+        this.cache = cache;
+
+        if (!this.getColor()) {
+            this.setColor(AnnotationColor.BOX_BLUE);
+        }
+    }
+
+    public getColor = (): string => this.cache.get(ANNOTATION_COLOR_KEY) as string;
```
jstoffan, line 33: Do we need to set the default value in the cache in the constructor or can we return it here with `|| bdlBoxBlue`?

jstoffan (approved):

src/lib/AnnotationModule.ts
ChenCodes: Nope, we can use the OR with `bdlBoxBlue`.

jstoffan (approved):

PR #1310, opened 2020-12-15
fix(drawing): fix accessibility for color picker components

ChenCodes:
Changes in this PR:
- [x] Move ColorPickerPalette element down below the button element so that when the user tabs from the button they are able to focus on the swatches in the palette
- [x] approved by design
- [x] cross browser testing

Note: We aren't adding any additional key handling (left / right arrows) as I think the current tabbing experience is satisfactory for complete navigation through the color picker elements


jstoffan:
Is this a chore or a fix?

jstoffan (approved):

jstoffan (approved):

PR #1314, opened 2021-01-08
feat(controls): Add react versions of time and volume scrubbers

jstoffan:


src/lib/viewers/controls/media/TimeControls.scss
```diff
@@ -0,0 +1,39 @@
+@import '../slider/styles';
+
+$bp-TimeControls-size: 18px; // Divisible by 3px (default) and 6px (hovered) for better vertical centering
```
ChenCodes, line 3: Is there some sort of grid styling that we can reuse here? For example, in box-ui-elements there is a variable called `$bdl-grid-unit` which is defaulted to 4px. Is the 3px above a constant that is being shared across multiple elements in the control bar?

src/lib/viewers/controls/media/TimeControls.tsx
```diff
@@ -0,0 +1,45 @@
...
+
+export type Props = {
+    bufferedRange?: TimeRanges;
+    currentTime?: number;
+    durationTime?: number;
+    onTimeChange: (volume: number) => void;
+};
+
+export const round = (value: number): number => {
+    return +value.toFixed(4);
```
ChenCodes, line 14: Could you explain what the `+` symbol is for here?

src/lib/viewers/controls/slider/__tests__/SliderControl-test.tsx
```diff
@@ -0,0 +1,43 @@
...
+import SliderControl from '../SliderControl';
+
+describe('SliderControl', () => {
+    const getWrapper = (props = {}): ShallowWrapper =>
+        shallow(<SliderControl onChange={jest.fn()} value={0} {...props} />);
+
+    describe('event handlers', () => {
+        test('should parse and return the current value on change', () => {
+            const onChange = jest.fn();
+            const wrapper = getWrapper({ step: 0.1, onChange });
```
ChenCodes, line 12: Nit: abc

src/lib/viewers/controls/media/TimeControls.scss
jstoffan: No, we don't have such a system for the preview controls. We're trying to keep the visual styles as close as possible to what's live today. We can investigate refactoring the styles if and when we do a visual refresh in the future.

src/lib/viewers/controls/media/TimeControls.tsx
jstoffan: It's a shortcut to convert a string to a number, as the former is what `toFixed` returns.

src/lib/viewers/controls/media/TimeControls.tsx
```diff
@@ -0,0 +1,45 @@
...
+
+    const handleChange = (newValue: number): void => {
+        onTimeChange(round(durationTime * (newValue / 100)));
+    };
+
+    return (
+        <SliderControl
+            className="bp-TimeControls"
+            onChange={handleChange}
+            step={0.1}
```
ChenCodes, line 39: It looks like we have this step property that is being passed in to SliderControl, but later in the SliderControl.tsx file, we're hardcoding step={1}. Does this prop still need to be passed down?

jstoffan: Yes, as it overrides the default value of `1` in SliderControl.

ChenCodes (approved):
LGTM

src/lib/viewers/controls/slider/__tests__/SliderControl-test.tsx
jstoffan: Will fix in a follow-up.

PR #1315, opened 2021-01-11
fix(drawing): emit color event when color picker is opened

ChenCodes:
**Issue:**
If the user updates the drawing color in the first browser tab (when there are two tabs open), the cursor color will be stale in the second tab. The proposed fix is to emit the color change event every time the Color Picker is opened - this will ensure that the cursor color matches the color in the Preview Toolbar.

**Changes in this PR:**
- [x] emit color change event whenever the Color Picker button is clicked that results in mode becoming `AnnotatonMode.DRAWING`
- [x] update unit tests

**Demo:**
ezgif com-gif-maker (5)

### DIFF
--- a/src/lib/viewers/BaseViewer.js
+++ b/src/lib/viewers/BaseViewer.js
@@ -1123,9 +1123,7 @@ class BaseViewer extends EventEmitter {
             }
         }
 
-        if (mode === AnnotationMode.DRAWING) {
-            this.annotator.emit(ANNOTATOR_EVENT.setColor, this.annotationModule.getColor());
-        }
+        this.annotator.emit(ANNOTATOR_EVENT.setColor, this.annotationModule.getColor());
     };
 
     /**
--- a/src/lib/viewers/BaseViewer.js
+++ b/src/lib/viewers/BaseViewer.js
@@ -1124,8 +1124,7 @@ class BaseViewer extends EventEmitter {
         }
 
         if (mode === AnnotationMode.DRAWING) {
-            const color = this.annotationModule.getColor();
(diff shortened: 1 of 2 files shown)


src/lib/viewers/BaseViewer.js
```diff
@@ -1122,6 +1122,11 @@ class BaseViewer extends EventEmitter {
                 this.containerEl.classList.add(className);
             }
         }
+
+        if (mode === AnnotationMode.DRAWING) {
+            const color = this.annotationModule.getColor();
+            this.annotator.emit(ANNOTATOR_EVENT.setColor, color);
```
mxiao6, line 1128: Can we remove the `color` variable and get the color inline?

src/lib/viewers/BaseViewer.js
```diff
@@ -1122,6 +1122,11 @@ class BaseViewer extends EventEmitter {
                 this.containerEl.classList.add(className);
             }
         }
+
+        if (mode === AnnotationMode.DRAWING) {
```
jstoffan, line 1126: Is this condition needed or can emit it always? Do we emit this event anywhere else that can be centralized here?

ChenCodes: We can emit it always.

We emit this event in `handleAnnotationColorChange` in DocBaseViewer and ImageViewer but I think those should be separate from this as `handleAnnotationColorChange` is tied to the user action of clicking on a color in the ColorPalette. The event that is emitted in BaseViewer is more so tied to the user action of when a user is clicking on the Drawing button to show the Color Picker.

mxiao6 (approved):

PR #1317, opened 2021-01-12
fix(annotations): Fix Palette doesn't close when blurred

mxiao6:


src/lib/viewers/controls/color-picker/ColorPickerControl.tsx
```diff
@@ -27,6 +27,7 @@ export default function ColorPickerControl({
             <button
                 className="bp-ColorPickerControl-button"
                 data-testid="bp-ColorPickerControl-button"
+                onBlur={(): void => setIsColorPickerToggled(false)}
```
jstoffan, line 30: Does clicking on a color still select it before the palette closes? If so, is line 21 still needed with this change?

src/lib/viewers/controls/color-picker/ColorPickerControl.tsx
```diff
@@ -16,28 +18,41 @@ export default function ColorPickerControl({
...
                 {...rest}
             >
                 <div className="bp-ColorPickerControl-swatch" style={{ backgroundColor: activeColor }} />
             </button>
-            {isColorPickerToggled && (
-                <div className="bp-ColorPickerControl-palette">
-                    <ColorPickerPalette colors={colors} data-testid="bp-ColorPickerPalette" onSelect={handleSelect} />
-                </div>
-            )}
+            <div
```
jstoffan, line 49: This changes the tab behavior so that the colors are included in the tab order even when the picker is closed. Is that desirable? It seems like we'd want to keep the `isColorPickerToggled &&` condition.

jstoffan: Actually, I'm mistaken, since `display: none` will cause them to be ignored, unlike other visual-only hiding techniques. This should work fine.

jstoffan (approved):
LGTM 👍

PR #1318, opened 2021-01-14
fix(drawing): show sub-shadow in color picker control button

ChenCodes:
**Changes in this PR:**
- [x] add sub shadow in color picker control button
- [x] cross browser testing

**Demo:**
ezgif com-gif-maker (6)

**Chrome**
chrome

**Safari**
safari

**FF**
firefox

**Edge**
edge

**IE**
ie

### DIFF
--- a/src/lib/viewers/controls/color-picker/ColorPickerControl.scss
+++ b/src/lib/viewers/controls/color-picker/ColorPickerControl.scss
@@ -7,7 +7,7 @@
     justify-content: center;
 }
 
-.bp-ColorPickerControl-button {
+.bp-ColorPickerControl-toggle {
     @include bp-ControlButton;
 
     &.bp-is-active,
@@ -48,7 +48,7 @@
     }
 }
 
-.bp-ColorPickerControl-swatch {
+.bp-ColorPickerControl-toggle-swatch {
     width: 18px;
     height: 18px;
     border: 2px solid #fff;
--- a/src/lib/viewers/controls/color-picker/ColorPickerControl.scss
+++ b/src/lib/viewers/controls/color-picker/ColorPickerControl.scss
@@ -1,11 +1,5 @@
 @import '../styles';
 
-@mixin bp-highlight-style {
-    padding: 8px;
-    background-color: $fours;
-    border-radius: 4px;
-}
-
 .bp-ColorPickerControl {
     position: relative;
     display: flex;
@@ -16,10 +10,13 @@
 .bp-ColorPickerControl-button {
     @include bp-ControlButton;
 
-    &:hover,
-    &:focus {
-        .bp-ColorPickerControl-highlight {
-            @include bp-highlight-style;
+    &.bp-is-toggled,
+    &:focus,
+    &:hover {
+        .bp-ColorPickerControl-toggle-background {
+            padding: 8px;
+            background-color: $fours;
+            border-radius: 4px;
         }
     }
 }
@@ -48,12 +45,6 @@
 
     &.bp-is-open {
         display: block;
-    }
-}
-
-.bp-ColorPickerControl-highlight {
-    &.bp-has-attention {
-        @include bp-highlight-style;
     }
 }
 
--- a/src/lib/viewers/controls/color-picker/ColorPickerControl.scss
+++ b/src/lib/viewers/controls/color-picker/ColorPickerControl.scss
@@ -9,6 +9,15 @@
 
 .bp-ColorPickerControl-button {
     @include bp-ControlButton;
+
+    &:hover,
+    &:focus {
+        .bp-ColorPickerControl-highlight {
+            padding: 8px;
+            background-color: $fours;
+            border-radius: 4px;
+        }
+    }
 }
 
 .bp-ColorPickerControl-palette {
@@ -36,12 +45,6 @@
     &.bp-is-open {
         display: block;
     }
-}
-
-.bp-ColorPickerControl-highlight {
-    padding: 8px;
-    background-color: $fours;
-    border-radius: 4px;
 }
 
 .bp-ColorPickerControl-swatch {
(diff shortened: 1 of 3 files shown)


ChenCodes:
Need to rebase

src/lib/viewers/controls/color-picker/ColorPickerControl.scss
```diff
@@ -9,6 +9,15 @@
...
     @include bp-ControlButton;
+
+    &:hover,
+    &:focus {
+        .bp-ColorPickerControl-highlight {
+            padding: 8px;
+            background-color: $fours;
+            border-radius: 4px;
+        }
+    }
```
ConradJChan, line 20: Is this enough to cover the case when we use keyboard tab to access the color palette and focus leaves the button?

ChenCodes: Good catch - I've added another change for reusing `isColorPickerToggled` to determine if we should also be showing the sub shadow when the palette is open.

src/lib/viewers/controls/color-picker/ColorPickerControl.scss
```diff
@@ -1,5 +1,11 @@
 @import '../styles';
 
+@mixin highlight-style {
```
ConradJChan, line 3: We've followed a convention of namespacing mixins with`ba-`

src/lib/viewers/controls/color-picker/ColorPickerControl.scss
```diff
@@ -9,6 +15,13 @@
 
 .bp-ColorPickerControl-button {
     @include bp-ControlButton;
+
+    &:hover,
+    &:focus {
+        .bp-ColorPickerControl-highlight {
+            @include highlight-style();
```
ConradJChan, line 22: I don't think you need the parens if no arguments are provided

src/lib/viewers/controls/color-picker/ColorPickerControl.scss
```diff
@@ -38,6 +51,12 @@
     }
 }
 
+.bp-ColorPickerControl-highlight {
+    &.bp-is-open {
```
ConradJChan, line 55: Open doesn't quite seem accurate, maybe `bp-has-attention` would be more appropriate?

src/lib/viewers/controls/color-picker/ColorPickerControl.scss
ChenCodes: Sounds good - I think you mean `bp-` instead?

src/lib/viewers/controls/color-picker/ColorPickerControl.scss
ChenCodes: Yes that makes sense in the context of the highlight div.

src/lib/viewers/controls/color-picker/ColorPickerControl.tsx
```diff
@@ -44,7 +44,13 @@ export default function ColorPickerControl({
                 type="button"
                 {...rest}
             >
-                <div className="bp-ColorPickerControl-swatch" style={{ backgroundColor: activeColor }} />
+                <div
+                    className={classNames('bp-ColorPickerControl-highlight', {
+                        'bp-has-attention': isColorPickerToggled,
```
jstoffan, line 49: The naming here is a little confusing, as we have a hook called `useAttention` that has different behavior. Can we name this `bp-is-active`? Should this class be on the `bp-ColorPickerControl-button` element, as well, since that is what actually is toggled?

src/lib/viewers/controls/color-picker/ColorPickerControl.tsx
```diff
@@ -44,7 +44,13 @@ export default function ColorPickerControl({
                 type="button"
                 {...rest}
             >
-                <div className="bp-ColorPickerControl-swatch" style={{ backgroundColor: activeColor }} />
+                <div
+                    className={classNames('bp-ColorPickerControl-highlight', {
```
jstoffan, line 48: Can we avoid the word "highlight" here, since it has an existing meaning for annotations (text highlights)? How about `bp-ColorPickerControl-toggle-background`?

src/lib/viewers/controls/color-picker/ColorPickerControl.scss
```diff
@@ -9,6 +15,13 @@
 
 .bp-ColorPickerControl-button {
     @include bp-ControlButton;
+
+    &:hover,
```
jstoffan, line 19: By moving the `bp-is-active` class to the button, it can be combined with the hover and focus pseudo-classes and the styles within the mixin can be inlined.

src/lib/viewers/controls/color-picker/ColorPickerControl.tsx
```diff
@@ -44,7 +44,13 @@ export default function ColorPickerControl({
                 type="button"
                 {...rest}
             >
-                <div className="bp-ColorPickerControl-swatch" style={{ backgroundColor: activeColor }} />
+                <div
```
jstoffan, line 47: Should we have a test?

src/lib/viewers/controls/color-picker/ColorPickerControl.tsx
ChenCodes: Yes that makes sense

src/lib/viewers/controls/color-picker/ColorPickerControl.tsx
ChenCodes: Yep

src/lib/viewers/controls/color-picker/ColorPickerControl.tsx
```diff
@@ -37,14 +37,16 @@ export default function ColorPickerControl({
     return (
         <div className="bp-ColorPickerControl">
             <button
-                className="bp-ColorPickerControl-button"
+                className={classNames('bp-ColorPickerControl-button', { 'bp-is-toggled': isColorPickerToggled })}
```
jstoffan, line 40: Should this also be renamed to `bp-ColorPickerControl-toggle`? Same for the swatch?

ChenCodes: Yes, will rename them

jstoffan (approved):

PR #1321, opened 2021-01-20
fix(drawing): Fix Safari doesn't focus and blur color picker button

mxiao6:
- [x] Unit test
- [x] Cross-browser testing


src/lib/viewers/controls/color-picker/ColorPickerControl.tsx
```diff
@@ -40,13 +40,24 @@ export default function ColorPickerControl({
 
     const handleClick = (): void => setIsColorPickerToggled(!isColorPickerToggled);
 
+    const handleMouseDown = (event: React.MouseEvent<HTMLButtonElement>): void => {
+        if (event.currentTarget.focus) {
+            // Buttons do not receive focus in Firefox and Safari on MacOS
+            event.currentTarget.focus();
+            // Stop propagation to higher level component which may trigger blur event
```
jstoffan, line 47: Is this needed? The comment makes it seem uncertain. We should avoid messing with propagation if possible.

mxiao6: Yeah, it's needed. The deselect mouse down handler in annotations blurs the button just after it gets focused.

ConradJChan: Does that mean clicking the color picker button doesn't deselect an active annotation?

src/lib/viewers/controls/color-picker/ColorPickerControl.tsx
```diff
@@ -40,13 +40,23 @@ export default function ColorPickerControl({
 
     const handleClick = (): void => setIsColorPickerToggled(!isColorPickerToggled);
 
+    const handleMouseDown = (event: React.MouseEvent<HTMLButtonElement>): void => {
+        if (event.currentTarget.focus) {
+            // Buttons do not receive focus in Firefox and Safari on MacOS
+            event.currentTarget.focus();
+            // When focus() is called, preventDefault must be called to keep the focus from leaving the target
```
jstoffan, line 47: `When focus is called within a mousedown handler`, just for clarity?

src/lib/viewers/controls/color-picker/__tests__/ColorPickerControl-test.tsx
```diff
@@ -75,6 +75,19 @@ describe('ColorPickerControl', () => {
...
         });
+
+        test('should call focus and stop propagation when mouse down', () => {
+            const mockEvent = {
+                currentTarget: {
+                    focus: jest.fn(),
+                },
+                preventDefault: jest.fn(),
+            };
+            (getToggleButton(getWrapper()).prop('onMouseDown') as Function)(mockEvent);
```
jstoffan, line 86: I've been favoring `simulate` because it avoids the need to cast. I've seen the Jest folks say not to use it, but it's there and it works well. Up to you.

jstoffan (dismissed):

src/lib/viewers/controls/color-picker/ColorPickerControl.tsx
ConradJChan: Could this be done in `handleClick` as well?

mxiao6: It doesn't affect the current flow, but affects the second click of the button. When you click the button the second time, handle blur happens before handle click, which sets the state to false and back to true again. so you cannot close the palette anymore.

mxiao6: The interesting thing here is that even if you click the same button which has the focus, it will still blur and re-focus again, except you handle the logic before everything in `handleMouseDown`

src/lib/viewers/controls/color-picker/__tests__/ColorPickerControl-test.tsx
mxiao6: Yeah I prefer using `simulate` too, and I tried it first but it doesn't work (I googled it why but it seems sometimes it happens and everyone just use props.onClick() etc. ). Then I have to use this fallback plan.

jstoffan (approved):

PR #1323, opened 2021-01-29
fix(thumbnails): Replace remove method that is unsupported in IE11

jstoffan:
I considered polyfilling the method on the Element prototype, but decided against it to avoid further altering global objects.

### DIFF
--- a/src/lib/__tests__/VirtualScroller-test.js
+++ b/src/lib/__tests__/VirtualScroller-test.js
@@ -33,15 +33,14 @@ describe('VirtualScroller', () => {
 
     describe('destroy()', () => {
         test('should remove the HTML element references', () => {
-            const scrollingEl = { remove: () => {} };
-            jest.spyOn(scrollingEl, 'remove').mockImplementation();
+            jest.spyOn(virtualScroller.anchorEl, 'removeChild').mockImplementation();
 
-            virtualScroller.scrollingEl = scrollingEl;
+            virtualScroller.scrollingEl = document.createElement('div');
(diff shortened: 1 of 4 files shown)


src/lib/__tests__/VirtualScroller-test.js
```diff
@@ -464,14 +463,15 @@ describe('VirtualScroller', () => {
...
-            scrollingEl = { remove: () => {}, dispatchEvent: stubs.dispatchEvent };
+            scrollingEl = { dispatchEvent: stubs.dispatchEvent };
 
             virtualScroller.totalItems = 10;
             virtualScroller.itemHeight = 10;
             virtualScroller.margin = 0;
             virtualScroller.scrollingEl = scrollingEl;
 
             stubs.isVisible = jest.spyOn(virtualScroller, 'isVisible').mockImplementation();
+            stubs.removeChild = jest.spyOn(virtualScroller.anchorEl, 'removeChild').mockImplementation();
```
ConradJChan, line 474: Is this needed? Not sure I see it being used in any test

ConradJChan (approved):

src/lib/__tests__/VirtualScroller-test.js
jstoffan: No, not really. I did it so it would fit in better with the other spy on the line above. I can remove it if it's concerning.

ConradJChan: Not concerning, already approved

PR #1324, opened 2021-02-08
build(highlightjs): Add upgrade script for highlightjs

ConradJChan:


build/upgrade_highlightjs.sh
```diff
@@ -0,0 +1,92 @@
+#!/bin/bash
+# Run with specific branch/tag (e.g. ./upgrade_highlightjs.sh 10.6.0) or with no arguments to use master
+
+# TODO: This list of languages should be made more dynamic based on the file extensions Preview supports
```
jstoffan, line 4: This seems like a likely place for future changes to break things. Should we also add a comment to where this list is maintained in the source code so both sides can hopefully stay in sync?

build/upgrade_highlightjs.sh
```diff
@@ -0,0 +1,92 @@
...
+TEXT_BASE_PATH="src/third-party/text"
+TEXT_STATIC_ASSETS_BRANCH=${1:-master}
+TEXT_STATIC_ASSETS_VERSION=$(./build/current_version.sh)
+TEXT_STATIC_ASSETS_PATH="${TEXT_BASE_PATH}/${TEXT_STATIC_ASSETS_VERSION}"
+
+build_custom_highlightjs() {
+    echo "-----------------------------------------------------------------------------------"
+    echo "Cloning higlightjs repo at branch: $TEXT_STATIC_ASSETS_BRANCH..."
+    echo "-----------------------------------------------------------------------------------"
+    rm -rf ./highlight.js
```
jstoffan, line 15: Minor, but at first glance, this looked like a file, which wouldn't need `-r`. Can we check out to a folder without an extension to minimize confusion?

build/upgrade_highlightjs.sh
```diff
@@ -0,0 +1,92 @@
...
+    git clone git@github.com:highlightjs/highlight.js.git --depth 1 --single-branch --branch ${TEXT_STATIC_ASSETS_BRANCH} || return 1
+
+    echo "-----------------------------------------------------------------------------------"
+    echo "Building higlightjs for browser for languages ${HIGHLIGHTJS_LANGUAGES}..."
+    echo "-----------------------------------------------------------------------------------"
+    cd ./highlight.js
+    yarn install || return 1
```
jstoffan, line 22: Should we use `--frozen-lockfile` here to avoid any potential issues with dependency drift?

build/upgrade_highlightjs.sh
```diff
@@ -0,0 +1,92 @@
...
+    node tools/build -t browser ${HIGHLIGHTJS_LANGUAGES} || return 1
+    cd ../
```
jstoffan, line 24: Is the trailing `/` needed here?

build/upgrade_highlightjs.sh
```diff
@@ -0,0 +1,92 @@
...
+    echo "-----------------------------------------------------------------------------------"
+    rm -rf ./highlight.js
+    git clone git@github.com:highlightjs/highlight.js.git --depth 1 --single-branch --branch ${TEXT_STATIC_ASSETS_BRANCH} || return 1
+
+    echo "-----------------------------------------------------------------------------------"
+    echo "Building higlightjs for browser for languages ${HIGHLIGHTJS_LANGUAGES}..."
+    echo "-----------------------------------------------------------------------------------"
+    cd ./highlight.js
+    yarn install || return 1
+    node tools/build -t browser ${HIGHLIGHTJS_LANGUAGES} || return 1
```
jstoffan, line 23: Can we simplify to `node ./highlight.js/tools/build ...` to avoid the `cd`s?

build/upgrade_highlightjs.sh
```diff
@@ -0,0 +1,92 @@
...
+
+    echo "-----------------------------------------------------------------------------------"
+    echo "Minifying github.css with cssnano"
+    echo "-----------------------------------------------------------------------------------"
+    ./node_modules/.bin/cssnano ${TEXT_STATIC_ASSETS_PATH}/github.css ${TEXT_STATIC_ASSETS_PATH}/github.min.css || return 1
+
+    echo "-----------------------------------------------------------------------------------"
+    echo "Cleaning up highlightjs repo"
+    echo "-----------------------------------------------------------------------------------"
+    rm -rf ./highlight.js
```
jstoffan, line 55: Should this live outside of the "process" step?

build/upgrade_highlightjs.sh
```diff
@@ -0,0 +1,92 @@
...
+}
+
+upgrade_highlightjs() {
+    if [[ $(git diff --shortstat 2> /dev/null | tail -n1) != "" ]] ; then
```
jstoffan, line 59: Is check this needed? Does having a dirty directory prevent the upgrade?

build/upgrade_highlightjs.sh
```diff
@@ -0,0 +1,90 @@
+#!/bin/bash
+# Run with specific branch/tag (e.g. ./upgrade_highlightjs.sh 10.6.0) or with no arguments to use master
+
+# TODO: This list of languages should be made more dynamic based on the file extensions Preview supports, see src/lib/extensions.js
```
jstoffan, line 4: Should we include a comment in `src/lib/extensions.js` that if the file types change we need to regenerate the highlightjs bundle?

jstoffan (approved):

build/upgrade_highlightjs.sh
ConradJChan: Good idea, I'll add a comment there

jstoffan (approved):

PR #1325, opened 2021-02-09
build: bump thirdparty dependencies during upgrade

ConradJChan:
To track the third party dependencies that are managed outside of npm and used for Preview viewer rendering, we are introducing a new `manifest.json` that lives in each `src/third-party/*/<version>/` directory. 

This `manifest.json` will track the versions of third-party libraries that it contains

### DIFF
--- a/package.json
+++ b/package.json
@@ -157,5 +157,13 @@
     },
     "resolutions": {
         "mojito-rb-gen/merge": "1.2.1"
+    },
+    "thirdparty-dependencies": {
+        "highlightjs": "9.7.0",
+        "papaparse": "4.1.2",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -157,5 +157,13 @@
     },
     "resolutions": {
         "mojito-rb-gen/merge": "1.2.1"
+    },
+    "thirdparty-dependencies": {
```
jstoffan, line 161: Is this a common pattern? Is it enforced by tooling or could be it be forgotten and drift when adding or modifying "out of band" libraries?

jstoffan (approved):

PR #1326, opened 2021-02-10
build: Add upgrade scripts for text assets

ConradJChan:
Add support for upgrading all text viewer dependencies. Added new package.json script target `upgrade:dependency`.

It can be used to upgrade:

- highlightjs such as `yarn upgrade:dependency highlightjs 9.18.1`
- remarkable such as `yarn upgrade:dependency remarkable v2.0.1`
- papaparse such as `yarn upgrade:dependency papaparse 5.3.0`


build/upgrade_utils.sh
```diff
@@ -0,0 +1,37 @@
+#!/bin/bash
+
+bump_manifest_version() {
+    echo "-----------------------------------------------------------------------------------"
+    echo "Bumping $REPO_SRC_DIR version in manifest.json"
+    echo "-----------------------------------------------------------------------------------"
+    VERSION=$(cd $REPO_SRC_DIR && ../build/current_version.sh) || return 1
```
jstoffan, line 7: Should these values be passed in as arguments rather than set as shared variables? Do these helpers provide enough utility to live separately and justify the extra cognitive overhead?

build/upgrade_remarkable.sh
```diff
@@ -0,0 +1,68 @@
+#!/bin/bash
+# Run with specific branch/tag (e.g. ./upgrade_remarkable.sh 2.0.1) or with no arguments to use master
+
+REMARKABLE_DIST="remarkable.min.js"
```
jstoffan, line 4: Should these all be combined into a single `upgrade:viewers:text` script?

package.json
```diff
@@ -137,8 +137,8 @@
         "test:e2e:open": "npm-run-all -p -r start:dev cy:open",
         "test:watch": "yarn test --watch",
         "upgrade:annotations": "./build/upgrade_annotations.sh",
-        "upgrade:highlightjs": "./build/upgrade_highlightjs.sh",
-        "upgrade:pdfjs": "./build/upgrade_pdfjs.sh"
+        "upgrade:pdfjs": "./build/upgrade_pdfjs.sh",
+        "upgrade:viewers:text": "./build/upgrade_dependency.sh"
```
jstoffan, line 141: Should there be any arguments here?

ConradJChan: I left it open for now, it runs like `yarn upgrade:viewers:text highlightjs 9.18.1`. I could add script names for `yarn upgrade:viewers:text:highlights` and the like but that seemed unnecessary

jstoffan: Should we rename it to `upgrade:dependency`, in that case? It doesn't seem specific to viewers at the moment.

jstoffan (approved):

PR #1328, opened 2021-02-12
build(deps): Upgrade highlightjs to 9.18.1

ConradJChan:
Tested against the Sanity folder for all browsers


src/third-party/text/2.64.0/remarkable.min.js
```diff
@@ -0,0 +1,4 @@
+/*! remarkable 1.7.1 https://github.com/jonschlinkert/remarkable @license MIT */
```
jstoffan, line 1: Looks like this is the only asset we don't have in a non-minified state. Does that matter?

jstoffan (approved):
Should the commit message be more general, since we're updating multiple libraries?

ConradJChan:
This commit is just for highlightjs, I'll do separate commits for remarkable and papaparse

jstoffan:
Gotcha. The assets copied over from the prior version bamboozled me. LGTM

PR #1329, opened 2021-02-22
build(deps): Upgrade remarkable to 2.0.1

ConradJChan:
Tested against sanity folder on all browsers


mxiao6 (approved):

src/third-party/text/2.65.0/manifest.json
```diff
@@ -5,6 +5,6 @@
     "dependencies": {
```
mxiao6, line 5: Do you also need to update the version here (the line above)?

ConradJChan: Good catch

mxiao6 (approved):

PR #1330, opened 2021-02-22
build(deps): Upgrade Papaparse to 5.0.0

ConradJChan:
Tested against sanity folder in all browsers


src/lib/viewers/text/CSVViewer.js
```diff
@@ -53,43 +52,53 @@ class CSVViewer extends TextBaseViewer {
...
+            file: { extension },
+            representation,
+        } = this.options;
         const template = representation.content.url_template;
-        const assetUrlCreator = createAssetUrlCreator(location);
-        const papaWorkerUrl = assetUrlCreator(JS[0]);
 
         return Promise.all([this.loadAssets(JS), this.getRepStatus().getPromise()])
             .then(() => {
-                this.api.get(papaWorkerUrl, { type: 'blob' }).then(papaWorkerBlob => {
```
ConradJChan, line 63: Papaparse 5.0.0 added the `worker: true` option as the way to load an inline worker

src/lib/viewers/text/CSVViewer.js
```diff
@@ -53,43 +52,53 @@ class CSVViewer extends TextBaseViewer {
...
+        // Papaparse will by default autodetect the delimiter but in cases where the content has many
+        // special characters, it can get confused. Since we have the file extension, rely on that for
+        // the explicit delimiter, otherwise default to the autodetection.
+        switch (extension) {
+            case 'csv':
+                return ',';
+            case 'tsv':
+                return '\t';
+            default:
+                return '';
```
mxiao6, line 98: Does returning empty string fall back to auto-detection? Since it's CSV (Comma-separated values) viewer, should the default be comma?

ConradJChan: Yeah, empty string falls back to auto-detection. Since we currently rely on auto-detection, I figured we would want to keep that default (even though we only support csv and tsv currently)

mxiao6 (dismissed):

src/lib/viewers/text/CSVViewer.js
jstoffan: How does it determine the asset url for the worker? Does it load it via a path relative to the main file?

ConradJChan: It appears to load the source code inline: https://github.com/mholt/PapaParse/blob/master/papaparse.js#L52

src/lib/viewers/text/CSVViewer.js
```diff
@@ -53,38 +52,31 @@ class CSVViewer extends TextBaseViewer {
...
-
-                            this.data = results.data;
-                            this.finishLoading();
-                            URL.revokeObjectURL(workerSrc);
-                        },
-                    });
+                this.startLoadTimer();
+                const urlWithAuth = this.createContentUrlWithAuthParams(template);
+                Papa.parse(urlWithAuth, {
+                    delimitersToGuess: [',', '\t'],
```
ConradJChan, line 63: I've added the option `delimitersToGuess` which will restrict the delimiters to comma and tab which technically was all that we supported given that we support the extensions CSV and TSV. I could remove this option entirely, but then there is an issue with the auto-detection of delimiter for XSS.csv with the excessive `;` as well as `,`. 

At the end of the day it feels like keeping the delimiters to what we support makes sense. Allowing users to hack the system by naming it csv but providing delimiters such as `|` or `;` doesn't make all that much sense, imo.

Thoughts?

jstoffan: Gotcha, so it sounds like we may need some new options to maintain consistency with the current experience. Should we update the value to `[',', '\t', Papa.RECORD_SEP, Papa.UNIT_SEP]` to retain support for the "true" delimiters mentioned in the docs?

src/lib/viewers/text/CSVViewer.js
```diff
@@ -53,38 +52,31 @@ class CSVViewer extends TextBaseViewer {
...
+                    download: true,
+                    error: (err, file, inputElem, reason) => {
```
jstoffan, line 65: Did we validate that these parameters are still accurate/available with the new version?

ConradJChan: Yeah, the new version still follows the same callback signature

src/lib/viewers/text/CSVViewer.js
ConradJChan: Changed version to 5.0.0 which doesn't contain the tweaks to the method which aims to guess the delimiter. Will follow up with Papaparse with the issue we're seeing

jstoffan (approved):

PR #1332, opened 2021-02-25
build(deps): Upgrade react and react-dom to 16.14.0

ConradJChan:
Affects the `CSVViewer` and `ArchiveViewer`


ConradJChan:
Will go for a React 17 upgrade instead

PR #1344, opened 2021-03-24
feat(loading): Add options for loading icon and progress bar

jstoffan:


ConradJChan (dismissed):
LGTM

src/lib/__tests__/Preview-test.js
```diff
@@ -1715,27 +1713,16 @@ describe('lib/Preview', () => {
 
         test('should show the loading download button if file can be downloaded', () => {
             stubs.canDownload.mockReturnValue(true);
-            preview.loadViewer();
+            preview.loadViewer({});
```
ConradJChan: Is the empty object argument needed?

ConradJChan (approved):

src/lib/__tests__/Preview-test.js
jstoffan: No, but that's how it is in master. I'm not making any changes in this area, so I reverted this change as well.

PR #1345, opened 2021-03-30
feat(loading): Update loading experience to latest designs

jstoffan:
This PR is dependent on https://github.com/box/box-ui-elements/pull/2498, which improves the loading experience when Preview is loaded via the ContentPreview Element. This change uses that same experience if Preview is loaded directly through the Preview SDK.

Todo
- [x] Cross-browser testing
- [x] Upgrade box-ui-elements


src/lib/viewers/error/PreviewErrorViewer.js
```diff
@@ -76,23 +80,12 @@ class PreviewErrorViewer extends BaseViewer {
 
         const { displayMessage, details, message } = error;
         const { file } = this.options;
+        const { extension } = file || {};
+        const iconExtension = extension === 'flv' || extension === 'tgz' || extension === 'zip' ? extension : undefined;
```
ConradJChan, line 84: Is the `zip` condition still needed?

ConradJChan (approved):

src/lib/viewers/error/PreviewErrorViewer.js
jstoffan: Good question. I left it as-is from the prior logic, but I'm not totally sure _why_ it's meant to work this way. Maybe we only expected this type of file in our error payloads at some point?

jstoffan: I'll leave it for now and follow-up with the product team about it as a follow-up.

PR #1350, opened 2021-04-07
fix(DocFindBar): Add aria-label to docfindbar input field

fsuarez-globant:
fix(DocFindBar): Add aria-label to docfindbar input field

### DIFF
--- a/src/i18n/en-US.properties
+++ b/src/i18n/en-US.properties
@@ -55,7 +55,7 @@ find_previous=Find previous
 # Button tooltip for closing the find bar
 find_close=Close
 # Search input Aria-label attribute
-search_input=Search input
+find_label=Search input
 
 # Error messages
 # Default preview error message
--- a/src/i18n/en-US.properties
+++ b/src/i18n/en-US.properties
@@ -54,6 +54,8 @@ find_next=Find next
 find_previous=Find previous
 # Button tooltip for closing the find bar
 find_close=Close
+# Search input Aria-label attribute
+search_input=Search input
 
 # Error messages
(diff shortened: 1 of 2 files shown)


src/lib/viewers/doc/DocFindBar.js
```diff
@@ -82,6 +82,7 @@ class DocFindBar extends EventEmitter {
         // Find input field
         this.findFieldEl = document.createElement('input');
         this.findFieldEl.classList.add('bp-doc-find-field');
+        this.findFieldEl.setAttribute('aria-label', 'Search input');
```
jstoffan, line 85: Hi @fsuarez-globant, thanks for submitting this change. We require all our user-facing strings to support internationalization. Can you please add this string to `src/i18n/en-US.properties` and use it via the `__` helper?

fsuarez-globant: @jstoffan of course, I'll add this string in the internationalization file.

src/i18n/en-US.properties
```diff
@@ -54,6 +54,8 @@ find_next=Find next
 find_previous=Find previous
 # Button tooltip for closing the find bar
 find_close=Close
+# Search input Aria-label attribute
+search_input=Search input
```
jstoffan, line 58: Can we use `find_label` to stay consistent with the other `find_` keys? That will keep them all together in the output JSON file, as well.

fsuarez-globant: @jstoffan I'll change this as well

jstoffan (approved):

PR #1352, opened 2021-04-08
fix(a11y): Add meaningful alt text to custom logo

tille:
- Missing alt attribute doesn't compliant a11y standards.

### DIFF
--- a/src/lib/PreviewUI.js
+++ b/src/lib/PreviewUI.js
@@ -48,9 +48,6 @@ class PreviewUI {
     /** @property {HTMLElement} - Preview container element which houses the sidebar and content */
     previewContainer;
 
-    /** @property {HTMLElement} - Preview container element which houses the custom-logo */
-    customLogoEl;
-
     /**
      * Destroy preview container content.
      *
@@ -115,10 +112,6 @@ class PreviewUI {
         this.container = this.container.querySelector(SELECTOR_BOX_PREVIEW_CONTAINER);
         this.previewContainer = this.container.querySelector(SELECTOR_BOX_PREVIEW);
         this.contentContainer = this.container.querySelector(SELECTOR_BOX_PREVIEW_CONTENT);
-        this.customLogoEl = this.container.querySelector(SELECTOR_BOX_PREVIEW_LOGO_CUSTOM);
-
-        // Set alternative accesible text
-        this.customLogoEl.setAttribute('alt', __('logo'));
 
         // Setup the header, buttons, and theme
         if (options.header !== 'none') {
@@ -376,11 +369,14 @@ class PreviewUI {
      */
     setupHeader(headerTheme, logoUrl) {
         const headerContainerEl = this.container.querySelector(SELECTOR_BOX_PREVIEW_HEADER_CONTAINER);
-        headerContainerEl.classList.remove(CLASS_HIDDEN);
-
+        const customLogoEl = this.container.querySelector(SELECTOR_BOX_PREVIEW_LOGO_CUSTOM);
         const headerEl = headerContainerEl.firstElementChild;
+        const defaultLogoEl = headerEl.querySelector(SELECTOR_BOX_PREVIEW_LOGO_DEFAULT);
+
+        headerContainerEl.classList.remove(CLASS_HIDDEN);
         headerEl.className = `${CLASS_BOX_PREVIEW_HEADER} ${CLASS_BOX_PREVIEW_BASE_HEADER}`;
         this.previewContainer.classList.add(CLASS_BOX_PREVIEW_HAS_HEADER);
+        defaultLogoEl.setAttribute('aria-label', __('logo'));
 
         // Setup theme, default is 'light'
         if (headerTheme === 'dark') {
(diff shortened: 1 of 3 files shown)


src/lib/PreviewUI.js
```diff
@@ -48,6 +48,9 @@ class PreviewUI {
     /** @property {HTMLElement} - Preview container element which houses the sidebar and content */
     previewContainer;
 
+    /** @property {HTMLElement} - Preview container element which houses the custom-logo */
+    customLogoEl;
```
ConradJChan, line 52: It doesn't seem like `customLogoEl` is referenced anywhere else, could the `alt` attribute be set as part of the `setupHeader` logic without introducing this class level property?

tille: the problem are these lines:

```
if (options.header !== 'none') {
     this.setupHeader(options.header, options.logoUrl);
```

if there's no `options.header` then `setupHeader` is not call,  so `customLogoEl` exists but it is hidden because the class `CLASS_HIDDEN`

however even if its hidden is an error for a11y, so that's why you need to be sure to always set the `alt` attribute

src/lib/PreviewUI.js
```diff
@@ -385,9 +392,8 @@ class PreviewUI {
             const defaultLogoEl = headerEl.querySelector(SELECTOR_BOX_PREVIEW_LOGO_DEFAULT);
             defaultLogoEl.classList.add(CLASS_HIDDEN);
 
-            const customLogoEl = headerEl.querySelector(SELECTOR_BOX_PREVIEW_LOGO_CUSTOM);
-            customLogoEl.src = logoUrl;
-            customLogoEl.classList.remove(CLASS_HIDDEN);
+            this.customLogoEl.src = logoUrl;
```
jstoffan, line 395: I don't think this covers all our use cases, since we have both a "default" element and a "custom" logo element. I suggest we rearrange `setupHeader` to apply the `alt` attribute to both.

src/lib/PreviewUI.js
```diff
@@ -112,6 +115,10 @@ class PreviewUI {
         this.container = this.container.querySelector(SELECTOR_BOX_PREVIEW_CONTAINER);
         this.previewContainer = this.container.querySelector(SELECTOR_BOX_PREVIEW);
         this.contentContainer = this.container.querySelector(SELECTOR_BOX_PREVIEW_CONTENT);
+        this.customLogoEl = this.container.querySelector(SELECTOR_BOX_PREVIEW_LOGO_CUSTOM);
```
jstoffan, line 118: We should try to avoid adding instance properties when possible. It seems like we should be able to handle all of this within `setupHeader`.

jstoffan (commented):
Thanks for the submission, @tille. I think we still need to iterate on the solution a bit.

src/lib/PreviewUI.js
jstoffan: @tille, can you please clarify how this affects the end user? The `CLASS_HIDDEN` class, `bp-is-hidden`, sets `display: none` on the element to which it's applied, which essentially removes it from the DOM as it is not accessible to the user.

src/lib/PreviewUI.js
```diff
@@ -369,11 +369,14 @@ class PreviewUI {
      */
     setupHeader(headerTheme, logoUrl) {
         const headerContainerEl = this.container.querySelector(SELECTOR_BOX_PREVIEW_HEADER_CONTAINER);
-        headerContainerEl.classList.remove(CLASS_HIDDEN);
-
+        const customLogoEl = this.container.querySelector(SELECTOR_BOX_PREVIEW_LOGO_CUSTOM);
```
jstoffan, line 372: Can we continue using `headerEl.querySelector` to retrieve this element?

src/lib/PreviewUI.js
```diff
@@ -369,11 +369,14 @@ class PreviewUI {
...
         const headerEl = headerContainerEl.firstElementChild;
+        const defaultLogoEl = headerEl.querySelector(SELECTOR_BOX_PREVIEW_LOGO_DEFAULT);
+
+        headerContainerEl.classList.remove(CLASS_HIDDEN);
```
jstoffan, line 376: Can we restore this line to its prior position? It doesn't seem related to the other changes.

src/lib/PreviewUI.js
```diff
@@ -369,11 +369,14 @@ class PreviewUI {
...
         headerEl.className = `${CLASS_BOX_PREVIEW_HEADER} ${CLASS_BOX_PREVIEW_BASE_HEADER}`;
         this.previewContainer.classList.add(CLASS_BOX_PREVIEW_HAS_HEADER);
+        defaultLogoEl.setAttribute('aria-label', __('logo'));
```
jstoffan, line 379: It looks like the attribute we're using is inconsistent. Should these both use `alt` or `aria-label`?

tille: for `defaultLogoEl` it has to be `aria-label` because its NOT an `img` element

tille: but is mandatory for ALL the `img` elements to have an `alt` attribute

jstoffan: Good catch. Should we add `role="img"` to the SVG, while we're at it?

jstoffan (approved):
LGTM, thanks for updating!

src/lib/PreviewUI.js
tille: not because ScreenReader would never reach the svg, they will focus the container element that's `defaultLogoEl`

tille: but we can add `role=img` to `<span class="bp-default-logo">`

jstoffan (approved):

PR #1353, opened 2021-04-08
feat(loading): Show ghost state for document preloaders and pages

jstoffan:
This change replaces the pulsing three blue bar crawler with a "shine" ghost state effect.

PR #1355, opened 2021-04-13
feat(loading): Update loading ux for pages, thumbnails, preloaders

jstoffan:
This change replaces the pulsing three blue bar crawler with a "shine" ghost state effect. Retry of #1353.

**Current**
loading-current

**Updated**
loading-placeholder


jstoffan:
I ran into an issue with this solution where the background of the pages flashes grey/white when resizing the window. Will investigate and push an update when I can.

src/lib/viewers/doc/DocPreloader.js
```diff
@@ -4,28 +4,26 @@ import {
...
-    PDFJS_HEIGHT_PADDING_PX,
 } from '../../constants';
 import { setDimensions, handleRepresentationBlobFetch } from '../../util';
 
 const EXIF_COMMENT_TAG_NAME = 'UserComment'; // Read EXIF data from 'UserComment' tag
 const EXIF_COMMENT_REGEX = /pdfWidth:([0-9.]+)pts,pdfHeight:([0-9.]+)pts,numPages:([0-9]+)/;
 
 const NUM_PAGES_DEFAULT = 2; // Default to 2 pages for preload if true number of pages cannot be read
-const NUM_PAGES_MAX = 500; // Don't show more than 500 placeholder pages
+const NUM_PAGES_MAX = 20;
```
ConradJChan, line 23: 500 seems like a lot, but are there any concerns with dropping it down to 20?

ConradJChan (approved):

src/lib/viewers/doc/DocPreloader.js
jstoffan: It will limit how much users can scroll down before the document fully loads. However, when the document loads, the scroll position is reset to the top anyway. It doesn't seem meaningful to have it set to such a high number.

PR #1357, opened 2021-04-16
feat(thumbnails): Update document thumbnails to latest designs

jstoffan:
I included the `custom-theme` styles in this PR to avoid any issues with unsynchronized releases. We can move them in a follow-up, if desired.

thumbnails-design

### DIFF
--- a/src/lib/viewers/doc/_theme.scss
+++ b/src/lib/viewers/doc/_theme.scss
@@ -1,0 +1,19 @@
+@import '~box-ui-elements/es/styles/constants/colors';
+
+.custom-theme {
+    .bp .bp-thumbnail.bp-thumbnail-is-selected {
+        .bp-thumbnail-nav {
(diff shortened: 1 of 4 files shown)


src/lib/viewers/doc/_theme.scss
```diff
@@ -0,0 +1,19 @@
+@import '~box-ui-elements/es/styles/constants/colors';
+
+.custom-theme {
+    .bp .bp-thumbnail.bp-thumbnail-is-selected {
+        .bp-thumbnail-nav {
+            box-shadow: 0 0 0 3px $bdl-gray, 0 4px 5px 0 rgba(0, 0, 0, .08);
```
ConradJChan, line 6: Should the offset-y and blur radius for the second shadow be variables since they're shared in multiple places?

ConradJChan (approved):

src/lib/viewers/doc/_theme.scss
jstoffan: Discussed offline.

PR #1358, opened 2021-04-20
feat(toolbar): Update control toolbars to latest design

jstoffan:
We may get some slightly different icons and colors from design, but this should be pretty much the final look otherwise.

**Documents**
controls-documents

**Images**
controls-images

**MultiImage**
controls-multiimage

**Text**
controls-text

**3D (Legacy Controls)**
controls-3d-legacy


ConradJChan (approved):
LGTM! 🚀

PR #1363, opened 2021-04-23
build(link buie): webpack config to use linked buie

megansmith-box:
Used this reference, to be able to use linked buie in content preview
https://github.com/box/box-ui-elements/blob/master/DEVELOPING.md#testing-ui-elements-in-a-parent-project
Also used EUA webpack/package.json config as a reference


package.json
```diff
@@ -54,7 +54,7 @@
         "eslint-plugin-cypress": "^2.8.1",
         "eslint-plugin-flowtype": "^4.5.2",
         "eslint-plugin-formatjs": "^1.5.4",
-        "eslint-plugin-import": "^2.19.1",
+        "eslint-plugin-import": "^2.22.1",
```
megansmith-box, line 57: Was running into an eslint issue when trying to commit and push up, used recommendation from this article to be able to commit without the --no-verify
https://github.com/airbnb/javascript/issues/2331

build/webpack.common.config.js
```diff
@@ -68,7 +83,14 @@ module.exports = language => {
...
                 'box-elements-messages': path.resolve(`node_modules/box-ui-elements/i18n/${language}`),
                 'react-intl-locale-data': path.resolve(`node_modules/react-intl/locale-data/${locale}`),
+                // force one copy of these packages:
+                'box-ui-elements/es': isBuieLinked
+                    ? path.resolve('node_modules/box-ui-elements/src')
+                    : path.resolve('node_modules/box-ui-elements/es'),
+                react: path.resolve('node_modules/react'),
+                'react-dom': path.resolve('node_modules/react-dom'),
             },
+            modules: ['src', 'node_modules'],
```
jstoffan, line 93: Does the change work without this line?

package.json
```diff
@@ -152,6 +152,8 @@
         }
     },
     "resolutions": {
-        "mojito-rb-gen/merge": "1.2.1"
+        "mojito-rb-gen/merge": "1.2.1",
+        "react": "17.0.1",
```
jstoffan, line 156: I don't think this is required since we don't use peer dependencies.

build/webpack.common.config.js
megansmith-box: It does, thank you!

package.json
```diff
@@ -124,9 +124,9 @@
         "release:patch": "./build/release.sh -p",
         "setup": "yarn install",
         "start": "yarn setup && yarn build:dev --watch",
-        "start:linked": "yarn link box-annotations && IS_LINKED=1 yarn start",
+        "start:linked": "f() { LINKED_LIBS=${@:-box-ui-elements} && yarn link $LINKED_LIBS && LINKED_LIBS=$LINKED_LIBS yarn start; }; f",
```
megansmith-box, line 127:
```suggestion
        "start:linked": "f() { LINKED_LIBS=${@:-box-annotations} && yarn link $LINKED_LIBS && LINKED_LIBS=$LINKED_LIBS yarn start; }; f",
```

yarn.lock
```diff
@@ -11996,7 +12022,7 @@ raw-loader@^3.1.0:
     loader-utils "^1.1.0"
     schema-utils "^2.0.1"
 
-react-dom@^17.0.1:
+react-dom@17.0.1, react-dom@^17.0.1:
```
jstoffan, line 12025: Hm, it looks like we still have some stray changes in here. Can we revert the `yarn.lock` file and re-run `yarn add -D eslint-plugin-import@^2.22.1`?

build/webpack.common.config.js
```diff
@@ -44,6 +46,19 @@ module.exports = language => {
                         name: '[name].[ext]',
                     },
                 },
+                // Transpile BUIE JS only if we are sym-linking
+                {
+                    test: /\.(js|mjs|ts|tsx)$/,
+                    include: /node_modules\/box-ui-elements/,
```
jstoffan, line 52: It's a little unusual to use both `include` and `exclude` for the same dependency. Does this affect our production bundle at all? Is it needed for the change to work?

megansmith-box: I copied this rule from EUA. It doesn't look like the exclude is needed. I do think it may affect production bundle as it's a common webpack file. I will investigate more the implications on production

jstoffan:
Please feel free to re-open if this is still needed once the concerns above are addressed.

PR #1364, opened 2021-04-26
feat(annotations): tooltip

megansmith-box:
Screen Shot 2021-04-25 at 11 07 10 AM

* I see that AnnotationsControls Component is imported for Images & Docs.
* `this.preview.updateExperiences` will be called whenever the experiences props passed in updates https://github.com/box/box-ui-elements/pull/2549/
    * this will updateExperiences in the viewer depending on the file
        * this will renderUI again & update ControlsRoot
            *   If the tooltip is showing we want the controls bar to show therefore I added `forceShow` prop to `ControlsLayer` Component
            * When the user closes the tooltip, we don't want the control bar to persist, so I added the callback `setWasClosedByUser`

### DIFF
--- a/src/lib/Preview.js
+++ b/src/lib/Preview.js
@@ -1787,7 +1787,7 @@ class Preview extends EventEmitter {
      * Updates experiences option after props have changed in parent app
      *
      * @public
-     * @param {Object} experiences - new experiences prop
+     * @param {Object} experiences - new experiences value
      * @return {void}
      */
     updateExperiences(experiences) {
--- a/src/lib/Preview.js
+++ b/src/lib/Preview.js
@@ -995,9 +995,6 @@ class Preview extends EventEmitter {
         // Add the response interceptor to the preview instance
         this.options.responseInterceptor = options.responseInterceptor;
 
-        // Add Targeting APIs for experiences in Preview to the preview instance
-        this.options.experiences = options.experiences || {};
-
         // Disable or enable viewers based on viewer options
         Object.keys(this.options.viewers).forEach(viewerName => {
             const isDisabled = this.options.viewers[viewerName].disabled;
@@ -1794,8 +1791,6 @@ class Preview extends EventEmitter {
      * @return {void}
      */
     updateExperiences(experiences) {
-        this.options.experiences = experiences;
-
         if (this.viewer && this.viewer.updateExperiences) {
             this.viewer.updateExperiences(experiences);
         }
--- a/src/lib/Preview.js
+++ b/src/lib/Preview.js
@@ -995,6 +995,9 @@ class Preview extends EventEmitter {
         // Add the response interceptor to the preview instance
         this.options.responseInterceptor = options.responseInterceptor;
 
+        // Add Targeting APIs for experiences in Preview to the preview instance
+        this.options.experiences = options.experiences || {};
+
         // Disable or enable viewers based on viewer options
         Object.keys(this.options.viewers).forEach(viewerName => {
             const isDisabled = this.options.viewers[viewerName].disabled;
@@ -1781,6 +1784,21 @@ class Preview extends EventEmitter {
             MOUSEMOVE_THROTTLE_MS - 500,
             true,
         );
+    }
+
+    /**
+     * Updates experiences option after props have changed in parent app
+     *
+     * @public
+     * @param {Object} experiences - new experiences prop
+     * @return {void}
+     */
+    updateExperiences(experiences) {
+        this.options.experiences = experiences;
+
+        if (this.viewer && this.viewer.updateExperiences) {
+            this.viewer.updateExperiences(experiences);
+        }
     }
 
     /**
--- a/src/lib/viewers/controls/annotations/AnnotationsButton.tsx
+++ b/src/lib/viewers/controls/annotations/AnnotationsButton.tsx
@@ -1,8 +1,7 @@
 import React from 'react';
 import classNames from 'classnames';
 import noop from 'lodash/noop';
-import TargetedClickThroughGuideTooltip from 'box-ui-elements/es/features/targeting/TargetedClickThroughGuideTooltip';
-import { AnnotationMode, TargetingApi } from '../../../types';
+import { AnnotationMode } from '../../../types';
 
 import './AnnotationsButton.scss';
 
@@ -13,8 +12,6 @@ export type Props = Omit<React.ButtonHTMLAttributes<HTMLButtonElement>, 'onClick
     isEnabled?: boolean;
     mode: AnnotationMode;
     onClick?: (mode: AnnotationMode) => void;
-    setWasClosedByUser?: (experienceName: string | undefined) => void;
-    tooltipFlowAnnotationsExperience?: TargetingApi;
 };
 
 export default function AnnotationsButton({
@@ -24,13 +21,6 @@ export default function AnnotationsButton({
     isEnabled = true,
     mode,
     onClick = noop,
-    setWasClosedByUser = noop,
-    tooltipFlowAnnotationsExperience = {
-        canShow: false,
-        onClose: noop,
-        onComplete: noop,
-        onShow: noop,
-    },
     ...rest
 }: Props): JSX.Element | null {
     if (!isEnabled) {
@@ -38,32 +28,15 @@ export default function AnnotationsButton({
     }
 
     return (
-        <TargetedClickThroughGuideTooltip
-            body={__('annotations_tooltip_body')}
-            className="preview-annotations-tooltip"
-            shouldTarget={tooltipFlowAnnotationsExperience.canShow}
-            showCloseButton
-            title={__('annotations_tooltip_title')}
-            useTargetingApi={(): TargetingApi => {
-                return {
-                    ...tooltipFlowAnnotationsExperience,
-                    onClose: (): void => {
-                        setWasClosedByUser('tooltipFlowAnnotationsExperience');
-                        tooltipFlowAnnotationsExperience.onClose();
-                    },
-                };
-            }}
+        <button
+            className={classNames('bp-AnnotationsButton', className, {
+                'bp-is-active': isActive,
+            })}
+            onClick={(): void => onClick(mode)}
+            type="button"
+            {...rest}
         >
-            <button
-                className={classNames('bp-AnnotationsButton', className, {
-                    'bp-is-active': isActive,
-                })}
-                onClick={(): void => onClick(mode)}
-                type="button"
-                {...rest}
-            >
-                {children}
-            </button>
-        </TargetedClickThroughGuideTooltip>
+            {children}
+        </button>
     );
 }
--- a/src/lib/viewers/controls/annotations/AnnotationsButton.tsx
+++ b/src/lib/viewers/controls/annotations/AnnotationsButton.tsx
@@ -2,7 +2,6 @@ import React from 'react';
 import classNames from 'classnames';
 import noop from 'lodash/noop';
 import { AnnotationMode } from '../../../types';
-
 import './AnnotationsButton.scss';
 
 export type Props = Omit<React.ButtonHTMLAttributes<HTMLButtonElement>, 'onClick'> & {
--- a/src/lib/viewers/controls/annotations/AnnotationsButton.tsx
+++ b/src/lib/viewers/controls/annotations/AnnotationsButton.tsx
@@ -1,7 +1,9 @@
 import React from 'react';
 import classNames from 'classnames';
 import noop from 'lodash/noop';
-import { AnnotationMode } from '../../../types';
+import TargetedClickThroughGuideTooltip from 'box-ui-elements/es/features/targeting/TargetedClickThroughGuideTooltip';
+import { AnnotationMode, TargetingApi } from '../../../types';
+
 import './AnnotationsButton.scss';
 
 export type Props = Omit<React.ButtonHTMLAttributes<HTMLButtonElement>, 'onClick'> & {
@@ -11,6 +13,8 @@ export type Props = Omit<React.ButtonHTMLAttributes<HTMLButtonElement>, 'onClick
     isEnabled?: boolean;
     mode: AnnotationMode;
     onClick?: (mode: AnnotationMode) => void;
+    setWasClosedByUser?: (experienceName: string | undefined) => void;
+    tooltipFlowAnnotationsExperience?: TargetingApi;
 };
 
 export default function AnnotationsButton({
@@ -20,6 +24,13 @@ export default function AnnotationsButton({
     isEnabled = true,
     mode,
     onClick = noop,
+    setWasClosedByUser = noop,
+    tooltipFlowAnnotationsExperience = {
+        canShow: false,
+        onClose: noop,
+        onComplete: noop,
+        onShow: noop,
+    },
     ...rest
 }: Props): JSX.Element | null {
     if (!isEnabled) {
@@ -27,15 +38,32 @@ export default function AnnotationsButton({
     }
 
     return (
-        <button
-            className={classNames('bp-AnnotationsButton', className, {
-                'bp-is-active': isActive,
-            })}
-            onClick={(): void => onClick(mode)}
-            type="button"
-            {...rest}
+        <TargetedClickThroughGuideTooltip
+            body={__('annotations_tooltip_body')}
+            className="preview-annotations-tooltip"
+            shouldTarget={tooltipFlowAnnotationsExperience.canShow}
+            showCloseButton
+            title={__('annotations_tooltip_title')}
+            useTargetingApi={(): TargetingApi => {
+                return {
+                    ...tooltipFlowAnnotationsExperience,
+                    onClose: (): void => {
+                        setWasClosedByUser('tooltipFlowAnnotationsExperience');
+                        tooltipFlowAnnotationsExperience.onClose();
+                    },
+                };
+            }}
         >
-            {children}
-        </button>
+            <button
+                className={classNames('bp-AnnotationsButton', className, {
+                    'bp-is-active': isActive,
+                })}
+                onClick={(): void => onClick(mode)}
+                type="button"
+                {...rest}
+            >
+                {children}
+            </button>
+        </TargetedClickThroughGuideTooltip>
     );
 }
--- a/src/lib/viewers/controls/annotations/AnnotationsControls.tsx
+++ b/src/lib/viewers/controls/annotations/AnnotationsControls.tsx
@@ -6,27 +6,33 @@ import IconDrawing24 from '../icons/IconDrawing24';
 import IconHighlightText16 from '../icons/IconHighlightText16';
 import IconRegion24 from '../icons/IconRegion24';
 import useFullscreen from '../hooks/useFullscreen';
-import { AnnotationMode } from '../../../types';
+import { AnnotationMode, TargetingApi } from '../../../types';
 import './AnnotationsControls.scss';
 
 export type Props = {
     annotationColor?: string;
     annotationMode?: AnnotationMode;
+    experiences: {
+        [name: string]: TargetingApi;
+    };
     hasDrawing?: boolean;
     hasHighlight?: boolean;
     hasRegion?: boolean;
     onAnnotationModeClick?: ({ mode }: { mode: AnnotationMode }) => void;
     onAnnotationModeEscape?: () => void;
+    setWasClosedByUser?: (experienceName: string | undefined) => void;
 };
 
 export default function AnnotationsControls({
     annotationColor = bdlBoxBlue,
     annotationMode = AnnotationMode.NONE,
+    experiences,
     hasDrawing = false,
     hasHighlight = false,
     hasRegion = false,
     onAnnotationModeClick = noop,
     onAnnotationModeEscape = noop,
+    setWasClosedByUser = noop,
 }: Props): JSX.Element | null {
     const isFullscreen = useFullscreen();
     const showDrawing = !isFullscreen && hasDrawing;
@@ -76,7 +82,9 @@ export default function AnnotationsControls({
                 isEnabled={showDrawing}
                 mode={AnnotationMode.DRAWING}
                 onClick={handleModeClick}
+                setWasClosedByUser={setWasClosedByUser}
                 title={__('drawing_comment')}
+                tooltipFlowAnnotationsExperience={experiences.tooltipFlowAnnotationsExperience}
             >
                 <IconDrawing24 fill={isDrawingActive ? annotationColor : '#fff'} />
             </AnnotationsButton>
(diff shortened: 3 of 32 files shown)


src/lib/viewers/controls/annotations/__tests__/AnnotationsControls-test.tsx
```diff
@@ -18,10 +18,14 @@ describe('AnnotationsControls', () => {
         let unmount = (): void => {
             // placeholder
         };
+        let found = false;
 
         beforeEach(() => {
             jest.spyOn(React, 'useEffect').mockImplementation(cb => {
-                unmount = cb() as () => void; // Enzyme unmount helper does not currently invoke useEffect cleanup
+                if (!found) {
```
megansmith-box, line 25: After adding `TargetedClickThroughGuideTooltip` there is now more than one useEffect spied on. The imported useEffect from BUIE that is being spied on is https://github.com/box/box-ui-elements/blob/master/src/features/targeting/utils/useOnClickBody.js.

src/lib/viewers/controls/controls-root/ControlsRoot.tsx
```diff
@@ -67,7 +82,32 @@ export default class ControlsRoot {
...
+    shouldForceShow(): boolean {
+        if (this.wasClosedByUser.tooltipFlowAnnotationsExperience) {
+            return false;
+        }
+
+        if (!this.experiences.tooltipFlowAnnotationsExperience) {
+            return false;
+        }
+
+        return this.experiences.tooltipFlowAnnotationsExperience.canShow;
```
ConradJChan, line 94: Is there a way this logic could be enhanced to scale to support more experiences?

src/lib/viewers/controls/controls-layer/ControlsLayer.tsx
```diff
@@ -74,12 +75,21 @@ export default function ControlsLayer({ children, onMount = noop }: Props): JSX.
...
+        }
+    }, [forceShow]);
+
     // Destroy timeouts on unmount
     React.useEffect(() => helpersRef.current.clean, []);
 
     return (
         <div
-            className={`bp-ControlsLayer ${isShown ? SHOW_CLASSNAME : ''}`}
+            className={`bp-ControlsLayer ${isShown || forceShow ? SHOW_CLASSNAME : ''}`}
```
ConradJChan, line 92: Does `forceShow` need to be maintained separately from `isShown` state? Or can it be rolled into the state value?

src/lib/viewers/controls/annotations/AnnotationsControls.tsx
```diff
@@ -76,7 +82,9 @@ export default function AnnotationsControls({
                 isEnabled={showDrawing}
                 mode={AnnotationMode.DRAWING}
                 onClick={handleModeClick}
+                setWasClosedByUser={setWasClosedByUser}
```
ConradJChan, line 85: Does `setWasClosedByUser` need to be passed down or can the `tooltipFlowAnnotationsExperience` be modified here to include `setWasClosedByUser` so that it can be passed transparently in the child component

src/lib/viewers/controls/controls-root/ControlsRoot.tsx
```diff
@@ -67,7 +82,32 @@ export default class ControlsRoot {
         this.controlsEl.classList.remove('bp-is-hidden');
     }
 
+    shouldForceShow(): boolean {
+        if (this.wasClosedByUser.tooltipFlowAnnotationsExperience) {
```
jstoffan, line 86: Can we move this logic to another component? ControlsRoot is currently pretty generic and shouldn't know about what controls/experiments it's rendering. I'm thinking something like:

```
<ControlsLayer onMount={this.handleMount}>
    <ControlsExperiments onClosed={this.helpers.reset()}>
        {controls}
    </ControlsExperiments>
</ControlsLayer>

File: src/lib/viewers/controls/controls-layer/ControlsLayer.tsx [hunk 6]
```diff
@@ -10,13 +10,14 @@ export type Helpers = {
 
 export type Props = {
     children: React.ReactNode;
+    forceShow?: boolean;
```
jstoffan, line 13: Can we avoid having a separate prop for showing/hiding? It could get difficult to keep things synced up.

src/lib/viewers/controls/annotations/AnnotationsControls.tsx
```diff
@@ -6,27 +6,33 @@ import IconDrawing24 from '../icons/IconDrawing24';
...
     hasRegion?: boolean;
     onAnnotationModeClick?: ({ mode }: { mode: AnnotationMode }) => void;
     onAnnotationModeEscape?: () => void;
+    setWasClosedByUser?: (experienceName: string | undefined) => void;
 };
 
 export default function AnnotationsControls({
     annotationColor = bdlBoxBlue,
     annotationMode = AnnotationMode.NONE,
+    experiences,
```
jstoffan, line 29: Can we avoid prop drilling by using context?

src/lib/viewers/controls/annotations/AnnotationsButton.tsx
```diff
@@ -20,22 +24,46 @@ export default function AnnotationsButton({
...
-        <button
-            className={classNames('bp-AnnotationsButton', className, {
-                'bp-is-active': isActive,
-            })}
-            onClick={(): void => onClick(mode)}
-            type="button"
-            {...rest}
+        <TargetedClickThroughGuideTooltip
+            body={__('annotations_tooltip_body')}
+            className="preview-annotations-tooltip"
```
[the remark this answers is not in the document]
jstoffan: We also use SUIT CSS naming conventions, so `bp-AnnotationsTooltip` would work here.

src/lib/viewers/controls/annotations/AnnotationsButton.tsx
```diff
@@ -20,22 +24,46 @@ export default function AnnotationsButton({
...
 
     return (
-        <button
-            className={classNames('bp-AnnotationsButton', className, {
-                'bp-is-active': isActive,
-            })}
-            onClick={(): void => onClick(mode)}
-            type="button"
-            {...rest}
+        <TargetedClickThroughGuideTooltip
```
jstoffan, line 41: Can this live in a separate component?

src/lib/Preview.js
```diff
@@ -1783,6 +1786,21 @@ class Preview extends EventEmitter {
...
 
+    /**
+     * Updates experiences option after props have changed in parent app
+     *
+     * @public
+     * @param {Object} experiences - new experiences prop
+     * @return {void}
+     */
+    updateExperiences(experiences) {
+        this.options.experiences = experiences;
```
jstoffan, line 1797: Do we need to keep a copy of this on the Preview instance, viewer instance, and controls instance? Keeping them all in sync seems to be adding some overhead.

src/lib/viewers/controls/annotations/AnnotationsButton.tsx
```diff
@@ -11,6 +13,8 @@ export type Props = Omit<React.ButtonHTMLAttributes<HTMLButtonElement>, 'onClick
     isEnabled?: boolean;
     mode: AnnotationMode;
     onClick?: (mode: AnnotationMode) => void;
+    setWasClosedByUser?: (experienceName: string | undefined) => void;
+    tooltipFlowAnnotationsExperience?: TargetingApi;
 };
 
 export default function AnnotationsButton({
```
jstoffan, line 20: Should every annotation button be wrapped in this tooltip? Will only one show?

src/lib/viewers/controls/annotations/AnnotationsButton.tsx
```diff
@@ -20,22 +24,46 @@ export default function AnnotationsButton({
     isEnabled = true,
     mode,
     onClick = noop,
+    setWasClosedByUser = noop,
+    tooltipFlowAnnotationsExperience = {
```
jstoffan, line 28: Can we check for this value being defined and not render the tooltip if it's missing?

megansmith-box: I've updated the shouldTarget prop for `TargetedClickThroughGuideTooltip` to
```
const { tooltipFlowAnnotationsExperience } = experiences;
shouldTarget={showDrawing && tooltipFlowAnnotationsExperience && tooltipFlowAnnotationsExperience.canShow}
```
if shouldTarget equals false it will not render the tooltip. Did you want me to be more explicit? such as:
```
if (!tooltipFlowAnnotationsExperience) {
    return children;
}

return (
    <tooltipFlowAnnotationsExperience>
        {children}
    </tooltipFlowAnnotationsExperience>
);
```

src/lib/viewers/controls/annotations/AnnotationsButton.tsx
megansmith-box: My first way of doing it was every AnnotationsButton was wrapped in this tooltip, but it was only ever shown for one due to the shouldTarget prop. I've moved the tooltip out to the Annotations Control component. Now it only wraps one button. Let me know what you think. I wasn't sure if creating another component was a good idea. Thank you!

src/lib/viewers/controls/controls-root/ControlsRoot.tsx
megansmith-box: Let me know what you think of my solution that can be used by more experiences once I push my changes!

src/lib/viewers/controls/annotations/AnnotationsControls.tsx
megansmith-box: Yes, I think that's a good idea!

src/lib/Preview.js
megansmith-box: Cool! It looks like we don't need it in Preview instance 👍 We do need it in viewer instance to pass it to the context. I think we need it in controls but will continue to look

megansmith-box:
Still WIP, I'm trying to figure out how to hide/show control bar when tooltip is showing & closed

megansmith-box:
Thank you @jstoffan for your help! I referenced your example of using 2 different contexts. The only thing I had to add was a `wasClosedByUser` state in the AnnotationsTaretedTooltip component. This makes sure that we don't call `setIsForced(true);` after the tooltip is closed. There is similar logic in BUIE to make sure onShow is only called once https://github.com/box/box-ui-elements/blob/master/src/features/targeting/useShowOne.js#L14

src/lib/viewers/controls/annotations/AnnotationsControls.tsx
```diff
@@ -2,11 +2,13 @@ import React from 'react';
 import noop from 'lodash/noop';
 import { bdlBoxBlue } from 'box-ui-elements/es/styles/variables';
 import AnnotationsButton from './AnnotationsButton';
+import AnnotationsTargetedTooltip from './AnnotationsTargetedTooltip';
 import IconDrawing24 from '../icons/IconDrawing24';
 import IconHighlightText16 from '../icons/IconHighlightText16';
 import IconRegion24 from '../icons/IconRegion24';
 import useFullscreen from '../hooks/useFullscreen';
 import { AnnotationMode } from '../../../types';
+
```
jstoffan, line 11: We typically group all imports together without whitespace.

src/lib/viewers/controls/annotations/AnnotationsControls.tsx
```diff
@@ -69,17 +71,19 @@ export default function AnnotationsControls({
...
-                data-testid="bp-AnnotationsControls-drawBtn"
-                isActive={isDrawingActive}
-                isEnabled={showDrawing}
-                mode={AnnotationMode.DRAWING}
-                onClick={handleModeClick}
-                title={__('drawing_comment')}
-            >
-                <IconDrawing24 fill={isDrawingActive ? annotationColor : '#fff'} />
-            </AnnotationsButton>
+            <AnnotationsTargetedTooltip isEnabled={showDrawing}>
```
jstoffan, line 74: Do we have a preference for which annotation type we want to draw the most attention to? It seems like the choice would be between drawing and regions, the latter of which may be a bit simpler of a user experience for a first-time user. It may be worth checking with the product team on this question.

src/lib/viewers/controls/annotations/AnnotationsTargetedTooltip.tsx
```diff
@@ -0,0 +1,57 @@
+import React from 'react';
+import TargetedClickThroughGuideTooltip from 'box-ui-elements/es/features/targeting/TargetedClickThroughGuideTooltip';
```
jstoffan, line 2: It looks like this relies on `react-tether`. Can we ensure that we're using the same version BUIE expects? Can we also do a comparison of the production bundle size before and after this change?

src/lib/viewers/controls/annotations/AnnotationsTargetedTooltip.tsx
```diff
@@ -0,0 +1,57 @@
...
+
+    const shouldTarget = !!(
+        isEnabled &&
+        experiences &&
+        experiences.tooltipFlowAnnotationsExperience &&
+        experiences.tooltipFlowAnnotationsExperience.canShow
+    );
+
+    if (!shouldTarget) {
+        return <div>{children}</div>;
```
jstoffan, line 24: Can we omit the wrapping `div` here?

src/lib/viewers/controls/annotations/__tests__/AnnotationsControls-test.tsx
jstoffan: Can we mock `AnnotationsTargetedTooltip`, instead? It doesn't look like it's under test at the moment.

src/lib/viewers/controls/experiences/ExperiencesProvider.tsx
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
+import ExperiencesContext from './ExperiencesContext';
+import { TargetingApi } from '../../../types';
+
+export type Props = React.PropsWithChildren<{
+    experiences?: {
```
jstoffan, line 6: Can we use the shared `Experiences` type here?

src/lib/viewers/controls/annotations/AnnotationsControls.tsx
megansmith-box: I will confirm with product 👍

megansmith-box: Thanks, it looks like they prefer the comment region button!

src/lib/viewers/controls/annotations/AnnotationsTargetedTooltip.tsx
megansmith-box: I updated react-tether to 1.0.5. I will compare the bundle sizes now...

megansmith-box: It looks like Preview.js went up about 100 Kib and Preview.css 10 Kib. Do you think that sounds reasonable?
Screen Shot 2021-05-03 at 2 17 54 PM

src/lib/viewers/controls/annotations/AnnotationsTargetedTooltip.tsx
```diff
@@ -0,0 +1,57 @@
...
+
+    const shouldTarget = !!(
+        isEnabled &&
+        experiences &&
+        experiences.tooltipFlowAnnotationsExperience &&
+        experiences.tooltipFlowAnnotationsExperience.canShow
+    );
+
+    if (!shouldTarget) {
+        return <>{children}</>;
```
jstoffan, line 24: We should be able to do `return children;` here.

src/lib/viewers/controls/annotations/__tests__/AnnotationsTargetedTooltip-test.tsx
```diff
@@ -0,0 +1,49 @@
...
+
+    describe('render', () => {
+        test('should return tooltip when is enabled', () => {
+            const wrapper = getWrapper({
+                isEnabled: true,
+            });
+
+            expect(wrapper.children().text()).not.toBe('Child');
+            expect(wrapper.children().prop('shouldTarget')).toBe(true);
+            expect(wrapper.children().prop('body')).toBeTruthy();
```
jstoffan, line 35: Can we use more explicit assertions here?

src/lib/viewers/controls/controls-layer/ControlsLayer.tsx
```diff
@@ -1,5 +1,7 @@
 import React from 'react';
 import noop from 'lodash/noop';
+import ControlsLayerContext from './ControlsLayerContext';
+
```
jstoffan, line 4: We typically group our imports without extra whitespace.

src/lib/viewers/controls/controls-layer/ControlsLayer.tsx
```diff
@@ -18,6 +20,7 @@ export const SHOW_CLASSNAME = 'bp-is-visible';
 
 export default function ControlsLayer({ children, onMount = noop }: Props): JSX.Element {
     const [isShown, setIsShown] = React.useState(false);
+    const [isForced, setIsForced] = React.useState(false);
```
jstoffan, line 23: Can we bump this up a line so it's alphabetized?

src/lib/viewers/controls/annotations/AnnotationsTargetedTooltip.tsx
jstoffan: Hm, that seems like a big jump given the nature of the change. Can we use the bundle analyzer to make sure we're not pulling in code we shouldn't be, like another version of React or something? 

We should also check the built CSS file to make sure we're not bringing in extra global styles from the Button components in GuideTooltip. I think we've run into issues around that before.

megansmith-box: Screen Shot 2021-05-04 at 8 59 06 AM
preview.js is now at 755 KiB. I ran the analyzer and see that the Tooltip Component in BUIE is about 15.8 KiB, therefore I think the Tooltip component + React Tether make up most of the 54 additional KiB. I'm not sure what's included in the 10 KiB extra for preview.css. Do you know how to see what styles are included in the webpack analyzer plugin?

src/lib/viewers/controls/annotations/AnnotationsTargetedTooltip.tsx
```diff
@@ -0,0 +1,77 @@
...
+            {children}
+        </TargetedClickThroughTooltip>
+    );
+}
+
+const TargetedClickThroughTooltip = withTargetedClickThrough(
+    ({ children, ...props }: { children: React.ReactNode }): JSX.Element => {
+        return (
+            <Tooltip {...props}>
+                <>{children}</>
```
jstoffan, line 71: Can we omit the `<>` wrapper here?

src/lib/viewers/controls/annotations/__tests__/AnnotationsTargetedTooltip-test.tsx
```diff
@@ -0,0 +1,48 @@
...
+            canShow: true,
+            onClose: jest.fn(),
+            onComplete: jest.fn(),
+            onShow: jest.fn(),
+        },
+    },
+    setIsForced: jest.fn(),
+}));
+
+React.useContext = mockUseContext;
```
jstoffan, line 17: Can we use `jest.spyOn(React, 'useContext')` in a `beforeEach`?

src/lib/viewers/controls/annotations/AnnotationsTargetedTooltip.tsx
```diff
@@ -0,0 +1,77 @@
...
+                    },
+                };
+            }}
+        >
+            {children}
+        </TargetedClickThroughTooltip>
+    );
+}
+
+const TargetedClickThroughTooltip = withTargetedClickThrough(
```
jstoffan, line 67: Can we move this to a separate file?

src/lib/viewers/controls/annotations/AnnotationsTargetedTooltip.tsx
```diff
@@ -0,0 +1,77 @@
+import React from 'react';
+import Tooltip from 'box-ui-elements/es/components/tooltip/Tooltip';
+import { withTargetedClickThrough } from 'box-ui-elements/es/features/targeting/hocs';
+import { ControlsLayerContext } from '../controls-layer';
+import { ExperiencesContext } from '../experiences';
+import { TargetingApi } from '../../../types';
+import './AnnotationsTargetedTooltip.scss';
+
+export type Props = {
+    children: React.ReactElement;
```
jstoffan, line 10: It looks like we're using React.ReactNode in some places and React.ReactElement in others. Should we use the `React.PropsWithChildren` helper?

src/lib/viewers/controls/annotations/AnnotationsTargetedTooltip.scss
```diff
@@ -0,0 +1,13 @@
...
+    padding: 18px 0 12px 18px;
+}
+
+.bp-AnnotationsTooltip-title {
+    font-size: 18px;
+}
+
+.bp-AnnotationsTooltip-body {
+    font-size: 14px;
+    line-height: 1.57;
```
jstoffan, line 12: Can we use `20px` here? It looks like that was the intended result.

src/lib/viewers/controls/annotations/AnnotationsTargetedTooltip.scss
```diff
@@ -0,0 +1,13 @@
+.bp-AnnotationsTooltip {
+    min-width: 408px;
```
jstoffan, line 2: Should this be max-width to match BUIE?

jstoffan (commented):
This is looking great!

src/lib/viewers/controls/annotations/AnnotationsTargetedTooltip.tsx
megansmith-box: I would like to. However when I use `React.PropsWithChildren` or `children: React.ReactNode` & `return children`. I get the following type error:
```
Type 'ReactNode' is not assignable to type 'Element | null'.
 Type 'undefined' is not assignable to type 'Element | null'.
```
It does work when I use `React.PropsWithChildren` & `return <>children</>`
It also is working when I use `children: React.ReactElement` & `return children`

src/lib/viewers/controls/annotations/__tests__/AnnotationsTargetedTooltip-test.tsx
megansmith-box: The problem when using `jest.spyOn(React, 'useContext')` versus how I have it, is with the spy version that experiences value is `{}` which results in shouldTarget always being false.

jstoffan (approved):
🚀

PR #1371, opened 2021-05-04
refactor(media): Renaming MediaSetting to Settings

ConradJChan:


src/lib/viewers/controls/settings/SettingsAutoplay.tsx
```diff
@@ -1,34 +1,34 @@
...
+import Settings, { Menu } from '.';
 
 export type Props = {
     autoplay: boolean;
     onAutoplayChange: (autoplay: boolean) => void;
 };
 
-export default function MediaSettingsMenuAutoplay({ autoplay, onAutoplayChange }: Props): JSX.Element {
-    const { setActiveMenu } = React.useContext(MediaSettings.Context);
+export default function SettingsMenuAutoplay({ autoplay, onAutoplayChange }: Props): JSX.Element {
```
jstoffan, line 9: Let's leave the consumers of the base components in the `media` folder, since they're specific to those viewers.

src/lib/viewers/media/MP3.scss
```diff
@@ -25,6 +25,7 @@
     .bp-media-gear-icon {
         position: absolute;
         right: 10px;
+        width: 36px;
```
jstoffan, line 28: Is this change intentional?

ConradJChan: Yes, because the `SettingsToggle` no longer uses the `bp-MediaButton` mixin, so the width needs to be adjusted

jstoffan (approved):

PR #1375, opened 2021-05-11
feat(annotations): update mode if necessary for experiences

megansmith-box:
For Images in Preview, we noticed that the Region is default Selected.
- We want to set the default selected to None, if this tooltip experience can show
We also want the tooltip to close when you click on the mode button
Screen Shot 2021-05-11 at 9 26 12 AM

### DIFF
--- a/src/lib/viewers/controls/annotations/AnnotationsTargetedTooltip.tsx
+++ b/src/lib/viewers/controls/annotations/AnnotationsTargetedTooltip.tsx
@@ -12,8 +12,8 @@ export type Props = React.PropsWithChildren<{
 function AnnotationsTargetedTooltip({ children, isEnabled = false }: Props): JSX.Element | null {
     const { experiences } = React.useContext(ExperiencesContext);
     const { setIsForced } = React.useContext(ControlsLayerContext);
+    const [shouldTargetAnnotationsTooltip, setShouldTargetAnnotationsTooltip] = React.useState(true);
     const [wasClosedByUser, setWasClosedByUser] = React.useState(false);
-    const [shouldTargetAnnotationsShareTooltip, setShouldTargetAnnotationsShareTooltip] = React.useState(true);
 
     const shouldTarget = !!(
         isEnabled &&
@@ -29,7 +29,7 @@ function AnnotationsTargetedTooltip({ children, isEnabled = false }: Props): JSX
     return (
         <TargetedClickThroughTooltip
             className="bp-AnnotationsTooltip"
-            shouldTarget={shouldTargetAnnotationsShareTooltip}
+            shouldTarget={shouldTargetAnnotationsTooltip}
             showCloseButton
             text={
                 <div>
@@ -47,7 +47,7 @@ function AnnotationsTargetedTooltip({ children, isEnabled = false }: Props): JSX
                     },
                     onComplete: (): void => {
                         experiences.tooltipFlowAnnotationsExperience.onComplete();
-                        setShouldTargetAnnotationsShareTooltip(false);
+                        setShouldTargetAnnotationsTooltip(false);
                     },
                     onShow: (): void => {
                         experiences.tooltipFlowAnnotationsExperience.onShow();
--- a/src/lib/viewers/controls/annotations/AnnotationsTargetedTooltip.tsx
+++ b/src/lib/viewers/controls/annotations/AnnotationsTargetedTooltip.tsx
@@ -10,9 +10,10 @@ export type Props = React.PropsWithChildren<{
 }>;
 
 function AnnotationsTargetedTooltip({ children, isEnabled = false }: Props): JSX.Element | null {
-    const { experiences } = React.useContext(ExperiencesContext);
+    const { experiences, updateModeIfNecessary } = React.useContext(ExperiencesContext);
     const { setIsForced } = React.useContext(ControlsLayerContext);
     const [wasClosedByUser, setWasClosedByUser] = React.useState(false);
+    const [shouldTargetAnnotationsShareTooltip, setShouldTargetAnnotationsShareTooltip] = React.useState(true);
 
     const shouldTarget = !!(
         isEnabled &&
@@ -28,7 +29,7 @@ function AnnotationsTargetedTooltip({ children, isEnabled = false }: Props): JSX
     return (
         <TargetedClickThroughTooltip
             className="bp-AnnotationsTooltip"
-            shouldTarget
+            shouldTarget={shouldTargetAnnotationsShareTooltip}
             showCloseButton
             text={
                 <div>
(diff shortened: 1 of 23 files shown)


jstoffan (changes requested):
I think we'll need to tie this into the "discoverability" conditions in both image and document annotators. Otherwise, the app could wind up in an inconsistent state. I'll reach out directly, as well.

src/lib/viewers/controls/annotations/AnnotationsTargetedTooltip.tsx
```diff
@@ -44,6 +45,10 @@ function AnnotationsTargetedTooltip({ children, isEnabled = false }: Props): JSX
                         experiences.tooltipFlowAnnotationsExperience.onClose();
                         setIsForced(false);
                     },
+                    onComplete: (): void => {
+                        experiences.tooltipFlowAnnotationsExperience.onComplete();
+                        setShouldTargetAnnotationsTooltip(false);
```
jstoffan, line 50: I'm a little confused as to how this is different than our `shouldTarget` variable above. Does calling `onComplete()` not set `canShow` to false?

src/lib/viewers/image/__tests__/ImageViewer-test.js
```diff
@@ -895,4 +895,26 @@ describe('lib/viewers/image/ImageViewer', () => {
...
         });
     });
+
+    describe('isDiscoverabilityEnabled()', () => {
+        [
+            [true, true],
+            [true, false],
+            [false, true],
+            [false, false],
+        ].forEach(([canShow, enableAnnotationsImageDiscoverability]) => {
```
jstoffan, line 905: Can we use `test.each` for this, instead?

src/lib/viewers/image/ImageViewer.js
```diff
@@ -37,7 +37,9 @@ class ImageViewer extends ImageBaseViewer {
         this.updatePannability = this.updatePannability.bind(this);
 
         this.annotationControlsFSM = new AnnotationControlsFSM(
-            this.options.enableAnnotationsImageDiscoverability ? AnnotationState.REGION_TEMP : AnnotationState.NONE,
```
jstoffan, line 40: It seems like we would need to update all usages of `this.options. enableAnnotationsImageDiscoverability` to use this new method?

jstoffan (commented):
Do we need a change for DocBaseViewer, as well? Or any other viewers?

src/lib/viewers/controls/annotations/AnnotationsTargetedTooltip.tsx
megansmith-box: That is correct, calling onComplete does not set canShow to false.

src/lib/viewers/image/ImageViewer.js
megansmith-box: good point 👍

src/lib/viewers/image/ImageBaseViewer.js
```diff
@@ -137,6 +138,20 @@ class ImageBaseViewer extends BaseViewer {
...
+     *
+     * @private
+     * @param {string} discoverabilityType
+     * @return {boolean} value of whether discoverability is enabled for given type
+     */
+    isDiscoverabilityEnabled(discoverabilityType) {
+        const { experiences = {} } = this.options;
+        const canShow = Object.values(experiences).some(experience => experience.canShow);
+
+        return !canShow && !!this.options[discoverabilityType];
```
jstoffan, line 152: Since we're inside ImageViewer, can we bake the check for `this.options.enableAnnotationsImageDiscoverability` inside this method to avoid having to pass in a string?

src/lib/viewers/image/__tests__/ImageViewer-test.js
```diff
@@ -895,4 +895,28 @@ describe('lib/viewers/image/ImageViewer', () => {
...
+            'should return correct value for isDiscoverabilityEnabled',
+            ({ canShow, enableAnnotationsImageDiscoverability }) => {
+                image.options.experiences = {
+                    tooltipFlowAnnotationsExperience: {
+                        canShow,
+                    },
+                };
+                image.options.enableAnnotationsImageDiscoverability = enableAnnotationsImageDiscoverability;
+
+                expect(image.isDiscoverabilityEnabled('enableAnnotationsImageDiscoverability')).toBe(
```
jstoffan, line 916: Can we remove the string parameter here?

megansmith-box: oops, fixed!

jstoffan (dismissed):

src/lib/viewers/image/ImageBaseViewer.js
```diff
@@ -137,6 +138,19 @@ class ImageBaseViewer extends BaseViewer {
...
     }
 
+    /**
+     * Determines whether discoverability is enabled
+     *
+     * @private
+     * @return {boolean} value of whether discoverability is enabled for given type
+     */
+    isDiscoverabilityEnabled() {
+        const { experiences = {} } = this.options;
```
ConradJChan, line 148: nit: We could destructure `enableAnnotationsImageDiscoverability` here as well

src/lib/viewers/image/ImageBaseViewer.js
```diff
@@ -137,6 +138,19 @@ class ImageBaseViewer extends BaseViewer {
...
+        const canShow = Object.values(experiences).some(experience => experience.canShow);
```
ConradJChan, line 149: iirc, annotations is the only experience we're passing through right now, but should this be so generic? What happens if there is a new preview thumbnails siderail experience (which doesn't apply to images) would that prevent image discoverability?

src/lib/viewers/image/__tests__/ImageViewer-test.js
```diff
@@ -895,4 +895,26 @@ describe('lib/viewers/image/ImageViewer', () => {
...
+            'should return correct value for isDiscoverabilityEnabled',
+            ({ canShow, enableAnnotationsImageDiscoverability }) => {
+                image.options.experiences = {
+                    tooltipFlowAnnotationsExperience: {
+                        canShow,
+                    },
+                };
+                image.options.enableAnnotationsImageDiscoverability = enableAnnotationsImageDiscoverability;
+
+                expect(image.isDiscoverabilityEnabled()).toBe(!canShow && !!enableAnnotationsImageDiscoverability);
```
ConradJChan, line 916: Instead of computing the expected result and embedding logic in the test, could it be passed in as a parameter?

src/lib/viewers/image/ImageBaseViewer.js
megansmith-box: @ConradJChan - That is a good point. I think it's most likely if any experience can show we would want to disable discoverability. I also think it would be more likely to forget to add than remember to add it, if we created some sort of eligible experiences list. I think it would make sense to address this, when the time comes, but open to make the suggested change. Let me know what you think!

jstoffan (approved):

src/lib/viewers/image/ImageBaseViewer.js
megansmith-box: I'll go ahead and merge this PR and we can make updates in a separate PR if needed!

PR #1377, opened 2021-05-12
feat(controls): Add Settings menu to Model3D

ConradJChan:
TODO:
- [x] Unit tests
- [x] Cross-browser testing

Before:
Screen Shot 2021-05-12 at 3 21 54 PM

After:
Screen Shot 2021-05-12 at 3 21 42 PM

### DIFF
--- a/src/lib/viewers/box3d/model3d/Model3DViewer.js
+++ b/src/lib/viewers/box3d/model3d/Model3DViewer.js
@@ -46,17 +46,17 @@ class Model3DViewer extends Box3DViewer {
     /** @property {boolean} - Boolean indicating whether the animation is playing */
     isAnimationPlaying = false;
 
-    /** @property {boolean} - Boolean indicating whether the grid is showing */
-    renderGrid = DEFAULT_RENDER_GRID;
-
     /** @property {string} - string indicating what the render mode is */
     renderMode = RENDER_MODE_LIT;
 
+    /** @property {boolean} - Boolean indicating whether the grid is showing */
+    showGrid = DEFAULT_RENDER_GRID;
+
     /** @property {boolean} - Boolean indicating whether the skeletons are showing */
-    renderSkeletons = false;
+    showSkeletons = false;
 
     /** @property {boolean} - Boolean indicating whether the wireframes are showing */
-    renderWireframes = false;
+    showWireframes = false;
 
     /** @inheritdoc */
     constructor(option) {
@@ -216,11 +216,11 @@ class Model3DViewer extends Box3DViewer {
         this.renderMode = defaults.defaultRenderMode || RENDER_MODE_LIT;
         this.projection = defaults.cameraProjection || CAMERA_PROJECTION_PERSPECTIVE;
         if (defaults.renderGrid === 'true') {
-            this.renderGrid = true;
+            this.showGrid = true;
         } else if (defaults.renderGrid === 'false') {
-            this.renderGrid = false;
+            this.showGrid = false;
         } else {
-            this.renderGrid = DEFAULT_RENDER_GRID;
+            this.showGrid = DEFAULT_RENDER_GRID;
         }
 
         if (this.axes.up !== DEFAULT_AXIS_UP || this.axes.forward !== DEFAULT_AXIS_FORWARD) {
@@ -346,7 +346,10 @@ class Model3DViewer extends Box3DViewer {
     handleReset() {
         super.handleReset();
 
-        this.isAnimationPlaying = false;
+        this.setAnimationState(false);
+        this.handleShowGrid(true);
+        this.handleShowSkeletons(false);
+        this.handleShowWireframes(false);
 
         if (this.controls) {
             if (this.getViewerOption('useReactControls')) {
@@ -356,7 +359,7 @@ class Model3DViewer extends Box3DViewer {
                 this.controls.setCurrentProjectionMode(this.projection);
                 this.controls.handleSetSkeletonsVisible(false);
                 this.controls.handleSetWireframesVisible(false);
-                this.controls.handleSetGridVisible(this.renderGrid);
+                this.controls.handleSetGridVisible(this.showGrid);
             }
         }
 
@@ -422,7 +425,7 @@ class Model3DViewer extends Box3DViewer {
         this.renderer.setSkeletonsVisible(visible);
 
         if (this.controls && this.getViewerOption('useReactControls')) {
-            this.renderSkeletons = visible;
+            this.showSkeletons = visible;
             this.renderUI();
         }
     }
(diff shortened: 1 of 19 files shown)


src/lib/viewers/controls/settings/Settings.tsx
```diff
@@ -15,11 +16,15 @@ import { decodeKeydown } from '../../../util';
 export type Props = React.PropsWithChildren<{
     className?: string;
     toggle?: React.ElementType;
+    onClose?: () => void;
```
jstoffan, line 19: Can we alphabetize these props?

src/lib/viewers/controls/model3d/Model3DSettings.scss
```diff
@@ -0,0 +1,24 @@
...
+        }
+    }
+
+    .bp-RotateAxisControls {
+        margin-bottom: 10px;
+    }
+}
+
+.bp-Model3DSettings-menu {
+    & > * {
```
jstoffan, line 21: We usually try to avoid `*` selectors. What's the reasoning here?

src/lib/viewers/box3d/model3d/Model3DViewer.js
```diff
@@ -43,9 +43,21 @@ class Model3DViewer extends Box3DViewer {
     /** @property {Object[]} - List of Box3D instances added to the scene */
     instances = [];
 
-    /** @property {boolean} - Boolean indicating whether the animation is playihng */
+    /** @property {boolean} - Boolean indicating whether the animation is playing */
     isAnimationPlaying = false;
 
+    /** @property {boolean} - Boolean indicating whether the grid is showing */
+    renderGrid = DEFAULT_RENDER_GRID;
```
jstoffan, line 50: The term `render` is usually used as a verb, which makes these look like methods at first. How about something like `currentGrid`?

jstoffan (commented):
LGTM overall, nice job.

src/lib/viewers/box3d/model3d/Model3DViewer.js
```diff
@@ -334,7 +349,10 @@ class Model3DViewer extends Box3DViewer {
     handleReset() {
         super.handleReset();
 
-        this.isAnimationPlaying = false;
+        this.setAnimationState(false);
+        this.handleShowGrid(true);
```
jstoffan, line 353: Does this interfere with the non-React controls at all?

src/lib/viewers/controls/model3d/Model3DSettings.scss
```diff
@@ -0,0 +1,25 @@
...
+        }
+    }
+
+    .bp-RotateAxisControls {
+        margin-bottom: 10px;
+    }
+}
+
+.bp-Model3DSettings-menu {
+    & .bp-SettingsCheckboxItem,
```
jstoffan, line 21: I think we should be able to omit the `&` here.

src/lib/viewers/controls/model3d/Model3DSettings.tsx
```diff
@@ -0,0 +1,96 @@
+import React from 'react';
+import Settings, { Menu, Props as SettingsProps } from '../settings';
```
jstoffan, line 2: Can we alphabetize the imports?

src/lib/viewers/controls/model3d/RotateAxisControls.scss
```diff
@@ -0,0 +1,36 @@
+@import '../styles';
+
+$bp-RotateAxisControls-spacing: 4px;
+
+.bp-RotateAxisControls {
+    display: flex;
+    flex-direction: column;
+    color: $bdl-gray-62;
+
+    .bp-RotateAxisControl + .bp-RotateAxisControl {
```
jstoffan, line 10: Does this need to be nested?

src/lib/viewers/controls/settings/Settings.tsx
```diff
@@ -34,14 +39,22 @@ export default function Settings({
         setActiveRect(undefined);
         setIsFocused(false);
         setIsOpen(false);
-    }, []);
+
+        onClose();
+    }, [onClose]);
```
jstoffan, line 44: It seems like any call to Model3dViewer's `renderUI` would reset the controls, since the handler is declared as an inline function. Is that not the case? Would it be safer to create a `useCallback` helper that can both call `setIsOpen` and `onClose` or `onOpen`?

src/lib/viewers/controls/settings/SettingsDropdown.tsx
```diff
@@ -7,20 +7,26 @@ import useClickOutside from '../hooks/useClickOutside';
...
 
-export type ListItem = {
+export type ListItem<V> = {
     label: string;
-    value: string;
+    value: V;
 };
 
-export type Props = {
+export type Props<V> = {
```
jstoffan, line 15: I think we can use `export type Props<V = string>` here to avoid having to specify types in our tests. Further, should we have an `extends` here to only allow certain types, such as boolean/number/string?

src/lib/viewers/box3d/model3d/Model3DViewer.js
ConradJChan: It shouldn't because it's essentially the same as what's being done below on L361-365. But now that you mention it, this can probably be simplified

src/lib/viewers/controls/settings/Settings.tsx
ConradJChan: No, that's not the case. Seems like any call to `renderUI` should probably create a new callback for `resetControls` but not necessarily call it

ConradJChan: Are you proposing something like:
```
const setIsOpenHelper = React.useCallback(() => {
    const nextIsOpen = !isOpen;
    setIsOpen(nextIsOpen);

    if (nextIsOpen) {
        onOpen();
    } else {
        onClose();
    }
}, [onClose, onOpen]);
```
And use that in place of `setIsOpen`?

src/lib/viewers/controls/settings/SettingsDropdown.scss
```diff
@@ -11,7 +11,7 @@ $bp-SettingsDropdown-spacing: 5px;
 }
 
 .bp-SettingsDropdown-flyout {
-    top: initial;
+    top: auto;
```
jstoffan, line 14: Does this affect the media controls at all?

jstoffan (approved):

src/lib/viewers/controls/settings/SettingsDropdown.scss
ConradJChan: No this is an override only for `SettingsDropdown`

src/lib/viewers/box3d/model3d/Model3DViewer.js
ConradJChan: Tried to remove the controls calls but there's some hidden magic to set the defaults so I'll wait til we migrate over fully and just delete it then

jstoffan (approved):

PR #1381, opened 2021-05-14
fix(controls): Add forwardRef to Dropdown and Checkbox

ConradJChan:
Fixes an issue where a console error occurs when opening the `Model3DSettings` menu because `SettingsList` attempts to set focus to the menu item that correlates to the `activeIndex`. Since `SettingsDropdown` and `SettingsCheckboxItem` are functional components, an error is logged.

TODO:
- [x] cross browser testing

### DIFF
--- a/src/lib/types/react-augment.d.ts
+++ b/src/lib/types/react-augment.d.ts
@@ -1,0 +1,7 @@
+import React from 'react';
+
+declare module 'react' {
+    function forwardRef<T, P = {}>(
+        render: (props: P, ref: React.ForwardedRef<T>) => React.ReactElement | null,
+    ): (props: P & RefAttributes<T>) => React.ReactElement | null;
+}
--- a/src/lib/viewers/controls/settings/SettingsDropdown.scss
+++ b/src/lib/viewers/controls/settings/SettingsDropdown.scss
@@ -2,13 +2,18 @@
 
 $bp-SettingsDropdown-spacing: 5px;
 
-.bp-SettingsDropdown,
-.bp-SettingsDropdown-container {
-    position: relative;
+.bp-SettingsDropdown {
     display: flex;
     flex: 1 1 auto;
     flex-direction: column;
     color: $bdl-gray-62;
+}
+
+.bp-SettingsDropdown-toggle {
+    position: relative;
+    display: flex;
+    flex: 1 1 auto;
+    flex-direction: column;
 }
 
 .bp-SettingsDropdown-flyout {
--- a/src/lib/viewers/controls/settings/SettingsDropdown.scss
+++ b/src/lib/viewers/controls/settings/SettingsDropdown.scss
@@ -2,7 +2,8 @@
 
 $bp-SettingsDropdown-spacing: 5px;
 
-.bp-SettingsDropdown {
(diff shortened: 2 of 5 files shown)


src/lib/types/react-augment.d.ts
```diff
@@ -0,0 +1,7 @@
+import React from 'react';
+
+declare module 'react' {
```
ConradJChan, line 3: This is needed because `React.forwardRef` doesn't support Generics typing

src/lib/viewers/controls/settings/SettingsDropdown.tsx
```diff
@@ -74,52 +71,69 @@ export default function SettingsDropdown<V extends Value = string>({
...
+                    buttonElRef.current.focus();
+                }
+            },
+        }),
+        [],
+    );
+
     return (
-        <div ref={dropdownElRef} className={classNames('bp-SettingsDropdown', className)}>
+        <div className={classNames('bp-SettingsDropdown', className)}>
```
ConradJChan, line 88: Moved the ref to a separate container on L92 so that clicking on the label is a valid way to close the dropdown

src/lib/viewers/controls/settings/SettingsDropdown.tsx
```diff
@@ -74,52 +71,69 @@ export default function SettingsDropdown<V extends Value = string>({
...
-            <SettingsFlyout className="bp-SettingsDropdown-flyout" isOpen={isOpen}>
-                <SettingsList
-                    ref={listElRef}
-                    aria-labelledby={`${id}-label`}
-                    className="bp-SettingsDropdown-list"
-                    isActive
-                    onKeyDown={handleKeyDown}
-                    role="listbox"
-                    tabIndex={-1}
+            <div ref={dropdownElRef} className="bp-SettingsDropdown-container">
```
jstoffan, line 92: We should try to avoid the use of `container`, since it can be confusing in SUIT naming. How about `bp-SettingsDropdown-toggle`?

src/lib/viewers/controls/settings/SettingsDropdown.scss
```diff
@@ -2,7 +2,8 @@
 
 $bp-SettingsDropdown-spacing: 5px;
 
-.bp-SettingsDropdown {
+.bp-SettingsDropdown,
+.bp-SettingsDropdown-container {
```
jstoffan, line 6: Should these have the same styles? Should they be inherited?

src/lib/viewers/controls/settings/SettingsDropdown.tsx
```diff
@@ -74,52 +71,69 @@ export default function SettingsDropdown<V extends Value = string>({
 
     useClickOutside(dropdownElRef, () => setIsOpen(false));
 
+    // Customize the forwarded ref to combine usage with the ref internal to this component
+    React.useImperativeHandle(
```
jstoffan, line 75: It seems unorthodox to expose the button as the component's ref, since it's a compound component. If we want to expose the dropdown itself later, that could get confusing.

jstoffan (approved):

PR #1385, opened 2021-05-17
feat(annotations tooltip): Add class for specific styles to element

megansmith-box:


src/lib/viewers/controls/annotations/AnnotationsTargetedTooltip.tsx
```diff
@@ -18,6 +18,7 @@ function AnnotationsTargetedTooltip({ children, isEnabled = false }: Props): JSX
             className="bp-AnnotationsTooltip"
             shouldTarget={isEnabled}
             showCloseButton
+            tetherElementClassName="bp-AnnotationsTooltip-element"
```
jstoffan, line 21: This class name is a little confusing because this element appears "above" `bp-AnnotationsTooltip` in the DOM and hyphens are used to indicate descendancy in SUIT CSS naming conventions. Could we use `bp-AnnotationsTooltipTether`?

src/lib/viewers/controls/annotations/AnnotationsTargetedTooltip.scss
```diff
@@ -12,3 +12,8 @@
     font-size: 14px;
     line-height: 20px;
 }
+
+// overwrite z-index
+.bp-AnnotationsTooltip-element.tooltip-element {
+    z-index: 160;
```
jstoffan, line 18: It seems like this value is very specific to the context where Preview is being used. Can we move the override to that repo?

jstoffan (approved):

PR #1387, opened 2021-05-21
feat(annotations tooltip): Close on click outside to avoid UI issues

megansmith-box:


src/lib/viewers/controls/annotations/AnnotationsTargetedTooltip.tsx
```diff
@@ -16,9 +16,9 @@ function AnnotationsTargetedTooltip({ children, isEnabled = false }: Props): JSX
     return (
         <TargetedClickThroughTooltip
             className="bp-AnnotationsTooltip"
+            closeOnClickOutside
             shouldTarget={isEnabled}
             showCloseButton
-            tetherElementClassName="bp-AnnotationsTooltipTether"
```
jstoffan, line 21: Just to confirm: Is this change intentional? Is the increased z-index no longer needed?

jstoffan (approved):

src/lib/viewers/controls/annotations/AnnotationsTargetedTooltip.tsx
megansmith-box, line 21: we no longer need this as the tooltip will close if the preview share modal is clicked

megansmith-box: @jstoffan - Correct, we no longer need the z-index, because if we click on the share button which opens the modal the tooltip will now close.

PR #1401, opened 2021-06-15
perf(presentation): Improve presentation rendering performance in Safari

jstoffan:
The Presentation viewer creates all pages of a document as a big stack of elements using absolute positioning. It controls the page elements' visibility via the `visibility` style directive. However, unlike other browsers, Safari 14 on MacOS Big Sur seems to frequently layout/paint every page of the document, even if the page has `visibility: hidden` set. Adding `opacity: 0` seems to resolve the problem.

### DIFF
--- a/src/lib/viewers/doc/_docBase.scss
+++ b/src/lib/viewers/doc/_docBase.scss
@@ -84,7 +84,6 @@ $thumbnail-sidebar-width: 191px; // Extra pixel to account for sidebar border
         background-color: $d-eight;
         border-radius: $thumbnail-border-radius;
         box-shadow: 0 0 0 1px $off-white, 0 1px 4px 0 rgba(0, 0, 0, .08);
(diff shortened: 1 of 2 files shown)


src/lib/viewers/doc/_docBase.scss
```diff
@@ -84,7 +84,6 @@ $thumbnail-sidebar-width: 191px; // Extra pixel to account for sidebar border
         background-color: $d-eight;
         border-radius: $thumbnail-border-radius;
         box-shadow: 0 0 0 1px $off-white, 0 1px 4px 0 rgba(0, 0, 0, .08);
-        transition: box-shadow 50ms ease;
```
jstoffan, line 87: I noticed some jank when mousing quickly over different thumbnails, so I removed this brief transition. It's perf-related, but I can move it to a separate PR if desired.

boxluecke (approved):

PR #1402, opened 2021-06-23
fix(controls): color correction on contols

hormaza:
Focus indicators have contrast issues:
image

This commit will increase the background alpha on the toolbar at 90% and increase the brightness of the focus indicators at #D3D3D3 (box-gray-20)


src/lib/viewers/controls/_styles.scss
```diff
@@ -2,8 +2,8 @@
 @import '~box-ui-elements/es/styles/deprecated';
 
 $bp-controls-background: fade-out($black, .2);
-$bp-controls-opacity: .8;
-$bp-controls-outline: inset 0 0 0 1px fade-out($white, .5), 0 1px 2px fade-out($black, .9);
+$bp-controls-opacity: .9;
+$bp-controls-outline: inset 0 0 0 1px fade-out($bdl-gray-20, 0), 0 1px 2px fade-out($black, .9);
```
jstoffan, line 6: We should be able to remove the `fade-out` directive here, since we're passing a zero value to it.

jstoffan:
@hormaza, can you please include both a "before" and "after" screenshot for all visual changes to preview? Thanks!

src/lib/viewers/controls/_styles.scss
hormaza: Thanks Jared. Shouldn't be zero. Updated!

hormaza:
Before:
image
After:
image

ConradJChan (approved):

PR #1410, opened 2021-07-14
feat(slider): Migrate to custom slider due to range input limitations

jstoffan:


src/lib/viewers/controls/slider/SliderControl.tsx
```diff
@@ -1,42 +1,151 @@
...
+    }, []);
+
+    const getPositionValue = React.useCallback(
+        (pageX: number) => {
+            const { current: sliderEl } = sliderElRef;
+
+            if (!sliderEl) return 0;
+
+            const { width: sliderWidth } = sliderEl.getBoundingClientRect();
+            const newValue = (getPosition(pageX) / sliderWidth) * max;
```
ConradJChan, line 52: Do we need to worry about decimal values here?

src/lib/viewers/controls/slider/SliderControl.tsx
```diff
@@ -1,42 +1,151 @@
...
+        const { current: sliderEl } = sliderElRef;
+
+        if (!sliderEl) return 0;
+
+        const { left: sliderLeft, width: sliderWidth } = sliderEl.getBoundingClientRect();
+        return Math.max(0, Math.min(pageX - sliderLeft, sliderWidth));
+    }, []);
+
+    const getPositionValue = React.useCallback(
+        (pageX: number) => {
```
ConradJChan, line 46: Does tslint not care about the return types for these anonymous functions used in `useCallback`?

src/lib/viewers/controls/slider/SliderControl.tsx
```diff
@@ -1,42 +1,151 @@
...
+
+        onMove(getPositionValue(pageX), getPosition(pageX), sliderWidth);
+    };
+
+    const handleTouchStart = ({ touches }: React.TouchEvent<Ref>): void => {
+        onUpdate(getPositionValue(touches[0].pageX));
+        setIsScrubbing(true);
+    };
+
+    React.useEffect(() => {
```
ConradJChan, line 91: Are there any issues with this when doing a quick click and drag and release interaction with the slider?

ConradJChan (approved):

src/lib/viewers/controls/slider/SliderControl.tsx
jstoffan: I think we'd want to let the consuming component handle rounding to avoid making assumptions about its use. Videos can have durations defined as decimal values, for whatever reason, for example.

src/lib/viewers/controls/slider/SliderControl.tsx
jstoffan: It doesn't. It looks like the return value is being inferred correctly in this case, though, so maybe that's why?

src/lib/viewers/controls/slider/SliderControl.tsx
jstoffan: I haven't seen any issues in my testing, but will keep an eye out as we build more features with this component.

PR #1414, opened 2021-07-29
fix(pdfjs): Use allowlist for exporting font data

ConradJChan:
Applying the fix found here: https://github.com/mozilla/pdf.js/issues/13771

Fixes the issue where PDF files using Courier font (could possibly apply to others as well) were copying properties that were incompatible with cloning between the pdf.js in the main thread and pdf.worker.js in the WebWorker by only copying properties found in an allow list

TODO:
- [x] cross-browser testing


src/lib/constants.js
```diff
@@ -100,7 +100,7 @@ export const PDFJS_MAX_AUTO_SCALE = 1.25; // Should match MAX_AUTO_SCALE in pdf_
 export const PDFJS_WIDTH_PADDING_PX = 40; // Should match SCROLLBAR_PADDING in pdf_viewer.js
 export const PDFJS_HEIGHT_PADDING_PX = 5; // Should match VERTICAL_PADDING in pdf_viewer.js
 
-export const DOC_STATIC_ASSETS_VERSION = '2.16.0'; // v2.53.0 has font rendering issues in large documents in MS Edge
+export const DOC_STATIC_ASSETS_VERSION = '2.76.0'; // v2.53.0 has font rendering issues in large documents in MS Edge
```
jstoffan, line 103: Can we probably remove the comment now.

src/third-party/doc/2.76.0/pdf.worker.js
```diff
@@ -35682,11 +35724,13 @@ var Font = function FontClosure() {
     },
 
     exportData: function Font_exportData() {
-      var data = {};
-
-      for (var i in this) {
-        if (this.hasOwnProperty(i)) {
-          data[i] = this[i];
+      const data = Object.create(null);
```
jstoffan, line 35727: Should we continue to use `var` here to stay consistent with the existing code?

jstoffan (approved):

jstoffan (approved):

src/third-party/doc/2.76.0/pdf.worker.js
```diff
@@ -35682,13 +35724,15 @@ var Font = function FontClosure() {
     },
 
     exportData: function Font_exportData() {
-      var data = {};
-
-      for (var i in this) {
-        if (this.hasOwnProperty(i)) {
-          data[i] = this[i];
+      var data = Object.create(null);
+      var that = this, value;
```
jstoffan, line 35728: Super minor, but in production code we would want either one `var` per variable or to use one `var` with a comma separated list of variables. We have both formats here.

src/third-party/doc/2.76.0/pdf.worker.js
```diff
@@ -35682,13 +35724,15 @@ var Font = function FontClosure() {
...
+      EXPORT_DATA_PROPERTIES.forEach(function (property) {
+        value = that[property];
```
jstoffan, line 35730: Does `value` need to be a closure or can it be used inline here?

jstoffan (approved):
Do we need to update our `upgrade_pdfjs` script with this change?

jstoffan (approved):

PR #1417, opened 2021-08-05
fix(ThumbnailsSidebar): A11Y - add aria-current to ThumbnailsSidebar

fsuarez-globant:
There is a visual indication of the current thumbnail that also presents the current page You should convey this information to screen reader with aria-current="true". For example:

<div class="bp-thumbnail bp-thumbnail-is-selected bp-thumbnail-image-loaded" data-bp-page-num="7" role="button" aria-current="true">

Screen Shot 2021-08-05 at 2 56 06 PM

### DIFF
--- a/src/lib/ThumbnailsSidebar.js
+++ b/src/lib/ThumbnailsSidebar.js
@@ -440,8 +440,10 @@ class ThumbnailsSidebar {
             const parsedPageNum = parseInt(thumbnailEl.dataset.bpPageNum, 10);
             if (parsedPageNum === this.currentPage) {
                 thumbnailEl.classList.add(CLASS_BOX_PREVIEW_THUMBNAIL_IS_SELECTED);
+                thumbnailEl.setAttribute('aria-current', true);
             } else {
                 thumbnailEl.classList.remove(CLASS_BOX_PREVIEW_THUMBNAIL_IS_SELECTED);
+                thumbnailEl.removeAttribute('aria-current');
             }
         });
     }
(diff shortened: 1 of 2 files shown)


src/lib/ThumbnailsSidebar.js
```diff
@@ -254,6 +254,7 @@ class ThumbnailsSidebar {
 
         if (pageNum === this.currentPage) {
             thumbnailEl.classList.add(CLASS_BOX_PREVIEW_THUMBNAIL_IS_SELECTED);
+            thumbnailEl.setAttribute('aria-current', true);
```
jstoffan, line 257: Does this value get updated as the user changes between pages?

fsuarez-globant: @jstoffan you are right. I added and removed the aria-current when user changes between pages.

jstoffan (commented):
Can we add tests, please?

fsuarez-globant:
@jstoffan I just added an unit test

src/lib/__tests__/ThumbnailsSidebar-test.js
```diff
@@ -475,6 +477,14 @@ describe('ThumbnailsSidebar', () => {
             expect(stubs.removeClass).toBeCalledTimes(2);
             expect(stubs.addClass).toBeCalledTimes(1);
         });
+
+        test('should add the aria-current attribute when thumbnail is selected', () => {
+            thumbnailsSidebar.currentPage = 2;
+
+            thumbnailsSidebar.applyCurrentPageSelection();
+
+            expect(stubs.addAriaAttribute).toBeCalledTimes(1);
```
jstoffan, line 486: Can we add an assertion for the `removeAttribute` call, as well, similar to the tests above?

jstoffan (approved):

src/lib/__tests__/ThumbnailsSidebar-test.js
fsuarez-globant: @jstoffan I just added another unit test case.

PR #1421, opened 2021-08-30
fix(ThumbnailSidebar): A11Y add descriptive attributes

tille:
Adds a couple aria-label attributes for screen-reader users

### DIFF
--- a/src/lib/viewers/doc/DocBaseViewer.js
+++ b/src/lib/viewers/doc/DocBaseViewer.js
@@ -163,9 +163,10 @@ class DocBaseViewer extends BaseViewer {
         this.startPageNum = this.getStartPage(this.startAt);
 
         if (this.options.enableThumbnailsSidebar) {
-            this.thumbnailsSidebarEl = document.createElement('div');
+            this.thumbnailsSidebarEl = document.createElement('nav');
             this.thumbnailsSidebarEl.className = `${CLASS_BOX_PREVIEW_THUMBNAILS_CONTAINER}`;
             this.thumbnailsSidebarEl.setAttribute('data-testid', 'thumbnails-sidebar');
+            this.thumbnailsSidebarEl.setAttribute('aria-label', __('thumbnail_label'));
             this.thumbnailsSidebarEl.tabIndex = 0;
             this.rootEl.insertBefore(this.thumbnailsSidebarEl, this.containerEl);
         }
--- a/src/i18n/en-US.properties
+++ b/src/i18n/en-US.properties
@@ -58,6 +58,10 @@ find_previous=Find previous
 find_close=Close
 # Search input Aria-label attribute
 find_label=Search input
+# Sidebar thumbnails aria-label attribute
+thumbnail_label=Page thumbnails
+# Document preview aria-label attribute
+file_preview_label=File preview
 
 # Error messages
 # Default preview error message
--- a/src/lib/shell.html
+++ b/src/lib/shell.html
@@ -59,7 +59,7 @@
         </div>
     </div>
     <div class="bp" data-testid="bp" tabindex="-1">
-        <div class="bp-content" data-testid="bp-content" tabindex="-1">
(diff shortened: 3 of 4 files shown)


src/lib/viewers/doc/DocBaseViewer.js
```diff
@@ -163,13 +163,19 @@ class DocBaseViewer extends BaseViewer {
...
+            this.thumbnailsSidebarEl = document.createElement('nav');
             this.thumbnailsSidebarEl.className = `${CLASS_BOX_PREVIEW_THUMBNAILS_CONTAINER}`;
             this.thumbnailsSidebarEl.setAttribute('data-testid', 'thumbnails-sidebar');
+            this.thumbnailsSidebarEl.setAttribute('aria-label', __('thumbnail_label'));
             this.thumbnailsSidebarEl.tabIndex = 0;
             this.rootEl.insertBefore(this.thumbnailsSidebarEl, this.containerEl);
         }
 
+        if (this.containerEl) {
+            this.containerEl.setAttribute('aria-label', __('document_preview_label'));
```
ConradJChan, line 175: Should this only be done for document previews? Should all file previews set the `aria-label` and `role` on `this.containerEl`?

tille: good question, yes we can set those attributes for all-files but changing the message from 'document preview' to something else ex. "file preview"

tille: makes sense?

ConradJChan: Yes, the type of file being previewed is known by the specific viewers. Does this label get read by the screen reader? 

Could we possibly put these `role` and `aria-label` directly in the [shell.html](https://github.com/box/box-content-preview/blob/master/src/lib/shell.html#L62) instead of adding it programmatically?

jstoffan: I agree it would be preferable to keep the message generic, such as "file preview," and put it in `shell.html` as suggested above.

jstoffan: @ConradJChan, it looks like the `aria-label` value needs to be injected via JS, since it's a translated string.

src/lib/viewers/doc/DocBaseViewer.js
```diff
@@ -163,13 +163,19 @@ class DocBaseViewer extends BaseViewer {
...
+            this.thumbnailsSidebarEl = document.createElement('nav');
             this.thumbnailsSidebarEl.className = `${CLASS_BOX_PREVIEW_THUMBNAILS_CONTAINER}`;
             this.thumbnailsSidebarEl.setAttribute('data-testid', 'thumbnails-sidebar');
+            this.thumbnailsSidebarEl.setAttribute('aria-label', __('thumbnail_label'));
             this.thumbnailsSidebarEl.tabIndex = 0;
             this.rootEl.insertBefore(this.thumbnailsSidebarEl, this.containerEl);
         }
 
+        if (this.containerEl) {
+            this.containerEl.setAttribute('aria-label', __('file_preview_label'));
```
jstoffan, line 175: Should this be added in `BaseViewer` so it is available for all file types?

src/lib/viewers/doc/DocBaseViewer.js
```diff
@@ -163,13 +163,19 @@ class DocBaseViewer extends BaseViewer {
...
+            this.containerEl.setAttribute('role', 'region');
```
jstoffan, line 176: This can be added in `shell.html`.

jstoffan (approved):

jstoffan:
@tille, it looks like you may need to squash, rebase, and force push to get past the commit message linter.

ConradJChan (dismissed):

tille:
@jstoffan squash, done

jstoffan (approved):

PR #1427, opened 2021-09-15
build(env): Upgrade node to v14 and yarn to v1.22

jstoffan:


ConradJChan (approved):
Do we need to update `package.json`?

jstoffan:
We've run into issues with updating `engines` in the past, so I figured I'd leave it be. It should still work with Node 10 in theory. Thoughts?

PR #1428, opened 2021-09-16
Color contrast correction on control focus

hormaza:
image

Whe  the findBar is closed the focus should go back to the find toggle button located on toolbar.

image


src/lib/viewers/doc/DocBaseViewer.js
```diff
@@ -410,8 +410,9 @@ class DocBaseViewer extends BaseViewer {
     }
 
     handleFindBarClose() {
-        if (this.docEl) {
-            this.docEl.focus(); // Prevent focus from transferring to the root document element
+        const el = document.querySelector('.bp-FindBarToggle'); // Transferring focus to the findBarToggle Button
```
jstoffan, line 413: We generally want to avoid reaching into other components via the DOM. We also want to scope our selectors to the `this.docEl` element, since multiple Preview instances can appear on the same page.

hormaza: Understood. Is there an interface get the element with the class mentioned in code?

jstoffan: No, not currently. It will be a little tricky to add because the button is rendered by React and its reference isn't (and shouldn't be) exposed for programmatic use.

I suggest updating DocBaseViewer's `toggleFindBar` method to accept and store a reference to the "originator" element, which can be passed from DocControls. Then, when the `findBar` instance is closed, we can set focus back to the stored "originator" element. Does that make sense?

hormaza: Hi Jared. I realized that commits was on a wrong branch, so I did a new pull request.
This is the new PR:
https://github.com/box/box-content-preview/pull/1429

PR #1435, opened 2021-11-03
fix(preview): Close color palette when tabbed out of the palette focus

JChan106:
Fixes the bug where focusing out of the color palette using keyboard navigation didn't close the palette. The palette now closes when a user tabs out of the color palette, or if the user is tabbed in the palette and the user clicks out of the palette.

### DIFF
--- a/src/lib/viewers/controls/color-picker/ColorPickerControl.tsx
+++ b/src/lib/viewers/controls/color-picker/ColorPickerControl.tsx
@@ -19,15 +19,23 @@ export default function ColorPickerControl({
 }: Props): JSX.Element | null {
     const paletteRef = React.useRef<HTMLDivElement>(null);
     const [isColorPickerToggled, setIsColorPickerToggled] = React.useState(false);
-    const [isPaletteActive, handlers] = useAttention();
+    const { current: paletteEl } = paletteRef;
+
+    const handlePaletteBlur = ({ relatedTarget }: React.FocusEvent<HTMLButtonElement>): void => {
+        const targetElement = relatedTarget as HTMLButtonElement;
+        if (targetElement.parentElement && targetElement.parentElement.parentElement !== paletteEl) {
+            setIsColorPickerToggled(false);
+        }
+    };
+
+    const [isPaletteActive, handlers] = useAttention({ onBlur: handlePaletteBlur });
 
     const handleSelect = (color: string): void => {
         setIsColorPickerToggled(false);
         onColorSelect(color);
     };
 
-    const handleBlur = ({ relatedTarget }: React.FocusEvent<HTMLButtonElement>): void => {
-        const { current: paletteEl } = paletteRef;
+    const handleToggleBlur = ({ relatedTarget }: React.FocusEvent<HTMLButtonElement>): void => {
         // IE11 does not have relatedTarget but update activeElement before blur
         const nextTarget = relatedTarget || document.activeElement;
 
@@ -55,7 +63,7 @@ export default function ColorPickerControl({
             <button
                 className={classNames('bp-ColorPickerControl-toggle', { 'bp-is-active': isColorPickerToggled })}
                 data-testid="bp-ColorPickerControl-toggle"
-                onBlur={handleBlur}
+                onBlur={handleToggleBlur}
                 onClick={handleClick}
                 onMouseDown={handleMouseDown}
                 type="button"
--- a/src/lib/viewers/controls/color-picker/ColorPickerControl.tsx
+++ b/src/lib/viewers/controls/color-picker/ColorPickerControl.tsx
@@ -18,35 +18,31 @@ export default function ColorPickerControl({
     ...rest
 }: Props): JSX.Element | null {
     const paletteRef = React.useRef<HTMLDivElement>(null);
+    const toggleRef = React.useRef<HTMLButtonElement>(null);
     const [isColorPickerToggled, setIsColorPickerToggled] = React.useState(false);
+    const [handlers] = useAttention();
     const { current: paletteEl } = paletteRef;
-
-    const handlePaletteBlur = ({ relatedTarget }: React.FocusEvent<HTMLButtonElement>): void => {
-        const targetElement = relatedTarget as HTMLButtonElement;
-        if (targetElement.parentElement && targetElement.parentElement.parentElement !== paletteEl) {
-            setIsColorPickerToggled(false);
-        }
-    };
-
-    const [isPaletteActive, handlers] = useAttention({ onBlur: handlePaletteBlur });
+    const { current: toggleEl } = toggleRef;
 
     const handleSelect = (color: string): void => {
         setIsColorPickerToggled(false);
         onColorSelect(color);
     };
 
-    const handleToggleBlur = ({ relatedTarget }: React.FocusEvent<HTMLButtonElement>): void => {
+    const handleBlur = ({ relatedTarget }: React.FocusEvent<HTMLButtonElement>): void => {
         // IE11 does not have relatedTarget but update activeElement before blur
         const nextTarget = relatedTarget || document.activeElement;
 
-        if (isPaletteActive || (nextTarget && paletteEl && paletteEl.contains(nextTarget as Node))) {
+        if ((nextTarget && paletteEl && paletteEl.contains(nextTarget as Node)) || toggleEl === (nextTarget as Node)) {
             return;
         }
 
         setIsColorPickerToggled(false);
     };
 
-    const handleClick = (): void => setIsColorPickerToggled(!isColorPickerToggled);
+    const handleClick = (): void => {
+        setIsColorPickerToggled(!isColorPickerToggled);
+    };
 
     const handleMouseDown = (event: React.MouseEvent<HTMLButtonElement>): void => {
         if (event.currentTarget.focus) {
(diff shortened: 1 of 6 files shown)


src/lib/viewers/controls/color-picker/ColorPickerControl.tsx
```diff
@@ -19,15 +19,23 @@ export default function ColorPickerControl({
 }: Props): JSX.Element | null {
     const paletteRef = React.useRef<HTMLDivElement>(null);
     const [isColorPickerToggled, setIsColorPickerToggled] = React.useState(false);
-    const [isPaletteActive, handlers] = useAttention();
+    const { current: paletteEl } = paletteRef;
+
+    const handlePaletteBlur = ({ relatedTarget }: React.FocusEvent<HTMLButtonElement>): void => {
+        const targetElement = relatedTarget as HTMLButtonElement;
```
jstoffan, line 25: I don't think IE11 supports `relatedTarget`, so we may need to fall back to `document.activeElement` here as in the `handleToggleBlur` function below.

src/lib/viewers/controls/color-picker/ColorPickerControl.tsx
```diff
@@ -19,15 +19,23 @@ export default function ColorPickerControl({
...
+        if (targetElement.parentElement && targetElement.parentElement.parentElement !== paletteEl) {
```
jstoffan, line 26: The `parentElement` calls here break encapsulation by making some assumptions about the DOM that may eventually change and break unexpectedly. Can we use `!paletteEl.contains(targetElement)`, instead?

src/lib/viewers/controls/color-picker/ColorPickerControl.tsx
```diff
@@ -19,15 +19,23 @@ export default function ColorPickerControl({
...
+            setIsColorPickerToggled(false);
+        }
+    };
+
+    const [isPaletteActive, handlers] = useAttention({ onBlur: handlePaletteBlur });
```
jstoffan, line 31: Can we add `onBlur` to `<ColorPickerPalette>` to avoid having to modify `useAttention` with separate callbacks?

src/lib/viewers/controls/color-picker/ColorPickerControl.tsx
```diff
@@ -19,15 +19,23 @@ export default function ColorPickerControl({
...
 
     const handleSelect = (color: string): void => {
         setIsColorPickerToggled(false);
         onColorSelect(color);
     };
 
-    const handleBlur = ({ relatedTarget }: React.FocusEvent<HTMLButtonElement>): void => {
-        const { current: paletteEl } = paletteRef;
+    const handleToggleBlur = ({ relatedTarget }: React.FocusEvent<HTMLButtonElement>): void => {
```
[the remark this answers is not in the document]
JChan106: Yeah just realized I can use the same handler if I take out the `isPaletteActive` conditional. I didn't see any focus issues in taking that conditional out.

src/lib/viewers/controls/color-picker/ColorPickerControl.tsx
```diff
@@ -18,27 +18,31 @@ export default function ColorPickerControl({
     ...rest
 }: Props): JSX.Element | null {
     const paletteRef = React.useRef<HTMLDivElement>(null);
+    const toggleRef = React.useRef<HTMLButtonElement>(null);
     const [isColorPickerToggled, setIsColorPickerToggled] = React.useState(false);
-    const [isPaletteActive, handlers] = useAttention();
+    const [handlers] = useAttention();
+    const { current: paletteEl } = paletteRef;
```
jstoffan, line 24: Can we revert this change? We typically destructure these inside the event handler where they're used, since the value could theoretically change before they're invoked.

src/lib/viewers/controls/color-picker/ColorPickerControl.tsx
```diff
@@ -18,27 +18,31 @@ export default function ColorPickerControl({
...
         onColorSelect(color);
     };
 
     const handleBlur = ({ relatedTarget }: React.FocusEvent<HTMLButtonElement>): void => {
-        const { current: paletteEl } = paletteRef;
         // IE11 does not have relatedTarget but update activeElement before blur
         const nextTarget = relatedTarget || document.activeElement;
 
-        if (isPaletteActive || (nextTarget && paletteEl && paletteEl.contains(nextTarget as Node))) {
+        if ((nextTarget && paletteEl && paletteEl.contains(nextTarget as Node)) || toggleEl === (nextTarget as Node)) {
```
jstoffan, line 36: Can we make this a bit more readable by using some additional variables?

```
const { current: paletteEl } = paletteRef;
const { current: toggleEl } = toggleRef;

// IE11 does not have relatedTarget but update activeElement before blur
const nextTarget = (relatedTarget || document.activeElement) as Node;
const isNextInPalette = paletteEl && paletteEl.contains(nextTarget);
const isNextToggle = toggleEl && toggleEl === nextTarget;

if (isNextInPalette || isNextToggle) {
    return;
}

setIsColorPickerToggled(false);

File: src/lib/viewers/controls/color-picker/ColorPickerControl.tsx [hunk 7]
```diff
@@ -18,27 +18,31 @@ export default function ColorPickerControl({
...
             return;
         }
 
         setIsColorPickerToggled(false);
     };
 
-    const handleClick = (): void => setIsColorPickerToggled(!isColorPickerToggled);
+    const handleClick = (): void => {
```
jstoffan, line 43: Is this change needed or can we revert it?

src/lib/viewers/controls/color-picker/ColorPickerPalette.tsx
```diff
@@ -3,10 +3,11 @@ import './ColorPickerPalette.scss';
 
 export type Props = {
     colors: Array<string>;
+    onBlur: React.FocusEventHandler<HTMLButtonElement>;
     onSelect: (color: string) => void;
 };
 
-export default function ColorPickerPalette({ colors, onSelect }: Props): JSX.Element {
+export default function ColorPickerPalette({ colors, onSelect, onBlur }: Props): JSX.Element {
```
jstoffan, line 10: We typically alphabetize our props, attributes, etc. when possible.

src/lib/viewers/controls/color-picker/__tests__/ColorPickerControl-test.tsx
```diff
@@ -76,6 +76,21 @@ describe('ColorPickerControl', () => {
             expect(getColorPickerPalette(wrapper).hasClass('bp-is-open')).toBe(true);
         });
 
+        test('should close the palette when focus is outside palette', () => {
+            const wrapper = getWrapper();
+            const toggleButton = getToggleButton(wrapper);
+            const divNode = document.createElement('div');
```
jstoffan, line 82: We would typically refer to this as `divEl` or `divElement`, since we know it's an Element object rather than a generic Node.

src/lib/viewers/controls/color-picker/__tests__/ColorPickerControl-test.tsx
```diff
@@ -76,6 +76,21 @@ describe('ColorPickerControl', () => {
...
+
+            toggleButton.simulate('click');
+            expect(getColorPickerPalette(wrapper).hasClass('bp-is-open')).toBe(true);
+
+            toggleButton.simulate('blur', {
```
jstoffan, line 87: Should we be testing the palette instead of the button? Do we also need a test for when the `nextTarget` after blur is the toggle button?

src/lib/viewers/controls/color-picker/ColorPickerControl.tsx
```diff
@@ -71,7 +76,12 @@ export default function ColorPickerControl({
                 data-testid="bp-ColorPickerControl-palette"
                 {...handlers}
             >
-                <ColorPickerPalette colors={colors} data-testid="bp-ColorPickerPalette" onSelect={handleSelect} />
+                <ColorPickerPalette
+                    colors={colors}
+                    data-testid="bp-ColorPickerPalette"
```
jstoffan, line 81: This isn't related to your change, but does this get applied to the palette? It's not a listed prop and we don't have a `...rest` operator anywhere. Can we remove it safely?

src/lib/viewers/controls/color-picker/ColorPickerControl.tsx
```diff
@@ -18,27 +18,31 @@ export default function ColorPickerControl({
     ...rest
 }: Props): JSX.Element | null {
     const paletteRef = React.useRef<HTMLDivElement>(null);
+    const toggleRef = React.useRef<HTMLButtonElement>(null);
     const [isColorPickerToggled, setIsColorPickerToggled] = React.useState(false);
-    const [isPaletteActive, handlers] = useAttention();
+    const [handlers] = useAttention();
```
jstoffan, line 23: Do we still need `useAttention` if we're not using the state it returns?

src/lib/viewers/controls/color-picker/ColorPickerControl.tsx
JChan106: forgot that, thanks

src/lib/viewers/controls/color-picker/ColorPickerControl.tsx
JChan106: Good point, will remove

jstoffan (approved):

PR #1436, opened 2021-11-17
feat(viewer): Add support for disabling .boxdicom file type

JChan106:
Removes preview support for boxdicom files. This is a prerequisite for modifying the default behavior to opening boxdicom files with Ambra.

### DIFF
--- a/src/lib/viewers/iframe/IFrameLoader.js
+++ b/src/lib/viewers/iframe/IFrameLoader.js
@@ -28,10 +28,10 @@ class IFrameLoader extends AssetLoader {
      * @inheritdoc
      */
     determineViewer(file, disabledViewers = [], viewerOptions = {}) {
-        const isDicomFile = file.name === 'Dicom.boxdicom' || file.extension === 'boxdicom';
-        const openWithAmbraEnabled = getProp(viewerOptions, 'IFrame.openWithAmbra');
-        // The IFrame viewer is disabled when the file is a Boxdicom file and Open_with_Ambra FF is enabled
-        if (openWithAmbraEnabled && isDicomFile) {
+        const isDicomFile = file.extension === 'boxdicom';
+        const disableDicom = getProp(viewerOptions, 'IFrame.disableDicom');
+        // The IFrame viewer is disabled when the file is a Boxdicom file and the disableDicom viewer option is enabled
+        if (disableDicom && isDicomFile) {
             disabledViewers.push('IFrame');
         }
 
--- a/src/lib/viewers/iframe/IFrameLoader.js
+++ b/src/lib/viewers/iframe/IFrameLoader.js
@@ -1,3 +1,4 @@
+import getProp from 'lodash/get';
 import AssetLoader from '../AssetLoader';
 import IFrameViewer from './IFrameViewer';
 import { ORIGINAL_REP_NAME } from '../../constants';
@@ -21,6 +22,20 @@ class IFrameLoader extends AssetLoader {
     constructor() {
         super();
         this.viewers = VIEWERS;
+    }
+
+    /**
+     * @inheritdoc
+     */
+    determineViewer(file, disabledViewers = [], viewerOptions = {}) {
+        const isDicomFile = file.name === 'Dicom.boxdicom' || file.extension === 'boxdicom';
+        const openWithAmbraEnabled = getProp(viewerOptions, 'IFrame.openWithAmbra');
+        // The IFrame viewer is disabled when the file is a Boxdicom file and Open_with_Ambra FF is enabled
+        if (openWithAmbraEnabled && isDicomFile) {
+            disabledViewers.push('IFrame');
+        }
+
+        return super.determineViewer(file, disabledViewers);
     }
 }
 
--- a/src/lib/viewers/iframe/__tests__/IFrameLoader-test.js
+++ b/src/lib/viewers/iframe/__tests__/IFrameLoader-test.js
@@ -1,8 +1,6 @@
 import IFrameLoader from '../IFrameLoader';
 import IFrameViewer from '../IFrameViewer';
 
-const sandbox = sinon.createSandbox();
-
 describe('lib/viewers/iframe/IFrameLoader', () => {
     const iframe = new IFrameViewer({
         file: {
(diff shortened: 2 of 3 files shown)


src/lib/viewers/iframe/IFrameLoader.js
```diff
@@ -22,6 +23,20 @@ class IFrameLoader extends AssetLoader {
         super();
         this.viewers = VIEWERS;
     }
+
+    /**
+     * @inheritdoc
+     */
+    determineViewer(file, disabledViewers = [], viewerOptions = {}) {
+        const isDicomFile = file.name === 'Dicom.boxdicom' || file.extension === 'boxdicom';
```
ConradJChan, line 31: The extension check is probably sufficient here

src/lib/viewers/iframe/IFrameLoader.js
```diff
@@ -22,6 +23,20 @@ class IFrameLoader extends AssetLoader {
...
+        const openWithAmbraEnabled = getProp(viewerOptions, 'IFrame.openWithAmbra');
```
ConradJChan, line 32: Instead of openWithAmbra, maybe the option should be `disableDicom` focused because preview SDK knows nothing about an Ambra integration

src/lib/viewers/iframe/IFrameLoader.js
```diff
@@ -22,6 +23,20 @@ class IFrameLoader extends AssetLoader {
...
+        // The IFrame viewer is disabled when the file is a Boxdicom file and Open_with_Ambra FF is enabled
```
ConradJChan, line 33: We shouldn't mention FF from other codebases here

src/lib/viewers/iframe/__tests__/IFrameViewer-test.js
```diff
@@ -1,5 +1,6 @@
 import IFrameViewer from '../IFrameViewer';
 import BaseViewer from '../../BaseViewer';
+import IFrameLoader from '../IFrameLoader';
```
ConradJChan, line 3: Sort the imports

src/lib/viewers/iframe/__tests__/IFrameViewer-test.js
```diff
@@ -82,6 +90,42 @@ describe('lib/viewers/iframe/IFrameViewer', () => {
...
+            const viewerOptions = {
+                IFrame: {
+                    openWithAmbra: true,
+                },
+            };
+            const viewer = IFrameLoader.determineViewer(iframe, [], viewerOptions);
+            expect(viewer).toBeUndefined();
+        });
+
+        test('should return a viewer if file is a boxnote and open_with_ambra FF is enabled', () => {
```
ConradJChan, line 105: Could lump these together with a `test.each` https://jestjs.io/docs/api#1-testeachtablename-fn-timeout

src/lib/viewers/iframe/IFrameLoader.js
```diff
@@ -22,6 +23,20 @@ class IFrameLoader extends AssetLoader {
...
+        if (openWithAmbraEnabled && isDicomFile) {
+            disabledViewers.push('IFrame');
+        }
+
+        return super.determineViewer(file, disabledViewers);
```
jstoffan, line 38: Should we pass `viewerOptions` in here as the third argument?

jstoffan:
@JChan106, we'll probably want to go with `feat` rather than `chore` for the commit message and PR title since this change includes new functionality.

src/lib/viewers/iframe/IFrameLoader.js
JChan106: It's not needed in this case since we are only disabling the IFrame viewer. However from my understanding if we were to have viewer options that modifies logic in both the IFrame loader and Asset loader then we would.

jstoffan: I was mainly thinking about maintaining consistency with what the base class expects, but it looks like many other viewers fail to pass `viewerOptions` down, as well. Not a big deal in that case.

jstoffan (approved):

src/lib/viewers/iframe/__tests__/IFrameViewer-test.js
```diff
@@ -82,6 +90,66 @@ describe('lib/viewers/iframe/IFrameViewer', () => {
...
+            (disableDicom, fileType, viewerDefined) => {
+                iframe.options.file.extension = fileType;
+
+                const viewerOptions = {
+                    IFrame: {
+                        disableDicom,
+                    },
+                };
+                const viewer = IFrameLoader.determineViewer(iframe.options.file, [], viewerOptions);
+                if (viewerDefined) {
```
jstoffan, line 145: We typically want to avoid having logic and conditions in our tests. We could do something like this, though:

```
test.each`
    disableDicom    | fileType        | viewerInstance
    ${false}        | ${'boxdicom"}   | ${IFrameViewer}
    ${true}         | ${'boxdicom"}   | ${undefined}
`

**Review by jstoffan — CHANGES_REQUESTED** [review]
Requesting changes to cancel the prior approval. Sorry for the confusion.

File: src/lib/viewers/iframe/__tests__/IFrameLoader-test.js [hunk 8]
```diff
@@ -18,4 +31,24 @@ describe('lib/viewers/iframe/IFrameLoader', () => {
             EXT: ['boxnote', 'boxdicom'],
         });
     });
+
+    test.each([
+        // disableDicom, fileType, viewerInstance
```
jstoffan, line 36: In the future, consider using the [tagged template literal syntax](https://github.com/box/box-content-preview/pull/1436#discussion_r751671279) to incorporate these variable names directly into the inputs.

src/lib/viewers/iframe/__tests__/IFrameLoader-test.js
```diff
@@ -1,10 +1,23 @@
...
+                    {
+                        representation: 'ORIGINAL',
+                    },
+                ],
+            },
+        },
+    });
+
     afterEach(() => {
         sandbox.verifyAndRestore();
```
jstoffan, line 22: This isn't directly related to your change, but we should remove these when we edit a test file, if possible. It's not actually being used in any tests in this case, either.

jstoffan (dismissed):

ConradJChan (approved):
Maybe instead of `Remove support...` in the PR title, it is `Add support for disabling .boxdicom`

src/lib/viewers/iframe/__tests__/IFrameLoader-test.js
JChan106: Ohh, my bad I didn't know that actual syntax. Will update.

jstoffan (approved):
Looks great. Thanks for working through the feedback!

PR #1437, opened 2021-11-19
feat(viewer): use unsupported file type error message for .boxdicom

JChan106:
Uses the unsupported file type error message instead of the account error message for Box Dicom files.

### DIFF
--- a/src/lib/viewers/iframe/IFrameLoader.js
+++ b/src/lib/viewers/iframe/IFrameLoader.js
@@ -34,10 +34,9 @@ class IFrameLoader extends AssetLoader {
      * @inheritdoc
      */
     determineViewer(file, disabledViewers = [], viewerOptions = {}) {
-        const isDicomFile = file.extension === 'boxdicom';
         const disableDicom = getProp(viewerOptions, 'IFrame.disableDicom');
         // Removes boxdicom as a supported extension when the file is a Boxdicom file and the disableDicom viewer option is enabled
-        if (disableDicom && isDicomFile) {
+        if (disableDicom) {
             this.viewers = this.viewers.filter(viewer => !viewer.EXT.includes('boxdicom'));
         }
 
--- a/src/lib/viewers/iframe/IFrameLoader.js
+++ b/src/lib/viewers/iframe/IFrameLoader.js
@@ -35,7 +35,7 @@ class IFrameLoader extends AssetLoader {
      */
     determineViewer(file, disabledViewers = [], viewerOptions = {}) {
         const disableDicom = getProp(viewerOptions, 'IFrame.disableDicom');
-        // Removes boxdicom as a supported extension when the file is a Boxdicom file and the disableDicom viewer option is enabled
+        // Removes boxdicom as a supported extension when the disableDicom viewer option is enabled
         if (disableDicom) {
             this.viewers = this.viewers.filter(viewer => !viewer.EXT.includes('boxdicom'));
         }
--- a/src/lib/viewers/iframe/IFrameLoader.js
+++ b/src/lib/viewers/iframe/IFrameLoader.js
@@ -9,7 +9,13 @@ const VIEWERS = [
         NAME: 'IFrame',
         CONSTRUCTOR: IFrameViewer,
         REP: ORIGINAL_REP_NAME,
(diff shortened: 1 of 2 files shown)


src/lib/viewers/iframe/IFrameLoader.js
```diff
@@ -33,6 +33,12 @@ class IFrameLoader extends AssetLoader {
         // The IFrame viewer is disabled when the file is a Boxdicom file and the disableDicom viewer option is enabled
         if (disableDicom && isDicomFile) {
             disabledViewers.push('IFrame');
```
jstoffan, line 35: Do we still need this logic now that the extension is removed?

src/lib/viewers/iframe/IFrameLoader.js
```diff
@@ -33,6 +33,12 @@ class IFrameLoader extends AssetLoader {
...
+
+            // Removes boxdicom as a supported extension
+            const iframeViewer = this.viewers[0].EXT;
```
jstoffan, line 38: Can we simplify things a bit by breaking out the viewers into multiple entries? Maybe something like:


```
const VIEWERS = [
    {
        NAME: 'IFrame',
        CONSTRUCTOR: IFrameViewer,
        REP: ORIGINAL_REP_NAME,
        EXT: ['boxdicom'],
    },
    {
        NAME: 'IFrame',
        CONSTRUCTOR: IFrameViewer,
        REP: ORIGINAL_REP_NAME,
        EXT: ['boxnote'],
    },
];

...

if (disableDicom && isDicomFile) {
    this.viewers = this.viewers.filter(viewer => !viewer.EXT.includes('boxdicom'));
}
```

src/lib/viewers/iframe/IFrameLoader.js
JChan106: Yeah I don't think we need it anymore, will remove.

src/lib/viewers/iframe/__tests__/IFrameLoader-test.js
```diff
@@ -43,6 +51,7 @@ describe('lib/viewers/iframe/IFrameLoader', () => {
             };
             const viewer = IFrameLoader.determineViewer(iframe.options.file, [], viewerOptions);
             expect(viewer).toEqual(viewerInstance);
+            expect(IFrameLoader.getViewers().some(v => v.EXT.includes('boxdicom'))).toEqual(!disableDicom);
```
jstoffan, line 54: Do we need this assertion if we already have the one above it?

jstoffan (dismissed):

src/lib/viewers/iframe/IFrameLoader.js
```diff
@@ -30,9 +36,9 @@ class IFrameLoader extends AssetLoader {
     determineViewer(file, disabledViewers = [], viewerOptions = {}) {
         const isDicomFile = file.extension === 'boxdicom';
         const disableDicom = getProp(viewerOptions, 'IFrame.disableDicom');
-        // The IFrame viewer is disabled when the file is a Boxdicom file and the disableDicom viewer option is enabled
+        // Removes boxdicom as a supported extension when the file is a Boxdicom file and the disableDicom viewer option is enabled
```
ConradJChan, line 39: Do we even need to check if `isDicomFile` is true anymore? If `disableDicom` option is passed in, would it work if we just remove it from the `this.viewers`?

JChan106: I think you're right, it doesn't really matter about the current file, we should just remove the dicom viewer from the supported viewers if the disableDicom is true.

src/lib/viewers/iframe/__tests__/IFrameLoader-test.js
JChan106: We could, the assertion checks if the list of viewers is correct based on the disableDicom viewer option. While the above assertion checks if the viewer instance is correct.

src/lib/viewers/iframe/IFrameLoader.js
ConradJChan: May want to update the comment as well to be consistent with the code

ConradJChan (dismissed):

jstoffan (approved):

PR #1439, opened 2021-12-15
feat(viewer): Add focus trap for fullscreen

ConradJChan:
Adds `FocusTrap` class which takes an element and allows you to `enable` or `disable` focus trapping.

This class is used for when Preview enters fullscreen to constrain tab navigation to within the visible previewed content. This can also be applied to the print modal.

TODO:
- [x] add/update unit tests
- [x] cross-browser testing

### DIFF
--- a/src/lib/FocusTrap.ts
+++ b/src/lib/FocusTrap.ts
@@ -1,0 +1,137 @@
+import { decodeKeydown } from './util';
+
+const FOCUSABLE_ELEMENTS = [
+    'a[href]:not([disabled])',
+    'button:not([disabled])',
+    'textarea:not([disabled])',
+    'input[type="text"]:not([disabled])',
+    'input[type="radio"]:not([disabled])',
+    'input[type="checkbox"]:not([disabled])',
+    'select:not([disabled])',
+];
+
+function isButton(element: HTMLElement): boolean {
+    return element.tagName.toLowerCase() === 'button';
+}
+function isVisible(element: HTMLElement): boolean {
+    return element.offsetHeight > 0 && element.offsetWidth > 0;
+}
+function createFocusAnchor(): HTMLElement {
+    const element = document.createElement('i');
+    element.setAttribute('aria-hidden', 'true');
+    element.tabIndex = 0;
+
+    return element;
+}
+
+class FocusTrap {
+    element: HTMLElement;
+
+    firstFocusableElement: HTMLElement | null = null;
+
+    lastFocusableElement: HTMLElement | null = null;
+
+    trapFocusableElement: HTMLElement | null = null;
+
+    constructor(element: HTMLElement) {
+        this.element = element;
+    }
+
+    destroy(): void {
+        this.disable();
+    }
+
+    focusFirstElement = (): void => {
+        if (!this.element) {
+            return;
+        }
+
+        const focusableElements = this.getFocusableElements();
+        if (focusableElements.length > 0) {
+            focusableElements[0].focus();
+        } else if (this.trapFocusableElement) {
+            this.trapFocusableElement.focus();
+        }
+    };
+
+    focusLastElement = (): void => {
+        if (!this.element) {
+            return;
+        }
+
+        const focusableElements = this.getFocusableElements();
+        if (focusableElements.length > 0) {
+            focusableElements[focusableElements.length - 1].focus();
+        } else if (this.trapFocusableElement) {
+            this.trapFocusableElement.focus();
+        }
+    };
+
+    getFocusableElements = (): Array<HTMLElement> => {
+        // Look for focusable elements
+        const foundElements = this.element.querySelectorAll(FOCUSABLE_ELEMENTS.join(', '));
+        // Filter out the elements that are preview control buttons and not visible
+        return Array.prototype.slice.call(foundElements).filter(el => (isButton(el) && isVisible(el)) || !isButton(el));
+    };
+
+    handleTrapKeydown = (event: KeyboardEvent): void => {
+        const key = decodeKeydown(event);
+        const isTabPressed = key === 'Tab' || key === 'Shift+Tab';
+
+        if (!isTabPressed) {
+            return;
+        }
+        event.stopPropagation();
+        event.preventDefault();
+    };
+
+    handleKeydown = (event: KeyboardEvent): void => {
+        const key = decodeKeydown(event);
+        const isTabPressed = key === 'Tab' || key === 'Shift+Tab';
+
+        if (this.element === document.activeElement && isTabPressed) {
+            this.focusFirstElement();
+            event.stopPropagation();
+            event.preventDefault();
+        }
+    };
+
+    enable(): void {
+        this.element.addEventListener('keydown', this.handleKeydown);
+
+        // Create focus anchors (beginning, end and trap)
+        this.firstFocusableElement = createFocusAnchor();
+        this.lastFocusableElement = createFocusAnchor();
+        this.trapFocusableElement = createFocusAnchor();
+
+        this.firstFocusableElement.addEventListener('focus', this.focusLastElement);
+        this.lastFocusableElement.addEventListener('focus', this.focusFirstElement);
+        this.trapFocusableElement.addEventListener('keydown', this.handleTrapKeydown);
+
+        this.element.prepend(this.firstFocusableElement);
+        this.element.append(this.lastFocusableElement);
+        this.element.append(this.trapFocusableElement);
+    }
+
+    disable(): void {
+        this.element.removeEventListener('keydown', this.handleKeydown);
+
+        if (this.firstFocusableElement) {
+            this.element.removeChild(this.firstFocusableElement);
+        }
+
+        if (this.lastFocusableElement) {
+            this.element.removeChild(this.lastFocusableElement);
+        }
+
+        if (this.trapFocusableElement) {
+            this.element.removeChild(this.trapFocusableElement);
+        }
+
+        this.firstFocusableElement = null;
+        this.lastFocusableElement = null;
+        this.trapFocusableElement = null;
+    }
+}
+
+export default FocusTrap;
--- a/src/lib/FocusTrap.ts
+++ b/src/lib/FocusTrap.ts
@@ -1,14 +1,22 @@
 import { decodeKeydown } from './util';
 
 const FOCUSABLE_ELEMENTS = [
-    'a[href]:not([disabled])',
-    'button:not([disabled])',
-    'textarea:not([disabled])',
-    'input[type="text"]:not([disabled])',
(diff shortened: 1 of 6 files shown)


src/lib/viewers/BaseViewer.js
```diff
@@ -277,11 +281,16 @@ class BaseViewer extends EventEmitter {
             }
         }
 
+        if (this.focusTrap) {
```
jstoffan, line 284: Should we destroy the focus trap before we destroy `this.containerEl` up above?

src/lib/viewers/BaseViewer.js
```diff
@@ -579,6 +588,13 @@ class BaseViewer extends EventEmitter {
         if (this.fullscreenToggleEl && this.fullscreenToggleEl.focus) {
             this.fullscreenToggleEl.focus();
         }
+
+        if (this.focusTrap) {
+            this.focusTrap.destroy();
```
jstoffan, line 593: Why do we need to destroy/recreate the focus trap each time we enter fullscreen mode?

src/lib/FocusTrap.ts
```diff
@@ -0,0 +1,137 @@
+import { decodeKeydown } from './util';
+
+const FOCUSABLE_ELEMENTS = [
```
jstoffan, line 3: Do we need to check for any other attributes like `tabindex="-1"` or `aria-disabled`?

src/lib/FocusTrap.ts
```diff
@@ -0,0 +1,137 @@
...
+        this.lastFocusableElement.addEventListener('focus', this.focusFirstElement);
+        this.trapFocusableElement.addEventListener('keydown', this.handleTrapKeydown);
+
+        this.element.prepend(this.firstFocusableElement);
+        this.element.append(this.lastFocusableElement);
+        this.element.append(this.trapFocusableElement);
+    }
+
+    disable(): void {
+        this.element.removeEventListener('keydown', this.handleKeydown);
```
jstoffan, line 117: We have guard clauses for `if (!this.element)` in some of our other methods. Do we need them everywhere?

src/lib/FocusTrap.ts
```diff
@@ -0,0 +1,137 @@
...
+        // Create focus anchors (beginning, end and trap)
+        this.firstFocusableElement = createFocusAnchor();
+        this.lastFocusableElement = createFocusAnchor();
+        this.trapFocusableElement = createFocusAnchor();
+
+        this.firstFocusableElement.addEventListener('focus', this.focusLastElement);
+        this.lastFocusableElement.addEventListener('focus', this.focusFirstElement);
+        this.trapFocusableElement.addEventListener('keydown', this.handleTrapKeydown);
+
+        this.element.prepend(this.firstFocusableElement);
```
jstoffan, line 111: I don't think `Element.prepend` is supported in IE11 by default. Do we have it polyfilled?

src/lib/FocusTrap.ts
```diff
@@ -0,0 +1,137 @@
...
+        } else if (this.trapFocusableElement) {
+            this.trapFocusableElement.focus();
+        }
+    };
+
+    getFocusableElements = (): Array<HTMLElement> => {
+        // Look for focusable elements
+        const foundElements = this.element.querySelectorAll(FOCUSABLE_ELEMENTS.join(', '));
+        // Filter out the elements that are preview control buttons and not visible
+        return Array.prototype.slice.call(foundElements).filter(el => (isButton(el) && isVisible(el)) || !isButton(el));
```
jstoffan, line 74: Do we have `Array.from` available and polyfilled in IE11? If so, we could use that here?

src/lib/FocusTrap.ts
ConradJChan: `tabindex="-1"` doesn't seem necessary since those elements wouldn't be keyboard navigable. `aria-disabled` also doesn't seem relevant because the control is disabled anyway and can't be interacted with

src/lib/FocusTrap.ts
ConradJChan: It does look like we have `Array.from` polyfilled, I'll switch to that

src/lib/FocusTrap.ts
ConradJChan: negative, I'll do it another way thanks for the catch

src/lib/FocusTrap.ts
ConradJChan: I'll simplify it by throwing an error in the constructor if we don't get an element so I can remove all guard clauses

src/lib/FocusTrap.ts
```diff
@@ -0,0 +1,141 @@
...
+        }
+        event.stopPropagation();
+        event.preventDefault();
+    };
+
+    handleKeydown = (event: KeyboardEvent): void => {
+        const key = decodeKeydown(event);
+        const isTabPressed = key === 'Tab' || key === 'Shift+Tab';
+
+        if (this.element === document.activeElement && isTabPressed) {
```
jstoffan, line 96: Should this be a guard clause, similar to `handleTrapKeydown` above?

src/lib/FocusTrap.ts
```diff
@@ -0,0 +1,141 @@
+import { decodeKeydown } from './util';
+
+const FOCUSABLE_ELEMENTS = [
```
jstoffan, line 3: How about `FOCUSABLE_SELECTORS`, since we usually use "element" to mean something else?

src/lib/FocusTrap.ts
```diff
@@ -0,0 +1,141 @@
...
+    'a[href]',
+    'button',
+    'textarea',
+    'input[type="text"]',
+    'input[type="radio"]',
+    'input[type="checkbox"]',
+    'select',
+];
+const ATTRIBUTES = ['disabled', 'tabindex="-1"', 'aria-disabled="true"'];
```
jstoffan, line 12: Should we clarify the variable name?

src/lib/FocusTrap.ts
```diff
@@ -0,0 +1,141 @@
...
+            focusableElements[focusableElements.length - 1].focus();
+        } else if (this.trapFocusableElement) {
+            this.trapFocusableElement.focus();
+        }
+    };
+
+    getFocusableElements = (): Array<HTMLElement> => {
+        // Look for focusable elements
+        const foundElements = this.element.querySelectorAll<HTMLElement>(FOCUSABLE_ELEMENTS_SELECTOR);
+        // Filter out the elements that are preview control buttons and not visible
```
jstoffan, line 77: Is this behavior only for preview controls or would it work for modals and such, as well?

jstoffan (commented):
LGTM apart from some super minor comments, just needs tests.

src/lib/viewers/__tests__/BaseViewer-test.js
```diff
@@ -534,6 +540,28 @@ describe('lib/viewers/BaseViewer', () => {
             expect(base.disableAnnotationControls).toBeCalled();
             expect(base.processAnnotationModeChange).toBeCalledWith(AnnotationMode.NONE);
         });
+
+        test('should enable the focus trap', () => {
+            jest.spyOn(FocusTrap.prototype, 'constructor');
```
jstoffan, line 545: It's a little unusual to spy on the prototype. Can we instead make sure that `base.focusTrap` is defined, then  use `jest.spyOn(base.focusTrap, 'enable')`? That way we're testing the instance rather than the prototype.

src/lib/__tests__/FocusTrap-test.ts
```diff
@@ -0,0 +1,147 @@
...
+            expect(document.activeElement).toBe(children[1]);
+        });
+
+        test('should keep focus trapped on trap anchor when Tab is pressed', () => {
+            children[4].focus(); // trap focus anchor element
+            expect(document.activeElement).toBe(children[4]);
+
+            children[4].dispatchEvent(new KeyboardEvent('keydown', { key: 'Tab' }));
+            expect(document.activeElement).toBe(children[4]);
+            expect(Event.prototype.stopPropagation).toHaveBeenCalled();
```
jstoffan, line 87: Can we spy on the specific event that was fired rather than the prototype?

src/lib/__tests__/FocusTrap-test.ts
```diff
@@ -0,0 +1,147 @@
...
+        test('should redirect focus from first anchor to last focusable element', () => {
+            children[1].focus(); // first input
+            expect(document.activeElement).toBe(children[1]);
+
+            children[0].focus(); // first focus anchor element
+            expect(document.activeElement).toBe(children[2]);
+        });
+
+        test('should redirect focus from last anchor to first focusable element', () => {
+            children[2].focus(); // second input
```
jstoffan, line 74: Accessing the children via index is a little fragile. Should we use class names or `data-testid`?

src/lib/FocusTrap.ts
```diff
@@ -0,0 +1,148 @@
...
+            this.focusLastElement();
+        }
+        event.stopPropagation();
+        event.preventDefault();
+    };
+
+    enable(): void {
+        this.element.addEventListener('keydown', this.handleKeydown);
+
+        // Create focus anchors (beginning, end and trap)
```
jstoffan, line 113: Do we need this comment?

src/lib/FocusTrap.ts
```diff
@@ -0,0 +1,148 @@
...
+
+    firstFocusableElement: HTMLElement | null = null;
+
+    lastFocusableElement: HTMLElement | null = null;
+
+    trapFocusableElement: HTMLElement | null = null;
+
+    constructor(element: HTMLElement) {
+        if (!element) {
+            throw new Error('FocusTrap needs an HTMLElement passed into the constructor');
```
jstoffan, line 47: Do we need to catch this exception in `Preview.js`?

src/lib/viewers/__tests__/BaseViewer-test.js
ConradJChan: Yeah, I don't prefer spying the prototype but I went this route because I wanted to test the case where we actually create the FocusTrap vs reuse it. I went with the approach you suggested below when reusing the existing focus trap

src/lib/FocusTrap.ts
```diff
@@ -0,0 +1,148 @@
...
+        if (this.element !== document.activeElement || !isTabPressed) {
+            return;
+        }
+
+        if (key === 'Tab') {
+            this.focusFirstElement();
+        } else {
+            this.focusLastElement();
+        }
+        event.stopPropagation();
```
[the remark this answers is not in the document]
ConradJChan: Seems like if we're controlling the focus because of the tab key we should stop propagation. This is also how it's done in box-ui-elements

jstoffan (approved):

PR #1444, opened 2022-03-23
feat(docuworks): Add support for docuworks file types

ConradJChan:


package.json
```diff
@@ -17,6 +17,7 @@
         "@babel/preset-typescript": "^7.7.7",
         "@box/frontend": "^6.4.0",
         "@box/languages": "^1.0.0",
+        "@box/react-virtualized": "9.22.3-rc-box.2",
```
jstoffan, line 20: Should we remove the regular `react-virtualized`?

package.json
```diff
@@ -124,9 +125,9 @@
         "release:patch": "./build/release.sh -p",
         "setup": "yarn install",
         "start": "yarn setup && yarn build:dev --watch",
-        "start:linked": "yarn link box-annotations && IS_LINKED=1 yarn start",
+        "start:linked": "f() { LINKED_LIBS=${@:-box-annotations} && yarn link $LINKED_LIBS && LINKED_LIBS=$LINKED_LIBS IS_LINKED=1 yarn start; }; f",
```
jstoffan, line 128: I don't think `LINKED_LIBS` is supposed by box-content-previews Webpack config.

src/lib/extensions.js
```diff
@@ -51,6 +51,7 @@ export const CODE_EXTENSIONS = [
     'yaml',
 ];
 
+const DOCUWORKS_EXTENSIONS = ['xdw', 'xbd'];
```
jstoffan, line 54: Can we export this const to match the others in the file?

package.json
```diff
@@ -35,7 +36,7 @@
         "babel-loader": "^8.0.6",
         "babel-plugin-transform-require-ignore": "^0.1.1",
         "box-annotations": "^2.3.0",
-        "box-ui-elements": "^13.1.0-beta.62",
+        "box-ui-elements": "^15.0.0-beta.49",
```
jstoffan, line 39: This is a big jump. Do all our BUIE usages still work as expected?

ConradJChan: Will do some regression testing

ConradJChan: Looks good, tested CSV viewer, archive viewer, annotations

src/lib/viewers/text/BoxCSV.js
```diff
@@ -1,6 +1,6 @@
 import React from 'react';
 import ReactDOM from 'react-dom';
-import Grid from 'react-virtualized/dist/es/Grid/Grid';
+import { Grid } from '@box/react-virtualized/dist/es/Grid';
```
jstoffan, line 3: Do we need the object notation? It looks like the Grid barrel file re-exports `Grid` as its default.

ConradJChan: Yeah I tried without and the unit tests failed, I think because of the jest mapping to the `commonjs` directory might not have the same exports

PR #1445, opened 2022-05-08
feat(preview): Tooltip for Persistent Onboarding

robertjk:
## Description

Implement Annotations tooltip for Persistent Onboarding Box Edit:
- Create a new component `PersistentOnboardingBoxEditAnnotationsTooltip`
  * It mostly duplicates code of `PersistentOnboardingTooltip` from `EndUserApp`
- As `PersistentOnboardingBoxEditAnnotationsTooltip` displays on the same element as `AnnotationsTargetedTooltip`, I created a wrapper `AnnotationsButtonRegionTooltip` that controls which tooltip to show. `AnnotationsTargetedTooltip` was moved inside it and renamed to `AnnotationsFlowTooltip` (to match it's experience name: Tooltip Flow Annotations)
- ~Modify existing `experiences` option:~
  * ~Extend `experiences` option to have 2 additional properties necessary for the new tooltip:~
    + ~`isLocaleLanguageEnglish`~
    + ~`onPersistentOnboardingTooltipBack`~
  * ~Rename the whole wrapping option from `experiences` to `dependencies`, while moving `experiences` to be inner property on `dependencies`~
- Add Persistent Onboarding Targeting API type with additional `onPrevious()` callback property
- Implement getting locale language

## TODO

Few things still to do before I convert it out of Draft:
- [x] Implement tests for tooltip components
- [x] Rename `dependencies` property ~to something more specific~ back to `experiences`
- [ ] Investigate the bundle size after importing `<TargetedClickThroughGuideTooltip>` from `box-ui-elements`

## Big size of the PR

**Sorry for such a huge PR!** Initially it was planned to be much smaller, but I was encountering issues along the way and it grew a lot. I'll gladly split it in few PR-s, if anyone would prefer that.


src/lib/viewers/controls/annotations/annotations-button-region-tooltips/AnnotationsButtonRegionTooltips.tsx
```diff
@@ -0,0 +1,24 @@
...
+
+export type Props = PropsWithChildren<{
+    isEnabled?: boolean;
+}>;
+
+export default function AnnotationsButtonRegionTooltips({ children, isEnabled }: Props): JSX.Element {
+    const { experiences } = useContext(DependenciesContext);
+    const { persistentOnboardingBoxEditAnnotations: persistentOnboardingBoxEditAnnotationsExperience } = experiences;
+
+    if (persistentOnboardingBoxEditAnnotationsExperience && persistentOnboardingBoxEditAnnotationsExperience.canShow) {
```
wpiesiak, line 16: I think we could use optional chaining here: 

```suggestion
    if (persistentOnboardingBoxEditAnnotationsExperience?.canShow) {
```

src/lib/viewers/controls/annotations/annotations-button-region-tooltips/PersistentOnboardingBoxEditAnnotationsTooltip.tsx
```diff
@@ -0,0 +1,81 @@
...
+        },
+    });
+
+    const title = __('persistent_onboarding_box_edit_annotations_tooltip_title');
+    const body = __('persistent_onboarding_box_edit_annotations_tooltip_body');
+
+    const image = (
+        <div
+            className={classNames('bp-PersistentOnboardingBoxEditAnnotationsTooltip-image', {
+                isLocalized: isLocaleLanguageEnglish,
```
jstoffan, line 48: Should this condition be inverted?

src/lib/viewers/controls/annotations/annotations-button-region-tooltips/PersistentOnboardingBoxEditAnnotationsTooltip.tsx
```diff
@@ -0,0 +1,81 @@
+import React, { useContext, PropsWithChildren } from 'react';
+import classNames from 'classnames';
+
+// TODO: Compare bundle sizes and possibly re-write here
```
jstoffan, line 4: I agree, we should do this before proceeding. It looks like this component has many other dependencies.

src/lib/viewers/controls/annotations/annotations-button-region-tooltips/PersistentOnboardingBoxEditAnnotationsTooltip.tsx
```diff
@@ -0,0 +1,81 @@
...
+    );
+
+    const primaryButtonProps = {
+        children: __('persistent_onboarding_box_edit_annotations_tooltip_button_next'),
+    };
+
+    const backButtonProps = {
+        children: __('persistent_onboarding_box_edit_annotations_tooltip_button_previous'),
+        onClick(): void {
+            onPersistentOnboardingTooltipBack();
```
jstoffan, line 61: The targeting API should give us many different callbacks. Can we use those rather than passing in a net new one?

jstoffan (changes requested):

src/lib/viewers/controls/annotations/annotations-button-region-tooltips/AnnotationsButtonRegionTooltips.tsx
robertjk: I actually tried using optional chaining, but got Babel compilation error that this feature is off in this project and would need to be turned on. Very useful feature though, so it might be worthwhile considering turning it on.

I'm resolving this.

src/lib/Preview.js
```diff
@@ -980,8 +980,8 @@ class Preview extends EventEmitter {
         // have preview permissions (any collaboration role except for `Uploader`).
         this.options.downloadWM = !!options.downloadWM;
 
-        // Object with ftux experience data that can determine whether specific experiences can show
-        this.options.experiences = options.experiences || {};
+        // Object with ftux experience data that can determine whether specific dependencies can show
+        this.options.dependencies = options.dependencies || {};
```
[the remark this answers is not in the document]
robertjk: I renamed the property, as it's no longer just experiences that we're passing in - it's experiences + other stuff. I felt the name `experiences` is no longer valid in this context.

That being said `dependencies` is not a good name neither, way too general. I thought `experiencesDependencies` or `ftuxDependencies` (FTUX = Firs Time User Experience) might be better.

But if you feel that keeping the old `experiences` is the way to go, alright. Just tell me how do you think I should nest `onPersistentOnboardingTooltipBack` property there? Just nest the experiences and `onPersistentOnboardingTooltipBack` on the top level? (below)

```
experiences: {
  persistentOnboardingBoxEditAnnotationsExperience,
  tooltipFlowAnnotationsExperience,
  onPersistentOnboardingTooltipBack
}
```

src/lib/viewers/image/ImageBaseViewer.js
```diff
@@ -145,10 +145,14 @@ class ImageBaseViewer extends BaseViewer {
      * @return {boolean} value of whether discoverability is enabled for given type
      */
     isDiscoverabilityEnabled() {
-        const { enableAnnotationsImageDiscoverability, experiences = {} } = this.options;
-        const canShow = Object.values(experiences).some(experience => experience.canShow);
```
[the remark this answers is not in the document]
robertjk: This change was needed because:
- `experiences` structure changed
- We have another experience now, and I don't think this new experience's `canShow` should affect this method, while it'd do with the old implementation. So I refactored the method to depend only on `tooltipFlowAnnotationsExperience.canShow` (but not on `persistentOnboardingBoxEditAnnotationsExperience.canShow`)

src/lib/viewers/controls/annotations/annotations-button-region-tooltips/PersistentOnboardingBoxEditAnnotationsTooltip.tsx
robertjk: Good catch. I have actually already fixed that. Am resolving.

src/lib/viewers/controls/annotations/annotations-button-region-tooltips/PersistentOnboardingBoxEditAnnotationsTooltip.tsx
jstoffan: Does the Targeting API need to evolve to support this new capability? It seems like it could be a common need.

robertjk: I'm not 100% sure about that, but I think we currently don't have any plans of integrating such a feature into our Targeting API. @wpiesiak @megansmith-box Can you confirm?

It's not a bad idea though. It definitely feels limiting our Experiences API allows you to proceed forward with the experience but not to go back.

src/lib/viewers/controls/dependencies/DependenciesContext.ts
```diff
@@ -0,0 +1,18 @@
+import noop from 'lodash/noop';
+import React from 'react';
+
+import { Experiences } from '../../../types';
+
+export type Dependencies = {
+    experiences: Experiences;
+    isLocaleLanguageEnglish: boolean;
```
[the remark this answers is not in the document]
robertjk: Can you point me in the right direction on how to do this? I see we have `lib/i18n.js` file, which provides `locale` property. I see it's used in `BaseViewer.js`, but I'm not sure how to access it in React app under `<AnnotationsControls>`.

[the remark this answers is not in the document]
jstoffan: You can access it via the `__LANGUAGE__` global variable.

src/lib/viewers/image/ImageBaseViewer.js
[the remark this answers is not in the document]
jstoffan: @megansmith-box, can you comment on the existing logic and if/how the new experience should integrate with it?

src/lib/viewers/controls/annotations/annotations-button-region-tooltips/PersistentOnboardingBoxEditAnnotationsTooltip.tsx
jstoffan: I think we'd like each experience to be as self-contained as possible. Let's solve this in the framework rather than adding yet more options to the Preview SDK.

robertjk: > I think we'd like each experience to be as self-contained as possible. 

I agree. That's why my first approach was to not put this experience-related code into `box-content-preview` at all. But that turned out to be impossible.

> Let's solve this in the framework rather than adding yet more options to the Preview SDK.

I don't see a way to do this within the current framework. Do you?

src/lib/viewers/controls/dependencies/DependenciesContext.ts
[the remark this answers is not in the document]
robertjk: Thanks. `__LANGUAGE__` returns the `language-TERRITORY` form. I see in `src/lib/i18n.js` we already have a method `getLocale()` which returns only the language value, which is what I need. Is there a way to access this somehow?

src/lib/viewers/controls/annotations/annotations-button-region-tooltips/PersistentOnboardingBoxEditAnnotationsTooltip.tsx
robertjk: I'll do that, when other issues mentioned here are resolved. I added that to TODO section in PR's description.

src/lib/viewers/controls/dependencies/DependenciesContext.ts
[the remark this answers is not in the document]
jstoffan: You can create and export a `getCurrentLocale` method in `lib/i18n` that derives the locale from the underlying `__LANGUAGE__` variable. That should solve both problems.

src/lib/viewers/controls/annotations/annotations-button-region-tooltips/PersistentOnboardingBoxEditAnnotationsTooltip.tsx
jstoffan: My ask is to see if we can extend the framework with new functionality for this new behavior.

src/lib/viewers/controls/dependencies/DependenciesContext.ts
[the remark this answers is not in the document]
robertjk: I'll gladly do that. And then I'd just import `getCurrentLocale` from `lib/i18n.js` in my component file, right?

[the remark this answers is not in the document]
jstoffan: Yes, which will also allow you to mock that method to avoid needing to assign to the global constant.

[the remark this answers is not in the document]
robertjk: Done.

I did it using a static variable `currentLocale` and not as a method `getCurrentLocale()`. Did it this way as the rest of the implementation of `i18n.js` seems to be static as well - `language` value is initialized when the module is executed for the first time and doesn't change after that.

Let me know if there is a reason to make this a method instead.

src/lib/viewers/image/ImageBaseViewer.js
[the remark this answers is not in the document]
megansmith-box: @robertjk - This method was updated because for Images in Preview, we noticed that the Region is default Selected. We want to set the default selected to None, if this tooltip experience can show. We also want the tooltip to close when you click on the mode button. Is the new tooltip placed on the control bar? If so, I believe we do want to include it in this method, so we can leave it how it was. (If we rename dependencies back to experiences).

src/lib/Preview.js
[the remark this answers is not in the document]
robertjk: That's reverted back to `experiences`. Am resolving this.

src/lib/viewers/controls/annotations/annotations-button-region-tooltips/PersistentOnboardingBoxEditAnnotationsTooltip.tsx
robertjk: I extended `TargetingApi` type with `onPrevious()` callback, as we discussed on Slack. Am resolving this. Please create a new thread to discuss the new implementation.

src/lib/viewers/image/ImageBaseViewer.js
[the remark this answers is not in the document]
robertjk: @megansmith-box Thanks for that explanation.

> Is the new tooltip placed on the control bar? If so, I believe we do want to include it in this method, so we can leave it how it was.

Yes, the new tooltip displays on the same element as the previous one. So I reverted back to the old implementation, so that it affects the result of this method as well. Am resolving this.

src/lib/i18n.js
```diff
@@ -26,3 +28,4 @@ const createAnnotatorIntl = () => {
 };
 
 export default { createAnnotatorIntl };
+export { currentLocale };
```
jstoffan, line 31: Please add this to the object in the default export above to keep the module exports consistent.

src/lib/viewers/controls/annotations/annotations-button-region-tooltips/AnnotationsButtonRegionTooltips.tsx
```diff
@@ -0,0 +1,26 @@
+import React, { useContext, PropsWithChildren } from 'react';
```
jstoffan, line 1: Can we remain consistent with the existing codebase conventions? We typically use the `React` namespace for these helpers.

src/lib/viewers/controls/annotations/annotations-button-region-tooltips/AnnotationsFlowTooltip.scss
```diff
@@ -1,14 +1,14 @@
-.bp-AnnotationsTooltip {
+.bp-AnnotationsFlowTooltip {
```
jstoffan, line 1: It doesn't seem like these changes are strictly necessary. Can we revert?

src/lib/viewers/controls/annotations/annotations-button-region-tooltips/PersistentOnboardingBoxEditAnnotationsTooltip.scss
```diff
@@ -0,0 +1,39 @@
...
+        margin-top: 25px;
+        background-image: url('https://cdn01.boxcdn.net/app-assets/product-adoption/persistent-onboarding/annotations-english.gif');
+        background-repeat: no-repeat;
+        background-position: top center;
+        background-size: cover;
+        border-radius: ($bdl-grid-unit * 4) ($bdl-grid-unit * 4) 0 0;
+        box-shadow: 0 8px 24px 0 rgba(0, 0, 0, .08);
+        content: '';
+
+        &.is-localized {
```
jstoffan, line 35: Please prefix with `bp-`

src/lib/viewers/controls/annotations/annotations-button-region-tooltips/PersistentOnboardingBoxEditAnnotationsTooltip.tsx
```diff
@@ -0,0 +1,103 @@
...
+            })}
+            data-testid="tooltipImage"
+        />
+    );
+
+    const primaryButtonProps = {
+        children: __('persistent_onboarding_box_edit_annotations_tooltip_button_next'),
+    };
+
+    const backButtonProps = {
```
jstoffan, line 80: Is this still needed? It seems like the `TargetedClickThroughGuideTooltip` component should handle it.

src/lib/viewers/controls/experiences/ExperiencesContext.ts
```diff
@@ -5,6 +5,6 @@ export type Context = {
     experiences: Experiences;
 };
 
-export default React.createContext<Context>({
-    experiences: {},
-});
+const ExperiencesContext = React.createContext<Context>({ experiences: {} });
```
jstoffan, line 8: Why was this changed?

src/lib/viewers/image/ImageBaseViewer.js
```diff
@@ -148,7 +148,7 @@ class ImageBaseViewer extends BaseViewer {
         const { enableAnnotationsImageDiscoverability, experiences = {} } = this.options;
         const canShow = Object.values(experiences).some(experience => experience.canShow);
 
-        return !canShow && !!enableAnnotationsImageDiscoverability;
+        return !canShow && Boolean(enableAnnotationsImageDiscoverability);
```
jstoffan, line 151: Please revert.

jstoffan (commented):
Let's scope the changes down to what's needed.

src/lib/viewers/controls/annotations/annotations-button-region-tooltips/AnnotationsButtonRegionTooltips.tsx
```diff
@@ -0,0 +1,26 @@
...
+    isEnabled?: boolean;
+}>;
+
+export default function AnnotationsButtonRegionTooltips({ children, isEnabled }: Props): JSX.Element {
+    const { experiences } = useContext(ExperiencesContext);
+
+    const { persistentOnboardingBoxEditAnnotations: experience } = experiences;
+    const canShow = experience ? experience.canShow : false;
+
+    if (canShow) {
```
megansmith-box, line 18: I'm concerned about this conditionally rendering tooltip, the useTargetingApi will display the tooltip accordingly based on the canShow property and they will never both be true at the same time if they are wrapped correctly in EUA in the makeUseShowOne hook. We've seen problems in the past were the methods in the targeting api aren't called correctly or the children component event listeners aren't called correctly when conditionally rendering tooltips. Can we instead leave the original `AnnotationsTargetedTooltip` as it is and have nested tooltips?

src/lib/viewers/controls/annotations/annotations-button-region-tooltips/PersistentOnboardingBoxEditAnnotationsTooltip.tsx
```diff
@@ -0,0 +1,103 @@
...
+
+export default function PersistentOnboardingBoxEditAnnotationsTooltip({
+    children,
+    isEnabled = false,
+}: Props): JSX.Element {
+    const { experiences } = useContext(ExperiencesContext);
+    const { setIsForced } = useContext(ControlsLayerContext);
+
+    const {
+        persistentOnboardingBoxEditAnnotations: experience = {
```
megansmith-box, line 30: Why do we need this default value?

src/lib/i18n.js
robertjk: This is already in that object (under `locale` property). The problem is that I don't know how to access this object in my React components. I see it declared in `BaseViewer.js`, but don't see it being used in any of React components.

src/lib/viewers/controls/annotations/AnnotationsTargetedTooltip.tsx
```diff
@@ -1,52 +0,0 @@
...
-import { ControlsLayerContext } from '../controls-layer';
-import { ExperiencesContext } from '../experiences';
-import { TargetingApi } from '../../../types';
-import './AnnotationsTargetedTooltip.scss';
-
-export type Props = React.PropsWithChildren<{
-    isEnabled?: boolean;
-}>;
-
-function AnnotationsTargetedTooltip({ children, isEnabled = false }: Props): JSX.Element | null {
```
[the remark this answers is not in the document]
robertjk: We could do without changing it, but I feel it's better to keep it as it is. When this component was defined there was only a single tooltip on Annotations. Now there are 2. IMO it makes sense to:
- Rename it, as not it's not the only Annotations-targeted tooltip anymore, so it's old name is misleading
- As now there are 2 tooltips, I created a wrapper `AnnotationsButtonRegionTooltips` component to switch between them. `AnnotationsTargetedTooltip` is used only locally by that new wrapper component. I moved it to the directory of that component because of that.

src/lib/viewers/controls/annotations/annotations-button-region-tooltips/AnnotationsButtonRegionTooltips.tsx
```diff
@@ -0,0 +1,26 @@
...
+
+export type Props = PropsWithChildren<{
+    isEnabled?: boolean;
+}>;
+
+export default function AnnotationsButtonRegionTooltips({ children, isEnabled }: Props): JSX.Element {
+    const { experiences } = useContext(ExperiencesContext);
+
+    const { persistentOnboardingBoxEditAnnotations: experience } = experiences;
+    const canShow = experience ? experience.canShow : false;
```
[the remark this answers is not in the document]
robertjk: Only because `experience` is possibly `undefined`. I assumed that when `box-content-preview` will be used outside of `EndUserApp` contexts, `previewExperiences` will be `undefined`, so I have to guard against that.

I saw 2 possible approaches:
- Define the `Experiences` type with possibly `undefined` properties and then guard against that in the components using them
- Define the `Experiences` type with required properties and declare some mock experiences at `ExperiencesProvider` level

I chose the first path. Let me know if you think this would be handled better somehow differently.

src/lib/viewers/controls/annotations/annotations-button-region-tooltips/AnnotationsButtonRegionTooltips.tsx
robertjk: Sure. I have always found named imports to be a cleaner pattern, but you're right that being consistent is even more important.

robertjk: Done. Resolving this.

src/lib/viewers/controls/annotations/annotations-button-region-tooltips/PersistentOnboardingBoxEditAnnotationsTooltip.scss
robertjk: Done. Resolving this.

src/lib/viewers/controls/annotations/annotations-button-region-tooltips/AnnotationsFlowTooltip.scss
robertjk: This connects to https://github.com/box/box-content-preview/pull/1445#discussion_r871310068 where I explained why I renamed this component.

src/lib/viewers/controls/annotations/annotations-button-region-tooltips/AnnotationsFlowTooltip.tsx
```diff
@@ -0,0 +1,65 @@
...
+export type Props = React.PropsWithChildren<{
+    isEnabled?: boolean;
+}>;
+
+export default function AnnotationsFlowTooltip({ children, isEnabled = false }: Props): JSX.Element | null {
+    const { experiences } = React.useContext(ExperiencesContext);
+    const { setIsForced } = React.useContext(ControlsLayerContext);
+
+    const {
+        tooltipFlowAnnotations: experience = {
```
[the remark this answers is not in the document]
robertjk: This is connected to the change I did to `Experiences` type. It used to be specified using index signatures as:

```typescript
export type Experiences = {
    [key: string]: TargetingApi;
};
```

That's not very type-safe, as it doesn't give any guarantee that `tooltipFlowAnnotationsExperience` exists. As we know the name of the experiences upfront I thought it's safer to list them explicitly:

```typescript
export interface Experiences {
    persistentOnboardingBoxEditAnnotations?: PersistentOnboardingTargetingApi;
    tooltipFlowAnnotations?: TargetingApi;
}
```

One more reason for doing that was that `persistentOnboardingBoxEditAnnotations` has a little different type, as it has extra `onPrevious` property.

src/lib/viewers/controls/annotations/annotations-button-region-tooltips/PersistentOnboardingBoxEditAnnotationsTooltip.scss
```diff
@@ -0,0 +1,39 @@
+@import '~box-ui-elements/es/styles/variables';
+
+// Size of the actual image file
+$image-file-width: 704px;
```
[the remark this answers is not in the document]
robertjk: You're right. I forgot Sass files declare variables in global scope as well. I'll definitely move them inside.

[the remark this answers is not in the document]
robertjk: Done. Resolving.

src/lib/viewers/controls/experiences/ExperiencesContext.ts
robertjk: Just didn't notice the change after coming back from trying other approaches here. I'll revert it.

robertjk: Done. Resolving.

src/lib/viewers/controls/annotations/annotations-button-region-tooltips/PersistentOnboardingBoxEditAnnotationsTooltip.tsx
robertjk: It is needed to declare the back button and bind the experience's `onPrevious()` callback to it.

src/lib/viewers/image/ImageBaseViewer.js
robertjk: Done. Resolving.

src/lib/viewers/controls/annotations/annotations-button-region-tooltips/PersistentOnboardingBoxEditAnnotationsTooltip.tsx
robertjk: Because I marked each experience as optional on `Experiences` type. Likely it wasn't a good decision. I explained it more here: https://github.com/box/box-content-preview/pull/1445#discussion_r871424732

src/lib/i18n.js
jstoffan: To clarify, I'm just asking that we move the export to the default:

```
export default const { createAnnotatorIntl, currentLocale };`

File: src/lib/viewers/controls/experiences/index.ts [hunk 21]
```diff
@@ -1,3 +1,2 @@
-export * from './ExperiencesProvider';
-export { default } from './ExperiencesProvider';
+export { default, Props as ExperiencesProviderProps } from './ExperiencesProvider';
```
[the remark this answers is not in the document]
megansmith-box: I see a warning
`"export 'Props' (reexported as 'ExperiencesProviderProps') was not found in './ExperiencesProvider'`

src/lib/viewers/controls/annotations/annotations-button-region-tooltips/AnnotationsButtonRegionTooltips.tsx
megansmith-box: @robertjk - I wonder if it's possible to use just one component and modify props? I find there is the double click issue with the annotations tooltip. I also notice another problem where `isDiscoverabilityEnabled` isn't being updated after the initial page load. Here are some videos to demonstrate the issues:

1. https://cloud.box.com/s/turwgqzglk2x2gbbo6c8syklt7zww3jv - you will notice that I click next on the preview open button tooltip before the control bar loads, therefore when it loads, canShow is true for the annotations tooltip, and `isDiscoverabilityEnabled` is properly returning false so there is no default selected. When I click on the annotations region button, it closes the tooltip but doesn't select the region button, therefore you need to click it twice (this is the bug that always happens when conditionally rendering tooltips like this). Therefore, I believe we need to use the same tooltip component and modify props if we want to avoid this issue. And also, we should update `isDiscoverabilityEnabled` to change when experiences canShow value changes, since this tooltip doesn't typically appear on page load.

2. https://cloud.box.com/s/vy87li6yoxqfy9qiz9v17xtb1qp3ib8j - you will notice if I wait for the control bar to load and I don't click next to view the annotations tooltip, the annotations region will be default selected. When I click next and the annotations tooltip appears, it doesn't change, even though canShow is now true. When I click on the button, the tooltip closes and it doesn't deselect. (In this case the bug isn't bad because we still have the annotations region selected) However, this is not the expected behavior.

src/lib/i18n.js
robertjk: Then I misunderstood. I'll do that then.

PR #1446, opened 2022-06-07
feat(annotations): Add annotations-only controls

ivanthai:
Added an annotations-only controls mode when selecting an annotations mode. This is to make more efficient use of the screen width real estate when viewing preview on mobile or embed experiences. 

**Implementation**
This feature is currently under a responsive-web feature flip which is controlled by EUA. This is why I chose to add it via configuration as to not force any changes onto any consumers of this library. This functionality can be toggled on by setting `enableAnnotationsOnlyControls` option to true when instantiating `new Preview()`

I also opted in to making most of the changes via CSS rather than via React props (which was my first attempt). This to me is a bit more elegant and cleaner as it doesn't require massive changes to the React components. 

Also open to any changes to config naming

**Design**

https://zpl.io/dxDDLEp
image

**Demo**
* New "Exit Annotations" button

**DocViewer**

annotation-only-controls-mobile-demo-final

**ImageViewer**

annotations-only-controls-image-viewer-demo

**With old "Exit Annotation" Button**

annotation-only-mode-demo

Keyboard navigation - a11y
annotations-only-controls-a11y-keyboard

### DIFF
--- a/src/lib/constants.js
+++ b/src/lib/constants.js
@@ -3,6 +3,7 @@ export const CLASS_ANNOTATIONS_CREATE_DRAWING = 'bp-annotations-create--drawing'
 export const CLASS_ANNOTATIONS_CREATE_HIGHLIGHT = 'bp-annotations-create--highlight';
 export const CLASS_ANNOTATIONS_CREATE_REGION = 'bp-annotations-create--region';
 export const CLASS_ANNOTATIONS_DISCOVERABLE = 'bp-annotations-discoverable';
+export const CLASS_ANNOTATIONS_ONLY_CONTROLS = 'bp-annotations-only-controls';
 export const CLASS_ANNOTATIONS_IMAGE_FTUX_CURSOR_SEEN = 'bp-annotations-ftux-image-cursor-seen';
 export const CLASS_ANNOTATIONS_DOCUMENT_FTUX_CURSOR_SEEN = 'bp-annotations-ftux-document-cursor-seen';
 export const CLASS_NAVIGATION_VISIBILITY = 'bp-is-navigation-visible';
--- a/src/lib/viewers/BaseViewer.js
+++ b/src/lib/viewers/BaseViewer.js
@@ -24,6 +24,7 @@ import {
     CLASS_ANNOTATIONS_CREATE_HIGHLIGHT,
     CLASS_ANNOTATIONS_CREATE_REGION,
     CLASS_ANNOTATIONS_DISCOVERABLE,
+    CLASS_ANNOTATIONS_ONLY_CONTROLS,
     CLASS_BOX_PREVIEW_MOBILE,
     FILE_OPTION_START,
     SELECTOR_BOX_PREVIEW_BTN_ANNOTATE_DRAW,
@@ -227,6 +228,10 @@ class BaseViewer extends EventEmitter {
             this.containerEl.classList.add(CLASS_ANNOTATIONS_DISCOVERABLE);
         }
 
+        if (this.options.enableAnnotationsOnlyControls && this.containerEl) {
+            this.containerEl.classList.add(CLASS_ANNOTATIONS_ONLY_CONTROLS);
+        }
+
         this.isSetup = true;
     }
 
@@ -273,6 +278,7 @@ class BaseViewer extends EventEmitter {
             this.containerEl.removeEventListener('contextmenu', this.preventDefault);
             this.containerEl.innerHTML = '';
             this.containerEl.classList.remove(CLASS_ANNOTATIONS_DISCOVERABLE);
+            this.containerEl.classList.remove(CLASS_ANNOTATIONS_ONLY_CONTROLS);
         }
 
         // Destroy the annotator
--- a/src/lib/viewers/controls/annotations/AnnotationsButton.tsx
+++ b/src/lib/viewers/controls/annotations/AnnotationsButton.tsx
@@ -13,10 +13,9 @@ export type Props = Omit<React.ButtonHTMLAttributes<HTMLButtonElement>, 'onClick
     onClick?: (mode: AnnotationMode) => void;
 };
 
-function AnnotationsButton(
-    { children, className, isActive = false, isEnabled = true, mode, onClick = noop, ...rest }: Props,
-    ref: React.Ref<HTMLButtonElement>,
-): JSX.Element | null {
+function AnnotationsButton(props: Props, ref: React.Ref<HTMLButtonElement>): JSX.Element | null {
+    const { children, className, isActive = false, isEnabled = true, mode, onClick = noop, ...rest } = props;
+
     if (!isEnabled) {
         return null;
     }
--- a/src/lib/viewers/controls/annotations/AnnotationsButton.tsx
+++ b/src/lib/viewers/controls/annotations/AnnotationsButton.tsx
@@ -13,21 +13,17 @@ export type Props = Omit<React.ButtonHTMLAttributes<HTMLButtonElement>, 'onClick
     onClick?: (mode: AnnotationMode) => void;
 };
 
-export default function AnnotationsButton({
-    children,
-    className,
-    isActive = false,
-    isEnabled = true,
-    mode,
-    onClick = noop,
-    ...rest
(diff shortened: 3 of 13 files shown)


src/lib/viewers/controls/annotations/AnnotationsControls.tsx
```diff
@@ -70,6 +70,16 @@ export default function AnnotationsControls({
...
         <div className="bp-AnnotationsControls">
+            <AnnotationsButton
+                className="bp-AnnotationsControls-exitBtn"
+                data-resin-target="exit"
+                data-testid="bp-AnnotationsControls-exitBtn"
+                mode={AnnotationMode.NONE}
+                onClick={handleModeClick}
+                title={__('exit_annotations')}
+            >
+                Exit
```
jstoffan, line 81: We'll want to translate this string, as well.

jstoffan (commented):
@ivanthai, this seems like a nice approach; the simplicity is great. How is focus state handled when the control groups are shown/hidden, such as for keyboard users?

src/lib/constants.js
```diff
@@ -3,6 +3,7 @@ export const CLASS_ANNOTATIONS_CREATE_DRAWING = 'bp-annotations-create--drawing'
 export const CLASS_ANNOTATIONS_CREATE_HIGHLIGHT = 'bp-annotations-create--highlight';
 export const CLASS_ANNOTATIONS_CREATE_REGION = 'bp-annotations-create--region';
 export const CLASS_ANNOTATIONS_DISCOVERABLE = 'bp-annotations-discoverable';
+export const CLASS_ANNOTATIONS_ONLY_CONTROLS_ENABLED = 'bp-annotations-only-controls--enabled';
```
ConradJChan, line 6: There seems to be a little bit of a mismatch between the classname style (SUIT modifier) and how it's named. `enabled` seems to indicate the state of the system, that this feature is on, but it gets toggled on and off as and when the mode changes. 

Maybe we want either `bp-annotations-only-controls` to be set once and we adjust the styles as an when other recognized annotation modes classes are added or we toggle a `.bp-annotations-only-controls.is-shown` state class to hide and show. Thoughts?

src/lib/viewers/doc/_docBase.scss
```diff
@@ -400,6 +400,16 @@ $thumbnail-sidebar-width: 191px; // Extra pixel to account for sidebar border
         }
     }
 
+    &.bp-annotations-only-controls--enabled {
```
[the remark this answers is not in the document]
ivanthai: No. _docBase is included in the Document and Presentation scss files

src/lib/constants.js
ivanthai: I like this idea better.

src/lib/viewers/controls/annotations/AnnotationsControls.tsx
ivanthai: I've convinced design to adopt an icon instead. We'll just need to translate the tooltip/label

ivanthai:
I've added a gif to illustrate the keyboard navigation behavior in the description above

src/lib/viewers/doc/_docBase.scss
```diff
@@ -400,6 +400,26 @@ $thumbnail-sidebar-width: 191px; // Extra pixel to account for sidebar border
...
+            // Show exit button when annotation-only controls is active
+            .bp-AnnotationsControls-exitBtn {
+                display: block;
+            }
+
+            // Hide all ControlBarGroups except for last instance which contains AnnotationsControls
+            // and hide all direct children of the group that isn't AnnotationControls
+            .bp-ControlsBarGroup {
+                &:not(:last-child),
+                & > *:not(.bp-AnnotationsControls) {
```
jstoffan, line 416: We generally want to avoid `*` selectors and complex `:not` selectors, but I'm coming up empty on alternative solutions that don't involve more changes to the other controls. I don't think we need the `&` here, though.

ivanthai: Alternatively, we could go back to my previous implementation where we use new class names to be a bit more explicit

ivanthai:
I've updated the description with new gifs
- DocViewer
- ImageViewer

src/lib/viewers/doc/_docBase.scss
jstoffan: Yeah, why did we move away from below? Was it not covering all the use cases?

```
.bp-ControlsBarGroup:not(.bp-ControlsBarGroup-annotations) {
    display: none;
}

File: src/lib/viewers/doc/_docBase.scss (hunk 3 above)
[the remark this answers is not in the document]
Comment (ConradJChan) [reply src/lib/viewers/doc/_docBase.scss]: I was trying to wrack my brain to figure out why it works for Image viewers but it turns out all of box-content-preview stylesheets get compiled into a single preview.css which is why the `docBase` styles here end up applying to the image viewer as well 😅

File: src/lib/viewers/controls/annotations/AnnotationsButton.tsx [hunk 5]
```diff
@@ -13,21 +13,17 @@ export type Props = Omit<React.ButtonHTMLAttributes<HTMLButtonElement>, 'onClick
...
-    children,
-    className,
-    isActive = false,
-    isEnabled = true,
-    mode,
-    onClick = noop,
-    ...rest
-}: Props): JSX.Element | null {
+function AnnotationsButton(
+    { children, className, isActive = false, isEnabled = true, mode, onClick = noop, ...rest }: Props,
```
jstoffan, line 17: Super minor, but we generally use this function signature for forwarding refs, since it can get awkward:

```
function AnnotationsButton(props: Props, ref: React.Ref<HTMLButtonElement>) {
    const { children, className, isActive = false, isEnabled = true, mode, onClick = noop, ...rest } = props;
    // ...etc
}

File: src/lib/viewers/controls/annotations/AnnotationsControls.tsx [hunk 6]
```diff
@@ -33,12 +34,27 @@ export default function AnnotationsControls({
     const showDrawing = !isFullscreen && hasDrawing;
     const showHighlight = !isFullscreen && hasHighlight;
     const showRegion = !isFullscreen && hasRegion;
+    const annotationBtnRefs = {
```
jstoffan, line 37: Minor, but can we alphabetize the variables here? We can also skip the abbreviation via `buttonRefs` or something.

src/lib/viewers/BaseViewer.js
```diff
@@ -273,6 +278,7 @@ class BaseViewer extends EventEmitter {
             this.containerEl.removeEventListener('contextmenu', this.preventDefault);
             this.containerEl.innerHTML = '';
             this.containerEl.classList.remove(CLASS_ANNOTATIONS_DISCOVERABLE);
+            this.containerEl.classList.remove(CLASS_ANNOTATIONS_ONLY_CONTROLS);
```
jstoffan, line 281: Can we add an assertion for this logic?

jstoffan (approved):
LGTM with minor feedback.

src/lib/viewers/doc/_docBase.scss
ivanthai: There was some styling that wasn't ideal and that I had to add additional definitions to cover. I tried to find a way to not have to do that

src/lib/viewers/controls/annotations/__tests__/AnnotationsControls-test.tsx
```diff
@@ -107,6 +107,32 @@ describe('AnnotationsControls', () => {
...
+            ({ current, selector }) => {
+                const props = {
+                    annotationMode: current,
+                    hasDrawing: true,
+                    hasHighlight: true,
+                    hasRegion: true,
+                };
+                const parentEl = document.createElement('div');
+                document.body.appendChild(parentEl);
+                const wrapper = mount(<AnnotationsControls {...props} />, { attachTo: parentEl });
```
jstoffan, line 127: Can we use `getWrapper` here like for the other tests?

jstoffan (approved):

jstoffan (approved):

PR #1448, opened 2022-06-11
feat(annotations): responsive media queries for annotations-only ctrls

ivanthai:
moved media query definitions to Preview. 

annotations-only-controls-media-query

PR #1449, opened 2022-06-11
feat(thumbnails): remove thumbnail toggle on small modalities

ivanthai:
Note: Please approve https://github.com/box/box-content-preview/pull/1448 first

This is hidden on smaller modalities in order to prioritize other functions

https://zpl.io/dxDDLEp

preview-controls-hide-thumbnail-toggle

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,7 +38,7 @@
         "babel-loader": "^8.0.6",
         "babel-plugin-transform-require-ignore": "^0.1.1",
         "box-annotations": "^2.3.0",
-        "box-ui-elements": "15.0.0-beta.50",
+        "box-ui-elements": "15.1.0-beta.21",
         "chai": "^4.2.0",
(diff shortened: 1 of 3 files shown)


jstoffan (changes requested):
@ivanthai, we have [styling](https://github.com/box/box-content-preview/blob/master/src/lib/viewers/doc/_docBase.scss#L141) that handles responsive behavior for this element today. We'll need to make sure this new code is compatible.

ivanthai:
Seems like it would make sense to convert the other thumbnail component CSS to the new breakpoints...probably should define that breakpoint number in BUIE

jstoffan:
@ivanthai, are we planning to unify the breakpoints? I think we'll run into some odd behaviors otherwise.

jstoffan (approved):

PR #1451, opened 2022-06-14
feat(annotations): remove zoom ctrls on mobile

ivanthai:
Remove zoom controls on touch devices with small-medium screen widths (<767px). Touch devices in this case refer to mobile. However, mobile seems more like a subset of touch devices to me. Perhaps this is an edge case we can ignore. 

https://zpl.io/dxDDLEp

image

annotations-remove-zoom-mobile


bfoxx1906 (approved):
looks good

PR #1452, opened 2022-06-14
feat(annotations): hide fullscreen toggle if browser API not supported

ivanthai:
[Fullscreen API](https://developer.mozilla.org/en-US/docs/Web/API/Fullscreen_API) is not supported on all browsers, most notably iOS browsers (e.g. chrome, firefox, safari, etc) on iPhone. The API's are supported on iPad however.

https://zpl.io/dxDDLEp

image

**Desktop**

Screen Shot 2022-06-14 at 1 59 45 PM

**iPhone**

Screen Shot 2022-06-14 at 2 00 19 PM

**iPad**
Fullscreen API are supported on iOS iPad

Screen Shot 2022-06-14 at 2 09 29 PM
Screen Shot 2022-06-14 at 2 09 34 PM

### DIFF
--- a/src/lib/constants.js
+++ b/src/lib/constants.js
@@ -14,6 +14,7 @@ export const CLASS_BOX_PREVIEW_BUTTON = 'bp-btn';
 export const CLASS_BOX_PREVIEW_CONTAINER = 'bp-container';
 export const CLASS_BOX_PREVIEW_CONTENT = 'bp-content';
 export const CLASS_BOX_PREVIEW_FIND_BAR = 'bp-find-bar';
+export const CLASS_BOX_PREVIEW_FULLSCREEN = 'bp-has-fullscreen';
 export const CLASS_BOX_PREVIEW_HAS_HEADER = 'bp-has-header';
 export const CLASS_BOX_PREVIEW_HAS_NAVIGATION = 'bp-has-navigation';
 export const CLASS_BOX_PREVIEW_HEADER = 'bp-header';
--- a/src/lib/viewers/BaseViewer.js
+++ b/src/lib/viewers/BaseViewer.js
@@ -25,6 +25,7 @@ import {
     CLASS_ANNOTATIONS_CREATE_REGION,
     CLASS_ANNOTATIONS_DISCOVERABLE,
     CLASS_ANNOTATIONS_ONLY_CONTROLS,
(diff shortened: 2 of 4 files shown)


jstoffan (changes requested):
@ivanthai, we have logic to handle fullscreen on iOS devices, despite its lack of support for the official API. This change will likely be seen as a defect.

jstoffan:
If we want to remove this feature, we should remove all the supporting logic for the existing iOS experience from the codebase. What effect does this change have on our image and video viewers?

ivanthai:
I'll discuss with design. This was only done, less because of reducing width use, but because the existing pseudo functionality didn't appear to be practical as all it appeared to do was fit-to-screen.

ivanthai:
Looking over the code, it seems like this pseudo fullscreen functionality wasn't really working. As far as I can tell...the CSS class selector did not match the structure and consequently did not apply the styling. 

The [following](https://github.com/ivanthai/box-content-preview/blob/master/src/lib/Fullscreen.js#L102-L102) applied the CSS class name to the `.bp-content` component instead of the `.bp` component. 
```
// Fullscreen.js
fullscreenEnterHandler(el) {
        this.fullscreenElement = el;
        this.fullscreenElement.classList.add(CLASS_FULLSCREEN);
        this.fullscreenElement.focus();

        if (!this.isSupported()) {
            this.fullscreenElement.classList.add(CLASS_FULLSCREEN_UNSUPPORTED);
        }

        this.emit('enter');
    }
```

And the following CSS selector applies to `.bp .bp-fullscreen-unsupported` which never happens
```
// _common.scss
.bp {
    ...

    // this never gets applied
    &.bp-fullscreen-unsupported {
        position: fixed;
        top: 0;
        right: 0;
        bottom: 0;
        left: 0;
    }

}
```

So you get something like this which just appears to just do nothing other than resetting the zoom (fit-to-zoom).

annotations-fullscreen-bug-safari-ios

jstoffan:
Understood. We may have changed the behavior when we converted the controls to React. That said, if we're going to remove the behavior, let's remove all of its supporting code. What happens in our image and media viewers?

ivanthai:
I'll double check them. But the question is if we should fix this pseudo-full screen functionality or remove them

ivanthai:
Closing as this is not needed. Superceded by https://github.com/box/box-content-preview/pull/1453

PR #1453, opened 2022-06-16
fix(fullscreen): fix fullscreen mode for ios/iphone combo

ivanthai:
Fixed an issue with the pseudo-fullscreen mode on ios iphone. Fullscreen APIs are not supported on that os+device combo. The pseudo-fullscreen mode broke some time ago and was no longer usable

The Fullscreen class adds "bp-fullscreen-unsupported" class and the corresponding CSS that applied the fullscreen styling did not select the correct element. Additionally, it would also appear under elements such as the header and sidebar elements, so I had to bump the z-index up

iOS iPhone psuedo-fullscreen mode bug
annotations-fullscreen-bug-safari-ios

Fixed
preview-fullscreen-ios-fix

This supercedes https://github.com/box/box-content-preview/pull/1452 which hides the Fullscreen node toggle for unsupported devices

### DIFF
--- a/src/lib/_common.scss
+++ b/src/lib/_common.scss
@@ -1,3 +1,4 @@
+@import '~box-ui-elements/es/styles/variables';
 @import 'boxui';
 
 $header-height: 48px;
@@ -125,15 +126,6 @@ $header-height: 48px;
     outline: none;
     -webkit-overflow-scrolling: touch;
 
-    // Pseudo fullscreen for iOS Safari which doesn't support the fullscreen API
-    &.bp-fullscreen-unsupported {
-        position: fixed;
-        top: 0;
-        right: 0;
-        bottom: 0;
-        left: 0;
-    }
-
     &.bp-dark {
         background-color: $black;
     }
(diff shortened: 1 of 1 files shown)


jstoffan (approved):
Thanks for catching this, @ivanthai. Did we test the image/video viewers?

src/lib/_common.scss
```diff
@@ -148,6 +140,12 @@ $header-height: 48px;
         flex: 1 1 auto;
         align-items: center;
         outline: none;
+
+        // Pseudo fullscreen for iOS Safari which doesn't support the fullscreen API
+        &.bp-fullscreen-unsupported {
+            position: fixed;
+            z-index: $overlay-z-index;
```
ivanthai, line 147: set z-index to the highest z-index standard defined in BUIE which is 190 defined for overlays. This acts like an overlay when in full-screen mode

ivanthai:
I forgot to mention. Yes to image and kind of to video. I wasn't able to play videos on the iphone hitting my devvm for some reason. Seemed like it was specific to iphones with devvm. Without figuring out why that was the case, I instead tested it by forcing the pseudo-fullscreen mode on desktop.

PR #1455, opened 2022-06-21
feat(content-insights): Added ACI support to track per page time spent

marcoartaviaq:
Added support to detect page-changing events on documents viewed with the PDF.js. Also,  I added a new class 'PageTracker' that sends data to a new endpoint with the page and time spent.

This feature will be available under a feature flip and will be managed by the EUA by setting the enableAdvancedContentInsights option when creating a new Preview instance.

The UnitTests are still missing because the idea is to merge the PoC code into the main branch without breaking anything, so it would need  regression testing.


src/lib/PageTracker.js
```diff
@@ -0,0 +1,310 @@
...
+
+    /** @property {Date} -  */
+    startTime = null;
+
+    /** @property {Object} - Reference to content insights object */
+    contentInsightsProps = {
+        thumbnailImageCache: new BoundedCache(),
+    };
+
+    contentInsightsOptions = {
```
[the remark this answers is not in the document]
marcoartaviaq: My bad.... those variables aren't used anymore. I remove them from the file

src/lib/PageTracker.js
```diff
@@ -0,0 +1,310 @@
...
+     * @param {Object} data - Data Object to set the current page and file length
+     * @return {void}
+     */
+    initContentInsightsLog(data) {
+        if (data.currentPage) {
+            this.currentPage = data.currentPage;
+            this.fileLength = data.numPages;
+        }
+        this.setIdleTimer();
+        this.reportingStarted = true;
```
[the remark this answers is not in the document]
marcoartaviaq: In a prior version `reportingStarted` was neccesary as another control flag, but I refactored it to not use it anymore

src/lib/PageTracker.js
```diff
@@ -0,0 +1,310 @@
...
+    startContentInsightsInterval() {
+        if (!this.reportingStarted) {
+            return;
+        }
+        if (this.reportingInterval) {
+            this.stopContentInsightsInterval();
+        }
+        this.startTime = Date.now();
+
+        this.reportingInterval = setInterval(() => {
```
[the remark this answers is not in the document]
marcoartaviaq: With our experience, the page unload events weren't always reliable to send the data back to an API. We tried looking up also with websockets but they were failing too. In the case of sendBeacon, a year ago we read some articles that this new Api is not recommended because sometimes it fails to fire the events, so we stick with the setInterval method. In case we will stuck in a weird situation where executions start backing up, I think it won't be a big Issue because we don't necessary rely on the events order.

src/lib/Preview.js
```diff
@@ -1245,6 +1267,9 @@ class Preview extends EventEmitter {
         this.viewer.addListener('error', this.triggerError);
         this.viewer.addListener(VIEWER_EVENT.default, this.handleViewerEvents);
         this.viewer.addListener(VIEWER_EVENT.metric, this.handleViewerMetrics);
+        if (this.options.enableAdvancedContentInsights) {
```
ConradJChan, line 1270: There's already a handler for the various viewer events `handleViewerEvents`, is there a reason to separate it out here?

src/lib/Preview.js
```diff
@@ -1360,6 +1398,9 @@ class Preview extends EventEmitter {
 
             // If there wasn't an error and event logging is not disabled, use Events API to log a preview
             if (!this.options.disableEventLog) {
+                if (this.options.enableAdvancedContentInsights) {
```
ConradJChan, line 1401: Does this need to be tied to `disableEventLog`?

src/lib/Preview.js
```diff
@@ -1402,8 +1443,17 @@ class Preview extends EventEmitter {
                 id: fileId,
             },
         };
+
         const headers = getHeaders({}, token, sharedLink, sharedLinkPassword);
 
+        if (this.options.enableAdvancedContentInsights) {
+            this.pageTracker.init(this.previewOptions.contentInsightsOptions, { apiHost, headers }, this.file);
```
ConradJChan, line 1450: Should this move into the `finishLoading` method in Preview.js?

src/lib/Preview.js
```diff
@@ -1402,8 +1443,17 @@ class Preview extends EventEmitter {
...
+            data.additional_information = {
+                view_session: {
```
ConradJChan, line 1452: Does this require a schema change in the analytics table?

src/lib/Preview.js
```diff
@@ -1802,6 +1852,19 @@ class Preview extends EventEmitter {
...
     }
 
+    /**
+     * Updates advanced content insights options after props have changed in parent app
+     *
+     * @public
+     * @param {Object} contentInsightsOptions - new content insights options value
+     * @return {void}
+     */
+    updateContentInsightsOptions(contentInsightsOptions) {
```
ConradJChan, line 1862: The naming of the argument seems redundant, maybe just `options`?

src/lib/Preview.js
```diff
@@ -1802,6 +1852,19 @@ class Preview extends EventEmitter {
...
+        if (this.options.enableAdvancedContentInsights) {
+            this.previewOptions.contentInsightsOptions = contentInsightsOptions;
```
ConradJChan, line 1864: Instead of mutating the object, should we create a new object?
```
this.previewOptions = { ...this.previewOptions, contentInsights: options };
```
Also should we name the options related to content insights as simply `contentInsights` within the `previewOptions`?

src/lib/Preview.js
```diff
@@ -1943,6 +2006,19 @@ class Preview extends EventEmitter {
...
         );
     };
+
+    /**
+     * Get a thumbnail image element
+     *
+     * @param {number} pageNumber - the page number
+     * @return {Promise} - promise resolves with the image HTMLElement or null if generation is in progress
+     */
+    getThumbnail(pageNumber) {
```
ConradJChan, line 2016: Is this method currently used anywhere?

src/lib/Preview.js
```diff
@@ -1943,6 +2006,19 @@ class Preview extends EventEmitter {
...
+        if (this.viewer.getThumbnail) {
+            return this.viewer.getThumbnail(pageNumber);
+        }
+        return null;
```
ConradJChan, line 2020: Should this be wrapped in a `Promise.resolve(null)`?

src/lib/viewers/doc/DocBaseViewer.js
```diff
@@ -1699,6 +1705,23 @@ class DocBaseViewer extends BaseViewer {
...
+     * Get a thumbnail image element
+     *
+     * @param {number} pageNumber - the page number
+     * @return {Promise} - promise resolves with the image HTMLElement or null if generation is in progress
+     */
+    getThumbnail(pageNumber) {
+        if (!this.thumbnailsSidebar) {
+            this.thumbnailsSidebar = new ThumbnailsSidebar(this.thumbnailsSidebarEl, this.pdfViewer);
+        }
+        return this.thumbnailsSidebar.createThumbnailImage(pageNumber - 1, {
```
ConradJChan, line 1719: Should we instead make `createThumbnailImage` a static method on the `ThumbnailsSidebar` class?

src/lib/PageTracker.js
```diff
@@ -0,0 +1,347 @@
...
+            this.container.removeEventListener('click', this.throttledActiveUserHandler);
+            this.container.removeEventListener('wheel', this.throttledActiveUserHandler);
+            this.container.removeEventListener('keydown', this.throttledActiveUserHandler);
+            this.container.removeEventListener('touchstart', this.throttledActiveUserHandler);
+            this.container.removeEventListener('touchmove', this.throttledActiveUserHandler);
+        }
+    }
+
+    /**
+     *  Construct and send the payload object to send to the BFF endpoint.
```
ConradJChan, line 290: Since this is a OSS library, we don't like to expose what our proprietary endpoints are, could this follow the existing pattern of emitting an event that is consumed by a parent application?

src/lib/PageTracker.js
```diff
@@ -0,0 +1,347 @@
...
+
+            const payload = {
+                fileEvents: [
+                    {
+                        events,
+                        fileId: this.file.id,
+                        fileLength: this.fileLength,
+                        fileOwnerEid: this.fileOwnerEid,
+                        fileType: this.fileType,
+                        fileVersionId: this.file.file_version ? this.file.file_version.id : '',
```
ConradJChan, line 321: Could we use destructuring to avoid the repetitive `this.file.*`?

src/lib/PageTracker.js
```diff
@@ -0,0 +1,347 @@
...
+     *
+     * @public
+     * @return {void}
+     */
+    startTracking() {
+        this.stopTracking();
+        this.startTime = Date.now();
+
+        this.reportingIntervalId = setInterval(() => {
+            this.reportContentInsight();
```
ConradJChan, line 152: So if I'm reading this correctly, if the user stays on a page and is active (does something every 30s) this page tracker will continue to emit a report saying the user is on the page for 5s, on the page for 5s, on the page for 5s, etc...?

src/lib/PageTracker.js
```diff
@@ -0,0 +1,347 @@
...
+    }
+
+    /**
+     *  Set current page data and length from the event payload
+     *
+     * @private
+     * @param {Object} eventData - Event payload
+     * @return {void}
+     */
+    setCurrentPageData(eventData) {
```
ConradJChan, line 341: We could destructure in the method argument
```
setCurrentPageData({ currentPage = 0, numPages = 0 })
```

src/lib/Preview.js
marcoartaviaq: As per our requirements yes, because we won’t report the ACI if the access-stats are disabled, because we need to correlate the data.

src/lib/Preview.js
marcoartaviaq: Similar to the disableEventLog question, the initialization of the tracker is tied to the event log, so I would prefer to keep it here, but I could move it to the finishLoading if you think its better

src/lib/Preview.js
marcoartaviaq: Yes, because we need to correlate data from ACI to Access Stats. The monolith changes and schema will be updated as well

src/lib/Preview.js
marcoartaviaq: It would be called by BUIE to get a Thumb representation from a page

src/lib/viewers/doc/DocBaseViewer.js
marcoartaviaq: I think it’s better to not make it static. We need to show page thumbnails on the CI modal, and to accomplish this, we need to use PDFJS to fetch those. We still need to get proper thumbnail representation (WiP). I used the ThumbnailSidebar to avoid repeating the code for fetching a thumb, so my approach is to not change this class a lot because it is being used as a workaround at this moment. (Changing the createThumbnailImage to a static method would need other methods to be updated as well)

src/lib/PageTracker.js
```diff
@@ -0,0 +1,347 @@
...
+
+    /**
+     * Start the content insights interval, to report data each number of seconds defined on a Const
+     *
+     * @public
+     * @return {void}
+     */
+    startTracking() {
+        this.stopTracking();
+        this.startTime = Date.now();
```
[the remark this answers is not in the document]
marcoartaviaq: On ACI, the time spent viewing an asset shouldn’t need to be that precise. For an user, it’s not an issue if a viewer spent 20 seconds or 20.5 seconds on a page.

src/lib/PageTracker.js
marcoartaviaq: Yes. ACI tracking is an optional feature (will be turned off for a file as default), so we won’t have a lot of files sending these requests. In our previous experience, we found that this approach was the best way to report these insights, because using browser-events to detect page unload or using the beacon API weren’t always reliable. Also when we looked how people accomplish this type of reporting, we found that they were doing the same back then.

src/lib/PageTracker.js
```diff
@@ -0,0 +1,410 @@
+import uuidv4 from 'uuid/v4';
+import throttle from 'lodash/throttle';
+import EventEmitter from 'events';
+import { REPORT_ACI } from './events';
+
+export const REPORTING_INTERVAL = 5000; // number of seconds elapsed between each content insights reporting
```
jstoffan, line 6: Seconds or milliseconds?

src/lib/PageTracker.js
```diff
@@ -0,0 +1,410 @@
+import uuidv4 from 'uuid/v4';
```
jstoffan, line 1: I think this is a new dependency and is already outdated. Do we need it or can we reuse the `uuidv4` utility in `lib/logUtils`?

src/lib/PageTracker.js
```diff
@@ -0,0 +1,410 @@
...
+import { REPORT_ACI } from './events';
+
+export const REPORTING_INTERVAL = 5000; // number of seconds elapsed between each content insights reporting
+export const REPORT_TRESHOLD = 1500;
+export const IDLE_TIMER = 30000;
+const ACTIVE_USER_THROTTLE = 500;
+
+class PageTracker extends EventEmitter {
+    /** @property {boolean} - Whether Advanced Content Insights is on for the file that we are tracking */
+    active = false;
```
jstoffan, line 13: We typically use `is` or `has` for boolean variables, so `isActive`.

src/lib/PageTracker.js
```diff
@@ -0,0 +1,410 @@
...
+    events = [];
+
+    /** @property {Object} - Current file being previewed */
+    file = {};
+
+    /** @property {number} - File Length. On Document file types will be the number of pages. */
+    fileLength;
+
+    /** @property {string} - File Owner Enterprise ID. */
+    fileOwnerEId = '';
```
jstoffan, line 31: Should this be undefined by default? Same for userEId?

src/lib/PageTracker.js
```diff
@@ -0,0 +1,410 @@
...
+
+    /**
+     * Call the interval starting function and set the idle timer
+     *
+     * @public
+     * @return {void}
+     */
+    init() {
+        // Generate a new session id. This would need to be later changed to be generated on server side
+        if (!this.sessionId) {
```
jstoffan, line 109: Preview is often destroyed/recreated frequently based on user actions. Does this session id persist adequately or does it need to be cached somehow?

src/lib/Preview.js
```diff
@@ -877,6 +891,10 @@ class Preview extends EventEmitter {
         // Start the preview duration timer when the user starts to perceive preview's load
         const previewDurationTag = Timer.createTag(this.file.id, DURATION_METRIC);
         Timer.start(previewDurationTag);
+
+        if (this.pageTracker) {
```
jstoffan, line 895: The page tracking logic isn't really UI. Can we find a better home for our analytics setup/teardown?

src/lib/PageTracker.js
```diff
@@ -0,0 +1,410 @@
...
+    }
+
+    /**
+     *  Update content insights options
+     *
+     * @private
+     * @param {Object} options - Content Insights options
+     * @return {void}
+     */
+    updateOptions(options = {}) {
```
jstoffan, line 394: This seems similar to `init`. When would we need to update the options without reinitializing?

src/lib/PageTracker.js
```diff
@@ -0,0 +1,410 @@
...
+     *  Set content insights options
+     *
+     * @private
+     * @param {Object} options - Content Insights options
+     * @return {void}
+     */
+    setOptions(options) {
+        // Set User and file information
+        ({
+            ownerEId: this.fileOwnerEId = '',
```
jstoffan, line 380: This is an unusual pattern in our codebases. Can we destructure `options` and assign the values explicitly?

```
setOptions({ ownerEId }) {
    this.ownerEId = ownerEId;
}

File: src/lib/PageTracker.js [hunk 26]
```diff
@@ -0,0 +1,410 @@
...
+
+        // Report only time with a minimum threshold for this moment. This is to avoid a situation that the
+        // user is just spamming the scroll event to change pages and not send a lot of requests.
+        // Ideally, we can just batch that data and send it later...
+        if (timeToBeReported > REPORT_TRESHOLD) {
+            const { id: fileId, file_version: { id: fileVersionId } = null } = this.file;
+            // Add the current event to the events Array.
+            this.events.push({
+                userId: Number(this.userId),
+                timestamp: pageViewUTCTimestamp,
```
jstoffan, line 299: Did we resolve the question around what timecode format we wanted to use?

src/lib/PageTracker.js
```diff
@@ -0,0 +1,410 @@
...
+     * @return {void}
+     */
+    report() {
+        const pageViewUTCTimestamp = Date.now();
+        const timeToBeReported = pageViewUTCTimestamp - this.startTime;
+
+        // Report only time with a minimum threshold for this moment. This is to avoid a situation that the
+        // user is just spamming the scroll event to change pages and not send a lot of requests.
+        // Ideally, we can just batch that data and send it later...
+        if (timeToBeReported > REPORT_TRESHOLD) {
```
jstoffan, line 294: Can we make this a guard clause?

src/lib/PageTracker.js
```diff
@@ -0,0 +1,410 @@
...
+        this.stopTracking();
+    }
+
+    /**
+     * Binds DOM listeners to a container to manage if the user goes active or inactive.
+     *
+     * @private
+     * @return {void}
+     */
+    bindIdleUserEvents() {
```
jstoffan, line 241: Do we need to add a listener for tab visibility changes?

src/lib/PageTracker.js
```diff
@@ -0,0 +1,410 @@
...
+        this.setIdleTimer();
+    }
+
+    /**
+     * Call the resetIdleTimer if the user is active
+     *
+     * @private
+     * @return {void}
+     */
+    activeUserHandler() {
```
jstoffan, line 221: Minor, but our convention is to use "handle" (e.g. `handleUserActive`) for internal event handlers.

src/lib/PageTracker.js
```diff
@@ -0,0 +1,410 @@
...
+
+    /**
+     * Handle the page change event. At this moment, it would stop the current timer interval,
+     * report the data for the previous page, and start the interval again
+     *
+     * @public
+     * @param {Object} eventData - Event data payload received by a Page change event
+     * @return {void}
+     */
+    handleViewerPageChange(eventData) {
```
jstoffan, line 169: We're duplicating a fair bit of state by tracking the current and total pages here as well as the viewer. Can we pass in those values from the viewer event handlers, instead?

src/lib/viewers/doc/DocBaseViewer.js
```diff
@@ -1699,6 +1705,24 @@ class DocBaseViewer extends BaseViewer {
...
+     * @param {number} pageNumber - the page number
+     * @return {Promise} - promise resolves with the image HTMLElement or null if generation is in progress
+     */
+    getThumbnail(pageNumber) {
+        if (!this.thumbnailsSidebar) {
+            this.thumbnailsSidebar = new ThumbnailsSidebar(this.thumbnailsSidebarEl, this.pdfViewer);
+        }
+
+        return this.thumbnailsSidebar.createThumbnailImage(pageNumber - 1, {
+            cacheKey: `single_thumb_${pageNumber}`,
```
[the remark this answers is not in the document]
marcoartaviaq: Yes, they don't share the instance cache.

src/lib/PageTracker.js
marcoartaviaq: I think the sessionId doesn't need to be persisted. As far as I can see, each time Preview is destroyed or recreated, a new preview event request is made to Access Stats. Since a new preview event was sent, we also need to send the new sessionId. Do we have any use cases where Preview is destroyed and a new access stats request is not being made?

src/lib/PageTracker.js
marcoartaviaq: I'm thinking of a case where the content-insights options request was completed after the initialization of the tracker.

src/lib/PageTracker.js
marcoartaviaq: Is not final, but DP is still using timestamps

src/lib/PageTracker.js
marcoartaviaq: Yes, because if the viewer leaves the page to see another tab, they are not actually viewing the document, so we need to stop tracking. The same goes if the viewer comes back and we need to continue.

src/lib/PageTracker.js
```diff
@@ -0,0 +1,382 @@
...
+        return this.isAdvancedInsightsActive;
+    }
+
+    /**
+     * Start the content insights interval, to report data each number of seconds defined on a Const
+     *
+     * @public
+     * @return {void}
+     */
+    startTracking() {
```
JChan106, line 130: Is there a max number of seconds we should track?

src/lib/PageTracker.js
```diff
@@ -0,0 +1,382 @@
...
+
+    /**
+     * Handle the page change event. At this moment, it would stop the current timer interval,
+     * report the data for the previous page, and start the interval again
+     *
+     * @public
+     * @param {number} pageNumber - The new page number
+     * @return {void}
+     */
+    handleViewerPageChange(pageNumber) {
```
JChan106, line 162: What if theres an error, do we keep tracking?

src/lib/PageTracker.js
marcoartaviaq: There is no upper limit, but we stop emitting events after about 30 seconds (defined in this class as IDLE_TIMER).

src/lib/PageTracker.js
marcoartaviaq: If there is an error, the parent would stop sending the requests. I can add a function inside docBaseViewer that destroys the tracker that could be called from the parent.

src/lib/PageTracker.js
```diff
@@ -0,0 +1,382 @@
...
+    }
+
+    /**
+     * Binds DOM listeners to the document to manage if the user goes active or inactive.
+     *
+     * @private
+     * @return {void}
+     */
+    bindIdleUserEvents() {
+        window.addEventListener('blur', this.handleUserInactive);
```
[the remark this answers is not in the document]
marcoartaviaq: I did some tests and I got that same issue too. I added a new listener (visibilitychange) that captures that

src/lib/Preview.js
```diff
@@ -1404,17 +1407,30 @@ class Preview extends EventEmitter {
...
+            if (sessionId) {
+                data.additional_information = {
+                    view_session: {
+                        session_id: sessionId,
+                    },
+                };
+            }
+        }
+
         this.api
```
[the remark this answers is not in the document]
marcoartaviaq: We follow the same approach as Access-Stats. We batch unsuccessful attempts on the parent, and if we reach the retry limit, we would stop trying doing the requests

bfoxx1906 (approved):

PR #1458, opened 2022-06-28
feat(pdf): Update closure compiler to Babel

karelee7:
Updated compiler from Google Closure Compiler to Babel for pdfjs files


package.json
```diff
@@ -9,8 +9,10 @@
         "url": "git@github.com:box/box-content-preview.git"
     },
     "devDependencies": {
-        "@babel/core": "^7.5.5",
+        "@babel/cli": "^7.18.6",
+        "@babel/core": "^7.18.6",
         "@babel/plugin-proposal-class-properties": "^7.5.5",
```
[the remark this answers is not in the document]
karelee7: I think we can downgrade the new ones; I just needed the @babel/cli dependency which we didn't have before 👍

babel.config.js
```diff
@@ -6,6 +6,8 @@ module.exports = api => {
             [
                 '@babel/preset-env',
                 {
+                    exclude: ['@babel/plugin-transform-unicode-regex'],
```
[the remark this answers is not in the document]
karelee7: Maybe I'm missing something? Moving these into the presets section of the overrides doesn't seem to work; looks like they don't get applied which is strange..

babel.config.js
```diff
@@ -29,6 +30,18 @@ module.exports = api => {
                     ],
                 ],
             },
+            {
+                test: ['./src/third-party/doc/2.84.0/**'],
```
ConradJChan, line 34: We probably don't want to tie it to a specific version, can we have the regex match the contents of `./src/third-party/doc/**`?

build/upgrade_pdfjs.sh
```diff
@@ -56,13 +55,11 @@ sed -e 's@;r.setFlags(o.AnnotationFlag.HIDDEN)@@' -i '' ${DOC_STATIC_ASSETS_PATH
 # This fix may not need to be applied if next upgrade is >= 2.5.x
 # See https://github.com/box/box-content-preview/pull/1414 for more details
 
-# Minify using Google Closure Compiler, options:
-# Output to ES5 (Box supports Chrome, Edge, IE11, Firefox, Safari, and newer versions of iOS, Android)
-# Do not minify pdf.js or pdf.worker.js, as the closure compiler will mangle function names and cause bugs
+# Minify using Babels
```
ConradJChan, line 58: `Babels` -> `babel-minify`?

package.json
```diff
@@ -9,8 +9,10 @@
         "url": "git@github.com:box/box-content-preview.git"
     },
     "devDependencies": {
+        "@babel/cli": "^7.5.5",
         "@babel/core": "^7.5.5",
         "@babel/plugin-proposal-class-properties": "^7.5.5",
+        "@babel/plugin-proposal-private-methods": "^7.18.6",
```
ConradJChan, line 15: Should this match the 7.5.5 version as well?

build/upgrade_pdfjs.sh
karelee7: Oh actually, this is just a babel command, but I should fix the typo; I ran into some issues using the babel-minify command aha...

babel.config.js
```diff
@@ -14,6 +14,7 @@ module.exports = api => {
         plugins: [
             '@babel/plugin-proposal-class-properties',
             '@babel/plugin-proposal-object-rest-spread',
+            '@babel/plugin-proposal-private-methods',
```
ConradJChan, line 17: Should this move to the overrides as well?

karelee7: That's a good idea, I'll update it!

yarn.lock
```diff
@@ -2,6 +2,30 @@
...
+"@ampproject/remapping@^2.1.0":
+  version "2.2.0"
+  resolved "https://registry.yarnpkg.com/@ampproject/remapping/-/remapping-2.2.0.tgz#56c133824780de3174aed5ab6834f3026790154d"
+  integrity sha512-qRmjj8nj9qmLTQXXmaR1cck3UXSRMPrbsLJAasZpF+t3riI71BXed5ebIOYwQntykeZuhjsdweEc9BxH5Jc26w==
+  dependencies:
+    "@jridgewell/gen-mapping" "^0.1.0"
+    "@jridgewell/trace-mapping" "^0.3.9"
+
+"@babel/cli@^7.5.5":
+  version "7.18.6"
```
jstoffan, line 14: We're still seeing 7.18.6, since that's valid with the semantic version we've chosen. Should we upgrade the other dependencies, too?

karelee7: Oh interesting, is it supposed to match up?? And for the other dependencies, are you referring to the `dependencies`/`optionalDependencies` listed under this `@babel/cli`?

jstoffan: Ideally, the versions of all the `@babel/*` dependencies should match up. Otherwise, the CLI may be using a different version/binary than the main build.

karelee7: Ohhhhh I see....let me see...

karelee7: @jstoffan Huh..it seems that installing the `@babel/cli@^7.5.5` version results in this version mismatch...but installing `@babel/cli@7.5.5` matches everything up.

package.json
jstoffan, line 14: Can we update the remaining plugins and presets, as well?

karelee7: Yup! Updated! :)

PR #1466, opened 2023-01-24
fix(pdf): fix for blurry/pixelated images

bfoxx1906:
Altering pdf.js to ignore image interpolation values set to false. This was causing images to show up blurry. This code in question was added in this  [PR](https://github.com/mozilla/pdf.js/pull/13991l) . The change is in the method [getImageSmoothingEnabled ](https://github.com/box/box-content-preview/blob/66dd665777f916246c8cb6e5bf253a5c97bde8f2/src/third-party/doc/2.90.0/pdf.js#L12268)in pdf.js.  Additionally, I added a new version of pdf.js(2.90.0) in the third party directory to match this upgraded version of preview.


karelee7 (dismissed):

package.json
```diff
@@ -74,7 +74,10 @@
         "mini-css-extract-plugin": "^0.8.0",
         "mock-local-storage": "^1.1.15",
         "mojito-rb-gen": "^0.0.1",
-        "node-sass": "4.13.1",
+        "sass": "1.34.1",
```
jstoffan, line 77: @bfoxx1906, I'd suggest doing this in a separate PR. Did we validate the CSS output is identical between the old and new libraries?

jstoffan:
@bfoxx1906, can you add some comments in the diff to explain what the actual change was? Do we plan to add a `sed` command to the `upgrade_pdfjs.sh` script to prevent this issue in the future?

package.json
bfoxx1906: This is what what we're using in buie right now but that's a good point. I'll remove that and then verify the css.

jstoffan:
@bfoxx1906, did we update `pdf.min.js`, as well? It looks like the E2E suite is failing at the moment.

karelee7 (approved):

PR #1472, opened 2023-04-20
fix(content-insights): events recurrent request issue

arturfrombox:
Issue was introduced with latest 2.91 release

How to reproduce:

1. Open a Preview for any file.
2. Wait until the 2.0/events  request start, but go back before the request is finished
3. The code will enter in a state of sending those events in a loop state

Proposed change should fix this behaviour.

### DIFF
--- a/src/lib/Preview.js
+++ b/src/lib/Preview.js
@@ -1385,6 +1385,19 @@ class Preview extends EventEmitter {
     }
 
     /**
+     * Emits event in order to notify PageTracker if the Access Stats preview event was successfully reported
+     *
+     * @private
+     * @param {boolean} success - state of the request
+     * @return {void}
+     */
+    pageTrackerReporter(success = true) {
+        if (this.viewer && typeof this.viewer.emit === 'function') {
+            this.viewer.emit('preview_event_report', success);
+        }
+    }
+
+    /**
      * Logs 'preview' event via the Events API. This is used for logging that a
      * preview happened for access stats, unlike the Logger, which logs preview
      * errors and performance metrics.
@@ -1421,20 +1434,16 @@ class Preview extends EventEmitter {
         this.api
             .post(`${apiHost}/2.0/events`, data, { headers })
             .then(() => {
+                this.pageTrackerReporter(true);
                 // Reset retry count after successfully logging
                 this.logRetryCount = 0;
-                if (this.viewer) {
-                    this.viewer.emit('preview_event_report', true);
-                }
             })
             .catch(() => {
                 // Don't retry more than the retry limit
                 this.logRetryCount += 1;
                 if (this.logRetryCount > LOG_RETRY_COUNT) {
+                    this.pageTrackerReporter(false);
                     this.logRetryCount = 0;
-                    if (this.viewer) {
-                        this.viewer.emit('preview_event_report', false);
-                    }
                     return;
                 }
 
--- a/src/lib/Preview.js
+++ b/src/lib/Preview.js
@@ -1444,6 +1444,7 @@ class Preview extends EventEmitter {
                 if (this.logRetryCount > LOG_RETRY_COUNT) {
                     this.pageTrackerReporter(false);
                     this.logRetryCount = 0;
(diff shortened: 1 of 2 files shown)


marcoartaviaq (commented):
Should we add new unit tests to ensure we are not calling the emit function if the viewer is destroyed?

src/lib/Preview.js
```diff
@@ -1423,14 +1423,18 @@ class Preview extends EventEmitter {
             .then(() => {
                 // Reset retry count after successfully logging
                 this.logRetryCount = 0;
-                this.viewer.emit('preview_event_report', true);
+                if (this.viewer) {
```
jstoffan, line 1426: I'm not super familiar with the issue at hand, but I'm a bit confused by this change. If `this.viewer` is falsy, then `this.viewer.emit` couldn't be called in the first place and would throw rather than causing a loop. Have we actually prevented the issue here?

arturfrombox: due to reference error when this.viewer is undefined, it was falling into catch block where we have recurrent call of logPreviewEvent and on next call of logPreviewEvent again reference error .... with this if check we should be safe

karelee7:
Is there any way to put a gif or screenshots showing that this fix works on this PR? Not sure if there's a way to capture the fix in action, but it'd be nice to see!

src/lib/Preview.js
arturfrombox: good question, there some small delay between document is loaded and /2.0/events request is finished so preview is happening for about 400ms
and by that do you mean to terminate pending an axios request or abandon the execution in _then_ block if this.viewer is undefined ?

src/lib/Preview.js
```diff
@@ -1423,14 +1423,18 @@ class Preview extends EventEmitter {
             .then(() => {
                 // Reset retry count after successfully logging
                 this.logRetryCount = 0;
```
arturfrombox, line 1425: @bfoxx1906 actually i spotted one more bug on this line. this logRetryCount is reset to 0 every time after we execute logPreviewEvent again. so we need to remove this line in order to fix retry logic. because originally request should stop after logRetryCount reach the threshold of 3 retries but instead retry counter resets on every execution

arturfrombox:
> Is there any way to put a gif or screenshots showing that this fix works on this PR? Not sure if there's a way to capture the fix in action, but it'd be nice to see!

sure, link to video recording https://cloud.box.com/s/z222w90pym8thububfipldcaknmxrj1a

src/lib/Preview.js
bfoxx1906: Not seeing where we reset it to 0? That said this looks like it should prevent the infinite recursion. Have we verified that there are no negative impacts when the event is emitted for non document previews? It probably is't a problem but just wanted to make sure we thought about that.

src/lib/Preview.js
bfoxx1906: @arturfrombox disregard that last comment. I see what it's doing now.

src/lib/Preview.js
arturfrombox: I checked non document files and did not notice bugs with event. Here is video recording https://cloud.box.com/s/heb6e7c2g6u5ogmu0dbu9ny51dknq10f

@bfoxx1906 Let me know if there any more changes is needed

src/lib/Preview.js
```diff
@@ -1421,16 +1434,16 @@ class Preview extends EventEmitter {
...
             })
             .catch(() => {
                 // Don't retry more than the retry limit
                 this.logRetryCount += 1;
                 if (this.logRetryCount > LOG_RETRY_COUNT) {
+                    this.pageTrackerReporter(false);
                     this.logRetryCount = 0;
-                    this.viewer.emit('preview_event_report', false);
                     return;
                 }
```
bfoxx1906, line 1449: Should we clear the timeout every time in the catch block. Looks like if we exit after the retry count is hit we will leave the timeout in an uncleared state?

bfoxx1906 (approved):

PR #1473, opened 2023-05-05
fix(media): upgrade shaka player version

bfoxx1906:
Upgrading to the latest shake player. Version 113.0.5672.63 of Chrome broke the old player.


jstoffan (approved):
@bfoxx1906, does this still work in IE11?

PR #1476, opened 2023-05-26
fix(pdf): upgrade pdfjs to 3.6.172

bfoxx1906:
fixes issue with client custom conversion issue causing text layer on pdf's to not be visible


src/lib/viewers/controls/controls-root/ControlsRoot.scss
```diff
@@ -4,6 +4,7 @@
     position: absolute;
     bottom: 25px;
     left: 50%;
+    z-index: 3;
```
jstoffan, line 7: Why is this needed?

bfoxx1906: because the text layer z-index was set to 2 in this release and it was covering the buttons

bfoxx1906: because the text layer z-index was set to 2 in this release and it was covering the buttons

bfoxx1906: because the text layer z-index was set to 2 in this release and it was covering the buttons

JChan106 (approved):

PR #1477, opened 2023-06-01
feat(feature-toggles): Add feature object to standardize feature toggles

JChan106:
Purpose is to enhance the Preview SDK to standardize the usage of feature toggles. This takes in a features object passed in from the options object and sets to the Preview options field. This allows Preview items to have access to the features passed in from the Preview SDK caller.


jstoffan (changes requested):
We should avoid introducing global variables into the SDK. Let's discuss the approach.

src/lib/Preview.js
```diff
@@ -1015,6 +1015,11 @@ class Preview extends EventEmitter {
                 this.enableViewers(viewerName);
             }
         });
+
+        // Features
+        // This makes features available everywhere that features is passed in, which includes
+        // all of the viewers for different files
+        this.options.features = options.features;
```
jstoffan, line 1022: Can we move this before the enable/disable viewer logic? Should it have a default value, such as an empty object?

src/lib/__tests__/Preview-test.js
```diff
@@ -38,6 +38,8 @@ let preview;
 let containerEl;
 
 describe('lib/Preview', () => {
+    const features = { shouldUseBatman: { enabled: true } };
```
jstoffan, line 41: Probably best to avoid copyrighted characters. 😄

src/lib/__tests__/featureChecking-test.js
```diff
@@ -0,0 +1,34 @@
+import { getFeatureConfig, isFeatureEnabled } from '../featureChecking';
```
jstoffan, line 1: Can this be converted to a TS file, as well?

src/lib/__tests__/featureChecking-test.js
```diff
@@ -0,0 +1,34 @@
...
+
+const Box = {
```
jstoffan, line 3: Can we remove the `Box` wrapper?

src/lib/featureChecking.ts
```diff
@@ -0,0 +1,19 @@
...
+
+const isFeatureEnabled = (features: FeatureConfig, featureName: string): boolean => {
+    return !!get(features, featureName, false);
+};
+
+const getFeatureConfig = (features: FeatureConfig, featureName: string): FeatureConfig => {
+    return get(features, featureName, {});
+};
+
+export { isFeatureEnabled, getFeatureConfig };
```
jstoffan, line 19: Can we export inline?

```
export const getFeatureConfig ...
```

src/lib/Preview.js
JChan106: Good point, will do

src/lib/__tests__/featureChecking-test.ts
```diff
@@ -0,0 +1,31 @@
...
+const features = {
+    shouldUseCar: {
+        enabled: true,
+    },
+    shouldUsePlane: {
+        enabled: false,
+    },
+};
+
+describe('isFeatureEnabled', () => {
```
jstoffan, line 12: Minor, but we typically maintain one top-level `describe` per test file. Can we wrap these two?

jstoffan (approved):

src/lib/constants.js
```diff
@@ -102,7 +102,7 @@ export const PDFJS_WIDTH_PADDING_PX = 40; // Should match SCROLLBAR_PADDING in p
 export const PDFJS_HEIGHT_PADDING_PX = 5; // Should match VERTICAL_PADDING in pdf_viewer.js
 
 export const DOC_LEGACY_STATIC_ASSETS_VERSION = '2.76.0'; // IE11 only
-export const DOC_STATIC_ASSETS_VERSION = '2.93.0'; // Modern browsers
+export const DOC_STATIC_ASSETS_VERSION = '2.90.0'; // Modern browsers
```
jstoffan, line 105: Can we make this change as a separate PR?

JChan106: https://github.com/box/box-content-preview/pull/1478

jstoffan (approved):

PR #1482, opened 2023-06-15
feat(pdf): Get PDFjs upgrade ready for prod

karelee7:


src/third-party/doc/2.93.0/pdf.sandbox.min.js
bfoxx1906, line 1: why'd you add this ?

bfoxx1906 (changes requested):

src/third-party/doc/2.93.0/pdf.sandbox.min.js
karelee7: Connected on Slack, removing it now!

bfoxx1906 (dismissed):
lgtm

src/lib/constants.js
```diff
@@ -102,7 +102,7 @@ export const PDFJS_WIDTH_PADDING_PX = 40; // Should match SCROLLBAR_PADDING in p
 export const PDFJS_HEIGHT_PADDING_PX = 5; // Should match VERTICAL_PADDING in pdf_viewer.js
 
 export const DOC_LEGACY_STATIC_ASSETS_VERSION = '2.76.0'; // IE11 only
-export const DOC_STATIC_ASSETS_VERSION = '2.90.0'; // Modern browsers
+export const DOC_STATIC_ASSETS_VERSION = '2.93.0'; // Modern browsers
```
jstoffan, line 105: Can we move the Cypress upgrade to its own PR?

package.json
```diff
@@ -11,6 +11,7 @@
     "devDependencies": {
         "@babel/core": "^7.18.6",
         "@babel/plugin-proposal-class-properties": "^7.18.6",
+        "@babel/plugin-proposal-object-rest-spread": "^7.20.7",
```
jstoffan, line 14: Should we use `^7.18.6` to stay consistent with the other plugins?

package.json
```diff
@@ -112,7 +113,7 @@
         "build:prod": "BABEL_ENV=production NODE_ENV=production node --max_old_space_size=4096 ./node_modules/webpack/bin/webpack.js --progress --colors --config build/webpack.config.js",
         "clean": "rm -rf dist && rm -rf reports/coverage && rm -rf src/i18n/json",
         "cy:open": "yarn cy:wait; yarn cypress open",
-        "cy:run": "yarn cy:wait; yarn cypress run --spec test/integration/**/*.test.js",
+        "cy:run": "yarn cy:wait; yarn cypress run --spec test/integration/**/*.cy.js",
```
jstoffan, line 116: Can we add the `specPattern` option in our Cypress config file rather than renaming all our files?

test/support/commands.js
```diff
@@ -30,6 +29,10 @@ Cypress.Commands.add('showPreview', (token, fileId, options) => {
 Cypress.Commands.add('showControls', () => {
     cy.getByTestId('bp').trigger('mouseover');
     cy.getByTestId('bp-controls').should('be.visible');
+    // Elements with an opacity of 0 are now considered not visible in 6.0+
+    // This interaction forces the controls to be visble
+    // See https://github.com/cypress-io/cypress/issues/14191#issuecomment-745730897
+    cy.getByTitle('Click to enter page number').click();
```
jstoffan, line 35: We should move these back, as the controls pane is reused across all viewers, some of which don't use page controls. Mousing over the controls should set their opacity to 1, so it seems like there's an underlying issue with the steps above.

cypress.config.ts
```diff
@@ -0,0 +1,22 @@
...
+  fileServerFolder: 'test',
+  fixturesFolder: 'test/fixtures',
+  screenshotsFolder: 'test/screenshots',
+  video: false,
+  videosFolder: 'test/videos',
+  viewportHeight: 1260,
+  viewportWidth: 1600,
+  e2e: {
+    // We've imported your old cypress plugins here.
+    // You may want to clean this up later by importing these.
```
jstoffan, line 14: Should we take care of this now?

src/lib/constants.js
karelee7: I'll move this pdfjs upgrade to its own PR and keep the cypress upgrade here since it changed so many files

cypress.config.ts
karelee7: I added the import statement at the top! I think that was all the linter was complaining about; let me know if I missed anything here

.eslintrc.js
```diff
@@ -19,12 +19,16 @@ module.exports = {
         remarkable: false,
         sinon: false,
     },
+    parser: '@babel/eslint-parser',
     rules: {
         'class-methods-use-this': 0, // fixme
+        'flowtype/no-types-missing-file-annotation': 0,
```
jstoffan, line 25: We should try to avoid disabling any linting rules. Let's chat about the need for these in the team channel.

cypress.config.ts
```diff
@@ -0,0 +1,23 @@
...
+    defaultCommandTimeout: 15000,
+    fileServerFolder: 'test',
+    fixturesFolder: 'test/fixtures',
+    screenshotsFolder: 'test/screenshots',
+    video: false,
+    videosFolder: 'test/videos',
+    viewportHeight: 1260,
+    viewportWidth: 1600,
+    e2e: {
+        // We've imported your old cypress plugins here.
```
jstoffan, line 14: Should this comment be removed now or are we still doing something weird?

package.json
```diff
@@ -153,5 +154,8 @@
     },
     "resolutions": {
         "mojito-rb-gen/merge": "1.2.1"
+    },
+    "dependencies": {
+        "@babel/plugin-syntax-flow": "^7.22.5"
```
jstoffan, line 159: Let's move this to `devDependencies`.

package.json
```diff
@@ -92,7 +93,7 @@
         "stylelint-config-rational-order": "^0.1.2",
         "stylelint-config-standard": "^19.0.0",
         "stylelint-order": "^3.1.1",
-        "typescript": "^3.7.4",
+        "typescript": "5.0.4",
```
jstoffan, line 96: Can we retain the `^` on our dependencies or do we need this exact version?

src/lib/featureChecking.ts
```diff
@@ -1,7 +1,7 @@
 import get from 'lodash/get';
 
 type FeatureOptions = {
-    [key: string]: {};
+    [key: string]: NonNullable<unknown>;
```
jstoffan, line 4: Would `unknown` work here? It seems like this value could easily be `null`.

src/lib/viewers/controls/box3d/AnimationClipsControl.tsx
```diff
@@ -6,7 +6,7 @@ import './AnimationClipsControl.scss';
 type AnimationClip = {
     duration: number;
     id: string;
-    name: string;
```
jstoffan, line 9: Can you give some context on the need for this change? We should try to leave the source as-is if we can.

src/lib/viewers/controls/controls-bar/ControlsBarGroup.tsx
```diff
@@ -4,6 +4,7 @@ import './ControlsBarGroup.scss';
 
 export type Props = React.HTMLAttributes<HTMLDivElement> & {
     isDistinct?: boolean;
+    className?: string;
 };
```
jstoffan, line 8: Can we alphabetize these props?

src/lib/viewers/controls/settings/SettingsList.tsx
```diff
@@ -3,7 +3,11 @@ import classNames from 'classnames';
 import noop from 'lodash/noop';
 import { decodeKeydown } from '../../../util';
 
-export type Props = React.HTMLAttributes<HTMLDivElement> & { isActive?: boolean };
+export type Props = React.HTMLAttributes<HTMLDivElement> & {
+    className?: string;
```
jstoffan, line 7: It looks like the `React.HTMLAttributes` type no longer contain these common React attributes? Can we use a different type, instead, to avoid having to declare them explicitly?

.eslintrc.js
karelee7: Sure, here's some more context for this error: https://github.com/gajus/eslint-plugin-flowtype/issues/225

cypress.config.ts
karelee7: Oh yeah, we can remove it now; I'll update that

.eslintrc.js
jstoffan: We don't use flow in this repo, though. Should we remove the flow linting entirely? The original comment applies to the other overrides, as well.

src/lib/viewers/controls/box3d/AnimationClipsControl.tsx
karelee7: Yes, it seems that `name` is a global variable name so it's clashing - https://eslint.org/docs/latest/rules/no-restricted-globals

src/lib/featureChecking.ts
karelee7: Hmm, it looks like when using `unknown`, the linter pops this error message up  on line 16 of this file:
```
Type 'FeatureOptions | {}' is not assignable to type 'FeatureConfig'.
  Type 'FeatureOptions' is not assignable to type 'FeatureConfig'.
    'string' index signatures are incompatible.
      Type 'unknown' is not assignable to type 'FeatureOptions'.
```

src/lib/viewers/controls/settings/SettingsList.tsx
karelee7: Hm, on closer inspection it looks like classname should be on this type - https://github.com/DefinitelyTyped/DefinitelyTyped/blob/4e99c4c87d11cf464312b6150a1e5001be4f809b/types/react/index.d.ts#L1923C1-L1924C1. Seems like other people also ran into this issue and found a workaround: https://github.com/jsx-eslint/eslint-plugin-react/issues/3284. I'll give this a try..

.eslintrc.js
karelee7: As discussed on Slack, I've kept the flowtype rule disabling where it currently is and moved the rest of the rule disabling into the overrides (targeted specifically to TS files)!

package.json
```diff
@@ -10,7 +10,10 @@
     },
     "devDependencies": {
         "@babel/core": "^7.18.6",
+        "@babel/eslint-parser": "^7.5.4",
         "@babel/plugin-proposal-class-properties": "^7.18.6",
+        "@babel/plugin-proposal-object-rest-spread": "^7.18.6",
+        "@babel/plugin-syntax-flow": "^7.22.5",
```
jstoffan, line 16: Do we still need this plugin and, if so, should it be `^7.18.6`?

package.json
```diff
@@ -44,16 +46,16 @@
...
-        "eslint": "^6.7.2",
+        "eslint": "^8.43.0",
         "eslint-config-airbnb": "^18.0.1",
         "eslint-config-prettier": "^6.7.0",
         "eslint-import-resolver-webpack": "^0.11.1",
         "eslint-plugin-babel": "^5.3.0",
         "eslint-plugin-chai-friendly": "^0.4.1",
         "eslint-plugin-cypress": "^2.8.1",
-        "eslint-plugin-flowtype": "^4.5.2",
+        "eslint-plugin-flowtype": "8.0.3",
```
jstoffan, line 58: Can this use `^8.0.3` or do we need this specific version?

src/lib/viewers/controls/box3d/AnimationClipsControl.tsx
jstoffan: These aren't global variables, though, they're properties of an object. Is this another case of some kind of parser oddity for TypeScript files?

karelee7: Huh, it looks like if I remove `parser: '@babel/eslint-parser'` from the eslint config, this error goes away

src/lib/viewers/controls/fullscreen/FullscreenToggle.tsx
```diff
@@ -6,7 +6,7 @@ import './FullscreenToggle.scss';
 
 export type Props = {
     onFullscreenToggle: (isFullscreen: boolean, toggleFullscreenEl: EventTarget | null) => void;
-    onKeyDown?: (event: React.KeyboardEvent<HTMLButtonElement>) => void;
+    onKeyDown?: (keyboardEvent: React.KeyboardEvent<HTMLButtonElement>) => void;
```
jstoffan, line 9: Can we revert this change now that the `no-unrestricted-globals` issue is fixed?

src/lib/viewers/controls/icons/IconDrawing24.tsx
```diff
@@ -1,6 +1,10 @@
 import * as React from 'react';
 
-function IconDrawing24({ fill, ...rest }: React.SVGProps<SVGSVGElement>): JSX.Element {
+type Props = React.SVGProps<SVGSVGElement> & {
+    fill?: string;
```
jstoffan, line 4: I think `fill` is a valid SVG prop, so we shouldn't need to redeclare it. Did the underlying type change?

test/integration/sanity/Sanity.e2e.test.js
```diff
@@ -17,10 +17,15 @@ describe('Preview Sanity', () => {
         cy.showPreview(token, fileId);
         cy.getPreviewPage(1);
         cy.contains('The Content Platform for Your Apps');
+        cy.get('.bp-thumbnail').each($ele => {
```
jstoffan, line 20: Should this be moved to a helper, like `cy.hasThumbnails`, since it's used more than once?

test/support/commands.js
```diff
@@ -28,7 +27,9 @@ Cypress.Commands.add('showPreview', (token, fileId, options) => {
 });
 
 Cypress.Commands.add('showControls', () => {
-    cy.getByTestId('bp').trigger('mouseover');
+    cy.getByTestId('bp-controls')
+        .invoke('show')
```
jstoffan, line 31: Is the mouseover event no longer working for the controls?

jstoffan (commented):
Looking good, just a few loose ends to tie up. We want to try to avoid changing the "logic" of our tests, since consistency between runs is important.

test/support/commands.js
karelee7: Yes, it seems that the `trigger` and `mouseover` doesn't bring up the controls panel anymore.

src/lib/viewers/controls/icons/IconDrawing24.tsx
karelee7: Ah, hm, looks like `SVGProps` is an interface, perhaps it's similar to the other props issue that came up

test/support/commands.js
jstoffan: Does that reflect an actual problem with the controls or does Cypress use a different mechanism to trigger mouseover events now? We want the end-to-end tests to act as much like a user as possible, so we try to avoid invoking functions directly from them.

karelee7: Hmm, no, doesn't seem like a problem with the controls; I think something changed with the trigger command - I see that if you don't pass it an option of `force: true` it [waits for actionability](https://docs.cypress.io/api/commands/trigger#Assertions). Perhaps that has something to do with it...I'll give it a try

karelee7: Huh, I was curious what happened on the old version of cypress with the old version of pdfjs, and it looks like the hover never happens but the test passes - i.e. `Thumbnails.e2e.test.js`.

src/lib/viewers/controls/icons/IconDrawing24.tsx
karelee7: Per the slack discussion, we can disable this rule for TS files for now - seems like it's a known issue and after looking at interfaces, it seems like we do have the fill property; so it looks like this is a false positive.

test/support/commands.js
karelee7: Per slack discussion, `mousemove` works!

.eslintrc.js
```diff
@@ -19,8 +19,20 @@ module.exports = {
         remarkable: false,
         sinon: false,
     },
+    overrides: [
+        {
+            files: ['**/*.ts', '**/*.tsx'],
+            rules: {
+                'no-shadow': 'off',
+                'react/prop-types': 0,
```
jstoffan, line 27: Minor, but let's stick to either `'off'` or `0`?

test/support/mediaSettingsTests.js
```diff
@@ -175,12 +181,11 @@ export function runSubtitlesTests() {
             cy.getByTestId('bp-media-settings-subtitles').contains('Off');
 
             // Toggling CC back on should restore the previously used text track
-            cy.getByTitle('Subtitles/Closed Captions')
-                .as('subtitlesBtn')
-                .should('be.visible')
-                .should('have.attr', 'aria-pressed', 'false')
-                .click()
-                .should('have.attr', 'aria-pressed', 'true');
+            cy.getByTitle('Subtitles/Closed Captions').as('subtitlesBtn');
```
jstoffan, line 184: Minor, but we could alias this button using `.as('subtitlesBtn')` and then use `cy.get('@subtitlesBtn')` to stay consistent with the lines above.

jstoffan (commented):
@karelee7, I'm seeing some upgraded dependencies (e.g. Babel) in the yarn.lock file. Is that intentional? If not, let's reset it to the version from master, then run `yarn add -D eslint@^8.43.0 cypress@^12.15.0` and whatever others we want to upgrade.

karelee7:
@jstoffan Took a look at the changes in the lock file and it looks correct! Just to double check, I also tried restoring everything to the version on master and running the `yarn add -D eslint@^8.43.0 cypress@^12.15.0...` command - after running that command, it produces the same yarn.lock file!

jstoffan (approved):
Nice job! Have a great weekend!

PR #1484, opened 2023-06-18
fix(e2e-tests): fixing hd video player test

bfoxx1906:


test/support/mediaSettingsTests.js
```diff
@@ -61,9 +61,17 @@ export function runQualityMenuTests() {
                 .contains('Auto')
                 .click();
 
+            cy.get('[role="menuitem"]').contains('Quality');
```
JChan106, line 64: What was error in this case? Why do we need to move this line up?

test/support/mediaSettingsTests.js
```diff
@@ -61,9 +61,17 @@ export function runQualityMenuTests() {
                 .contains('Auto')
```
jstoffan, line 61: Do we know what changed in the underlying Shaka player that caused the "Auto" quality to no longer return HD video?

bfoxx1906: So based on what I saw testing manually there is a delay between when the video loads and when the HD button appears at times. Additionally, the automatic display of the HD icon is not consistent as it shows sometimes but not others. Looking under the hood I think the way events are passed to our settings components changed with the new shake player version which is what I think is ultimately what's causing the problem. What we're doing with this PR is making sure that the basic functionality of the badge works but as part of our updates to the SDK  we need to update our code to account for some of these event differences.

jstoffan: Got it, thanks. Given that this has already been released, I think this is safe to merge with the expectation that we revert the change once we figure out why "Auto" no longer works how it did previously.

test/support/mediaSettingsTests.js
bfoxx1906: No. There's no specific reason for that. I just moved it when I was updating the test.

PR #1486, opened 2023-07-07
fix(a11ty): video player thumnail has no text alternative

arturfrombox:
### Description of Changes
Recommendation was to set alt="", but since this is an svg and not an img, aria-labelledby and title element was added instead.
Div was changed to button to make it accessibility-compliant.

SVG accessibility: https://www.deque.com/blog/creating-accessible-svgs/

Screenshot 2023-07-07 at 17 39 41


src/lib/icons/play_48px.svg
```diff
@@ -1,4 +1,5 @@
-<svg fill="#fff" height="48" viewBox="0 0 24 24" width="48" focusable="false">
+<svg fill="#fff" height="48" viewBox="0 0 24 24" width="48" focusable="false" role="img" aria-labelledby="play-button-icon-01">
+    <title id="play-button-icon-01"></title>
```
jstoffan, line 2: We should avoid using `id` for selectors since multiple instances of Preview can be rendered in a document at once.

src/lib/icons/play_48px.svg
```diff
@@ -1,4 +1,5 @@
-<svg fill="#fff" height="48" viewBox="0 0 24 24" width="48" focusable="false">
+<svg fill="#fff" height="48" viewBox="0 0 24 24" width="48" focusable="false" role="img" aria-labelledby="play-button-icon-01">
```
jstoffan, line 1: Icons should be unaware of how they're used. It looks like we need to add a hidden span inside the button rather than in the SVG itself.

src/lib/viewers/media/VideoBaseViewer.js
```diff
@@ -43,10 +43,11 @@ class VideoBaseViewer extends MediaBaseViewer {
         this.mediaEl.setAttribute('playsinline', '');
 
         // Play button
-        this.playButtonEl = this.mediaContainerEl.appendChild(document.createElement('div'));
+        this.playButtonEl = this.mediaContainerEl.appendChild(document.createElement('button'));
```
jstoffan, line 46: Do we need to add any styling resets with this change? Buttons often have default styles.

jstoffan (changes requested):
Do we have other a11y fixes for these components? If so, we should consider launching the React versions, instead, since the [play/pause toggle](https://github.com/box/box-content-preview/blob/master/src/lib/viewers/controls/media/PlayPauseToggle.tsx) already supports it.

arturfrombox:
> Do we have other a11y fixes for these components? If so, we should consider launching the React versions, instead, since the [play/pause toggle](https://github.com/box/box-content-preview/blob/master/src/lib/viewers/controls/media/PlayPauseToggle.tsx) already supports it.

Thanks @jstoffan for reviewing this pr
I did not found other fixes for video controls. Let me know if we still should consider launching react versions

arturfrombox:
aria-hidden was set to true so the svg image is no longer meaningful for screen readers and title attribute was added to button instead.

src/lib/viewers/media/_mediaBase.scss
```diff
@@ -31,12 +41,13 @@
...
 
 .bp-media-play-button {
+    @include bp-reset-default-styling-button;
+
     position: absolute;
     top: 50%;
     left: 50%;
     width: 80px;
     height: 80px;
-    padding-top: 12px;
```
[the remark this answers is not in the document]
arturfrombox: before 
Screenshot 2023-07-10 at 17 22 41
after
Screenshot 2023-07-10 at 17 26 47

jstoffan:
Note: We would use `a11y` rather than `a11ty`, as the "11" is meant to indicate the number of letters between "a" and "y".

PR #1488, opened 2023-07-10
fix(a11y): remove instructions

arturfrombox:
### Description of Changes
Recommendation was to remove tabindex and aria label 'Press alt + arrow keys...' text since it is not accurate for Windows computers, and it will not help keyboard users who are not using a screen reader.

### DIFF
--- a/src/lib/viewers/doc/DocBaseViewer.js
+++ b/src/lib/viewers/doc/DocBaseViewer.js
@@ -158,7 +158,9 @@ class DocBaseViewer extends BaseViewer {
         super.setup();
 
         this.docEl = this.createViewer(document.createElement('div'));
+        this.docEl.setAttribute('aria-label', __('document_label'));
         this.docEl.classList.add('bp-doc');
+        this.docEl.tabIndex = '0';
 
         if (Browser.getName() === 'Safari') {
             this.docEl.classList.add(IS_SAFARI_CLASS);
--- a/src/lib/viewers/doc/DocBaseViewer.js
+++ b/src/lib/viewers/doc/DocBaseViewer.js
@@ -158,9 +158,7 @@ class DocBaseViewer extends BaseViewer {
         super.setup();
 
         this.docEl = this.createViewer(document.createElement('div'));
-        this.docEl.setAttribute('aria-label', __('document_label'));
         this.docEl.classList.add('bp-doc');
-        this.docEl.tabIndex = '0';
 
         if (Browser.getName() === 'Safari') {
             this.docEl.classList.add(IS_SAFARI_CLASS);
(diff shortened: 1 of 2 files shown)


src/lib/viewers/doc/DocBaseViewer.js
```diff
@@ -158,9 +158,7 @@ class DocBaseViewer extends BaseViewer {
         super.setup();
 
         this.docEl = this.createViewer(document.createElement('div'));
-        this.docEl.setAttribute('aria-label', __('document_label'));
```
jstoffan, line 161: We added this label to address a previous accessibility concern around [caret navigation](https://en.wikipedia.org/wiki/Caret_navigation). It seems unwise to remove it without a replacement solution.

src/lib/viewers/doc/DocBaseViewer.js
```diff
@@ -158,9 +158,7 @@ class DocBaseViewer extends BaseViewer {
...
         this.docEl.classList.add('bp-doc');
-        this.docEl.tabIndex = '0';
```
jstoffan, line 163: This logic has been here since 2019. Its removal will likely break keyboard scrolling and possibly other desired behaviors.

jstoffan (changes requested):

src/lib/viewers/doc/DocBaseViewer.js
arturfrombox: @jstoffan thanks for pointing this out
So the recommendations would be to change the aria label in this case to say "Document Viewer"
this is slightly redundant with the next element up in the DOM, which is a region called "File Preview." The main thing we need is to remove these confusing instructions

jstoffan: I don't think that addresses the issue I mentioned with caret navigation. Can you sync up with the person requesting these changes to make sure we don't regress?

arturfrombox: in short, recommendation stays the same (maybe we should consider updating box keyboard docs). Added more detailed explanation from person requesting changes below.

_as a practice we should not have very specific, hidden help text in our UIs.n this case, our accessibility auditors called out that this text is inaccurate and counter productive. nothing about the current text tells a user how to successfully use caret browsing. caret browsing also differs between browsers and OSes, so if we were to have in-UI instructions, we would have very bloated instructions that can (and have) become obsolete as commands change over time.
it is important to note that caret browsing is generally useful for users with vision, as screen reader users can linearly navigate through a document without caret browsing, using default screen reader browse commands (although screen reader users might use caret browsing for text selection). users with vision are not helped by hidden text in the UI.
what we can do is update our [keyboard commands documentation](https://support.box.com/hc/en-us/articles/360044194873-Using-Keyboard-Shortcuts-for-Box) to include a section on how to use caret browsing with Preview. I don't think this needs to be a blocker for this bug fix though - the first order of business is to remove the confusing instructions_

jstoffan (approved):

PR #1489, opened 2023-07-10
fix(copy-paste-issue): PDFJS copy paste fix

JChan106:
We noticed when upgrading to PDFJS v3.6.172, copy/paste did not work for text on pdfs. This is because in this new PDFJS version, the enum we were using to pass into the `textLayerMode`, had changed from `ENABLE_ENHANCE` to `ENABLE_PERMISSION`. In previous PDFJS versions, `ENABLE_ENHANCE` would be an option to enable textLayerMode and enhance the text selection (improve text selection across multiple lines). However they have deprecated this option as of `v3.0.279` as noted here: 
- https://github.com/mozilla/pdf.js/pull/15259
- https://github.com/mozilla/pdf.js/issues/15138
- https://github.com/mozilla/pdf.js/pull/15145

They have replaced this enum with a new option called `ENABLE_PERMISSION` which if passed in for `textLayerMode`, checks the permission flags set for the PDF document. So if the PDF document doesn't have the `COPY` flag set, and `ENABLE_PERMISSION` is passed in, the user will not be able to copy the document. This is discussed here:
- https://github.com/mozilla/pdf.js/pull/16333
- https://github.com/mozilla/pdf.js/pull/16320

Because `ENABLE_PERMISSION` uses the same enum as the deprecated `ENABLE_ENHANCE`, we were using `ENABLE_PERMISSION` mistakenly when upgrading versions. To solve this, we can just use the `ENABLE` enum instead, which doesn't check for the documents permissions. We use this enum for mobile and it is also what PDFJS uses as the default `textLayerMode` as well, so I don't believe there's much risk.

The Enhanced Text Selection seems to be the drawback to this change, i'm not sure if the pdfjs team has plans to bring it back, but this comment mentioned that "in general it's often too slow to be usable in practice". 
- https://github.com/mozilla/pdf.js/issues/15138#issuecomment-1177206433

### DIFF
--- a/src/lib/viewers/doc/DocBaseViewer.js
+++ b/src/lib/viewers/doc/DocBaseViewer.js
@@ -80,7 +80,7 @@ const PAGES_UNIT_NAME = 'pages';
 const PDFJS_TEXT_LAYER_MODE = {
     DISABLE: 0, // Should match TextLayerMode enum in pdf_viewer.js
     ENABLE: 1,
-    ENABLE_ENHANCE: 2,
+    ENABLE_PERMISSIONS: 2,
 };
 const PINCH_PAGE_CLASS = 'pinch-page';
 const PINCHING_CLASS = 'pinching';
@@ -796,7 +796,6 @@ class DocBaseViewer extends BaseViewer {
         const assetUrlCreator = createAssetUrlCreator(location);
         const hasDownload = checkPermission(file, PERMISSION_DOWNLOAD);
         const hasTextLayer = hasDownload && !this.getViewerOption('disableTextLayer');
-        const textLayerMode = this.isMobile ? PDFJS_TEXT_LAYER_MODE.ENABLE : PDFJS_TEXT_LAYER_MODE.ENABLE_ENHANCE;
 
         return new PdfViewerClass({
             annotationMode: PDFAnnotationMode.ENABLE, // Show annotations, but not forms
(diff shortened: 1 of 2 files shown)


jstoffan:
@JChan106, this seems like a pretty meaningful downgrade in functionality. We may need to retrofit our own solution to avoid this being seen as a regression. What effect does it have on the creation of text highlight annotations?

src/lib/viewers/doc/DocBaseViewer.js
```diff
@@ -80,7 +80,7 @@ const PAGES_UNIT_NAME = 'pages';
 const PDFJS_TEXT_LAYER_MODE = {
     DISABLE: 0, // Should match TextLayerMode enum in pdf_viewer.js
     ENABLE: 1,
-    ENABLE_ENHANCE: 2,
+    ENABLE_PERMISSIONS: 2,
```
bfoxx1906, line 83: Do we use this value anywhere? If not do we even need this?

JChan106:
> @JChan106, this seems like a pretty meaningful downgrade in functionality. We may need to retrofit our own solution to avoid this being seen as a regression. What effect does it have on the creation of text highlight annotations?

Good point, I've been testing the differences between enhanced text selection and normal text selection and I think we may have to retrofit. When selecting multiple elements with large spaces between them, It seems like the enhanced text selection shows the selection through the space. While as normal text selection only shows the selection of the actual elements.

Normal:
image

Enhanced:
image

The actual content selected seems to be the same, however if we want the same look and experience that users of box-content-preview, we must retrofit.

src/lib/viewers/doc/DocBaseViewer.js
JChan106: I think if we're going to retrofit, we may need to add an additional ENUM for `ENABLE_ENHANCE`, so we may want to keep `ENABLE_PERMISSIONS` there as 2 and have `ENABLE_ENHANCE` at 3 if that make sense. That way we don't have to modify the logic to replace PERMISSIONS.

JChan106:
Sent an issue to the pdfjs team regarding the enhanced text selection removal: https://github.com/mozilla/pdf.js/issues/16684. Will update this PR once we get a response.

bfoxx1906 (approved):

bfoxx1906 (approved):

bfoxx1906 (approved):

PR #1490, opened 2023-07-18
feat(pdf): Fix search and highlight + comment tooltip

karelee7:
Search (controls are clickable):
search

Highlight and Comment tooltip clickable:

Drawing Annotation works on text layer:

### DIFF
--- a/src/lib/viewers/controls/annotations/AnnotationsControls.scss
+++ b/src/lib/viewers/controls/annotations/AnnotationsControls.scss
@@ -9,3 +9,9 @@
 .bp-AnnotationsControls-exitBtn {
     display: none;
 }
+
+.ba-Layer--popup,
+.ba-Layer--highlight,
+.ba-Layer--drawing {
+    z-index: 3;
+}
--- a/src/lib/viewers/doc/DocFindBar.scss
+++ b/src/lib/viewers/doc/DocFindBar.scss
@@ -4,6 +4,7 @@
     position: absolute;
     top: 0;
     right: 17px; // Prevents overlap with scroll-bar
+    z-index: 3;
     display: flex;
     align-items: center;
     padding: 6px;


src/lib/viewers/controls/annotations/AnnotationsControls.scss
```diff
@@ -9,3 +9,8 @@
 .bp-AnnotationsControls-exitBtn {
     display: none;
 }
+
+.ba-Layer--popup,
+.ba-Layer--highlight {
+    z-index: 3;
```
JChan106, line 15: Why do we need to set it to 3? And is there any other elements that need to be changed now that this z-index has changed?

JChan106 (commented):
Can we get some gifs of the fix?

src/lib/viewers/controls/annotations/AnnotationsControls.scss
karelee7: We need it to be above the text layer, which is now at z-index 2 in this new version of PDFjs (see https://cdnjs.cloudflare.com/ajax/libs/pdf.js/3.6.172/pdf_viewer.css - the z-index in .textLayer)

karelee7: I don't think so; this will move the search controls, the tooltip, and the modal that pops up when clicking on the tooltip

bfoxx1906: I had to do something similar when I first merged the changes for this pdf.js version.

karelee7: I think the drawing layer is a separate layer so that may also need to move up, but should we also capture that in this ticket? Or should we have a separate ticket for that?

jcamden (approved):
Looks like you got the drawing layer in there too, per discussion. Nice!

JChan106 (approved):
lgtm!

PR #1491, opened 2023-07-25
feat(node-18-upgrade): Node 18 Upgrade

JChan106:
This is to upgrade to Node 18. Tested locally and did not find any issues. We use the same workaround BUIE uses for the OpenSSL issue.


package.json
```diff
@@ -101,15 +101,15 @@
...
     "engines": {
-        "node": ">=10.x",
+        "node": ">=18.x",
         "yarn": ">=1.19.x"
     },
     "scripts": {
         "build": "yarn setup && yarn clean && yarn build:i18n && yarn build:dev",
         "build:ci": "yarn setup && yarn webpack --progress --colors --config build/webpack.config.js",
-        "build:dev": "BABEL_ENV=dev NODE_ENV=dev yarn webpack --progress --colors --config build/webpack.config.js",
+        "build:dev": "BABEL_ENV=dev NODE_ENV=dev NODE_OPTIONS=--openssl-legacy-provider yarn webpack --progress --colors --config build/webpack.config.js",
```
jstoffan, line 110: Did we verify that this is needed? If so, can we upgrade Webpack to avoid it?

JChan106: It is needed if we don't upgrade Webpack. I was thinking it would be better to upgrade Webpack separately as there could be more issues that can occur with upgrading to Webpack 5?

jstoffan: Gotcha. We should make sure to do the upgrade at some point, since the option will be deprecated eventually.

yarn.lock
```diff
@@ -3079,10 +3079,10 @@ box-annotations@^2.3.0:
   resolved "https://registry.yarnpkg.com/box-annotations/-/box-annotations-2.3.0.tgz#5cac38171f7f8d9283659e2b243310f19d5ab7d3"
   integrity sha512-Ea7tPgyJjX7vcnmZIfCorbzHd6oYx/OHVMPnZVQL/dUHR5vRKhLM0610xqwmVlUpk627sqHw5x/APaa+kt4SXg==
 
-box-ui-elements@15.1.0-beta.21:
-  version "15.1.0-beta.21"
-  resolved "https://registry.yarnpkg.com/box-ui-elements/-/box-ui-elements-15.1.0-beta.21.tgz#4fa34bf6bf3b10d79f95b0e56349c057926573e3"
-  integrity sha512-ey7lwuv+ds9rq3nVfVcneedWaptw3nYdlF32antUxw0IkkuAYft3LX4XxRHDqNQSv5LzUBqxd5wd1caxVpiuWQ==
+box-ui-elements@18.1.0:
```
jstoffan, line 3082: Pretty big jump. Did anything meaningful change that Preview SDK is using?

jstoffan (approved):

yarn.lock
JChan106: Good point i'll check all the imports see if theres anything funky from those.

package.json
JChan106: Will create ticket for that

.travis.yml
```diff
@@ -13,6 +14,7 @@ cache:
 before_install:
     - curl -o- -L https://yarnpkg.com/install.sh | bash -s -- --version 1.22.5
     - export PATH=$HOME/.yarn/bin:$PATH
+    - export NODE_OPTIONS=--openssl-legacy-provider
```
jstoffan, line 17: Can we drop this option from the scripts in package.json after this change? Or do the environment options get blown away during build execution?

jstoffan (approved):
Nice job!

.travis.yml
JChan106: Docs say enviornment variables can be accessed at any stage in the build process. But we still need the option there in the package.json for self hosting right?

jstoffan: Great point!

PR #1492, opened 2023-07-26
fix(preview): fix auto scroll issue with embed preview

Machoper:
## Issue
Scrolling jumps to bottom of page on embedded multi-page previews

## Summary of changes
- Add `{ block: 'nearest' }` as param to [`scrollIntoView()`](https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoView) to prevent the parent scrollbar from jumping around
- Fix a type to pass lint check

### Before
auto_scroll_before

### After
auto_scroll_after

### DIFF
--- a/src/lib/VirtualScroller.js
+++ b/src/lib/VirtualScroller.js
@@ -418,7 +418,7 @@ class VirtualScroller {
             // If it is already present and visible, do nothing, but if not visible
             // then scroll it into view
             if (!this.isVisible(foundItem)) {
-                foundItem.scrollIntoView();
+                foundItem.scrollIntoView({ block: 'nearest' });
             }
         } else {
             // If it is not present, then adjust the scrollTop so that the list item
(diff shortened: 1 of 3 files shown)


JChan106 (commented):
Did you test that scrolling works as usual non-embedded? Can you also test on different browser sizes?

karelee7:
Can we also test it out on different browsers just to be safe?

Machoper:
> Did you test that scrolling works as usual non-embedded? Can you also test on different browser sizes?

@JChan106 Yes, scrolling works well in the Webapp across all browser sizes.

> Can we also test it out on different browsers just to be safe?

@karelee7 Yes, I have tested on all supported browsers.

---

Adding two more gifs for reference

**Safari + embedded:**
auto_scroll_safari

**MS Edge + non-embedded:**
auto_scroll_edge

JChan106 (approved):
LGTM

karelee7 (approved):

PR #1497, opened 2023-08-02
feat(pdf): Move annotation styling to box-annotations

karelee7:
In v3.6.172 of PDFjs, a z-index styling was explicitly added to pdf_viewer.css: https://cdnjs.cloudflare.com/ajax/libs/pdf.js/3.6.172/pdf_viewer.css. This styling was not present in the previous version of PDFjs that we're using: https://cdnjs.cloudflare.com/ajax/libs/pdf.js/2.14.305/pdf_viewer.css and keeping this new styling causes issues with the other layers in preview (i.e. popup layer, highlighted layer, etc.). Updating styling for our layers resolves the issues that QA has discovered, but moving them to the box-annotations repo is preferable.

Have tested this change in the SDK with the stylings moved to box-annotations/linked with EUA.

### DIFF
--- a/src/lib/viewers/controls/annotations/AnnotationsControls.scss
+++ b/src/lib/viewers/controls/annotations/AnnotationsControls.scss
@@ -9,10 +9,3 @@
 .bp-AnnotationsControls-exitBtn {
     display: none;
 }
-
-.ba-Layer--popup,
-.ba-Layer--highlight,
-.ba-Layer--drawing,
-.ba-point-annotation-marker {
-    z-index: 3;
-}


src/lib/viewers/controls/annotations/AnnotationsControls.scss
```diff
@@ -9,10 +9,3 @@
 .bp-AnnotationsControls-exitBtn {
     display: none;
 }
-
-.ba-Layer--popup,
-.ba-Layer--highlight,
-.ba-Layer--drawing,
-.ba-point-annotation-marker {
```
bfoxx1906, line 16: Is this ba-point-annotation-marker class actually used anywhere? I didn't see this in box-annotations.

bfoxx1906 (approved):

src/lib/viewers/controls/annotations/AnnotationsControls.scss
```diff
@@ -9,10 +9,3 @@
 .bp-AnnotationsControls-exitBtn {
     display: none;
 }
```
bfoxx1906, line 11: was ba-point-annotation-marker used anywhere

src/lib/viewers/controls/annotations/AnnotationsControls.scss
karelee7: Yeah, I also couldn't find it in annotations, but this needs to move above the text layer to be clickable (it's a fix for expiring embed links - since it's prefixed with `ba` I figured that annotations would be the best place to put it)

src/lib/viewers/controls/annotations/AnnotationsControls.scss
karelee7: See above answer!

src/lib/viewers/controls/annotations/AnnotationsControls.scss
jstoffan: It's used in a legacy version of `box-annotations` that's only used by the Expiring Embed flow. So, unless we want to release a patch to v2 of `box-annotations`, this specific style will need to stay in `box-content-preview`.

PR #1499, opened 2023-08-04
fix(pdf): Move controls layer above popup layer

karelee7:
The popup annotation layer, which comes before this controls layer in the HTML hierarchy, now has a z-index of 4 due to changes needed for the PDFjs upgrade - because of this, the popup annotation layer was covering this controls layer. This causes the E2E tests in box-annotations to fail. So to resolve this, we need to also bring the controls layer up to a z-index of 4. 

Ran this change with the E2E tests for box-annotations locally and it was passing.

### DIFF
--- a/src/lib/viewers/controls/controls-root/ControlsRoot.scss
+++ b/src/lib/viewers/controls/controls-root/ControlsRoot.scss
@@ -4,7 +4,7 @@
     position: absolute;
     bottom: 25px;
     left: 50%;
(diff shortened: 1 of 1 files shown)


jstoffan (approved):
Can we rename the PR title prefix to `fix`?

karelee7:
Thanks for the review @jstoffan ! I also updated the commit message to say `fix` :)

PR #1503, opened 2023-08-23
fix(pdf): fix disappearing form data

bfoxx1906:
As a result of  this [PR](https://github.com/mozilla/pdf.js/pull/16387/files) for pdf.js going into version 3.6.172, read only text in form fields were set to be rendered in their own canvas. This in and of itself is not a problem but the underlying code required that ENABLE_FORMS by set for the annotationMode value that we pass to the viewer class in DocBaseViewer which conflicts work with our current policy of making form fields uneditable. To fix this I upgraded the pdf.worker.min.js script to set the value "this.data.hasOwnCanvas" as part of the TextWidgetAnnotation object to default to false. In addition to this, I added a new shell script that verifies that the pdf.js custom changes we've needed to make have not been forgotten as we update versions of pdf.js and the unit tests will fail if they are not present. Lastly I updated the copyright to 2023.

Pre Change:

Screenshot 2023-08-28 at 12 56 53 PM

Post Change:

post-change for issue


build/upgrade_pdfjs2.sh
```diff
@@ -0,0 +1,63 @@
+#!/bin/bash
```
karelee7, line 1: Do we need to keep the old upgrade_pdfjs file too?

bfoxx1906: Yea this is a good point. I think this PR is doing too much. I'm going to pull this file out and create a separate PR for it.

JChan106:
Do we have examples of pdfs with forms that disappear? If so can we get some screenshots?

jstoffan (changes requested):
@bfoxx1906, the assets in the `third-party` folder are typically cached for a very long time and aren't hashed. We'll need to release a new version of the document viewer assets to ensure the fix reaches all our users.

That's also one reason these steps are baked into the `upgrade_pdfjs.sh` script. Should we update that script as well or instead?

bfoxx1906:
> @bfoxx1906, the assets in the `third-party` folder are typically cached for a very long time and aren't hashed. We'll need to release a new version of the document viewer assets to ensure the fix reaches all our users.
> 
> That's also one reason these steps are baked into the `upgrade_pdfjs.sh` script. Should we update that script as well or instead?

Yea that's a good point about the third-party folder. Good catch. I'll change this. I did create a new upgrade script that included this but I felt like that too much for one PR. I'll create another PR with that script once this one is merged.

README.md
```diff
@@ -419,6 +419,10 @@ If you have any questions, please search our [issues list](https://github.com/bo
 
 For general Box Platform, API, and Box Element questions, please visit our [developer forum](https://community.box.com/t5/Developer-Forum/bd-p/DeveloperForum) or contact us via one of our [available support channels](https://community.box.com/t5/Community/ct-p/English).
 
+Effective September 1, 2023, Box is discontinuing support for Internet Explorer 11 (IE 11) in Preview SDK. If your application requires IE 11 support, we recommend using the release versions of Preview SDK that were made available before August 15th, 2023. Our team will address any critical blocker and security issues specifically related to IE 11 until October 15, 2023. Following this date, we will no longer provide active support for IE 11 in Preview SDK.
```
jstoffan, line 422: This and the copyright change would be best as a separate PR.

src/lib/constants.js
```diff
@@ -102,7 +102,7 @@ export const PDFJS_WIDTH_PADDING_PX = 40; // Should match SCROLLBAR_PADDING in p
 export const PDFJS_HEIGHT_PADDING_PX = 5; // Should match VERTICAL_PADDING in pdf_viewer.js
 
 export const DOC_LEGACY_STATIC_ASSETS_VERSION = '2.76.0'; // IE11 only
-export const DOC_STATIC_ASSETS_VERSION = '2.93.0'; // Modern browsers
+export const DOC_STATIC_ASSETS_VERSION = '2.94.0'; // Modern browsers
```
jstoffan, line 105: Current version is `2.100.0`, right?

bfoxx1906: Yup. I'll fix that

jstoffan (approved):

PR #1505, opened 2023-08-31
fix(pdf): override css missing asset

bfoxx1906:
Because of a change in a css rule in the latest version of [pdf_viewer.css ](https://github.com/mozilla/pdf.js/blob/v3.6.172/web/pdf_viewer.css#L162)we were not overriding the loadingIcon style's background image which caused the CDN to record a large number of 404's as a result.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
     "name": "box-content-preview",
-    "version": "2.101.0",
+    "version": "2.100.0",
     "description": "Box Content Preview UI Element",
     "author": "Box (https://www.box.com/)",
     "license": "SEE LICENSE IN LICENSE",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
     "name": "box-content-preview",
-    "version": "2.100.0",
+    "version": "2.101.0",
     "description": "Box Content Preview UI Element",
     "author": "Box (https://www.box.com/)",
     "license": "SEE LICENSE IN LICENSE",
--- a/src/lib/viewers/doc/_docBase.scss
+++ b/src/lib/viewers/doc/_docBase.scss
@@ -290,6 +290,8 @@ $thumbnail-sidebar-width: 191px; // Extra pixel to account for sidebar border
         }
 
         &.loadingIcon::after {
+            top: $pdfjs-page-padding;
+            bottom: $pdfjs-page-padding;
             background: none;
         }
 
--- a/src/lib/viewers/doc/_docBase.scss
+++ b/src/lib/viewers/doc/_docBase.scss
@@ -295,6 +295,17 @@ $thumbnail-sidebar-width: 191px; // Extra pixel to account for sidebar border
         }
     }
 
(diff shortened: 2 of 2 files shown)


karelee7:
Could we add a short description about what issue this fixes with the PR?

src/lib/viewers/doc/_docBase.scss
```diff
@@ -295,6 +295,17 @@ $thumbnail-sidebar-width: 191px; // Extra pixel to account for sidebar border
         }
     }
 
+    //overwriting background image in pdf_viewer.css. Was causing 404's
+    /* stylelint-disable declaration-no-important */
+    .pdfViewer .page.loadingIcon {
+        background-image: none !important;
```
karelee7, line 301: Is `!important` the only way to do this? I know in general we try to avoid using this

src/lib/viewers/doc/_docBase.scss
```diff
@@ -295,6 +295,17 @@ $thumbnail-sidebar-width: 191px; // Extra pixel to account for sidebar border
         }
     }
 
+    //overwriting background image in pdf_viewer.css. Was causing 404's
+    /* stylelint-disable declaration-no-important */
+    .pdfViewer .page.loadingIcon {
```
jstoffan, line 300: Should there be a space between `.page` and `.loadingIcon`? It's a child element, right?

src/lib/viewers/doc/_docBase.scss
```diff
@@ -295,6 +295,17 @@ $thumbnail-sidebar-width: 191px; // Extra pixel to account for sidebar border
...
+        background-image: none !important;
+    }
+
+    .pdfViewer .page.loadingIcon::after {
```
jstoffan, line 304: We have other `.page .loadingIcon` styles a few lines up. Can we move these rules into the nested rules?

src/lib/viewers/doc/_docBase.scss
bfoxx1906: Believe it or not no. That's how pdf.js v3.6.7.2 has it. I can take out line 300 but I think we need the other one. Check this out https://github.com/mozilla/pdf.js/blob/v3.6.172/web/pdf_viewer.css#L162

src/lib/viewers/doc/_docBase.scss
bfoxx1906: done

src/lib/viewers/doc/_docBase.scss
bfoxx1906: yup reworked this

package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "box-content-preview",
-    "version": "2.101.0",
+    "version": "2.100.0",
```
jstoffan, line 3: Is this intentional? Seems like it would cause problems.

jstoffan (approved):

PR #1506, opened 2023-09-21
feat(remote-sync): Enable remote syncing based on an API response

TylerGauntlett:
### Why?

The current rsync.json file only supports a statically defined location. This works find for statically defined host/IP addresses but doesn't work well if the remote source goes through another proxy layer (ie K8 ingress, reverse proxy, etc)

### How?

To resolve this, I added the ability to define a host/IP address from the result of an API call. This expands on the current `rsync.json` file while preserving backwards compatablity.

### Example

```json
{
    "apiLocation": {
        "url": "https://example-website/api/info",
        "user": "root",
        "ip": "serverIpAddress",
        "path": "/var/www/assets/example-website"
    }
}

```

### DIFF
--- a/build/webpack.config.js
+++ b/build/webpack.config.js
@@ -3,20 +3,25 @@ const isDev = NODE_ENV === 'dev';
 const isProd = NODE_ENV === 'production';
 
 const fs = require('fs');
+const get = require('lodash/get');
 const path = require('path');
 const locales = require('@box/languages');
 const OptimizeCssAssetsPlugin = require('optimize-css-assets-webpack-plugin');
 const UglifyJsPlugin = require('uglifyjs-webpack-plugin');
 
+const { execSync } = require('child_process');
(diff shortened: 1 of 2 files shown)


build/webpack.config.js
```diff
@@ -78,6 +83,20 @@ function updateConfig(conf, language, index) {
         if (rsyncLocation) {
             config.plugins.push(new RsyncPlugin('dist/.', rsyncLocation));
         }
+
+        if (rsyncApiLocation) {
```
bfoxx1906, line 87: Should this be an else if? Is there a circumstance where we would need to support both cases?

TylerGauntlett: I don't see a use case right now for syncing to a API based and non API based host at the same time but left it supported. I imagine the normal use case would be either or though so I could update. 

Should I leave it as is or support both?

bfoxx1906 (approved):

PR #1508, opened 2023-09-25
feat(aci): use feature check for aci

arturfrombox:
This pr introduces changes to cover ACI codebase with feature checking

### DIFF
--- a/src/lib/Preview.js
+++ b/src/lib/Preview.js
@@ -1843,15 +1843,15 @@ class Preview extends EventEmitter {
      * Updates advanced content insights options after props have changed in parent app
      *
      * @public
-     * @param {Object} options - new content insights options value
+     * @param {Object} newOptions - new content insights options value
      * @return {void}
      */
-    updateContentInsightsOptions(options) {
-        this.options.features = { ...this.options.features, advancedContentInsights: options };
+    updateContentInsightsOptions(newOptions) {
+        this.options.features = { ...this.options.features, advancedContentInsights: newOptions };
         if (this.viewer && this.viewer.pageTracker) {
             this.previewOptions = {
                 ...this.previewOptions,
-                features: { ...this.previewOptions.features, advancedContentInsights: options },
+                features: { ...this.previewOptions.features, advancedContentInsights: newOptions },
             };
             this.viewer.pageTracker.updateOptions(this.previewOptions.features.advancedContentInsights);
         }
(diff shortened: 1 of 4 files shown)


src/lib/viewers/doc/DocBaseViewer.js
```diff
@@ -192,8 +194,11 @@ class DocBaseViewer extends BaseViewer {
         }
 
         this.updateDiscoverabilityResinTag();
-        if (this.options.advancedContentInsights) {
-            this.pageTracker = new PageTracker(this.options.advancedContentInsights, this.options.file);
+
+        const advancedContentInsightsConfig = getFeatureConfig(this.options.features, 'advancedContentInsights');
```
ahorowitz123, line 198: Same comment as above about `isFeatureEnabled`

src/lib/Preview.js
```diff
@@ -1845,9 +1851,13 @@ class Preview extends EventEmitter {
...
     updateContentInsightsOptions(options) {
+        this.options.features = { ...this.options.features, advancedContentInsights: options };
         if (this.viewer && this.viewer.pageTracker) {
-            this.previewOptions = { ...this.previewOptions, contentInsights: options };
-            this.viewer.pageTracker.updateOptions(this.previewOptions.contentInsights);
+            this.previewOptions = {
+                ...this.previewOptions,
+                features: { ...this.previewOptions.features, advancedContentInsights: options },
+            };
+            this.viewer.pageTracker.updateOptions(this.previewOptions.features.advancedContentInsights);
```
ahorowitz123, line 1860: Can't we just use `options` as the argument to the function here?

jstoffan (commented):
What's benefit of these changes if we already have a configuration object for these options? We appear to be increasing complexity at every layer.

src/lib/Preview.js
arturfrombox: can you explain why we should do so ?

arturfrombox:
> What's benefit of these changes if we already have a configuration object for these options? We appear to be increasing complexity at every layer.

It was one of PMAI and intention behind this change was to stick to "feature checking" mechanism implemented for Preview and to hide rest of ACI codebase under the feature flip.

arturfrombox:
@ahorowitz123 @jstoffan  thanks for your comments. I updated this PR

src/lib/Preview.js
loonskai: nitpick: I also found it a bit difficult to grasp having both `options` and `this.options`

ahorowitz123: I still find this confusing as well. Why not just call `this.viewer.pageTracker.updateOptions(newOptions)` here? It's much easier to read in my option and stays more consistent with the rest of this code block

ahorowitz123 (approved):

loonskai (approved):

ahorowitz123 (approved):

PR #1511, opened 2023-12-05
feat(error-codes): Add additional conversion error codes

patlm:
Support new messages for conversion error codes

### DIFF
--- a/src/i18n/en-US.properties
+++ b/src/i18n/en-US.properties
@@ -99,7 +99,7 @@ error_internet_explorer_office_online=Microsoft 365 apps and services no longer 
 # Preview error message shown when the preview fails because the file is too large for its extension type
 error_large_file=File exceeds supported size for preview. Open or download the file to view.
 # Preview error message shown when the preview fails because the file does not contain data
-error_zero_byte_file=Cannot preview 0kb file. Open or download the file to view.
+error_zero_byte_file=Cannot preview 0 KB file. Open or download the file to view.
 
 # Archive Preview
 # Label for filename column name
--- a/src/i18n/en-US.properties
+++ b/src/i18n/en-US.properties
@@ -96,6 +96,10 @@ error_not_downloadable=Oops! It looks like something is wrong with this file. We
 error_flash_not_enabled=We’re sorry, the preview didn’t load because Flash is not enabled on your browser. If possible, please enable Flash and refresh the page.
 # Preview error message shown when loading what would be an Office Excel Online file extension in Internet Explorer
 error_internet_explorer_office_online=Microsoft 365 apps and services no longer support Internet Explorer 11, therefore, Box users may have a degraded experience.
+# Preview error message shown when the preview fails because the file is too large for its extension type
+error_large_file=File exceeds supported size for preview. Open or download the file to view.
+# Preview error message shown when the preview fails because the file does not contain data
+error_zero_byte_file=Cannot preview 0kb file. Open or download the file to view.
 
 # Archive Preview
 # Label for filename column name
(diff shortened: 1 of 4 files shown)


src/lib/RepStatus.js
```diff
@@ -137,6 +137,12 @@ class RepStatus extends EventEmitter {
                     case ERROR_CODE.CONVERSION_FAILED:
                         errorMessage = __('error_reupload');
                         break;
+                    case ERROR_CODE.LARGE_FILE_SIZE:
+                        errorMessage = 'Placeholder - file too large';
```
jstoffan, line 141: Let's get final strings prior to merging so users don't see strange errors.

src/lib/events.js
```diff
@@ -32,6 +32,7 @@ export const ERROR_CODE = {
     GENERIC: 'error_generic',
     IMAGE_SIZING: 'error_image_sizing',
     INVALID_CACHE_ATTEMPT: 'error_invalid_file_for_cache',
+    LARGE_FILE_SIZE: 'error_large_size_file',
```
jstoffan, line 35: Is the issue file size or representation size? If the latter, should this be `CONVERSION_TOO_LARGE` or something?

src/lib/events.js
```diff
@@ -47,6 +48,7 @@ export const ERROR_CODE = {
     TOKEN_NOT_VALID: 'error_token_function_not_valid',
     UNSUPPORTED_FILE_TYPE: 'error_unsupported_file_type',
     VIEWER_LOAD_TIMEOUT: 'error_viewer_load_timeout',
+    ZERO_BYTE_FILE: 'error_zero_byte_file',
```
jstoffan, line 51: Same question as above. Is this a conversion issue or a file issue?

jstoffan (changes requested):

src/lib/events.js
patlm: This is an issue with the size of the original file. If it is larger than a certain size, a conversion to a PDF representation will not be attempted.

src/lib/events.js
patlm: This is also an issue with the original file. An example of this case would be creating an empty file with an extension like `empty.pptx`. If the file size is 0 bytes, an attempt to convert it to a different type won't take place.

src/lib/events.js
patlm: I will add the prefix, I think that makes sense.

src/i18n/en-US.properties
```diff
@@ -96,6 +96,10 @@ error_not_downloadable=Oops! It looks like something is wrong with this file. We
 error_flash_not_enabled=We’re sorry, the preview didn’t load because Flash is not enabled on your browser. If possible, please enable Flash and refresh the page.
 # Preview error message shown when loading what would be an Office Excel Online file extension in Internet Explorer
 error_internet_explorer_office_online=Microsoft 365 apps and services no longer support Internet Explorer 11, therefore, Box users may have a degraded experience.
+# Preview error message shown when the preview fails because the file is too large for its extension type
+error_large_file=File exceeds supported size for preview. Open or download the file to view.
```
jstoffan, line 100: Minor, but the language and tone of these messages isn't consistent with the existing errors. Something like the following would work:

```
We're sorry, the file size is too large to preview. Please open or download the file to view it.

File: src/i18n/en-US.properties [hunk 5]
```diff
@@ -96,6 +96,10 @@ error_not_downloadable=Oops! It looks like something is wrong with this file. We
...
+# Preview error message shown when the preview fails because the file does not contain data
+error_zero_byte_file=Cannot preview 0kb file. Open or download the file to view.
```
jstoffan, line 102: Minor, but the abbreviation should be `kB`.

jstoffan (dismissed):
Approving, but the language of the errors could be improved.

jstoffan (approved):

PR #1513, opened 2023-12-20
fix(build): Update package json build params

marcoartaviaq:
Upgrade webpack and babel-loader so we can get rid of `NODE_OPTIONS=--openssl-legacy-provider` flag


jstoffan (approved):
Thanks, @marcoartaviaq. Do we need to upgrade Jest as well?

marcoartaviaq:
> Thanks, @marcoartaviaq. Do we need to upgrade Jest as well?

Apologies for the delayed response, @jstoffan. The process of upgrading Jest, webpack (to version 5), and the Babel loader has proven to be more extensive than initially anticipated. To mitigate potential issues with local development, we're currently implementing a smaller-scale upgrade. The comprehensive upgrades will follow later. In the interim, we've incorporated the directive in the JPL to maintain functionality.

PR #1515, opened 2024-05-06
fix(polyfill): Remove IE11 polyfills

JChan106:
IE11 is no longer supported. This removes some polyfills needed for IE11 as those polyfills can cause other issues as we move forward.


jstoffan (commented):
Can we remove the entire `polyfill.js` file?

JChan106:
> Can we remove the entire `polyfill.js` file?

Looking at the PR's for the other shims, they don't seem to specify that they were only for IE11. I see that Number.isNaN, includes, and other functions arent unsupported by older Chrome, Firefox, Safari browsers. Need to check what specific browser versions we support.

JChan106:
> > Can we remove the entire `polyfill.js` file?
> 
> Looking at the PR's for the other shims, they don't seem to specify that they were only for IE11. I see that Number.isNaN, includes, and other functions arent unsupported by older Chrome, Firefox, Safari browsers. Need to check what specific browser versions we support.

Reflect.construct isn't supported on the older versions of Chrome, Firefox, and Safari as well https://caniuse.com/mdn-javascript_builtins_reflect_construct. Given that, if we move forward with the removal, we should be okay with removing the other polyfills as well.

JChan106:
Investigated the polyfilled functions using `caniuse.com`. The browsers (Chrome, Edge, Safari, Firefox) that don't support the functions all have less than a .1% global usage. Some examples:

Reflect.construct:
Chrome 34 - .01% usage
Chrome 38 - .02% usage
Chrome 47 - .01% usage
Chrome 48 - .02% usage
Safari 9.1 - .01% usage

Array.find:
Chrome 34 - .01% usage
Chrome 38 - .02% usage

Array.findIndex:
Chrome 34 - .01% usage
Chrome 38 - .02% usage

Array.includes:
Chrome 34 - .01% usage
Chrome 38 - .02% usage

String.endsWith:
Chrome 34 - .01% usage
Chrome 38 - .02% usage
Safari 8 - 0.1% usage

Number.isNaN:
Safari 8 - 0.1% usage

PerformanceObserver:
Edge 18 - .03% usage

Given this information, I think we are safe to remove the polyfill file. We'll just have to make sure we thoroughly QA test.

jstoffan:
I agree, we can likely safely remove the polyfill.

src/lib/__tests__/Preview-test.js
```diff
@@ -50,6 +50,10 @@ describe('lib/Preview', () => {
         preview.logger = new Logger();
         stubs = {};
         stubs.api = new Api();
+        window.PerformanceObserver = jest.fn(() => ({
```
JChan106, line 53: Mock needed, similar to PreviewPerf-test.js

jstoffan (approved):
LGTM. Please check with the team on if they want to release this as a major version.

bfoxx1906 (approved):
Looks good. Don't see any problems with this.  Let's just make sure all the tests pass. There could be some tests that rely on the polyfills.

PR #1517, opened 2024-05-20
feat(pdfjs): upgrade pdf.js to 4.3.136

ahorowitz123:
- Version 4 and greater use mjs files. Added support across the code base for this file type
- Updated pdf.js upgrade script
- Use legacy version of app to avoid errors due to the use of [promise.withResolvers](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/withResolvers) which was only recently supported by modern browsers
- Removed manual patches to new pdf.js versions which are no longer needed
- Added manual patch to fix a z-index issue that was conflicting with box-annotations and cause the first link on a pdf to be unaccessible. The issue here results from [this code](https://github.com/mozilla/pdf.js/blob/24e12d515dce2f11b392410dd6e6e9926270ac65/src/display/annotation_layer.js#L280) applying an increasing z index to each annotation layer. This causes the first annotation layer to be 0. When you are in region drawing mode, the `box-annotations` code applies a overlapping div over the whole pdf which sets user events to none. This div also has a z index of 0. When you have sibling divs with the same index, the div which is last in the DOM gets precedence. By ensuring that the z index of the annotation layer begins at 1, we can avoid this collision.
- Removed support for IE11
- Code has been manually tested with entire preview regression folder

Comparison of bundle sizes:
_Previous version_
279K Sep 18

_New version_
Screenshot 2024-05-28 at 4 47 01 PM


src/lib/viewers/doc/docAssets.js
```diff
@@ -3,11 +3,16 @@ import { DOC_LEGACY_STATIC_ASSETS_VERSION, DOC_STATIC_ASSETS_VERSION } from '../
 
 const STATIC_VERSION = Browser.isIE() ? DOC_LEGACY_STATIC_ASSETS_VERSION : DOC_STATIC_ASSETS_VERSION;
```
jstoffan, line 4: I realize it's not directly related to this change, but we can probably remove these legacy assets since we dropped support for IE11.

src/lib/constants.js
```diff
@@ -102,7 +102,7 @@ export const PDFJS_WIDTH_PADDING_PX = 40; // Should match SCROLLBAR_PADDING in p
 export const PDFJS_HEIGHT_PADDING_PX = 5; // Should match VERTICAL_PADDING in pdf_viewer.js
 
 export const DOC_LEGACY_STATIC_ASSETS_VERSION = '2.76.0'; // IE11 only
-export const DOC_STATIC_ASSETS_VERSION = '2.100.0'; // Modern browsers
+export const DOC_STATIC_ASSETS_VERSION = '4.2.67'; // Modern browsers
```
jstoffan, line 105: Historically, we've based these versions on the version of Preview SDK for which the new assets are being introduced.

build/upgrade_pdfjs.sh
```diff
@@ -24,52 +23,34 @@ mkdir ${DOC_STATIC_ASSETS_PATH}
...
-\cp -rf pdfjs-dist/legacy/build/pdf.js ${DOC_STATIC_ASSETS_PATH}
-\cp -rf pdfjs-dist/legacy/build/pdf.min.js ${DOC_STATIC_ASSETS_PATH}
-\cp -rf pdfjs-dist/legacy/build/pdf.worker.js ${DOC_STATIC_ASSETS_PATH}
-\cp -rf pdfjs-dist/legacy/build/pdf.worker.min.js ${DOC_STATIC_ASSETS_PATH}
-\cp -rf pdfjs-dist/legacy/web/pdf_viewer.css ${DOC_STATIC_ASSETS_PATH}
-\cp -rf pdfjs-dist/legacy/web/pdf_viewer.js ${DOC_STATIC_ASSETS_PATH}
-\cp -rf pdfjs-dist/cmaps ${DOC_STATIC_ASSETS_PATH}/cmaps
-rm -rf ./pdfjs-dist/
+
+curl https://cdnjs.cloudflare.com/ajax/libs/pdf.js/${DOC_STATIC_ASSETS_PDFJS_VERSION}/pdf.${JS_FILE_EXTENSION} -o  ${DOC_STATIC_ASSETS_PATH}/pdf.${JS_FILE_EXTENSION}
```
jstoffan, line 31: Could we download from the Github [assets](https://github.com/mozilla/pdf.js/releases/download/v4.2.67/pdfjs-4.2.67-dist.zip) of the canonical repository, instead?

build/upgrade_pdfjs.sh
```diff
@@ -5,10 +5,9 @@
 #this fix or fixes it on it's own
 
 
-DOC_COMPILER_BINARY="build/closure-compiler-v20200719.jar"
-DOC_STATIC_ASSETS_BRANCH=${1:-master}
-DOC_STATIC_ASSETS_VERSION=$(./build/current_version.sh)
```
jstoffan, line 10: This is the script that generates the correct asset version.

build/upgrade_pdfjs.sh
```diff
@@ -5,10 +5,9 @@
...
-DOC_STATIC_ASSETS_PATH="src/third-party/doc/${DOC_STATIC_ASSETS_VERSION}"
+DOC_STATIC_ASSETS_PDFJS_VERSION=${1}
+DOC_STATIC_ASSETS_PATH="src/third-party/doc/${DOC_STATIC_ASSETS_PDFJS_VERSION}"
+if  [[ ${1:0:1} -ge "4" ]]; then JS_FILE_EXTENSION="mjs"; else JS_FILE_EXTENSION="js"; fi
```
jstoffan, line 10: I don't think we need to support `.js` anymore. It's probably safe to assume the use of `mjs` going forward.

build/upgrade_pdfjs.sh
```diff
@@ -24,52 +23,34 @@ mkdir ${DOC_STATIC_ASSETS_PATH}
...
 echo "-----------------------------------------------------------------------------------"
-sed -e 's@var DEFAULT_CACHE_SIZE = 10;@var DEFAULT_CACHE_SIZE = /iphone|ipad|ipod|android|blackberry|bb10|mini|windows\sce|palm/i.test(navigator.userAgent) ? 5 : 10;@' -i '' ${DOC_STATIC_ASSETS_PATH}/pdf_viewer.js
+sed -e 's@var DEFAULT_CACHE_SIZE = 10;@var DEFAULT_CACHE_SIZE = /iphone|ipad|ipod|android|blackberry|bb10|mini|windows\sce|palm/i.test(navigator.userAgent) ? 5 : 10;@' -i '' ${DOC_STATIC_ASSETS_PATH}/pdf_viewer.${JS_FILE_EXTENSION}
 
 # Render e-signatures without validation
 echo "-----------------------------------------------------------------------------------"
 echo "Enabling e-signature rendering without validation..."
 echo "-----------------------------------------------------------------------------------"
-sed -e 's@;r.setFlags(o.AnnotationFlag.HIDDEN)@@' -i '' ${DOC_STATIC_ASSETS_PATH}/pdf.worker.min.js
+sed -e 's@;r.setFlags(o.AnnotationFlag.HIDDEN)@@' -i '' ${DOC_STATIC_ASSETS_PATH}/pdf.worker.min.${JS_FILE_EXTENSION}
```
jstoffan, line 49: Do all our patches still work?

jstoffan (changes requested):
We should figure out what impact this change is going to have on performance, especially if the new viewer assets are significantly larger than the old ones.

build/upgrade_pdfjs.sh
ahorowitz123: Only issue here is the build doesn't contain the minified versions of the build files. So we'd have to handle that ourselves

build/verify_custom_updates.sh
```diff
@@ -1,8 +1,9 @@
 #!/bin/bash
 
 VERSION="`grep -o 'DOC_STATIC_ASSETS_VERSION.*' ./src/lib/constants.js | grep -o "'.*'" | sed -e "s/'//g"`"
-PIXEL_FIX_MISSING="`grep -c '.PixelsPerInch.PDF_TO_CSS_UNITS);return void 0!==' ./src/third-party/doc/$VERSION/pdf.min.js`"
-READ_ONLY_FIELD_FIX="`grep -c 'class TextWidgetAnnotation extends WidgetAnnotation{constructor(e){super(e);this.data.hasOwnCanvas=this.data.readOnly' ./src/third-party/doc/$VERSION/pdf.worker.min.js`"
+if  [[ ${VERSION:0:1} -ge "4" ]]; then JS_FILE_EXTENSION="mjs"; else JS_FILE_EXTENSION="js"; fi
+PIXEL_FIX_MISSING="`grep -c '.PixelsPerInch.PDF_TO_CSS_UNITS);return void 0!==' ./src/third-party/doc/$VERSION/pdf.min.${JS_FILE_EXTENSION}`"
```
bfoxx1906, line 5: Did we fix this in the code?

src/lib/constants.js
bfoxx1906: Yes sorry about that. What I sent was not for production. Yes let's stick with the way we currently do it.

build/upgrade_pdfjs.sh
```diff
@@ -1,12 +1,11 @@
...
+# Run with specific version and branch/tag (e.g. ./upgrade_pdfjs.sh <version> tags/v2.2.228) or with no second argument to use master
 
 #IMPORTANT!!!! See https://github.com/box/box-content-preview/pull/1466 and make sure any updated version of pdf.js includes
 #this fix or fixes it on it's own
 
-
-DOC_COMPILER_BINARY="build/closure-compiler-v20200719.jar"
-DOC_STATIC_ASSETS_BRANCH=${1:-master}
+DOC_STATIC_ASSETS_PDFJS_VERSION=${1}
+DOC_STATIC_ASSETS_BRANCH=${2:-master}
```
jstoffan, line 8: Is this unused now? Can we update the comments in this document, as well?

build/upgrade_pdfjs.sh
```diff
@@ -24,52 +23,40 @@ mkdir ${DOC_STATIC_ASSETS_PATH}
...
 \cp -R src/third-party/doc/${DOC_CURRENT_ASSETS_VERSIONS}/exif.min.js ${DOC_STATIC_ASSETS_PATH}/
 \cp -R src/third-party/doc/${DOC_CURRENT_ASSETS_VERSIONS}/images ${DOC_STATIC_ASSETS_PATH}/images
-
-echo "-----------------------------------------------------------------------------------"
-echo "Cloning pdfjs-dist repo at branch: $DOC_STATIC_ASSETS_BRANCH..."
-echo "-----------------------------------------------------------------------------------"
-rm -rf ./pdfjs-dist/
-git clone https://github.com/mozilla/pdfjs-dist.git --depth 1 --single-branch --branch ${DOC_STATIC_ASSETS_BRANCH}
-
+\cp -R src/third-party/doc/${DOC_CURRENT_ASSETS_VERSIONS}/cmaps ${DOC_STATIC_ASSETS_PATH}/cmaps
```
jstoffan, line 26: Do we need a solution to retrieve updated `cmaps`, as well? We're only copying over the previous version at the moment.

package.json
```diff
@@ -9,7 +9,6 @@
         "url": "git@github.com:box/box-content-preview.git"
     },
     "devDependencies": {
-        "@babel/core": "^7.18.6",
```
jstoffan, line 12: Do we not still need this for Preview's build?

src/lib/util.js
```diff
@@ -136,7 +140,11 @@ export function createPrefetch(url, preload = false) {
     link.href = url;
 
     if (preload) {
-        link.as = url.indexOf('.js') !== -1 ? 'script' : 'style';
+        link.as = url.indexOf('.css') !== -1 ? 'style' : 'script';
```
jstoffan, line 143: Minor, but should we keep this as-is and add `.mjs` as a second condition, instead?

package.json
ahorowitz123: Good catch, I thought I added this when I was testing out babel CLI to help minify the file. I'll add back.

build/upgrade_pdfjs.sh
ahorowitz123: Did some more research and it looks like https://cdn.jsdelivr.net/npm/pdfjs-dist/ offers everything we need including `cmaps` and a minified version of `pdf_viewer`. I'm going to switch to that

jstoffan: I'm concerned about introducing a net new third-party into our artifact chain of custody. Do we use that service anywhere else?

src/lib/viewers/doc/DocBaseViewer.js
```diff
@@ -738,7 +738,6 @@ class DocBaseViewer extends BaseViewer {
             disableFontFace,
             disableRange,
             disableStream,
-            isEvalSupported: false,
```
ahorowitz123, line 741: This was a workaround for [this](https://github.com/advisories/GHSA-wgrm-67xf-hhpq) security issue which is patched in version 4.2.67.

src/lib/viewers/doc/_docBase.scss
```diff
@@ -380,14 +380,15 @@ $thumbnail-sidebar-width: 191px; // Extra pixel to account for sidebar border
...
-            display: none;
-        }
-
         .popup {
             font-size: inherit;
+
+            .header {
+                height: unset;
+                /* stylelint-disable declaration-no-important */
+                padding: unset !important;
```
ahorowitz123, line 389: The css this is conflicting with contains a lot of nested classes. In order to not muddy up this CSS I opted to go with `!important`

build/upgrade_pdfjs.sh
ahorowitz123: I've opted to go with https://cdn.jsdelivr.net/npm/pdfjs-dist/ here. cdn.cloudflare didn't offer the legacy build package, and the prebuild from github didn't include `pdf_viewer.mjs`.

I think its relatively safe given that its a recommended cdn to use by pdf.js and we only use it during the upgrade script. So we are not dependent on it to run the preview sdk

ahorowitz123:
> We should figure out what impact this change is going to have on performance, especially if the new viewer assets are significantly larger than the old ones.

I've found this difficult to test in dev due to the large fluctuations in loading times due to dev infrastructure. I think we'll get a better idea in staging

src/lib/viewers/doc/DocBaseViewer.js
```diff
@@ -738,7 +738,6 @@ class DocBaseViewer extends BaseViewer {
             disableFontFace,
             disableRange,
             disableStream,
```
bfoxx1906, line 740: I would leave isEvalSupported:false here. It will protect us from any other vulnerabilities that might popup because of this. It's a supported parameter and it should always be false. Did the default value change to false in pdf.js?

src/lib/viewers/doc/docAssets.js
```diff
@@ -1,13 +1,18 @@
-import Browser from '../../Browser';
-import { DOC_LEGACY_STATIC_ASSETS_VERSION, DOC_STATIC_ASSETS_VERSION } from '../../constants';
+import { DOC_STATIC_ASSETS_VERSION } from '../../constants';
 
-const STATIC_VERSION = Browser.isIE() ? DOC_LEGACY_STATIC_ASSETS_VERSION : DOC_STATIC_ASSETS_VERSION;
-const STATIC_URI = `third-party/doc/${STATIC_VERSION}`;
+const STATIC_URI = `third-party/doc/${DOC_STATIC_ASSETS_VERSION}`;
+const versionParts = DOC_STATIC_ASSETS_VERSION.split('.').map(part => parseInt(part, 10));
+const JS_EXTENSTION =
```
bfoxx1906, line 5: What if we set JS_EXTENSTION to 'mjs' by default and just add a comment saying why? If we had to revert to a previous version of pdf.js we would have to test and push out a new version anyway and tests would not pass if this value was not set correctly e.g. back to .js . This just eliminates some of the logic and simplifies the code a bit. Thoughts?

ahorowitz123: Yeah I like this approach. I don't love the version check either because the version numbers are kind of arbitrary. I'll just remove the `JS_EXTENSION` variable completely

bfoxx1906 (dismissed):
Looks good. Let's make let jstoffan take a last look too but I think we should be good to go.

build/upgrade_pdfjs.sh
```diff
@@ -1,77 +1,74 @@
...
-./node_modules/.bin/cssnano ${DOC_STATIC_ASSETS_PATH}/pdf_viewer.css ${DOC_STATIC_ASSETS_PATH}/pdf_viewer.min.css
+curl -L https://github.com/mozilla/pdf.js/releases/download/v${DOC_STATIC_ASSETS_PDFJS_VERSION}/pdfjs-${DOC_STATIC_ASSETS_PDFJS_VERSION}-legacy-dist.zip -o pdfjs-dist.zip
+unzip -q -d pdfjs-dist pdfjs-dist.zip
+\cp -rf pdfjs-dist/web/cmaps ${DOC_STATIC_ASSETS_PATH}/cmaps
+rm -rf pdfjs-dist
+rm pdfjs-dist.zip
 
 echo "-----------------------------------------------------------------------------------"
-echo "Successfully updated and minified pdf.js files!"
+echo "Successfully updated pdf.js files!"
```
bfoxx1906, line 73: Do we need to add a reminder to run the  command for the external link fix?

build/upgrade_pdfjs.sh
```diff
@@ -1,77 +1,74 @@
 #!/bin/bash
-# Run with specific branch/tag (e.g. ./upgrade_pdfjs.sh tags/v2.2.228) or with no arguments to use master
+# Run with specific version (e.g. ./upgrade_pdfjs.sh <version>)
 
 #IMPORTANT!!!! See https://github.com/box/box-content-preview/pull/1466 and make sure any updated version of pdf.js includes
```
bfoxx1906, line 4: We might not need this anymore since pdf.js fixed it.

build/verify_custom_updates.sh
```diff
@@ -1,20 +1,15 @@
...
-    echo "-----------------------------------------"
-    echo "Read only form field fix not implemented."
-    echo "Run: sed -i '' -e 's/class TextWidgetAnnotation extends WidgetAnnotation{constructor(e){super(e);this.data.hasOwnCanvas=this.data.readOnly&&!this.data.noHTML/class TextWidgetAnnotation extends WidgetAnnotation{constructor(e){super(e);this.data.hasOwnCanvas=false/' ./src/third-party/doc/<version>/pdf.worker.min.js"
-    echo "------------------------------------------"
-    FAILED=true
+if [[ $EXTERNAL_LINK_FIX_MISSING != 0 ]];then
+     echo "-----------------------------------------"
+        echo "External link fix not implemented."
+        echo "Verify that the fix is still needed by loading a pdf with multiple external links and confirm the first one is clickable."
+        echo "To fix, run: sed -i '' -e 's/this instanceof WidgetAnnotationElement||(n.tabIndex=Qt);const{style:r}=n;r.zIndex=this.parent.zIndex++;/this instanceof WidgetAnnotationElement||(n.tabIndex=Qt);const{style:r}=n;r.zIndex=++this.parent.zIndex;/' ./src/third-party/doc/${VERSION}/pdf.min.mjs"
```
bfoxx1906, line 11: This command still works with the terser minified version right?

bfoxx1906 (commented):
Looks good. Just left a few comments.

build/upgrade_pdfjs.sh
ahorowitz123: Yeah I can add that. The test also gets run as a commit hook

build/verify_custom_updates.sh
ahorowitz123: Yes

bfoxx1906 (approved):
approved

PR #1518, opened 2024-05-20
feat: Disable eval support for documents by default

patlm:


bfoxx1906:
@Mergifyio queue default

PR #1521, opened 2024-05-29
fix(annotations): Fix annotation region in mobile

marcoartaviaq:
Fix the creation of annotation regions when enableAnnotationsDiscoverability is set to false. When `enableAnnotationsDiscoverability` is true, the class `bp-annotations-discoverable` is added. This class sets `pointer-events: none` on the` .textLayer `element (see [source](https://github.com/box/box-content-preview/blob/master/src/lib/viewers/doc/_docBase.scss#L419)) to prevent the text layer from intercepting pointer events intended for the region creator. This PR introduces `pointer-events: none` to the `.textLayer `element when the region creation button is enabled (this causes the class `bp-annotations-create--region` to be [added](https://github.com/box/box-content-preview/blob/master/src/lib/viewers/BaseViewer.js#L1148)), allowing users to comment by region even when discoverability is disabled.

Behavior before the update:

https://github.com/box/box-content-preview/assets/2131354/928e21ab-845d-411e-89d4-1f960cbe83f5

Behavior after the update:

https://github.com/box/box-content-preview/assets/2131354/24515117-8412-4d06-8cda-37eb7f447275

Behavior in other non mobile browsers:

Firefox:

https://github.com/box/box-content-preview/assets/2131354/de7f72e4-e73e-4700-914a-2e6ea215c294

Chrome: 

https://github.com/box/box-content-preview/assets/2131354/2db2f96f-a460-4d93-9c2d-d7802fbcf46d

Safari:

https://github.com/box/box-content-preview/assets/2131354/e9fd43f2-5807-4539-91f1-26fdb38c8a6a

Edge:

https://github.com/box/box-content-preview/assets/2131354/782a48e4-8505-4917-8484-02ee6d381893


bfoxx1906 (approved):
Fix makes sens. Approving. Let's just make sure that we do regression testing for annotation creation on the other browsers in a non-mobile context as well.

ahorowitz123 (approved):

PR #1531, opened 2024-06-28
fix(preview): Fix highlighting text when performing click and drag

bkepka-box:
Fix highlighting text in Preview. This bug was introduced as a regression during PDF.js upgrade.

### Root cause description
Inside this [commit](https://github.com/mozilla/pdf.js/commit/1b00511301840d413f3369f834999ee44d3dfa8d#diff-90b9c7e02c41b4051095250577f8dd385455429d67abe6bf360e13ef17d7fe19L26) PDF.js team removed `z-index:2` from `.textLayer` class which is causing a bug that user cannot highlight text when annotations are enabled.

### Screenshots
2.108.0 version (current):
image

2.107.0 version:
image

### DIFF
--- a/src/lib/viewers/doc/_docBase.scss
+++ b/src/lib/viewers/doc/_docBase.scss
@@ -322,10 +322,11 @@ $thumbnail-sidebar-width: 191px; // Extra pixel to account for sidebar border
     }
 
     .textLayer {
-        caret-color: black; // Required for caret navigation on transparent text
         top: $pdfjs-page-padding;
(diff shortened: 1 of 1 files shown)


bkepka-box:
We need to set `z-index` to `1` because with `2` it breaks annotation links (for example internal link to page like in the test/integration/document/PresentationViewer.e2e.test.js)

dsporysz-box (approved):

loonskai (approved):

PR #1535, opened 2024-09-06
chore(tests): Migrate from Enzyme to RTL (1st batch)

loonskai:
### Description

Tests with Enzyme migrated: **23/61**

This PR is a part of Box-wide migration to the newer React v18. 
- The most of the scenarios were rewritten with consideration of RTL practices (thanks @nmkedziora for pointing this out). Some components required `fireEvent` executed on top of specific selectors instead of recommended `userEvent`. Some of these tradeoffs could be avoided (possibly) but would extend the timelines of this PR. I suggest improve these test iteratively as we work on them.
- Semantic selectors were used in most cases. `data-testid` attributes were added mostly for container elements to quickly validate if container is rendered.
- This PR doesn't introduce any runtime changes (except `data-testid` changes mentioned above)

This PR will be followed with the next steps:
- Rewrite the rest of test scenarios from Enzyme to RTL and cleanup Enzyme-related setup (dependencies, adapter etc.)
- Add [eslint-plugin-testing-library](https://github.com/testing-library/eslint-plugin-testing-library) and [eslint-plugin-jest-dom](https://github.com/testing-library/eslint-plugin-jest-dom) as ESLint rules

### Screenshots
image


src/lib/viewers/box3d/image360/__tests__/Image360Controls-test.tsx
```diff
@@ -11,11 +10,10 @@ describe('lib/viewers/box3d/image360/Image360Controls', () => {
         onVrToggle: jest.fn(),
     });
 
-    const getWrapper = (props: Partial<Props>): ShallowWrapper =>
-        shallow(<Image360Controls {...getDefaults()} {...props} />);
+    const getWrapper = (props: Partial<Props>) => render(<Image360Controls {...getDefaults()} {...props} />);
```
nmkedziora, line 13: Do not use `wrapper` concept when working with React Testing Library. 

> _The name `wrapper` is old cruft from enzyme and we don't need that here. The return value from `render` is not "wrapping" anything. It's simply a collection of utilities that (thanks to the next thing) you should actually not often need anyway._

Source:[ Common mistakes with React Testing Library](https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#using-wrapper-as-the-variable-name-for-the-return-value-from-render) by Kent C.Dodds

I highly recommend to get familiar with this article.

src/lib/viewers/box3d/image360/__tests__/Image360Controls-test.tsx
```diff
@@ -24,11 +22,10 @@ describe('lib/viewers/box3d/image360/Image360Controls', () => {
                 onVrToggle,
             });
 
-            expect(wrapper.find(VrToggleControl).props()).toMatchObject({
-                isVrShown: false,
-                onVrToggle,
-            });
-            expect(wrapper.find(FullscreenToggle).prop('onFullscreenToggle')).toEqual(onFullscreenToggle);
+            await expect(wrapper.queryByTitle('Toggle VR display')).toBe(null);
```
nmkedziora, line 25: Ditto. You do not need wrapper to find elements that were rendered - use `screen` instead. 
Also, I think all the `await` could probably omitted.

Source: [Not using screen](https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#not-using-screen)

src/lib/viewers/box3d/image360/__tests__/Image360Controls-test.tsx
```diff
@@ -24,11 +22,10 @@ describe('lib/viewers/box3d/image360/Image360Controls', () => {
...
+
+            await wrapper.queryByTitle('Enter fullscreen')?.click();
```
nmkedziora, line 27: What is `click()`?
Recommended way of working with events is to use `fireEvent` (a few use cases) and `userEvent` (majority of the time) [@testing-library/user-event](https://testing-library.com/docs/user-event/intro).

Source: [Firing Events at Testing Library docs](https://testing-library.com/docs/dom-testing-library/api-events/)

src/lib/viewers/box3d/image360/__tests__/Image360Controls-test.tsx
nmkedziora: > _The base queries from DOM Testing Library require you to pass a container as the first argument. Most framework-implementations of Testing Library provide a pre-bound version of these queries when you render your components with them which means you do not have to provide a container. In addition, if you just want to query document.body then you can use the [screen](https://testing-library.com/docs/queries/about/#screen) export as demonstrated below (**using screen is recommended**)._

Source: [Using Queries](https://testing-library.com/docs/queries/about/#using-queries)

src/lib/viewers/archive/__tests__/ArchiveExplorer-test-react.js
```diff
@@ -65,218 +78,124 @@ describe('lib/viewers/archive/ArchiveExplorer', () => {
...
+        test('should render breadcrumbs correctly', async () => {
+            renderView({ filename, itemCollection: data });
 
-    describe('getItemList()', () => {
-        test('should return correct item list', () => {
-            const component = getComponent({ filename, itemCollection: data });
+            await userEvent.click(screen.getByText('test'));
 
-            let itemList = component.instance().getItemList(data, ROOT_FOLDER);
+            expect(within(screen.getByLabelText('Breadcrumb')).getByText('test.zip')).toBeInTheDocument();
```
PaQ666, line 110: :thought_balloon: Isn't it enough to just search for this element in the document/screen which would fail if it does't exist?

src/lib/viewers/box3d/model3d/__tests__/Model3DControlsNew-test.tsx
```diff
@@ -68,36 +70,28 @@ describe('lib/viewers/box3d/model3d/Model3DControlsNew', () => {
...
-                onShowGridToggle,
-                onShowSkeletonsToggle,
-                onShowWireframesToggle,
-                renderMode: RenderMode.LIT,
-                showGrid: true,
-                showSkeletons: false,
-                showWireframes: false,
-            });
-            expect(wrapper.find(FullscreenToggle).prop('onFullscreenToggle')).toEqual(onFullscreenToggle);
+            await userEvent.click(screen.getByTitle('Reset'));
```
PaQ666, line 73: Tool creators suggest to initialise userEvent before tests: https://testing-library.com/docs/user-event/intro
Not sure if relevant, just noticed it while checking docs.

src/lib/viewers/controls/settings/__tests__/SettingsList-test.tsx
```diff
@@ -1,104 +1,68 @@
...
-        test('should return a valid wrapper', () => {
-            const onKeyDown = jest.fn();
-            const wrapper = getWrapper({ onKeyDown });
-            const element = wrapper.getDOMNode();
-
-            expect(element).toHaveClass('bp-SettingsList');
-            expect(element).toHaveAttribute('role', 'listbox');
-            expect(element).toHaveAttribute('tabIndex', '0');
+            expect(screen.getByTestId('test1')).toHaveFocus();
+            expect(screen.getByTestId('test2')).not.toHaveFocus();
```
rafalmaksymiuk, line 64: Isn't it redundant? I mean in DOM only one element can have focus?

src/lib/viewers/controls/settings/__tests__/SettingsDropdown-test.tsx
```diff
@@ -17,206 +15,115 @@ describe('SettingsDropdown', () => {
...
-            const element = wrapper.getDOMNode();
-
-            expect(element).toHaveClass('bp-SettingsDropdown');
-            expect(wrapper.find('.bp-SettingsDropdown-label').text()).toBe('Dropdown Label');
-            expect(wrapper.find('.bp-SettingsDropdown-button').text()).toBe('first');
-            expect(wrapper.exists(SettingsFlyout));
-            expect(wrapper.exists(SettingsList));
+            waitFor(() => {
+                expect(document.activeElement).toBe(screen.getByRole('button', { name: /first/i }));
+            });
```
rafalmaksymiuk, line 126: Maybe to properly test it we should actual user event like click that focuses on element as well? 
If you await for such event then `waitFor` might not be needed?

rafalmaksymiuk:
This is a massive PR - tremendous work but hard to review. I know it is difficult to push each PR separately, but is way easier to meaningfully review so maybe split more next time :)?

Another thought - or rather question. I do know that EUA uses special tooling based on RTL. Don't remember details, but it integrates Redux somehow. Do you know why it was needed there and if it might be here?

rafalmaksymiuk (dismissed):

src/lib/viewers/archive/__tests__/ArchiveExplorer-test-react.js
nmkedziora: Do you mean this 👇  @rafalmaksymiuk ?

```suggestion
            expect(screen.getByText('test.zip')).toBeInTheDocument();
```

If yes, I think this shortened version is much more readable.

src/lib/viewers/controls/settings/__tests__/SettingsDropdown-test.tsx
nmkedziora: I agree. Currently you test (native) focus implementation.
Instead, I'd do sth like in this [example](https://testing-library.com/docs/user-event/convenience/): 

```
 test('should be able to focus on the dropdown button', async() => {
            const user = userEvent.setup()
            renderView();

            const button = screen.getByRole('button', { name: /first/i });

           await user.tab() 

           expect(button).toHaveFocus();
        });
```

nmkedziora:
> This is a massive PR - tremendous work but hard to review. I know it is difficult to push each PR separately, but is way easier to meaningfully review so maybe split more next time :)?

+1 for me. 
Amazing job and great effort 🎉 

Splitting the PR by feature or folder would make a significant difference. It would not only make it easier to review but also quicker to merge bit by bit instead of blocking the merge for the entire PR due to comments on a single file.

src/lib/viewers/archive/__tests__/Breadcrumbs-test-react.js
```diff
@@ -1,10 +1,10 @@
 import React from 'react';
-import { shallow } from 'enzyme';
+import { render, screen } from '@testing-library/react';
 import Breadcrumbs from '../Breadcrumbs';
-import { ROOT_FOLDER, VIEWS } from '../constants';
+import { VIEWS } from '../constants';
 
 describe('lib/viewers/archive/Breadcrumbs', () => {
-    const getComponent = props => shallow(<Breadcrumbs {...props} />);
+    const renderView = props => render(<Breadcrumbs {...props} />);
```
nmkedziora, line 7: Nit: It's a small and simple render, so I think using `render` directly would be more readable. I highly recommend not using helper functions like that whenever possible or trying to name the helper function in a more meaningful manner, i.e. `renderBreadcrumb`. `renderView` is too generic IMHO, and because of that I need to check the helper implementation in order to understand the test itself.

src/lib/viewers/archive/__tests__/Breadcrumbs-test-react.js
```diff
@@ -20,52 +20,19 @@ describe('lib/viewers/archive/Breadcrumbs', () => {
...
-            const component = getComponent({ filename, fullPath, onClick, view });
+            renderView({ filename, fullPath, onClick, view });
 
-            expect(component.find('.bp-Breadcrumbs').length).toBe(1);
-            expect(component.find('Breadcrumb').length).toBe(1);
-            expect(component.find('PlainButton').length).toBe(3);
+            expect(screen.getByLabelText('Breadcrumb')).toBeInTheDocument();
+            expect(screen.getAllByRole('button')).toHaveLength(3);
+            expect(screen.getAllByRole('button').at(0).textContent).toContain('test.zip');
+            expect(screen.getAllByRole('button').at(1).textContent).toContain('test');
```
nmkedziora, line 28: Nit: you could do `screen.getAllByRole('button')` only once:
```suggestion
            const buttons = screen.getAllByRole('button')
            expect(buttons.at(1).textContent).toContain('test');
```

src/lib/viewers/archive/__tests__/ArchiveExplorer-test-react.js
rafalmaksymiuk: Rather simple `within(screen.getByLabelText('Breadcrumb')).getByText('test.zip')` which would serve same purpose, but then there is good argument here about the intention:
https://stackoverflow.com/questions/73281448/why-we-have-to-use-tobeinthedocument-in-testing-library

loonskai: @rafalmaksymiuk yes, I referred to the same recommendation

rafalmaksymiuk (approved):

src/lib/viewers/media/__tests__/DashControls-test.tsx
```diff
@@ -48,67 +19,189 @@ describe('DashControls', () => {
...
+            expect(onVolumeChange).toHaveBeenCalledTimes(1);
         });
 
-        test.each([true, false])('should set isHDSupported prop on MediaSettings as %s', isHDSupported => {
-            const wrapper = getWrapper({ isHDSupported });
-            expect(wrapper.find(MediaSettings).prop('isHDSupported')).toBe(isHDSupported);
+        test.each([true, false])('should set isHDSupported prop on MediaSettings as %s', async isHDSupported => {
+            const user = userEvent.setup();
+            render(
+                <DashControls
```
bkepka-box, line 62: Should we extract it in order to not duplicate setting props in every test case?

bkepka-box (approved):

PR #1536, opened 2024-10-03
feat(preview-error): Display file icon based on extension for errors

patlm:
**Description**
Remove restriction such that preview will fetch the icon based on the extension from box-ui-elements when there is a preview error. Will rely on the logic there to return a default icon when one does not exist for the extension.

See: https://github.com/box/box-ui-elements/blob/1850977829e4888139462a6f65794b1cffc52a08/src/components/preview/previewIcons.ts#L183

**Screenshots**
image
image-1

### DIFF
--- a/src/lib/viewers/error/PreviewErrorViewer.js
+++ b/src/lib/viewers/error/PreviewErrorViewer.js
@@ -80,8 +80,7 @@ class PreviewErrorViewer extends BaseViewer {
 
         const { displayMessage, details, message } = error;
         const { file } = this.options;
-        const { extension } = file || {};
-        const iconExtension = extension === 'flv' || extension === 'tgz' || extension === 'zip' ? extension : undefined;
+        const { extension: iconExtension } = file || {};
 
         // Display the default or file-specific error icon
         this.errorIcon = new ErrorIcon({ containerEl: this.iconEl });
--- a/src/lib/viewers/error/__tests__/PreviewErrorViewer-test.js
+++ b/src/lib/viewers/error/__tests__/PreviewErrorViewer-test.js
@@ -55,7 +55,8 @@ describe('lib/viewers/error/PreviewErrorViewer', () => {
             ['zip', 'zip'],
             ['tgz', 'tgz'],
             ['flv', 'flv'],
-            ['blah', undefined],
+            ['pdf', 'pdf'],
+            ['pptx', 'pptx'],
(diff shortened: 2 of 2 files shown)


src/lib/viewers/error/PreviewErrorViewer.js
```diff
@@ -80,8 +80,7 @@ class PreviewErrorViewer extends BaseViewer {
 
         const { displayMessage, details, message } = error;
         const { file } = this.options;
-        const { extension } = file || {};
-        const iconExtension = extension === 'flv' || extension === 'tgz' || extension === 'zip' ? extension : undefined;
+        const { extension: iconExtension } = file || {};
```
jmcbgaston, line 83: what happens if `iconExtension` is undefined?

src/lib/viewers/error/__tests__/PreviewErrorViewer-test.js
```diff
@@ -55,7 +55,8 @@ describe('lib/viewers/error/PreviewErrorViewer', () => {
             ['zip', 'zip'],
             ['tgz', 'tgz'],
             ['flv', 'flv'],
-            ['blah', undefined],
+            ['pdf', 'pdf'],
+            ['pptx', 'pptx'],
```
jmcbgaston, line 59: is it still possible for `[blah, undefined]` to happen?

patlm: It will no longer be possible. If 'blah' is the extension, it will be passed into box-ui-elements [getIcon](https://github.com/box/box-ui-elements/blob/1850977829e4888139462a6f65794b1cffc52a08/src/components/preview/previewIcons.ts#L183) function which in the case where it can not find it, it will return the default icon that was previous being used when iconExtension was passed in as undefined.

src/lib/viewers/error/PreviewErrorViewer.js
patlm: These scenarios will be handled by [getIcon](https://github.com/box/box-ui-elements/blob/1850977829e4888139462a6f65794b1cffc52a08/src/components/preview/previewIcons.ts#L183) in box-ui-elements. In the case of undefined, it will return the same default item as it returned before when iconExtension was explicitly set to undefined.

jmcbgaston (approved):
LGTM

jmcbgaston (approved):
LGTM

JChan106 (approved):

PR #1537, opened 2024-10-18
fix(docs): Remove gzip encoding query param 

minkovich:
Box will automatically use accept-encoding header

This fixes a bug where gzip encoding wasn't available in accept-encoding list

### DIFF
--- a/src/lib/viewers/doc/DocBaseViewer.js
+++ b/src/lib/viewers/doc/DocBaseViewer.js
@@ -26,20 +26,11 @@ import {
     CLASS_IS_SCROLLABLE,
     DISCOVERABILITY_ATTRIBUTE,
     DOCUMENT_FTUX_CURSOR_SEEN_KEY,
-    ENCODING_TYPES,
     PERMISSION_DOWNLOAD,
     PRELOAD_REP_NAME,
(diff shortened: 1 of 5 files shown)


src/lib/viewers/doc/DocBaseViewer.js
```diff
@@ -723,13 +718,6 @@ class DocBaseViewer extends BaseViewer {
         // Disable streaming by default unless it is explicitly enabled via options
         const disableStream = this.getViewerOption('disableStream') !== false;
 
-        // If range requests and streaming are disabled, request the gzip compressed version of the representation
-        this.encoding = disableRange && disableStream ? ENCODING_TYPES.GZIP : undefined;
-
-        if (this.encoding) {
-            queryParams[QUERY_PARAM_ENCODING] = this.encoding;
```
jstoffan, line 730: I think we can delete `queryParams` and remove the `appendQueryParams` call now, since it's a noop.

jstoffan (dismissed):
LGTM. I've updated the PR title, but you may need to update the commit message as well.

jstoffan (approved):
Thanks for submitting this change!

PR #1545, opened 2025-03-05
feat(nodejs20): WEBAPP-30462 Node updated to 20

rafalmaksymiuk:
Updated project to work with Nodejs 20.

Bumped resolution of `acorn` to most recent as with node update it's older version stops working properly with optional chaining operators. The proper solution is to update Webpack to next version which should be handled in another task.

Removed IE 11 compatibility. Due to side effect, constructor mocks with arrow functions stopped working (were transcribed before).

### DIFF
--- a/.nvmrc
+++ b/.nvmrc
@@ -1,1 +1,1 @@
-lts/hydrogen
+lts/iron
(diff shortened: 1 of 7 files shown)


loonskai:
@rafalmaksymiuk looks like the commits lint has failed, I think the format should be different https://git.dev.box.net/Box/frontend/tree/master/commitlint#commit-message-format

loonskai (approved):

PR #1547, opened 2025-04-14
Add thumbnail click functionality

bfoxx1906:
This PR adds click handling to the thumbnails generated by the preloader. When I thumbnail is clicked it will scroll the viewing pane to the corrsponding doc first page mimicking the behavior when the pdf is actually loaded with pdf.js. The scroll position is maintained so the jump between the doc first pages and the actual pdf content should not be very noticeable.

### DIFF
--- a/src/lib/ThumbnailsSidebar.js
+++ b/src/lib/ThumbnailsSidebar.js
@@ -138,7 +138,7 @@ class ThumbnailsSidebar {
         this.pdfViewer = null;
         this.currentThumbnails = [];
         this.currentPage = null;
-
+        this.preloader = null;
         this.anchorEl.removeEventListener('click', this.thumbnailClickHandler);
     }
 
--- a/src/lib/ThumbnailsSidebar.js
+++ b/src/lib/ThumbnailsSidebar.js
@@ -162,6 +162,10 @@ class ThumbnailsSidebar {
             this.isOpen = !!options.isOpen;
         }
 
+        /* If the thubmnail sidebar exists remove it so that we an recreate it
+           this is needed to remove the old virtual scroller that was created by the    
+           doc first preloader
+        */
         const element = document.getElementsByClassName('bp-vs')[0];
         if (element) {
             element.remove();
--- a/src/lib/ThumbnailsSidebar.js
+++ b/src/lib/ThumbnailsSidebar.js
@@ -311,7 +311,7 @@ class ThumbnailsSidebar {
     setCurrentPage(pageNumber) {
         const parsedPageNumber = parseInt(pageNumber, 10);
         // don't use the virtual scroller if the pdfViewer is not ready but the preloader has opened the thumbnails
-        if (parsedPageNumber >= 1 && this.preloader?.thumbnailsOpen && this.pdfViewer.pagesCount === 0) {
+        if (parsedPageNumber >= 1 && this.preloader?.thumbnailsOpen && !this.pdfViewer?.pagesCount) {
             this.currentPage = parsedPageNumber;
             this.applyCurrentPageSelection();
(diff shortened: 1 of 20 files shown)


patlm (approved):

src/lib/viewers/doc/DocFirstPreloader.js
```diff
@@ -67,6 +67,10 @@ class DocFirstPreloader extends EventEmitter {
 
     preloadedImages = {};
 
+    thumbnailsOpen = false;
```
mickr, line 70: Does this cause in stuttering on the page?

src/lib/viewers/doc/DocBaseViewer.js
```diff
@@ -537,14 +537,24 @@ class DocBaseViewer extends BaseViewer {
      */
     setPage(pageNumber) {
         const parsedPageNumber = parseInt(pageNumber, 10);
-        if (!parsedPageNumber || parsedPageNumber < 1 || parsedPageNumber > this.pdfViewer.pagesCount) {
+        // if the preloader has opened the thumbnails and the pdf has not finished loading go to the doc first page
+        // for that thumbnail
+        if (this.thumbnailsSidebar && this.preloader?.thumbnailsOpen && this.pdfViewer.pagesCount === 0) {
+            this.cachePage(parsedPageNumber);
```
mickr, line 543: What does this do?

src/lib/viewers/doc/DocFirstPreloader.js
bfoxx1906: no it doesn't cause any issues

src/lib/viewers/doc/DocBaseViewer.js
bfoxx1906: it stores the page you are on in local storage so that the next time you come to the document it takes you where you left off.

src/lib/ThumbnailsSidebar.js
```diff
@@ -314,8 +310,11 @@ class ThumbnailsSidebar {
      */
     setCurrentPage(pageNumber) {
         const parsedPageNumber = parseInt(pageNumber, 10);
-
-        if (parsedPageNumber >= 1 && parsedPageNumber <= this.pdfViewer.pagesCount) {
+        // don't use the virtual scroller if the pdfViewer is not ready but the preloader has opened the thumbnails
+        if (parsedPageNumber >= 1 && this.preloader?.thumbnailsOpen && this.pdfViewer.pagesCount === 0) {
+            this.currentPage = parsedPageNumber;
+            this.applyCurrentPageSelection();
+        } else if (parsedPageNumber >= 1 && parsedPageNumber <= this.pdfViewer.pagesCount) {
```
JChan106, line 317: Should we use `this.pdfViewer?.pageCount` to be safe?

src/lib/ThumbnailsSidebar.js
```diff
@@ -314,8 +310,11 @@ class ThumbnailsSidebar {
      */
     setCurrentPage(pageNumber) {
         const parsedPageNumber = parseInt(pageNumber, 10);
-
-        if (parsedPageNumber >= 1 && parsedPageNumber <= this.pdfViewer.pagesCount) {
+        // don't use the virtual scroller if the pdfViewer is not ready but the preloader has opened the thumbnails
+        if (parsedPageNumber >= 1 && this.preloader?.thumbnailsOpen && this.pdfViewer.pagesCount === 0) {
```
JChan106, line 314: Would `!this.pdfViewer.pagesCount` work here to be safe?

src/lib/ThumbnailsSidebar.js
bfoxx1906: I actually left that part alone. If it gets there and pdfViewer is undefined then it should throw an error.

JChan106 (approved):

PR #1549, opened 2025-04-19
New exif implementation

bfoxx1906:


shahzadaziz:
Is the new library bundled with preview assets ? Any way to determine what we can expect in terms of overall increase ?

jmcbgaston (dismissed):

src/lib/viewers/doc/DocFirstPreloader.js
```diff
@@ -218,8 +217,11 @@ class DocFirstPreloader extends EventEmitter {
     }
 
     getPreloadImageRequestPromises(preloadUrlWithAuth, pages, pagedPreLoadUrlWithAuth) {
-        const promise1 = this.api.get(preloadUrlWithAuth, { type: 'blob' });
-        const promises = [promise1];
+        const firstPageUrl = !pagedPreLoadUrlWithAuth
+            ? preloadUrlWithAuth
+            : pagedPreLoadUrlWithAuth.replace(PAGED_URL_TEMPLATE_PAGE_NUMBER_HOLDER, '1.webp');
```
JChan106, line 222: Might be helpful to make `1.webp` a constant

src/lib/viewers/doc/DocFirstPreloader.js
```diff
@@ -218,8 +217,11 @@ class DocFirstPreloader extends EventEmitter {
...
+        const promise1 = this.api.get(firstPageUrl, { type: 'blob' });
+        const promises = [promise1.catch(e => e)];
```
JChan106, line 224: This looks like new logic to put caught errors into the promises array. Have we accounted for this?

ahorowitz123, line 224: Where are these promises used?

src/lib/viewers/doc/DocFirstPreloader.js
```diff
@@ -436,6 +437,73 @@ class DocFirstPreloader extends EventEmitter {
...
+                reader.onload = () => {
+                    const arrayBuffer = reader.result;
+                    /* global ExifReader */
+                    tags = ExifReader.load(arrayBuffer);
+
+                    const userComment = tags.UserComment.description || tags.UserComment.value;
+                    const match = EXIF_COMMENT_REGEX.exec(userComment);
+
+                    // There should be 3 pieces of metadata: PDF width, PDF height, and num pages
+                    if (!match || match.length !== 4) {
```
JChan106, line 454: comment says 3 pieces, but it's matching 4? Can we update the comment if this is incorrect

src/lib/viewers/doc/DocFirstPreloader.js
```diff
@@ -436,6 +437,73 @@ class DocFirstPreloader extends EventEmitter {
         });
     }
 
+    readEXIFNew(imageBlob, imageEl) {
```
ahorowitz123, line 440: Is the old `readEXIF` function still used? Why don't we remove it and replace it with this one?

src/lib/viewers/doc/DocFirstPreloader.js
```diff
@@ -436,6 +437,73 @@ class DocFirstPreloader extends EventEmitter {
...
+                    }
+
+                    // Convert PDF Units to CSS Pixels
+                    let pdfWidth = parseInt(match[1], 10) * PDFJS_CSS_UNITS;
+                    let pdfHeight = parseInt(match[2], 10) * PDFJS_CSS_UNITS;
+                    const numPages = parseInt(match[3], 10);
+
+                    // Validate number of pages
+                    if (numPages <= 0) {
+                        reject(new Error('EXIF num pages data is invalid'));
```
JChan106, line 466: A lot of thrown errors in this file, do we have tests for all of these scenarios?

ahorowitz123:
Do we have to update the upgrade script to account for the new exif file?

src/lib/viewers/doc/DocFirstPreloader.js
bfoxx1906: They are used in the showPreload function. They fetch the images

src/lib/viewers/doc/DocFirstPreloader.js
bfoxx1906: This was logic that was there before. The regex actually returns 4 matches with the input string being the first. So it looks like this. I didn't change the regex at all.

0 ='pdfWidth:612pts,pdfHeight:792pts,numPages:10'
1 ='612'
2 ='792'
3 ='10'

src/lib/viewers/doc/DocFirstPreloader.js
bfoxx1906: Yea the logic takes care of that in the showPreload method

src/lib/viewers/doc/DocFirstPreloader.js
bfoxx1906: We have a lot of constants already and since that is only used in one place I felt like this would be ok.

src/lib/viewers/doc/DocFirstPreloader.js
bfoxx1906: Good catch

bfoxx1906:
> Do we have to update the upgrade script to account for the new exif file?

Yes that's a good catch. Will need to do that.

build/upgrade_pdfjs.sh
```diff
@@ -29,6 +29,7 @@ echo "--------------------------------------------------------------------------
 echo "Copying legacy files/folders from version ${DOC_CURRENT_ASSETS_VERSIONS}..."
 echo "-----------------------------------------------------------------------------------"
 \cp -R src/third-party/doc/${DOC_CURRENT_ASSETS_VERSIONS}/exif.js ${DOC_STATIC_ASSETS_PATH}/
```
mickr, line 31: Do we still need this library?

bfoxx1906: We need that for the old preloader to work. I didn't change any of the logic in the old preloader in case we have to rollback the split for whatever reason. Technically, we could get rid of the library now and use a common readEXIF function but that would require me to also modify the current non doc-first preloader class. Once the new preloader is ga'd we can get rid of the old preloader and exif library completely.

mickr (approved):

3kushraps:
Hey what the fuck are you doing you want an exit route I got one for you
kill yourself noe get the fuck out here and talk to me already and take
this shit away from me

PR #1552, opened 2025-04-23
fix(doc-first): webp rep is not marked as success bug

bfoxx1906:
When the webp rep comes back with a a non success status such as the response below, the preloader still constructed a webp image url from the template and tried to request it. This returned a 404 which stopped the preloader. The end result was that preloading was essentially disabled in cases such as this when in actuality the jpeg url should be used instead.

```
 {
                 "representation": "webp",
                 "properties": { ......},
                 "info": {
                    "url": "........."
                  },
                  "status": {
                    "state": "error",
                    "code": "error_large_size_file"
                 },
                 "content": {
                    "url_template": "..........."
                }
  },
```

Additionally, I updating the positioning of the spinner based on whether or not thumbnail sidebar is open or not.

### DIFF
--- a/src/lib/viewers/doc/Document.scss
+++ b/src/lib/viewers/doc/Document.scss
@@ -34,6 +34,10 @@
     &.bp-sidebar-closed {
         left: 50%;
     }
+
+    &.bp-thumbnails-close {
+        left: 37%;
+    }
 }
 
 // Used for showing preload, aka Instant Preview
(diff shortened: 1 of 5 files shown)


src/lib/viewers/doc/Document.scss
```diff
@@ -34,6 +34,10 @@
     &.bp-sidebar-closed {
         left: 50%;
     }
+
+    &.bp-thumbnails-close {
+        left: 37%;
+    }
```
jmcbgaston, line 40: what's this for?

bfoxx1906: that is to space the spinner properly if the thumbnails are closed and the sidebar is open

jmcbgaston (approved):

PR #1553, opened 2025-04-24
fix(doc-first): adding test coverage

bfoxx1906:
Adding more unit tests

### DIFF
--- a/src/lib/viewers/doc/__tests__/DocFirstPreloader-test.js
+++ b/src/lib/viewers/doc/__tests__/DocFirstPreloader-test.js
@@ -562,4 +562,270 @@ describe('/lib/viewers/doc/DocFirstPreloader', () => {
             await expect(promise).rejects.toThrow('Error reading blob as ArrayBuffer');
         });
     });
+
+    describe('getScaledDimensions()', () => {
+        beforeEach(() => {
+            preloader.wrapperEl = document.createElement('div');
+            Object.defineProperty(preloader.wrapperEl, 'clientHeight', { value: 0, writable: true });
+            Object.defineProperty(preloader.wrapperEl, 'clientWidth', { value: 0, writable: true });
+        });
+
+        test('should scale up to a max defined by maxZoomScale', () => {
+            const clientWidth = 500;
+            const clientHeight = 500;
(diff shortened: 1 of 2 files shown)


src/lib/viewers/doc/__tests__/DocFirstPreloader-test.js
```diff
@@ -561,5 +561,286 @@ describe('/lib/viewers/doc/DocFirstPreloader', () => {
...
+            const clientHeight = 500;
+            preloader.wrapperEl.clientWidth = clientWidth;
+            preloader.wrapperEl.clientHeight = clientHeight;
+
+            const pdfWidth = 1000;
+            const pdfHeight = 600;
+            const scaledDimensions = preloader.getScaledDimensions(pdfWidth, pdfHeight);
+
+            // Expect height scale to be used
+            const expectedScale = (clientHeight - 5) / pdfHeight;
```
patlm, line 614: Would it make sense to pull `PDFJS_HEIGHT_PADDING_PX` and `PDFJS_WIDTH_PADDING_PX` from constants for 5 and 40 in these?

bfoxx1906: I typically don't like to include external constants in unit tests. If the constant changes in error then the tests can pass when there is actually a problem. If the constant actually needs to change then the tests will fail and can be easily be fixed.

src/lib/viewers/doc/__tests__/DocFirstPreloader-test.js
```diff
@@ -561,5 +561,289 @@ describe('/lib/viewers/doc/DocFirstPreloader', () => {
...
+            const promise = preloader.readEXIF(mockImageBlob, mockImageEl);
+            mockFileReader.onerror();
+
+            await expect(promise).rejects.toThrow('Error reading blob as ArrayBuffer');
+        });
+    });
+
+    describe('getScaledDimensions()', () => {
+        const pdfjsHeightCSSUnits = 5;
+        const pdfjsHWidthCSSUnits = 40;
```
patlm, line 583: nit - could clarify that these are for padding

patlm (approved):

PR #1554, opened 2025-04-29
fix(virtualscroller): accurately remove scrolled-out elements

mickr:
- Make logic clearer when deleting items out of view elements
- Update comments

### DIFF
--- a/src/lib/VirtualScroller.js
+++ b/src/lib/VirtualScroller.js
@@ -289,8 +289,8 @@ class VirtualScroller {
             newStartOffset = curEndOffset + 1;
             // Create elements from curEnd + 1 to newEndOffset
             this.createItems(fragment, newStartOffset, newEndOffset);
-            // Delete the elements from curStartOffset to newStartOffset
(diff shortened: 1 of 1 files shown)


src/lib/VirtualScroller.js
```diff
@@ -289,8 +289,8 @@ class VirtualScroller {
             newStartOffset = curEndOffset + 1;
             // Create elements from curEnd + 1 to newEndOffset
             this.createItems(fragment, newStartOffset, newEndOffset);
-            // Delete the elements from curStartOffset to newStartOffset
-            this.deleteItems(this.listEl, curStartOffset - curStartOffset, offset - curStartOffset);
+            // Delete the first (offset - curStartOffset) elements that scrolled out of view
+            this.deleteItems(this.listEl, 0, offset - curStartOffset);
```
bfoxx1906, line 293: I looked at this for 20 minutes trying to figure out if there was any possible reason why 'curStartOffset - curStartOffset' was used instead of 0 and came up with nothing. Good catch.

bfoxx1906 (approved):

PR #1555, opened 2025-04-30
fix(doc-first): code update

bfoxx1906:
Based on issues found when migrated to production the following fixes were made

* updating the preloader time diff event to only return the time diff as it's value
* changing the logic so that webp preloading will happen if there is no available jpeg rep but a webp rep is available
* adding/updating test cases 
* being more explicit about Presentations not using the new preloader. While not causing en error, presentations were going down that doc first codepath when the feature was enabled and should not be.


JChan106 (approved):
lgtm

jmcbgaston (approved):

PR #1556, opened 2025-05-07
feat(doc-first): adding doc first behavior to presentations

bfoxx1906:
* Adding doc first preloader to presentations
* renaming retrievedPages to retrievedPagesCount
* updating metrics event logic to only preload time diff data if there are webp representations
* extracting doc first specific css out into the document mixin


src/lib/viewers/doc/PresentationViewer.js
```diff
@@ -1,5 +1,7 @@
 import throttle from 'lodash/throttle';
 import DocBaseViewer from './DocBaseViewer';
+
```
jmcbgaston, line 3: nit: funk new line

patlm (approved):

PR #1560, opened 2025-06-01
feat(prefetch-images): prefetch images updates

bfoxx1906:


src/lib/util.js
```diff
@@ -759,3 +759,35 @@ export function handleRepresentationBlobFetch(response) {
...
+        return [];
+    }
+    const firstPageUrl = !pagedPreLoadUrlWithAuth
+        ? preloadUrlWithAuth
+        : pagedPreLoadUrlWithAuth.replace(PAGED_URL_TEMPLATE_PAGE_NUMBER_HOLDER, '1.webp');
+    const promise1 = api.get(firstPageUrl, { type: 'blob' });
+    const promises = [promise1.catch(e => e)];
+    const count = pages > MAX_PRELOAD_PAGES ? MAX_PRELOAD_PAGES : pages;
+    if (pagedPreLoadUrlWithAuth) {
+        for (let i = 2; i <= count; i += 1) {
```
jmcbgaston: why 2?

bfoxx1906: Because the first image is either going to be a jpeg or webp image if webp images are ready. I add those first and if there are more webp images that will start from page 2 and not page 1 because page 1 has already been accounted for.

ittaishay: Should we add a comment in the code explaining this for easier future reference?

jmcbgaston (approved):
PR looks good but would be good to add some comments in certain sections for context

src/lib/util.js
bfoxx1906: Updated the code to make it more self documenting.

src/lib/viewers/doc/DocBaseViewer.js
```diff
@@ -318,6 +325,56 @@ class DocBaseViewer extends BaseViewer {
...
+        const onlyJpegRepAvailable = jpegRepReady && !pagedWebpRepReady;
+
+        if (onlyJpegRepAvailable) {
+            const { url_template: jpegUrlTemplate = '' } = jpegPreloadRep.content;
+            const jpegUrlAuthTemplate = this.createContentUrlWithAuthParams(jpegUrlTemplate);
+            const promises = getPreloadImageRequestPromises(this.api, jpegUrlAuthTemplate, 1, '');
+            Promise.all(promises);
+        } else if (pagedWebpRepReady) {
+            const { url_template: pagedUrlTemplate = '' } = pagedWebpRep.content;
+            const pageCount = pagedWebpRep.metadata?.pages || 8;
```
jmcbgaston, line 368: nit:

this 8 should be a shared const across all the files

like this `export const MAX_PRELOAD_PAGES = 8;`

just not sure where

jmcbgaston (approved):

PR #1565, opened 2025-07-24
feat(pdfjs-preloading): add function to preload pdfjs assets

bfoxx1906:
- added a function that can be called externally e.g. from EUA to load viewer assets instead of just prefetching them
- updated unit tests
- Added the new exif reader to the existing asset prefetching logic.

### DIFF
--- a/src/lib/Preview.js
+++ b/src/lib/Preview.js
@@ -711,6 +711,29 @@ class Preview extends EventEmitter {
             });
     }
 
+    /**
+     * Prefetches static viewer assets for the specified viewers.
+     *
+     * @public
+     * @param {string[]} [viewerNames] - Names of viewers to load assets for, defaults to none
+     * @return {void}
+     */
+    loadViewers(viewerNames = []) {
+        this.getViewers()
+            .filter(viewer => viewerNames.indexOf(viewer.NAME) !== -1)
+            .forEach(viewer => {
+                const viewerInstance = new viewer.CONSTRUCTOR(
+                    this.createViewerOptions({
+                        viewer,
+                    }),
+                );
+
+                if (typeof viewerInstance.loadViewerAssets === 'function') {
+                    viewerInstance.loadViewerAssets();
+                }
+            });
+    }
+
     //--------------------------------------------------------------------------
     // Private
     //--------------------------------------------------------------------------
--- a/src/lib/Preview.js
+++ b/src/lib/Preview.js
@@ -693,7 +693,7 @@ class Preview extends EventEmitter {
      */
     prefetchViewers(viewerNames = []) {
         this.getViewers()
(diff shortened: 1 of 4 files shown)


src/lib/viewers/doc/DocFirstPreloader.js
```diff
@@ -150,10 +150,10 @@ class DocFirstPreloader extends EventEmitter {
      * @return {Promise} Promise to show preload
      */
     async showPreload(preloadUrlWithAuth, containerEl, pagedPreLoadUrlWithAuth, pages, docBaseViewer) {
+        this.hidePreviewMask();
```
ittaishay: how is this change related to the PR?

src/lib/viewers/doc/DocBaseViewer.js
```diff
@@ -388,7 +388,8 @@ class DocBaseViewer extends BaseViewer {
         const isWatermarked = file && file.watermark_info && file.watermark_info.is_watermarked;
 
         if (assets) {
-            this.prefetchAssets(JS, CSS);
+            const ASSETS = this.docFirstPagesEnabled ? [...JS_NO_EXIF, ...EXIF_READER] : JS;
+            this.prefetchAssets(ASSETS, CSS);
```
ittaishay, line 392: how is this change related to the PR?

bfoxx1906: updated the PR description

src/lib/viewers/doc/DocFirstPreloader.js
bfoxx1906: updated the PR description

src/lib/Preview.js
```diff
@@ -711,6 +711,29 @@ class Preview extends EventEmitter {
...
+    /**
+     * Prefetches static viewer assets for the specified viewers.
+     *
+     * @public
+     * @param {string[]} [viewerNames] - Names of viewers to load assets for, defaults to none
+     * @return {void}
+     */
+    loadViewers(viewerNames = []) {
+        this.getViewers()
+            .filter(viewer => viewerNames.indexOf(viewer.NAME) !== -1)
```
JChan106: Could be `viewer => viewerNames.includes(viewer.NAME)` for more readability

src/lib/Preview.js
```diff
@@ -711,6 +711,29 @@ class Preview extends EventEmitter {
...
+            .forEach(viewer => {
```
JChan106, line 724: Would it be better/less risky to use .map here so we only use immutable methods? .filter should create a new array, so forEach wouldn't affect the original, so should be fine regardless.

src/lib/Preview.js
bfoxx1906: good point

JChan106 (approved):
lgtm

PR #1567, opened 2025-07-31
chore(i18n): Update Translations

boxmoji:


ahorowitz123 (changes requested):
string isn't translated in other languages

PR #1570, opened 2025-08-06
fix(pdfjs-preload): adding null check to emmitter function

bfoxx1906:
Noticed that there were multiple errors being logged in the console because the emit function in BaseViewer was not protecting against cases where the file could potentially be undefined when the assets loaded event is emitted which will always be the case with the pdf.js asset pre-loading optimization we implemented.

### DIFF
--- a/src/lib/viewers/BaseViewer.js
+++ b/src/lib/viewers/BaseViewer.js
@@ -666,7 +666,7 @@ class BaseViewer extends EventEmitter {
             event,
             data,
             viewerName: viewer ? viewer.NAME : '',
-            fileId: file.id,
+            fileId: file ? file.id : '',
         });
     }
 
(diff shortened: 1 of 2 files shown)


src/lib/viewers/BaseViewer.js
```diff
@@ -666,7 +666,7 @@ class BaseViewer extends EventEmitter {
             event,
             data,
             viewerName: viewer ? viewer.NAME : '',
-            fileId: file.id,
+            fileId: file ? file.id : '',
```
ahorowitz123: This could be simplified to `file?.id ?? ''`

ahorowitz123: Do we still want to emit an event if we don't know what file it belongs to?

bfoxx1906: I think that's actually a good idea. Nothing wrong with keeping the null check but this event will also get emitted on every preview regardless of whether the assets are preloaded or not so it probably doesn't make sense to do that. Good catch.

ahorowitz123 (approved):

PR #1571, opened 2025-08-06
fix(pdfjs-preleloader): adding exif reader to preload

bfoxx1906:
The new pdfjs preload logic worked for all of the pdfjs related files but it did not add the proper exifReader library because the value this.docFirstPagesEnabled was not yet initialized when the loadViewer assets call was made. This didn't cause an error but correctly including this library should improve performance for that first preview

### DIFF
--- a/src/i18n/bn-IN.properties
+++ b/src/i18n/bn-IN.properties
@@ -99,7 +99,7 @@ error_internet_explorer_office_online=Microsoft 365 অ্যাপস এবং
 # Preview error message shown when the preview fails because the file is too large for its extension type
 error_large_file=ফাইলটি প্রিভিউ দেখার জন্য সমর্থিত জায়গা অতিক্রম করেছে। দেখার জন্য ফাইলটি খুলুন অথবা ডাউনলোড করুন।
 # Preview error message shown when the preview fails because the file is has too many operations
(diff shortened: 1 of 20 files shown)


src/lib/viewers/doc/DocBaseViewer.js
```diff
@@ -176,7 +178,7 @@ class DocBaseViewer extends BaseViewer {
 
         // Call super() to set up common layout
         super.setup();
-        this.docFirstPagesEnabled = this.featureEnabled('docFirstPages.enabled');
+        this.docFirstPagesEnabled = this.featureEnabled(DOC_FIRST_PAGES_ENABLED);
```
ahorowitz123, line 181: Do we still have a use for this instance variable? It looks like below we are removing references to it

ahorowitz123: Can we just move this line into the constructor rather then part of the setup function?

bfoxx1906: I wanted to minimize the changes to just focus just on this. We definitely need the variable but we should be able to move it to the constructor. I'll check that but originally I was trying to keep this consistent with how the other instance variables are set.

PR #1581, opened 2025-09-30
feat(video-ui-updates): adding vertical media volume controls.

bfoxx1906:
This PR adds the vertical volume slider to the video controls. The previous volume slider used the same component as the horizontal video playback slider and all of the volume calculations were based on the sliders width. Now those calculations are based on the sliders height. 

PR Notes:

* The slider height is determined based on the volume PR [Reference](https://github.com/box/box-content-preview/pull/1581/files#diff-637ca22bd00a86eb52104d57bfd67e88b359dea42264160a56980b9acf44d7d3R158-R160)

* The  callbacks in the new VolumeSliderControl are the same as in the original [Slider](https://github.com/box/box-content-preview/blob/master/src/lib/viewers/controls/slider/SliderControl.tsx) with the difference being that they use height values instead when they get the mouse position value to base the volume calculation off of. PR [Reference](https://github.com/box/box-content-preview/pull/1581/files#diff-637ca22bd00a86eb52104d57bfd67e88b359dea42264160a56980b9acf44d7d3R55-R65)

* Added a check in the handleVolume function in MediaBaseViewer to guard against an invalid stored volume level breaking the video preview. In this case, if the volume value is greater than 1 then an error is thrown and the preview is stuck in a spinning state. I discovered this while developing the vertical slider when I dragged the slider to the top of the track but still kept moving my mouse up passed the bounds of the slider. This should not happen as I've fixed that issue [here](https://github.com/box/box-content-preview/pull/1581/files#diff-2ae8121a06738669330b14bf4f2e8f0fccfd7024dbcdb230695078492135ec61R41-R46) but it's worth adding the check because this will cause all video previews not to work until that value is removed from local storage. The screenshot below shows this happening. PR [Reference](https://github.com/box/box-content-preview/pull/1581/files#diff-242d253d5790db64a7f3a5f5a65fe30ee9045543c45bd03b46c77271058002d0R413-R417)

     
Screenshot 2025-09-29 at 8 58 46 PM

### DIFF
--- a/src/lib/viewers/controls/media/VolumeControls.scss
+++ b/src/lib/viewers/controls/media/VolumeControls.scss
@@ -7,6 +7,9 @@
 }
 
 .bp-VolumeControls-flyout {
+    position: absolute;
+    top: -40px;
+    left: -5px;
     display: none;
     flex: 0 0 auto;
     max-width: 0;
@@ -15,9 +18,6 @@
     transition: max-width 200ms ease-in-out;
 
     &.bp-is-open {
-        position: absolute;
-        top: -40px;
-        left: -5px;
         display: block;
         overflow: visible;
     }
--- a/src/lib/viewers/controls/media/VolumeControls.scss
+++ b/src/lib/viewers/controls/media/VolumeControls.scss
@@ -1,11 +1,13 @@
 @import './styles';
 
 .bp-VolumeControls {
+    position: relative;
     display: flex;
     align-items: center;
 }
 
 .bp-VolumeControls-flyout {
+    display: none;
     flex: 0 0 auto;
     max-width: 0;
     height: $bp-MediaControl-height;
@@ -13,12 +15,16 @@
     transition: max-width 200ms ease-in-out;
 
     &.bp-is-open {
-        max-width: 100px;
+        position: absolute;
+        top: -40px;
+        left: -5px;
+        display: block;
+        overflow: visible;
     }
 }
 
 .bp-VolumeControls-slider {
-    width: 100px - $bp-SliderControl-thumb-size;
+    width: 15px;
     height: 100%;
     margin-right: $bp-SliderControl-thumb-radius;
     margin-left: $bp-SliderControl-thumb-radius;
--- a/src/lib/viewers/controls/media/VolumeControls.tsx
+++ b/src/lib/viewers/controls/media/VolumeControls.tsx
@@ -1,6 +1,5 @@
 import React from 'react';
 import classNames from 'classnames';
-import { bdlBoxBlue, white } from 'box-ui-elements/es/styles/variables';
 import IconVolumeMed24 from '../icons/IconVolumeMed24';
 import IconVolumeLow24 from '../icons/IconVolumeLow24';
 import IconVolumeMuted24 from '../icons/IconVolumeMuted24';
@@ -8,7 +7,7 @@ import MediaToggle from './MediaToggle';
 import useAttention from '../hooks/useAttention';
 import './VolumeControls.scss';
 import IconVolumeMax24 from '../icons/IconVolumeMax24';
-import SliderControl from '../slider/SliderControl';
+import VolumeSliderControl from '../slider/VolumeSliderControl';
 
 export type Props = {
     onMuteChange: (isMuted: boolean) => void;
@@ -39,13 +38,24 @@ export default function VolumeControls({ onMuteChange, onVolumeChange, volume = 
 
     const handleVolume = React.useCallback(
         (newValue: number): void => {
-            onVolumeChange(newValue / 100);
+            const newValueToUse = newValue <= 5 ? 0 : newValue;
+            // make sure the value is always between 0 and 1 since the value passed in could be
+            // greater than 100 if the user drags the mouse above the top of the sider when adjusting the volume
+            const volumeToUse = Math.min(1, newValueToUse / 100);
+            onVolumeChange(volumeToUse);
         },
         [onVolumeChange],
     );
 
     return (
-        <div className="bp-VolumeControls" data-testid="bp-volume-controls">
+        <div
+            className="bp-VolumeControls"
+            data-testid="bp-volume-controls"
+            onBlur={handlers.onBlur}
+            onFocus={handlers.onFocus}
+            onMouseOut={handlers.onMouseOut}
+            onMouseOver={handlers.onMouseOver}
+        >
             <MediaToggle
                 className="bp-VolumeControls-toggle"
                 onClick={(): void => onMuteChange(!isMuted)}
@@ -54,15 +64,14 @@ export default function VolumeControls({ onMuteChange, onVolumeChange, volume = 
             >
                 <Icon />
             </MediaToggle>
-
             <div className={classNames('bp-VolumeControls-flyout', { 'bp-is-open': isActive })}>
-                <SliderControl
+                <VolumeSliderControl
                     className="bp-VolumeControls-slider"
                     max={100}
                     onUpdate={handleVolume}
                     step={1}
                     title={__('media_volume_slider')}
-                    track={`linear-gradient(to right, ${bdlBoxBlue} ${value}%, ${white} ${value}%)`}
+                    track={`linear-gradient(to right, #0061d5 ${value}%, #fff ${value}%)`}
                     value={value}
                     {...handlers}
                 />
--- a/src/lib/viewers/controls/media/VolumeControls.tsx
+++ b/src/lib/viewers/controls/media/VolumeControls.tsx
@@ -71,7 +71,6 @@ export default function VolumeControls({ onMuteChange, onVolumeChange, volume = 
                     onUpdate={handleVolume}
                     step={1}
                     title={__('media_volume_slider')}
-                    track={`linear-gradient(to right, #0061d5 ${value}%, #fff ${value}%)`}
                     value={value}
                     {...handlers}
                 />
--- a/src/lib/viewers/controls/media/__tests__/VolumeControls-test.tsx
+++ b/src/lib/viewers/controls/media/__tests__/VolumeControls-test.tsx
@@ -54,23 +54,21 @@ describe('VolumeControls', () => {
         });
 
         test.each`
-            volume   | track                                                   | value
-            ${0}     | ${`linear-gradient(to right, #0061d5 0%, #fff 0%)`}     | ${0}
-            ${0.0}   | ${`linear-gradient(to right, #0061d5 0%, #fff 0%)`}     | ${0}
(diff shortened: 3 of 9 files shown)


src/lib/viewers/controls/media/VolumeControls.scss
```diff
@@ -1,24 +1,30 @@
...
     height: $bp-MediaControl-height;
     overflow: hidden;
     transition: max-width 200ms ease-in-out;
 
     &.bp-is-open {
-        max-width: 100px;
+        position: absolute;
+        top: -40px;
+        left: -5px;
+        display: block;
```
ahorowitz123, line 21: Can we simplify this by just having `display: block` in this bracket and everything else in `bp-VolumeControls-flyout`?

src/lib/viewers/controls/media/VolumeControls.tsx
```diff
@@ -39,13 +38,24 @@ export default function VolumeControls({ onMuteChange, onVolumeChange, volume =
 
     const handleVolume = React.useCallback(
         (newValue: number): void => {
-            onVolumeChange(newValue / 100);
+            const newValueToUse = newValue <= 5 ? 0 : newValue;
```
ahorowitz123, line 41: Why are we making values 5 or less 0?

src/lib/viewers/controls/media/__tests__/VolumeControls-test.tsx
```diff
@@ -54,23 +54,21 @@ describe('VolumeControls', () => {
...
-            ${1.0}   | ${`linear-gradient(to right, #0061d5 100%, #fff 100%)`} | ${100}
-        `('should render the correct track and value for volume $volume', async ({ track, value, volume }) => {
+            volume   | value
+            ${0}     | ${5}
+            ${0.0}   | ${5}
+            ${0.01}  | ${1}
+            ${0.25}  | ${25}
+            ${0.254} | ${25}
+            ${0.255} | ${26}
+            ${1.0}   | ${100}
```
ahorowitz123, line 64: Should we test for values that would greater than a volume value of 1?

src/lib/viewers/controls/slider/VolumeSliderControl.tsx
```diff
@@ -0,0 +1,167 @@
...
+        const { current: sliderEl } = sliderElRef;
+        if (!sliderEl) return 0;
+
+        const { top: sliderTop, height: sliderHeight } = sliderEl.getBoundingClientRect();
+
+        // Calculate distance from bottom of slider
+        const sliderBottom = sliderTop + sliderHeight;
+        const distanceFromBottom = sliderBottom - clientY;
+
+        return Math.max(5, distanceFromBottom);
```
ahorowitz123: I noticed this 5 in VolumeControls as well. Do we want to make it a constant? Assuming this is the lowest value we allow for the slider

src/lib/viewers/controls/slider/VolumeSliderControl.tsx
```diff
@@ -0,0 +1,167 @@
...
+    }, []);
+
+    const getPositionValue = React.useCallback(
+        (pageY: number, clientY: number) => {
+            const { current: sliderEl } = sliderElRef;
+            if (!sliderEl) return 0;
```
ahorowitz123, line 60: nit: can we add a space after this line to make it more readable?

src/lib/viewers/controls/slider/VolumeSliderControl.tsx
```diff
@@ -0,0 +1,167 @@
...
+            const { height: sliderHeight } = sliderEl.getBoundingClientRect();
+            const positionRelativeToSlider = getPositionRelativeToSlider(clientY);
+            const newValue = (positionRelativeToSlider / sliderHeight) * max;
+            return Math.max(min, Math.min(newValue, max));
```
ahorowitz123, line 65: Might make sense to turn `Math.max(min, Math.min(newValue, max))` into a function. I see it used across other functions

src/lib/viewers/controls/slider/VolumeSliderControl.tsx
```diff
@@ -0,0 +1,167 @@
...
+        onUpdate(getPositionValue(pageY, clientY));
+        setIsScrubbing(true);
+        // Prevent clicking on the slider from triggering the mouse down event on the parent slider track
+        event.stopPropagation();
+    };
+
+    const handleMouseMove = ({ pageY, clientY }: React.MouseEvent<Ref>): void => {
+        const { current: sliderEl } = sliderElRef;
+        const { height: sliderHeight } = sliderEl ? sliderEl.getBoundingClientRect() : { height: 0 };
+        onMove(getPositionValue(pageY, clientY), getPositionRelativeToSlider(clientY), sliderHeight);
```
ahorowitz123, line 95: Do we only want to do this if scrubbing is true?

src/lib/viewers/controls/slider/VolumeSliderControl.tsx
```diff
@@ -0,0 +1,167 @@
...
+            onUpdate(getPositionRelativeToSlider(event.clientY));
+        };
+        const handleDocumentTouchMove = (event: TouchEvent): void => {
+            if (!isScrubbing || !event.touches || !event.touches[0]) return;
+
+            event.preventDefault();
+            onUpdate(getPositionRelativeToSlider(event.touches[0].clientY));
+        };
+
+        if (isScrubbing) {
```
ahorowitz123, line 118: Should we be removing these if `isScrubbing` is false?

src/lib/viewers/controls/slider/VolumeSliderControl.tsx
```diff
@@ -0,0 +1,167 @@
...
+        const { height: sliderHeight } = sliderEl ? sliderEl.getBoundingClientRect() : { height: 0 };
+        onMove(getPositionValue(pageY, clientY), getPositionRelativeToSlider(clientY), sliderHeight);
+    };
+
+    const handleTouchStart = ({ touches }: React.TouchEvent<Ref>): void => {
+        onUpdate(getPositionValue(touches[0].pageY, touches[0].clientY));
+        setIsScrubbing(true);
+    };
+
+    React.useEffect(() => {
```
ahorowitz123, line 103: I think these all might be more readable and simpler outside of a useEffect. The `isScrubbing` logic can be inside of the functions

src/lib/viewers/media/MediaBaseViewer.js
```diff
@@ -625,6 +629,7 @@ class MediaBaseViewer extends BaseViewer {
      * @return {void}
      */
     setVolume(volume) {
+        console.log('setVolume', volume);
```
ahorowitz123: Remove console.log

bfoxx1906: Other repos actually catch this with their linter. I'm not sure why preview doesn't. We'll have to update this.

src/lib/viewers/controls/slider/VolumeSliderControl.tsx
bfoxx1906: Hey so this logic was actually copied from the previous slider. I didn't want to change this.

src/lib/viewers/controls/media/VolumeControls.tsx
bfoxx1906: Because we don't have a track ball like we do with the horizontal slider so when the user pulls the slider to the bottom of the track we need to keep some height so that they can later increase the volume. That's where 5 comes in.

src/lib/viewers/controls/slider/VolumeSliderControl.tsx
bfoxx1906: It's only used in the other slider and it's pretty straightforward. I don't think we necessarily need a new function for this.

src/lib/viewers/controls/media/__tests__/VolumeControls-test.tsx
bfoxx1906: The volume value can never be greater than 1

src/lib/viewers/controls/media/VolumeControls.scss
bfoxx1906: I like that :)

src/lib/viewers/controls/slider/VolumeSliderControl.tsx
bfoxx1906: Changing the scrubbing logic caused things to not work. I wanted to keep as much of the existing functionality as possible.

src/lib/viewers/controls/media/VolumeControls.scss
bfoxx1906: Updated this. Not sure why it's not showing as outdated.

src/lib/viewers/controls/slider/VolumeSliderControl.tsx
ahorowitz123: I was referring to the fact that its used in `getPositionRelativeToSlider`, `getPositionRelativeToSlider`, and `handleKeydown` within this file

bfoxx1906: I see what you mean. Even still with that we'd just be replacing Math.max(val,val2) with something like getMax(val,val2) right?

src/lib/viewers/controls/slider/VolumeSliderControl.tsx
bfoxx1906: Like the other ones. Didn't want to change this. Removing this actually broke it.

ahorowitz123: Oh I meant removing the listeners if isScrubbing is false. Not a big deal but right now every time `isScrubbing is true, it redefines these functions and adds them as listeners again

src/lib/viewers/controls/slider/VolumeSliderControl.tsx
ahorowitz123: Yeah basically. Its fine either way. Just took me a second to understand what it was doing so I thought a readable function name would be easier.

src/lib/viewers/controls/slider/VolumeSliderControl.scss
```diff
@@ -0,0 +1,78 @@
...
+    height: 140px;
+    margin: auto;
+    padding: 16px;
+    background: rgba(34, 34, 34, .9);
+    border: 1px solid rgba(255, 255, 255, .13);
+    border-radius: 16px;
+    box-shadow: 0 2px 5px 0 rgba(0, 0, 0, .25);
+    cursor: pointer;
+
+    // force this to bottom of it's parent
```
jmcbgaston, line 30: what's this in reference to?

PR #1639, opened 2026-05-07
feat(preview): add monitoring dimensions + preload outcome

ahorowitz123:
## Summary

Adds five new dimensions to preview metric and error events so a downstream performance dashboard can slice by them:

- **`access_pattern`** — how the user reached preview (`file_list` / `direct_link` / `shared_link`)
- **`preview_mode`** — which preview surface (`default` / `shared_file` / `shared_folder` / `editable_shared_file` / `inline_feed`)
- **`shared_link_auth`** — auth state for shared-link previews (`logged_in` / `logged_out` / `na`)
- **`preload_status`** — whether the host called `Preview.prefetch({preload: true})` for this fileId (`hit` / `miss`)
- **`prefetch_status`** — whether the SDK served this load from the file-metadata cache (`hit` / `miss`)

## Implementation

- All five dimensions accepted as options on `Preview.parseOptions`. `access_pattern`, `preview_mode`, `shared_link_auth` attach to every emitted event via `emitLogEvent`.
- `preload_status` is host-supplied (since `ContentPreview` creates a fresh `Preview` instance per session, the host owns prefetch bookkeeping).
- `prefetch_status` is derived inside the SDK from `Logger.log.cache.hit` — which reflects whether `loadFromCache()` skipped the file-info API call.
- `preload_status` and `prefetch_status` only attach to `load`, `preview_error`, and a new `preview_preload_outcome` counter event — not on every emitted event, to keep the dashboard denominator clean.

## Test plan

- [x] Unit tests for dimension tagging on `emitLogEvent`, `emitLoadMetrics`, `emitPreviewError`
- [x] Unit tests for `parseOptions` option storage
- [x] Unit tests for `preview_preload_outcome` counter emission
- [x] Full jest suite passes
- [x] ESLint clean on touched files


src/lib/Preview.js
```diff
@@ -1097,6 +1099,13 @@ class Preview extends EventEmitter {
         // Object with ftux experience data that can determine whether specific experiences can show
         this.options.experiences = options.experiences || {};
 
+        // Host-supplied monitoring dimensions attached to every emitted event by emitLogEvent
+        this.options.accessPattern = options.accessPattern;
+        this.options.previewMode = options.previewMode;
+        this.options.sharedLinkAuth = options.sharedLinkAuth;
+        // Whether the host called Preview.prefetch({preload: true}) for this fileId
+        this.options.preloadStatus = options.preloadStatus;
```
mrscobbler, line 1107: Based on the comment in line 1106 this looks like a boolean (true/false) but the preloadStatus option is an enum (hit/miss/na) - is there a mismatch of expected inputs?

mrscobbler (commented):
added a comment!

src/lib/Preview.js
```diff
@@ -1824,13 +1834,22 @@ class Preview extends EventEmitter {
...
             [LOAD_METRIC.fileInfoTime]: infoTime.elapsed || 0,
             [LOAD_METRIC.convertTime]: convertTime.elapsed || 0,
             [LOAD_METRIC.downloadResponseTime]: downloadTime.elapsed || 0,
             [LOAD_METRIC.contentLoadTime]: contentLoadTime.elapsed || 0,
+            ...loadStateTags,
+        });
+
+        this.emitLogEvent(PREVIEW_METRIC, {
+            event_name: PREVIEW_PRELOAD_OUTCOME_EVENT,
+            value: loadStateTags.preload_status,
```
jmcbgaston: 🟡 **`value` is a categorical string on a metric event.**

Other `PREVIEW_METRIC` events use `value` for numeric elapsed times (e.g., `previewLoadEvent`). Overloading it with `'hit'` / `'miss'` risks NaN-ing downstream sum/avg/percentile aggregations. It's also redundant — `preload_status` is already in the payload via `...loadStateTags`.

Two cleaner options:
- **Drop `value` entirely** — the `preload_status` tag carries the signal.
- **Use `value: 1`** (counter convention) and rely on `preload_status` for bucketing.

Worth confirming with the dashboard consumer that a categorical `value` is accepted before shipping.

src/lib/Preview.js
```diff
@@ -1882,10 +1902,28 @@ class Preview extends EventEmitter {
...
+     * the load event, preview_preload_outcome counter, and preview_error — exactly one
+     * of those fires per preview session so the dashboard denominator stays clean.
+     *
+     * @private
+     * @return {{ preload_status: string, prefetch_status: string }}
+     */
+    getLoadStateTags() {
+        return {
+            preload_status: this.options?.preloadStatus || PRELOAD_STATUS.MISS,
+            prefetch_status: getProp(this.logger, 'log.cache.hit', false) ? 'hit' : 'miss',
```
jmcbgaston: 🟡 **Inconsistent use of `PRELOAD_STATUS` constants.**

`preload_status` uses `PRELOAD_STATUS.MISS` on the line above, but `prefetch_status` uses bare string literals here. Either use the constants here too, or rename `PRELOAD_STATUS` to something that scopes to both preload and prefetch (since they share the same values).

The tests reflect the same split — `preload_status` asserts against the constant, `prefetch_status` asserts against `'hit'` / `'miss'` literals.

src/lib/events.js
```diff
@@ -79,6 +79,16 @@ export const FIRST_RENDER_METRIC = 'first_render_metric';
...
 
+// Event fired once per load with preload hit/miss outcome (counter for hit rate)
+export const PREVIEW_PRELOAD_OUTCOME_EVENT = 'preview_preload_outcome';
+
+// Preload status values attached to load events and preview_preload_outcome
+export const PRELOAD_STATUS = {
+    HIT: 'hit', // Viewer consumed a preloaded representation/asset before the full load completed
+    MISS: 'miss', // Preload was attempted/available but not consumed
+    NA: 'na', // Preload is not applicable for this viewer or file type
+};
```
jmcbgaston, line 90: 🟡 **`NA: 'na'` is dead code in this PR.**

Defined but never assigned or referenced. Either wire up a producer (viewers that can't preload should set it) or drop the constant until a caller exists — otherwise call sites will invent `'na'` and diverge from the enum.

src/lib/Preview.js
```diff
@@ -1882,10 +1902,28 @@ class Preview extends EventEmitter {
...
+     * Returns the prefetch/preload dimensions for this preview session. Only attached to
+     * the load event, preview_preload_outcome counter, and preview_error — exactly one
+     * of those fires per preview session so the dashboard denominator stays clean.
+     *
+     * @private
+     * @return {{ preload_status: string, prefetch_status: string }}
+     */
+    getLoadStateTags() {
+        return {
+            preload_status: this.options?.preloadStatus || PRELOAD_STATUS.MISS,
```
jmcbgaston: 🟢 Nit: `||` will silently coerce `''` / `0` / `null` to `MISS`. `??` is slightly stricter if you want to allow future falsy-but-valid values. Low priority — hosts are unlikely to pass those.

jmcbgaston (commented):
Nice, clean addition — scope is tight and the test coverage is solid. A few inline notes below; the `value: <string>` one is the only item with real impact, the rest are cleanup.

**Approve with minor changes.**

src/lib/Preview.js
ahorowitz123: Updated the comment to be more clear

src/lib/Preview.js
```diff
@@ -61,10 +61,12 @@ import {
     ERROR_CODE,
     FIRST_RENDER_METRIC,
     LOAD_METRIC,
+    CACHE_STATUS,
```
jackiejou: can we sort this?

src/lib/Preview.js
```diff
@@ -1097,6 +1100,14 @@ class Preview extends EventEmitter {
         // Object with ftux experience data that can determine whether specific experiences can show
         this.options.experiences = options.experiences || {};
 
+        // Host-supplied monitoring dimensions attached to every emitted event by emitLogEvent
+        this.options.accessPattern = options.accessPattern;
+        this.options.previewMode = options.previewMode;
+        this.options.sharedLinkAuth = options.sharedLinkAuth;
+        // Host-supplied 'hit' | 'miss' indicating whether Preview.prefetch({preload: true})
```
jackiejou: do we need this comment? since we have that overarching one on line 1103?

src/lib/Preview.js
```diff
@@ -1882,10 +1905,28 @@ class Preview extends EventEmitter {
             rep_type: getProp(this.viewer, 'options.representation.representation', '').toLowerCase(),
             timestamp: getISOTime(),
             total_pages: getProp(this.viewer, 'pdfViewer.pdfDocument.numPages', ''),
+            ...(accessPattern !== undefined && { access_pattern: accessPattern }),
```
jackiejou: do we need these undefined checks? would logging access_pattern: undefined cause data issues?

jackiejou (approved):

PR #1640, opened 2026-05-07
feat(controls): add resin tagging to toolbar elements

kduncanhsu:
## Summary                                                                                                                                                                                                                     
  - Add `data-resin-target` attributes to 31 interactive toolbar elements across document, media, settings, and 3D controls                                                                                                    
  - Enables click tracking via the resin analytics library which reads DOM `data-resin-*` attributes                                                                                                                             
  - No code logic changes, no new dependencies, no test changes required                                                                                                                                                       
                            
resin console logger:
<img width="387" height="92" alt="Screenshot 2026-05-08 at 2 37 27 PM" src="https://github.com/user-attachments/assets/a3f564ab-3601-49c1-a013-f033861608e7" />
                                                                                                                                                                                                     
  ## Details                                                                                                                                                                                                                   
  Previously only 4 annotation buttons had resin tagging. This PR instruments all remaining interactive toolbar elements:

  | Category | Targets Added |
  |----------|--------------|
  | Document/Image | `fullscreen`, `findBar`, `thumbnails`, `rotate` |
  | Zoom | `zoomIn`, `zoomOut` |
  | Page Navigation | `pagePrevious`, `pageNext`, `pageNumberOpen`, `pageNumberInput` |
  | Color Picker | `colorPickerOpen`, `colorSelect` |
  | Media Player | `playPause`, `skipBackward`, `skipForward`, `volumeToggle`, `volumeSlider`, `timeScrubber`, `subtitlesToggle`, `mediaFullscreen` |
  | Settings | `settingsOpen`, `settingsMenuItem`, `settingsRadioItem`, `settingsCheckbox`, `settingsDropdown`, `settingsBack` |
  | 3D Controls | `vrToggle`, `resetView`, `animationClips`, `rotateAxisLeft`, `rotateAxisRight` |

  ## Test plan
  - [x] TypeScript compilation passes
  - [x] All 1508 related tests pass
  - [x] ESLint passes with no warnings
  - [ ] Verify resin events fire correctly in a staging environment with resin enabled

### DIFF
--- a/src/lib/viewers/controls/media/MediaFullscreenToggle.tsx
+++ b/src/lib/viewers/controls/media/MediaFullscreenToggle.tsx
@@ -13,5 +13,5 @@ export default function MediaFullscreenToggle(props: Props): JSX.Element {
         }
     };
 
-    return <FullscreenToggle onKeyDown={handleKeydown} {...props} />;
+    return <FullscreenToggle data-resin-target="mediaFullscreen" onKeyDown={handleKeydown} {...props} />;
 }
--- a/src/lib/viewers/controls/media/TimeControls.tsx
+++ b/src/lib/viewers/controls/media/TimeControls.tsx
@@ -66,6 +66,7 @@ export default function TimeControls({
             <DurationLabels currentTime={currentTime} durationTime={durationTime} />
             <SliderControl
                 className="bp-TimeControls-slider"
+                data-resin-target="timeScrubber"
                 max={durationValue}
                 min={0}
                 onBlur={noop} // Filmstrip is not currently shown during keyboard navigation
--- a/src/lib/viewers/controls/box3d/AnimationClipsToggle.tsx
+++ b/src/lib/viewers/controls/box3d/AnimationClipsToggle.tsx
@@ -13,6 +13,7 @@ function AnimationClipsToggle(props: Props, ref: React.Ref<HTMLButtonElement>): 
         <button
             ref={ref}
             className="bp-AnimationClipsToggle"
+            data-resin-target="animationClips"
             onClick={onClick}
             title={__('box3d_animation_clips')}
             type="button"
--- a/src/lib/viewers/controls/box3d/ResetControl.tsx
+++ b/src/lib/viewers/controls/box3d/ResetControl.tsx
@@ -12,7 +12,13 @@ export default function ResetControl({ onReset }: Props): JSX.Element {
     };
 
     return (
-        <button className="bp-ResetControl" onClick={handleClick} title={__('box3d_reset')} type="button">
+        <button
+            className="bp-ResetControl"
+            data-resin-target="resetView"
+            onClick={handleClick}
+            title={__('box3d_reset')}
+            type="button"
+        >
             <IconReset24 />
         </button>
     );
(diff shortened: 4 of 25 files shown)


src/lib/viewers/controls/media/TimeControls.tsx
```diff
@@ -66,6 +66,7 @@ export default function TimeControls({
             <DurationLabels currentTime={currentTime} durationTime={durationTime} />
             <SliderControl
                 className="bp-TimeControls-slider"
+                data-resin-target="timeScrubber"
```
jmcbgaston, line 69: Silent-failure risk: `data-resin-target` is being passed to a React component (`SliderControl`), not a DOM element. It works today because `SliderControl` spreads `{...rest}` onto its root `<div>` ([SliderControl.tsx:135](https://github.com/box/box-content-preview/blob/master/src/lib/viewers/controls/slider/SliderControl.tsx#L135)), but that's an invisible contract — if anyone ever stops spreading rest (or reorders props), the attribute silently disappears from the DOM. No type error, no test failure, just a gap in the analytics dashboard weeks later.

Low-cost mitigations:
1. Make `data-resin-target` an explicit prop on `SliderControl` applied directly to the rendered element.
2. Add a one-line assertion in the `TimeControls` test: `expect(container.querySelector('[data-resin-target="timeScrubber"]')).toBeInTheDocument()`.

Not blocking — just the class of bug that's hardest to notice.

src/lib/viewers/controls/media/MediaFullscreenToggle.tsx
```diff
@@ -13,5 +13,5 @@ export default function MediaFullscreenToggle(props: Props): JSX.Element {
         }
     };
 
-    return <FullscreenToggle onKeyDown={handleKeydown} {...props} />;
+    return <FullscreenToggle data-resin-target="mediaFullscreen" onKeyDown={handleKeydown} {...props} />;
```
jmcbgaston, line 16: Same silent-failure risk as the `SliderControl` case in `TimeControls`, plus an extra wrinkle: `FullscreenToggle` already hardcodes `data-resin-target="fullscreen"`, and this override only works because `{...rest}` is spread *after* that attribute on the child's `<button>`. If anyone reorders props in `FullscreenToggle` so the hardcoded attribute comes after the spread, this value gets clobbered back to `"fullscreen"` — and every media fullscreen click will be logged under the wrong target name.

Suggest promoting `data-resin-target` to an explicit prop on `FullscreenToggle` so the override is visible in the component's API, or at minimum adding a test assertion here to catch the regression.

jmcbgaston (commented):
Flagging a silent-failure pattern on two sites where `data-resin-target` is passed to a React component rather than a DOM element. Both work today but depend on invisible prop-forwarding contracts — if that ever breaks, analytics silently stop firing with no test or type error to catch it. Inline comments on each site.

src/lib/viewers/controls/media/TimeControls.tsx
kduncanhsu: Good callout on defensive coding. That said, passing data-resin-target through component props via {...rest} is actually the established pattern in this codebase — the 4 existing resin tags all do exactly this (passed to AnnotationsButton, which spreads {...rest} onto its <button>).

Both SliderControl and MediaToggle explicitly type their props as extending React.HTMLAttributes / React.ButtonHTMLAttributes, so the {...rest} spread is part of their intentional public API (accepting and forwarding arbitrary HTML attributes), not an accidental implementation detail.

That said, happy to add test assertions for the cases that pass through component boundaries if you feel strongly. Let me know.

src/lib/viewers/controls/media/MediaFullscreenToggle.tsx
kduncanhsu: same as above

ahorowitz123 (approved):

PR #1643, opened 2026-05-12
feat(auth): accept optional features in prefetch()

jmcbgaston:
### Description

`prefetch()` relied on `this.options.features`, which is only set during `Preview.show()`. Callers that prefetch before `show()` (for example, folder-list hover prefetch in a host application) had no way to opt into auth-header routing for the access token, so `featureEnabled('migrateAccessTokenToHeader')` checks inside viewer `prefetch()` methods always read the default (`false`).

This change adds an optional top-level `isAccessTokenHeaderEnabled = false` param to `prefetch()`, matching the shape of the existing `isDocFirstPrefetchEnabled` param. When the prefetch is for `preload`, the per-call viewer options merge in `features.migrateAccessTokenToHeader` so viewer `prefetch()` methods see the threaded value — without mutating `this.options.features`.

- Callers that omit the flag get byte-identical behavior.
- Subsequent `show()` / `prefetch()` calls aren't affected by prior per-call state.

### Test plan

- [x] `yarn test src/lib/__tests__/Preview-test.js` — 271 passing, including new tests covering flag-on (feature surfaced to viewer options) and flag-off (feature surfaced as `false`) paths
- [x] Host app verified hover-prefetch requests include `Authorization: Bearer` header (no `access_token` in URL) when passing `isAccessTokenHeaderEnabled: true`
- [x] Host app verified omitting / passing `false` preserves existing behavior

### Semantic release type

- [x] `feat` - New feature

### DIFF
--- a/src/lib/Preview.js
+++ b/src/lib/Preview.js
@@ -693,6 +693,7 @@ class Preview extends EventEmitter {
      * @param {string} options.sharedLink - Shared link
      * @param {string} options.sharedLinkPassword - Shared link password
      * @param {boolean} options.preload - Is this prefetch for a preload
+     * @param {Object} [options.features] - Feature flags to merge into viewer options for this prefetch
      * @param {string} token - Access token
      * @return {void}
      */
@@ -705,6 +706,7 @@ class Preview extends EventEmitter {
         preload = false,
         isDocFirstPrefetchEnabled = false,
         docFirstPagesConfig = null,
+        features,
     }) {
         let file;
         let loader;
@@ -745,6 +747,13 @@ class Preview extends EventEmitter {
             options.sharedLinkPassword = sharedLinkPassword;
             options.isDocFirstPrefetchEnabled = isDocFirstPrefetchEnabled;
             options.docFirstPagesConfig = docFirstPagesConfig;
+        }
+
+        // If features are passed in for this prefetch call (e.g. from callers that
+        // invoke prefetch() before show() has set this.options.features), merge them
+        // into the viewer options for this call without mutating this.options.
+        if (features) {
+            options.features = { ...this.options.features, ...features };
         }
 
         const viewerInstance = new viewer.CONSTRUCTOR(this.createViewerOptions(options));
--- a/src/lib/Preview.js
+++ b/src/lib/Preview.js
@@ -693,7 +693,6 @@ class Preview extends EventEmitter {
      * @param {string} options.sharedLink - Shared link
      * @param {string} options.sharedLinkPassword - Shared link password
      * @param {boolean} options.preload - Is this prefetch for a preload
-     * @param {Object} [options.features] - Feature flags to merge into viewer options for this prefetch
      * @param {string} token - Access token
      * @return {void}
      */
@@ -706,7 +705,7 @@ class Preview extends EventEmitter {
         preload = false,
         isDocFirstPrefetchEnabled = false,
         docFirstPagesConfig = null,
-        features,
+        isAccessTokenHeaderEnabled = false,
     }) {
         let file;
         let loader;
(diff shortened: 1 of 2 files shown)


src/lib/Preview.js
```diff
@@ -747,6 +749,13 @@ class Preview extends EventEmitter {
             options.docFirstPagesConfig = docFirstPagesConfig;
         }
 
+        // If features are passed in for this prefetch call (e.g. from callers that
+        // invoke prefetch() before show() has set this.options.features), merge them
+        // into the viewer options for this call without mutating this.options.
+        if (features) {
```
ahorowitz123: we already have established patterns of feeding in feature flags so we shouldn't re-invent the wheel here

src/lib/Preview.js
```diff
@@ -747,6 +749,13 @@ class Preview extends EventEmitter {
             options.docFirstPagesConfig = docFirstPagesConfig;
         }
 
+        // If features are passed in for this prefetch call (e.g. from callers that
+        // invoke prefetch() before show() has set this.options.features), merge them
+        // into the viewer options for this call without mutating this.options.
```
jackiejou: we already have the docstring on line 696 that explains what features is for. Do we still need this one here?

jmcbgaston: good point - addressing

src/lib/Preview.js
jmcbgaston: good point - addressing

src/lib/Preview.js
```diff
@@ -745,6 +746,10 @@ class Preview extends EventEmitter {
             options.sharedLinkPassword = sharedLinkPassword;
             options.isDocFirstPrefetchEnabled = isDocFirstPrefetchEnabled;
             options.docFirstPagesConfig = docFirstPagesConfig;
+            options.features = {
+                ...this.options.features,
+                migrateAccessTokenToHeader: isAccessTokenHeaderEnabled,
+            };
```
jmcbgaston, line 752: Note on the asymmetry between `options.features.migrateAccessTokenToHeader = ...` here vs. `options.isDocFirstPrefetchEnabled = isDocFirstPrefetchEnabled` above:

The param shape matches `isDocFirstPrefetchEnabled` (top-level boolean coming into `prefetch()`), but the destination differs because the two flags are read differently by viewers:

- `isDocFirstPrefetchEnabled` → read directly off `this.options`: `const { isDocFirstPrefetchEnabled } = this.options` (`DocBaseViewer.js:464`)
- `migrateAccessTokenToHeader` → read via `this.featureEnabled('migrateAccessTokenToHeader')`, which pulls from `this.options.features`. Used in ~14 call sites across `DocBaseViewer`, `ImageViewer`, `MultiImageViewer`, `ArchiveViewer`, `PlainTextViewer`, `CSVViewer`, `BaseViewer`.

So the param matches the existing top-level shape, but the destination has to land in `options.features` to align with how viewers already consume it (and with how `PreviewContent.tsx` sets it on the `show()` path). Placing it top-level instead would require refactoring every viewer call site.

src/lib/__tests__/Preview-test.js
```diff
@@ -737,6 +737,7 @@ describe('lib/Preview', () => {
                 sharedLinkPassword,
                 isDocFirstPrefetchEnabled: false,
                 docFirstPagesConfig: null,
+                features: { migrateAccessTokenToHeader: false },
```
jackiejou, line 740: should we update this to isAccessTokenHeaderEnabled also?

src/lib/Preview.js
```diff
@@ -745,6 +746,10 @@ class Preview extends EventEmitter {
             options.sharedLinkPassword = sharedLinkPassword;
             options.isDocFirstPrefetchEnabled = isDocFirstPrefetchEnabled;
             options.docFirstPagesConfig = docFirstPagesConfig;
+            options.features = {
+                ...this.options.features,
+                migrateAccessTokenToHeader: isAccessTokenHeaderEnabled,
```
jackiejou, line 751: same here

src/lib/__tests__/Preview-test.js
jmcbgaston: see https://github.com/box/box-content-preview/pull/1643#discussion_r3230460253

src/lib/Preview.js
jmcbgaston: see https://github.com/box/box-content-preview/pull/1643#discussion_r3230460253

jackiejou (approved):

jmcbgaston:
Reopening from upstream branch so Mergify's strict-merge queue can operate (the queue check failed because this PR was from a fork). Replacement PR: #1645

PR #1645, opened 2026-05-14
feat(auth): accept optional features in prefetch()

jmcbgaston:
### Description

`prefetch()` relied on `this.options.features`, which is only set during `Preview.show()`. Callers that prefetch before `show()` (for example, folder-list hover prefetch in a host application) had no way to opt into auth-header routing for the access token, so `featureEnabled('migrateAccessTokenToHeader')` checks inside viewer `prefetch()` methods always read the default (`false`).

This change adds an optional top-level `isAccessTokenHeaderEnabled = false` param to `prefetch()`, matching the shape of the existing `isDocFirstPrefetchEnabled` param. When the prefetch is for `preload`, the per-call viewer options merge in `features.migrateAccessTokenToHeader` so viewer `prefetch()` methods see the threaded value — without mutating `this.options.features`.

- Callers that omit the flag get byte-identical behavior.
- Subsequent `show()` / `prefetch()` calls aren't affected by prior per-call state.

### Test plan

- [x] `yarn test src/lib/__tests__/Preview-test.js` — 271 passing, including new tests covering flag-on (feature surfaced to viewer options) and flag-off (feature surfaced as `false`) paths
- [x] Host app verified hover-prefetch requests include `Authorization: Bearer` header (no `access_token` in URL) when passing `isAccessTokenHeaderEnabled: true`
- [x] Host app verified omitting / passing `false` preserves existing behavior

### Semantic release type

- [x] `feat` - New feature

Replaces #1643 (was opened from a fork; Mergify strict-merge queue could not operate on it).

### DIFF
--- a/src/lib/Preview.js
+++ b/src/lib/Preview.js
@@ -693,6 +693,7 @@ class Preview extends EventEmitter {
      * @param {string} options.sharedLink - Shared link
      * @param {string} options.sharedLinkPassword - Shared link password
      * @param {boolean} options.preload - Is this prefetch for a preload
+     * @param {Object} [options.features] - Feature flags to merge into viewer options for this prefetch
      * @param {string} token - Access token
      * @return {void}
      */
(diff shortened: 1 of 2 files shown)


jackiejou (approved):

jmcbgaston:
@Mergifyio queue

PR #1655, opened 2026-05-27
fix(media): load filmstrip via blob URL when access-token is in header

jmcbgaston:
### Description

When `migrateAccessTokenToHeader` is enabled the DASH filmstrip URL is built without an `access_token` query param. The URL is consumed as `` (legacy `MediaControls` and React `Filmstrip`), so the request goes out with no `Authorization` header and 401s for unauthenticated viewers — there's no session-cookie fallback when the access token is the only credential.

Wait for `filmstripStatus` to resolve, fetch the filmstrip via `fetchContentAsBlobUrl()`, and assign the resulting `blob:` URL — the same pattern already used by `ImageViewer`, `MultiImageViewer`, and `MediaBaseViewer`. `filmstripUrl` is deferred until the blob is ready to avoid an intermediate `` render that would 401. The blob URL is revoked on `destroy()` and on destroy-mid-fetch. Filmstrip is a non-critical scrubbing enhancement, so blob-fetch failures log a `console.warn` and are otherwise swallowed rather than failing the whole viewer.

### Test plan

- [x] Open a DASH-eligible video preview via a shared link (unauthenticated viewer) with `migrateAccessTokenToHeader` ON. Hover the timeline scrubber and confirm filmstrip thumbnails appear.
- [x] Open the same preview with the flag OFF. Confirm filmstrip thumbnails still appear on hover (legacy URL-token path).
- [x] Open the preview as a logged-in user with the flag ON. Confirm scrubbing thumbnails render the same as the shared-link case.
- [x] Open the preview and immediately close it before the filmstrip has had time to load. Confirm the close completes cleanly with no visible error state.

### Semantic release type

- [x] `fix` - Bug fix


src/lib/viewers/media/DashViewer.js
```diff
@@ -119,6 +119,11 @@ class DashViewer extends VideoBaseViewer {
             this.filmstripStatus.destroy();
         }
 
+        // Release blob: URL allocated for the filmstrip when migrateAccessTokenToHeader is on
+        if (this.filmstripUrl && this.filmstripUrl.startsWith('blob:')) {
```
jmcbgaston, line 123: **`startsWith('"blob:"')` guard**: the flag-off path still stores a regular HTTPS URL in `filmstripUrl`. Only the headers path produces a `blob:` URL that needs revoking — the guard prevents `revokeObjectURL` from being called on a non-blob string.

src/lib/viewers/media/DashViewer.js
```diff
@@ -905,21 +910,56 @@ class DashViewer extends VideoBaseViewer {
         const filmstripInterval = filmstrip && filmstrip.metadata && filmstrip.metadata.interval;
 
         if (filmstripInterval > 0) {
-            const url = this.featureEnabled('migrateAccessTokenToHeader')
+            const useHeaders = this.featureEnabled('migrateAccessTokenToHeader');
+            const useReactControls = this.useReactControls();
```
jmcbgaston, line 914: **Caching `useReactControls()`**: read once at the top so the rest of the function sees a stable value. Two reasons: (1) avoids re-running `getViewerOption()` on every chain step, (2) defends against the `mockReturnValueOnce` pattern used in existing tests — without the cache, the second read returns `undefined` and the chain branches the wrong way.

src/lib/viewers/media/DashViewer.js
```diff
@@ -905,21 +910,56 @@ class DashViewer extends VideoBaseViewer {
...
+            const url = useHeaders
                 ? this.createContentUrlV2(filmstrip.content.url_template)
                 : this.createContentUrlWithAuthParams(filmstrip.content.url_template);
 
             this.filmstripInterval = filmstripInterval;
             this.filmstripStatus = this.getRepStatus(filmstrip);
-            this.filmstripUrl = url;
+            // When useHeaders is on, the URL has no token and would 401 if rendered as <img src>.
+            // Defer setting filmstripUrl until the blob: URL is ready below.
+            this.filmstripUrl = useHeaders ? null : url;
```
jmcbgaston, line 923: **`null` instead of the tokenless URL**: if we set `filmstripUrl = url` here and React re-rendered before the blob fetch completed, the `` would request the tokenless URL with no `Authorization` header and 401. Holding `null` blocks the render until the blob is ready.

src/lib/viewers/media/DashViewer.js
```diff
@@ -905,21 +910,56 @@ class DashViewer extends VideoBaseViewer {
...
 
-            if (this.useReactControls()) {
-                this.filmstripStatus.getPromise().then(() => {
-                    this.renderUI(); // Render once the filmstrip is ready
-                });
-            } else {
+            // Legacy controls have a synchronous init path when the URL already carries a token.
+            if (!useHeaders && !useReactControls) {
```
jmcbgaston, line 926: **Synchronous fast-path preserved**: the flag-off + legacy-controls combo still goes straight to `initFilmstrip(url, …)` with no microtask hops — byte-equivalent to the pre-PR behavior. This is still the dominant path until `migrateAccessTokenToHeader` rolls out everywhere.

src/lib/viewers/media/DashViewer.js
```diff
@@ -905,21 +910,56 @@ class DashViewer extends VideoBaseViewer {
...
                 this.mediaControls.initFilmstrip(url, this.filmstripStatus, this.aspect, filmstripInterval);
+                return;
             }
+
+            this.filmstripStatus
+                .getPromise()
+                .then(() => (useHeaders ? this.fetchContentAsBlobUrl(url) : url))
```
jmcbgaston, line 933: **Single chain, conditional swap**: `useHeaders ? fetchContentAsBlobUrl(url) : url` lets one chain serve both "need blob" and "already-tokenized URL" cases. Returning a non-promise `url` is fine — the next `.then()` resolves it normally.

src/lib/viewers/media/DashViewer.js
```diff
@@ -905,21 +910,56 @@ class DashViewer extends VideoBaseViewer {
...
+                .then(filmstripUrl => {
+                    // <img src> can't carry an Authorization header, so when the access-token has
+                    // been migrated out of the URL we expose the rep as a blob: URL instead.
+                    // If the viewer was destroyed while we were fetching, release the blob now —
+                    // destroy() can't see it because filmstripUrl was still null when it ran.
+                    if (this.destroyed) {
```
jmcbgaston, line 939: **Destroy-mid-fetch protection**: when `useHeaders` is on, `filmstripUrl` is `null` (line 923) until this resolved-handler runs. So if `destroy()` fires while we're awaiting the blob fetch, the destroy-side revoke (line 122) sees `null` and skips. This branch covers that gap.

src/lib/viewers/media/DashViewer.js
```diff
@@ -905,21 +910,56 @@ class DashViewer extends VideoBaseViewer {
...
+                        if (useHeaders) {
```
jmcbgaston, line 940: **`if (useHeaders)` is belt-and-suspenders**: only the headers path produces a blob URL, so revoking unconditionally would also be safe. The explicit guard makes the invariant readable rather than relying on the reader to trace the conditional in line 933.

src/lib/viewers/media/DashViewer.js
```diff
@@ -905,21 +910,56 @@ class DashViewer extends VideoBaseViewer {
...
+                            this.aspect,
+                            filmstripInterval,
+                        );
+                    }
+                })
+                // Filmstrip is a non-critical scrubbing-preview enhancement, so any failure
+                // (rep status reject, blob fetch reject) is swallowed rather than failing
+                // the whole viewer. Warn so prod 401s leave a breadcrumb in DevTools.
+                .catch(err => {
+                    console.warn('Filmstrip load failed', err); // eslint-disable-line no-console
```
jmcbgaston, line 961: **`console.warn`, not silent**: matches the convention at `DocBaseViewer.js:346` and `MediaBaseViewer.js:141` (with the same `// eslint-disable-line no-console`). Filmstrip is non-critical so we don't fail the viewer, but a prod 401 should still leave a breadcrumb in DevTools.

src/lib/viewers/media/__tests__/DashViewer-test.js
```diff
@@ -16,6 +16,8 @@ const CSS_CLASS_MEDIA = 'bp-media';
 const CSS_CLASS_HD = 'bp-media-controls-is-hd';
 const sandbox = sinon.createSandbox();
 
+const flushPromises = () => new Promise(resolve => process.nextTick(resolve));
```
jmcbgaston, line 19: **`flushPromises` helper**: the new chain is `getPromise().then().then()` — multiple microtask hops per `loadFilmStrip()` call. `process.nextTick` drains the queue, which a single `await mockPromise` can't do.

src/lib/viewers/media/__tests__/DashViewer-test.js
```diff
@@ -911,13 +913,11 @@ describe('lib/viewers/media/DashViewer', () => {
             expect(stubs.createUrl).toBeCalled();
         });
 
-        test('should render the controls again after the filmstrip is ready', done => {
-            const mockPromise = Promise.resolve();
-
+        test('should render the controls again after the filmstrip is ready', async () => {
```
jmcbgaston, line 916: **Converted from `done` callback to `async/await`**: the original `mockPromise.then(...)` only awaited the first link in the chain. With the refactor adding a second `.then()` (the `useHeaders ? fetch : url` step), the original test races and `renderUI` hasn't fired yet when the assertion runs. `flushPromises` drains both.

src/lib/viewers/media/__tests__/DashViewer-test.js
```diff
@@ -2093,28 +2091,122 @@ describe('lib/viewers/media/DashViewer', () => {
...
+            await flushPromises();
+            expect(dash.filmstripUrl).toBe('blob:http://localhost/abc');
+
+            dash.destroy();
+
+            expect(revokeSpy).toHaveBeenCalledWith('blob:http://localhost/abc');
+            revokeSpy.mockRestore();
+        });
+
+        test('should release the blob URL if the viewer is destroyed mid-fetch', async () => {
```
jmcbgaston: **Mid-fetch destroy pattern**: manually resolving `fetchPromise` via the `resolveFetch` closure is the only deterministic way to keep the fetch "in flight" across a `destroy()` call. With `mockResolvedValue`, the fetch would already be settled by the time we got here and the destroy ordering wouldn't be testable.

src/lib/viewers/media/__tests__/DashViewer-test.js
```diff
@@ -2093,28 +2091,122 @@ describe('lib/viewers/media/DashViewer', () => {
...
+            revokeSpy.mockRestore();
+        });
+
+        test('should swallow a fetchContentAsBlobUrl rejection without breaking the viewer', async () => {
+            jest.spyOn(dash, 'featureEnabled').mockReturnValue(true);
+            jest.spyOn(dash, 'createContentUrlV2').mockReturnValue('http://localhost/filmstrip.jpg');
+            jest.spyOn(dash, 'fetchContentAsBlobUrl').mockRejectedValue(new Error('boom'));
+            jest.spyOn(dash, 'useReactControls').mockReturnValue(false);
+            jest.spyOn(dash.mediaControls, 'initFilmstrip').mockImplementation();
+            const warnSpy = jest.spyOn(console, 'warn').mockImplementation();
```
jmcbgaston: **`warnSpy.mockImplementation()`**: silences the new `console.warn` so it doesn't pollute Jest's test output, while still letting us assert it was called with the expected message.

src/lib/viewers/media/__tests__/DashViewer-test.js
```diff
@@ -2093,28 +2091,122 @@ describe('lib/viewers/media/DashViewer', () => {
...
+            dash.loadFilmStrip();
+            await flushPromises();
+
+            expect(dash.filmstripUrl).toBeNull();
+            expect(dash.mediaControls.initFilmstrip).not.toHaveBeenCalled();
+            expect(warnSpy).toHaveBeenCalledWith('Filmstrip load failed', expect.any(Error));
+            warnSpy.mockRestore();
+        });
+
+        test('should fetch as blob and re-render with React controls when flag is enabled', async () => {
```
jmcbgaston, line 2194: **React-controls + headers gap**: the original five tests all set `useReactControls()` to `false`. This is the path the React `Filmstrip` consumer takes — without it, the `useHeaders && useReactControls` branch had no direct coverage.

JChan106 (approved):
lgtm

zhirongwang (approved):

PR #1661, opened 2026-05-30
feat(zoom): add pinch to zoom support for txt files

kduncanhsu:
## Summary
- Add trackpad pinch-to-zoom support for text file viewers (.txt, .json, .csv etc.), matching the existing behavior in image and document viewers
- Zoom is anchored to the line of text under the cursor using document.caretRangeFromPoint, so the line you're looking at stays in place as you zoom
- Gated behind the existing `pinchToZoom.enabled` feature flag

## Implementation

On Mac trackpads, pinch gestures fire wheel events with `ctrlKey=true`. We listen for these on the `.bp-text` element and:
1. Identify the anchor text (text where we want zoom to my anchored on): Call `document.caretRangeFromPoint(clientX, clientY)` to get a DOM `Range` pointing at the exact character under the cursor. This gives us a stable handle to a piece of text content, not a pixel position.
2. Record its screen position: Call `range.getBoundingClientRect()` to get where that line currently sits on screen, then compute its offset from the top of the scrollable container.
3. Apply zoom: Change `font-size` on the text element as a percentage (e.g. 110%). The browser reflows text — lines may wrap differently, content height changes.
4. Re-anchor: The `Range` still references the same text node + character offset (the DOM didn't change, only layout). Call `range.getBoundingClientRect()` again to find where that text moved to after reflow. Adjust `scrollTop` so it appears at the same vertical offset from the container top as before.

**Why line-level anchoring (not pixel/word-level)**

For images and PDFs, we can anchor zoom to the exact pixel under the cursor because content scales uniformly in both dimensions. Text is different — when font-size increases, lines reflow: a word that fit on one line may shift horizontally and wrap to the next line in an unpredictable way. Since the text container doesn't scroll horizontally (`white-space: pre-wrap` with `word-wrap: break-word`), there's no way to compensate for that horizontal drift.

Instead, we anchor to the line the cursor is on. After zoom, we adjust `scrollTop` so that the same line of text remains at the same vertical screen position. This keeps the user's reading context stable through the zoom — the text they were focused on doesn't jump off screen.

## Test plan
  - Enable pinchToZoom.enabled feature flag
  - Open a text file (.js, .py, .csv, etc.) in preview
  - Trackpad pinch to zoom in — verify font size increases and the line under cursor stays in place
  - Trackpad pinch to zoom out — verify font size decreases and line stays anchored
  - Verify zoom does not exceed min (0.1) or max (10) scale
  - Verify zoom controls (buttons, keyboard shortcuts) still work independently
  - Verify no effect when pinchToZoom.enabled is disabled
  - Verify no regression in image and document pinch-to-zoom

### DIFF
--- a/src/lib/viewers/text/PlainText.scss
+++ b/src/lib/viewers/text/PlainText.scss
@@ -9,6 +9,7 @@
     left: 0;
     margin: 0;
     overflow: auto;
+    line-height: 1.5;
     white-space: pre-wrap;
     text-align: left;
     word-wrap: break-word;
--- a/src/lib/viewers/text/TextBaseViewer.js
+++ b/src/lib/viewers/text/TextBaseViewer.js
@@ -9,6 +9,7 @@ const ZOOM_DEFAULT = 1.0;
 const ZOOM_MAX = 10;
 const ZOOM_MIN = 0.1;
 const ZOOM_STEP = 0.1;
+const WHEEL_ZOOM_SCALE_FACTOR = 0.01;
 
 class TextBaseViewer extends BaseViewer {
     /**
@@ -23,6 +24,7 @@ class TextBaseViewer extends BaseViewer {
         // Bind context for handlers;
         this.zoomOut = this.zoomOut.bind(this);
         this.zoomIn = this.zoomIn.bind(this);
+        this.wheelZoomHandler = this.wheelZoomHandler.bind(this);
     }
 
     /**
@@ -42,12 +44,98 @@ class TextBaseViewer extends BaseViewer {
      * @return {void}
      */
     destroy() {
+        this.unbindDOMListeners();
+
         // Destroy the controls
         if (this.controls && typeof this.controls.destroy === 'function') {
             this.controls.destroy();
         }
 
         super.destroy();
+    }
+
+    /**
+     * Binds DOM listeners for the text viewer.
+     *
+     * @protected
+     * @return {void}
+     */
+    bindDOMListeners() {
+        const textEl = this.containerEl && this.containerEl.querySelector('.bp-text');
+        if (textEl && this.featureEnabled('pinchToZoom.enabled')) {
+            textEl.addEventListener('wheel', this.wheelZoomHandler, { passive: false });
+        }
+    }
+
+    /**
+     * Unbinds DOM listeners for the text viewer.
+     *
+     * @protected
+     * @return {void}
+     */
+    unbindDOMListeners() {
+        const textEl = this.containerEl && this.containerEl.querySelector('.bp-text');
+        if (textEl) {
+            textEl.removeEventListener('wheel', this.wheelZoomHandler);
+        }
+    }
+
+    /**
+     * Handles trackpad pinch-to-zoom via wheel events with ctrlKey.
+     * On Mac trackpads, pinch gestures fire wheel events with ctrlKey set to true.
+     * Anchors zoom on the cursor position so content under the cursor stays in place.
+     *
+     * @protected
+     * @param {WheelEvent} event - wheel event object
+     * @return {void}
+     */
+    wheelZoomHandler(event) {
+        const textEl = this.containerEl && this.containerEl.querySelector('.bp-text');
+        if (!event.ctrlKey || !textEl) {
+            return;
+        }
+
+        event.preventDefault();
+
+        const scaleDelta = -event.deltaY * WHEEL_ZOOM_SCALE_FACTOR;
+        const newScale = Math.min(ZOOM_MAX, Math.max(ZOOM_MIN, this.scale + scaleDelta));
+
+        if (newScale === this.scale) {
+            return;
+        }
+
+        // Find the text position under the cursor so we can keep it anchored
+        const range = document.caretRangeFromPoint ? document.caretRangeFromPoint(event.clientX, event.clientY) : null;
+
+        let anchorScreenY = event.clientY;
+        if (range) {
+            const rangeRect = range.getBoundingClientRect();
+            anchorScreenY = rangeRect.top;
+        }
+
+        // Record where on screen the anchor line currently sits
+        const textRect = textEl.getBoundingClientRect();
+        const anchorViewportOffset = anchorScreenY - textRect.top;
+
+        // Apply zoom via font-size
+        textEl.style.fontSize = `${Math.round(newScale * 100)}%`;
+        this.scale = newScale;
+
+        // Find where the anchor text has moved to and adjust scroll
+        if (range) {
+            const newRangeRect = range.getBoundingClientRect();
+            const newTextRect = textEl.getBoundingClientRect();
+            const newPositionInContent = newRangeRect.top - newTextRect.top + textEl.scrollTop;
+            textEl.scrollTop = newPositionInContent - anchorViewportOffset;
+        }
+
+        this.emit('zoom', {
+            canZoomIn: newScale < ZOOM_MAX,
+            canZoomOut: newScale > ZOOM_MIN,
+            zoom: newScale,
+        });
+
+        this.renderUI();
     }
 
     /**
@@ -124,6 +212,7 @@ class TextBaseViewer extends BaseViewer {
      */
     loadUI() {
         this.controls = new ControlsRoot({ containerEl: this.containerEl, fileId: this.options.file.id });
+        this.bindDOMListeners();
         this.renderUI();
     }
 
--- a/src/lib/viewers/text/TextBaseViewer.js
+++ b/src/lib/viewers/text/TextBaseViewer.js
@@ -104,8 +104,19 @@ class TextBaseViewer extends BaseViewer {
             return;
         }
 
-        // Find the text position under the cursor so we can keep it anchored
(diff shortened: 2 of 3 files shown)


src/lib/viewers/text/__tests__/TextBaseViewer-test.js
```diff
@@ -174,4 +174,175 @@ describe('lib/viewers/text/TextBaseViewer', () => {
...
+        });
+
+        afterEach(() => {
+            textBase.containerEl.removeChild(textEl);
+        });
+
+        test('should remove wheel event listener', () => {
+            jest.spyOn(textEl, 'removeEventListener');
+
+            textBase.unbindDOMListeners();
```
jackiejou: is it possible to test destroy in unit test instead of directly calling unbindDOMListeners?

src/lib/viewers/text/TextBaseViewer.js
```diff
@@ -50,6 +54,90 @@ class TextBaseViewer extends BaseViewer {
...
     }
 
+    /**
+     * Binds DOM listeners for the text viewer.
+     *
+     * @protected
+     * @return {void}
+     */
+    bindDOMListeners() {
+        const textEl = this.containerEl && this.containerEl.querySelector('.bp-text');
```
jackiejou, line 64: should we consider a more specific selector (or add a new selector to use)? this name may be confused with blueprint text...

src/lib/viewers/text/TextBaseViewer.js
```diff
@@ -50,6 +54,90 @@ class TextBaseViewer extends BaseViewer {
...
+
+        const scaleDelta = -event.deltaY * WHEEL_ZOOM_SCALE_FACTOR;
+        const newScale = Math.min(ZOOM_MAX, Math.max(ZOOM_MIN, this.scale + scaleDelta));
+
+        if (newScale === this.scale) {
+            return;
+        }
+
+        // Find the text position under the cursor so we can keep it anchored
+        const range = document.caretRangeFromPoint ? document.caretRangeFromPoint(event.clientX, event.clientY) : null;
```
jmcbgaston: 🟠 `caretRangeFromPoint` is WebKit/Blink only — Firefox exposes `caretPositionFromPoint` with a different shape (`{ offsetNode, offset }`) and never matches the ternary, so `range` ends up `null`.

In Firefox the handler still zooms (good), but the `if (range) { ... }` re-anchor block at the bottom is skipped, so `scrollTop` isn't corrected and the line under the cursor jumps off-screen — exactly the UX this PR is trying to prevent.

Two options:
1. Add a Firefox path:
   ```js
   const range = document.caretRangeFromPoint?.(event.clientX, event.clientY)
       ?? caretPositionToRange(document.caretPositionFromPoint?.(event.clientX, event.clientY));
   ```
   (small helper to wrap `{ offsetNode, offset }` in a `Range`)
2. Document the Firefox limitation explicitly and degrade to the `event.clientY` fallback for both the read *and* the re-anchor — currently it's mixed (we use `clientY` for the initial anchor but skip correction entirely).

Either is fine, but the current state silently gives Firefox users a worse experience than the PR description implies.

src/lib/viewers/text/PlainText.scss
```diff
@@ -9,6 +9,7 @@
     left: 0;
     margin: 0;
     overflow: auto;
+    line-height: 1.5;
```
jmcbgaston, line 12: 🟡 This `line-height: 1.5` isn't gated by `pinchToZoom.enabled` and isn't called out in the PR body — it changes vertical rhythm for *every* `.bp-text-plain` preview on every browser, flag on or off.

If it's load-bearing for the anchor math (caret rects more stable when line-height is explicit instead of UA default), worth a comment so the next reader knows not to "clean it up". If it's a drive-by typography tweak, I'd split it into its own PR — visual regressions to all text previews shouldn't ride on a flagged feature, since the flag gate doesn't actually protect them.

src/lib/viewers/text/TextBaseViewer.js
```diff
@@ -50,6 +54,90 @@ class TextBaseViewer extends BaseViewer {
...
+            anchorScreenY = rangeRect.top;
+        }
+
+        // Record where on screen the anchor line currently sits
+        const textRect = textEl.getBoundingClientRect();
+        const anchorViewportOffset = anchorScreenY - textRect.top;
+
+        // Apply zoom via font-size
+        textEl.style.fontSize = `${Math.round(newScale * 100)}%`;
+        this.scale = newScale;
```
jmcbgaston, line 133: 🟡 Each wheel tick does `getBoundingClientRect → set fontSize → getBoundingClientRect → getBoundingClientRect` — a style mutation sandwiched between rect reads forces synchronous layout twice per tick.

Mac trackpad pinch fires wheel events at ~60Hz and a fast pinch can deliver dozens before the user lifts. On a 50k-line CSV (preview's not unusual at-scale case) this is the kind of pattern that visibly stutters.

A few options:
- rAF-throttle the zoom application — coalesce rapid wheel events into one paint
- Read both rects into locals before the style write (you can't avoid the second read since it depends on the new font-size, but you can at least batch the pre-read)
- Worst-case: benchmark on a large file before shipping and confirm it's fine

Not a blocker, but worth measuring before this lands.

src/lib/viewers/text/__tests__/TextBaseViewer-test.js
```diff
@@ -174,4 +174,175 @@ describe('lib/viewers/text/TextBaseViewer', () => {
...
+                clientX: 400,
+                clientY: 200,
+            });
+
+            textBase.wheelZoomHandler(event);
+
+            expect(document.caretRangeFromPoint).toBeCalledWith(400, 200);
+            expect(textBase.scale).toBeGreaterThan(1.0);
+
+            delete document.caretRangeFromPoint;
```
jmcbgaston: 🟡 `delete document.caretRangeFromPoint` at the end of the test body leaks if any earlier assertion throws — the cleanup never runs and the mock bleeds into subsequent tests in the file.

Move it to `afterEach` so cleanup is guaranteed:
```js
afterEach(() => {
    delete document.caretRangeFromPoint;
});
```

Or use `jest.spyOn(document, 'caretRangeFromPoint')` if jsdom lets you (depends on whether the property is configurable in your test env) — `mockRestore()` then handles it automatically.

Also worth adding a sibling test for the `document.caretRangeFromPoint = undefined` branch so the Firefox-style fallback path is exercised — even if behavior is degraded, the regression boundary is locked in.

src/lib/viewers/text/TextBaseViewer.js
kduncanhsu: this selector is already used throughout this codebase - it's the existing convention in the text viewers. It's also included in the existing `zoom()` function below. (I think "bp" here stands for "box preview")

though I agree it could be confusing... I think it makes more sense to follow up with a rename in a separate PR since this touches many different files

src/lib/viewers/text/TextBaseViewer.js
kduncanhsu: good catch. i was not able to test this myself as I couldn't get my local `box-content-preview` code to load on my devpod on firefox, but I digged into the [MDN docs](https://developer.mozilla.org/en-US/docs/Web/API/Document/caretRangeFromPoint?utm_source=chatgpt.com) which also recommend using `caretPositionFromPoint` for Firefox. will update accordingly

(from the docs `caretRangeFromPoint` support was actually introduced in April 2026 for Firefox 150, but since this is relatively new it still recommends using `caretPositionFromPoint` for maximum compatibility).

src/lib/viewers/text/__tests__/TextBaseViewer-test.js
kduncanhsu: good idea, have updated

src/lib/viewers/text/TextBaseViewer.js
kduncanhsu: have updated

src/lib/viewers/text/__tests__/TextBaseViewer-test.js
kduncanhsu: have updated - moved cleanup to `afterEach` and also grouped the three caret tests into a nested `describe('caret-based scroll anchoring')`

src/lib/viewers/text/PlainText.scss
kduncanhsu: true, though I added this style change not for the anchor math but to fix an existing style issue where txt files lines overlap when zoomed in significantly 

BEFORE (existing behavior on prod)
Screenshot 2026-06-01 at 6 25 01 PM

AFTER
Screenshot 2026-06-01 at 6 25 17 PM

src/lib/viewers/text/TextBaseViewer.js
kduncanhsu: i see what you're saying and am still looking into how i would implement this, but this optimization may be unnecessary. i was able to test this with a large csv file (~100k lines) and it performed just fine.

it may also make sense to handle this in a follow up PR, what do you think?

jackiejou (approved):

PR #1670, opened 2026-06-08
chore(doc): remove obsolete pdfDocument.destroy() call

ahorowitz123:
### Summary

Remove the `this.pdfViewer.pdfDocument.destroy()` call from `DocBaseViewer.destroy()`. `PDFDocumentProxy.destroy()` was removed in pdfjs >= 5.x; the `pdfLoadingTask` teardown earlier in the same path handles document-level cleanup (worker + transport).

### Why

After the bump to pdfjs 6.x, the call is a no-op today only because `pdfDocument` happens to no longer expose a `destroy` method — but the truthy guard means we never throw, we just silently skip. That's a footgun if the shape ever changes again. The loading-task `destroy()` is the canonical cleanup path in the pdfjs 6.x API.

### Changes

- `src/lib/viewers/doc/DocBaseViewer.js`: remove the `if (this.pdfViewer.pdfDocument) { this.pdfViewer.pdfDocument.destroy(); }` block; comment explains why only `pdfViewer.cleanup()` remains.
- `src/lib/viewers/doc/__tests__/DocBaseViewer-test.js`: update the `destroy()` test to no longer stub a `pdfDocument.destroy` (that stub would have masked a regression).

### Test plan

- [x] `yarn jest src/lib/viewers/doc/__tests__/DocBaseViewer-test.js -t destroy` passes
- [ ] Open and close several PDF previews; verify no console errors and no leaked workers in DevTools

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/lib/viewers/doc/DocBaseViewer.js
+++ b/src/lib/viewers/doc/DocBaseViewer.js
@@ -266,13 +266,11 @@ class DocBaseViewer extends BaseViewer {
             }
         }
 
-        // Clean up viewer and PDF document object
+        // Clean up viewer and PDF document object. Note: pdfjs >= 5.x removed
+        // PDFDocumentProxy.destroy(); the loading task above tears down the
+        // worker + transport, which is the only document-level cleanup needed.
         if (this.pdfViewer) {
             this.pdfViewer.cleanup();
-
-            if (this.pdfViewer.pdfDocument) {
-                this.pdfViewer.pdfDocument.destroy();
-            }
         }
 
         if (this.printPopup) {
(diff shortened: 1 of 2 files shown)


src/lib/viewers/doc/DocBaseViewer.js
```diff
@@ -266,13 +266,11 @@ class DocBaseViewer extends BaseViewer {
             }
         }
 
-        // Clean up viewer and PDF document object
+        // Clean up viewer and PDF document object. Note: pdfjs >= 5.x removed
+        // PDFDocumentProxy.destroy(); the loading task above tears down the
+        // worker + transport, which is the only document-level cleanup needed.
```
jackiejou: do we need this comment at all? since the code is removed, future devs wont know about the destroy anyway

jackiejou (approved):

PR #1673, opened 2026-06-11
feat(video): update V2 player layout and narrow viewport handling

kduncanhsu:
## Summary
  - Mount V2 controls on the wrapper element so their width is constrained by the viewport rather than the video element
  - Add padding, radial background gradient, and subtle border styles for the V2 video container
  - Hide skip forward/backward buttons on narrow viewports to prevent controls overflow

## Known Bug

There is a known issue where the preload thumbnail of the video (i.e. the still image when user opens video but not does play the video) does not match the exact dimensions/orientation of the video when played. This will be fixed in a future PR
  
  ## Test plan
  - [ ] Verify V2 video player controls span the viewport width, not the video width
  - [ ] Verify padding and background gradient render correctly around the video
  - [ ] Resize viewport to narrow width and confirm skip buttons are hidden
  - [ ] Confirm V1 player (feature flag off) is unaffected

### DIFF
--- a/src/lib/viewers/media/VideoBaseViewer.js
+++ b/src/lib/viewers/media/VideoBaseViewer.js
@@ -105,6 +105,11 @@ class VideoBaseViewer extends MediaBaseViewer {
             this.bufferingSpinnerEl.style.marginTop = `-${VIDEO_PLAYER_CONTROL_BAR_HEIGHT / 2 + SPINNER_HALF_SIZE}px`;
         }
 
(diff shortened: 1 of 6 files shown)


ahorowitz123:
Do we need any tests here?

kduncanhsu:
> Do we need any tests here?

added them! changes were stashed and i didn't realize they weren't committed 😅

ahorowitz123 (approved):

PR #1674, opened 2026-06-12
feat(video): add time format dropdown to V2 player timestamp

kduncanhsu:
## Summary

Makes the timestamp in the V2 video controls bar clickable, opening a dropdown with three time format options:

- **Standard time** — e.g. `1:05 / 2:00` (default)
- **Timecode** — e.g. `00:01:05:00 / 00:02:00:00`
- **Frame numbers** — e.g. `Frame 157 / 1567`
  
The selected format persists for the duration of the viewing session (React state). Timecode and frame formats use the actual video FPS from the Shaka manifest when available, falling back to 24fps. In these formats the current time label updates on animation frames for frame-accurate display during playback.
  
### Feature flag changes
  
FPS is now decoupled from `frameStep.enabled`. New `DashViewer.getFps()`:
- `videoPlayerV2.enabled` → always passes fps (manifest value or 24 fallback), so the format dropdown and frame-accurate scrubber stepping work for all V2 users
- Old player → unchanged (fps only when `frameStep.enabled` and manifest fps is available)
- `frameStep.enabled` now only gates `,` / `.` frame-by-frame keyboard navigation
    
Note: since fps is now always set under V2, the scrubber's arrow-key step is one frame instead of 5 seconds for all V2 users.
  
  ### Other changes
  
- `FilmstripV2` hover preview now always shows timecode instead of standard time
- New `TimestampControl` component (dropdown UI matching designs); replaces the static timestamp span in `VideoControlsV2` 
- New i18n strings: `media_time_format`, `media_time_format_standard`,`media_time_format_timecode`, `media_time_format_frames`,`media_frame_number`

## Screenshots

Screenshot 2026-06-12 at 4 57 13 PM
Screenshot 2026-06-12 at 4 57 04 PM
Screenshot 2026-06-12 at 4 57 20 PM
Screenshot 2026-06-12 at 4 57 28 PM
Screenshot 2026-06-12 at 4 57 40 PM
    
## Test plan
- [ ] 12 new `TimestampControl` unit tests (formats, open/close, outside click, Escape, fps calculation)
- [ ] 5 new `DashViewer.getFps()` tests covering flag combinations
- [ ] Updated `VideoControlsV2` timestamp tests
- [ ] All media + controls suites pass (69 suites, 1107 tests); `lint:ts`,`lint:js`, `lint:css` clean
- [ ] Manual: verify dropdown styling against Figma, format switching during playback, narrow-viewport behavior (current time only), and that `,`/`.` keys remain gated behind `frameStep.enabled`


src/lib/viewers/controls/media/TimestampControl.tsx
```diff
@@ -0,0 +1,179 @@
...
+    useClickOutside(containerElRef, () => setIsOpen(false));
+
+    React.useEffect(() => {
+        // Re-render via timeupdate events is too slow (~4Hz) for frame-accurate
+        // formats, so update the current time text on animation frames instead
+        if (format === TimeFormat.STANDARD || !mediaEl || !currentElRef.current) {
+            return undefined;
+        }
+
+        const tick = (): void => {
```
zhirongwang, line 68: Is there a way to pause this loop if the video is paused ?

src/lib/viewers/controls/media/TimestampControl.tsx
```diff
@@ -0,0 +1,179 @@
...
+
+export type Props = {
+    currentTime?: number;
+    durationTime?: number;
+    fps?: number;
+    isNarrowWidth?: boolean;
+    mediaEl?: HTMLVideoElement | null;
+};
+
+export function formatLabel(time: number, format: TimeFormat, fps: number): string {
```
zhirongwang, line 27: Is it worthy to add `isFinite` check for fps ?
```

export function formatLabel(time: number, format: TimeFormat, fps: number): string {
      const safeFps = isFinite(fps) && fps > 0 ? fps : DEFAULT_FPS;
      switch (format) {
          case TimeFormat.TIMECODE:
              return formatTimecode(time, safeFps);
          case TimeFormat.FRAMES:
              return replacePlaceholders(__('media_frame_number'), [Math.floor(time * safeFps).toString()]);
          case TimeFormat.STANDARD:
          default:
              return formatTime(time);
      }
  }
```

src/lib/viewers/controls/media/TimestampControl.tsx
kduncanhsu: yes we can, we can listen to play/pause events on the video element to start/stop the loop

thanks for the suggestion, this seems like a good optimization as we don't need to run the loop when the time isn't changing

src/lib/viewers/controls/media/TimestampControl.tsx
kduncanhsu: I think this is unnecessary. The source (`getVideoFps`) reads from the Shaka manifest which always provides a strict finite number

src/lib/viewers/controls/media/TimestampControl.tsx
```diff
@@ -0,0 +1,194 @@
...
+        }
+
+        const tick = (): void => {
+            if (currentElRef.current && mediaEl) {
+                currentElRef.current.textContent = formatLabel(mediaEl.currentTime, format, fps);
+            }
+            rafIdRef.current = requestAnimationFrame(tick);
+        };
+
+        const startLoop = (): void => {
```
zhirongwang, line 75: We may need to avoid multiple loop running at a time. 

> 1. Duplicate rAF loop leak — Add cancelAnimationFrame(rafIdRef.current) at the start of startLoop (one-line fix, prevents CPU leak)

kduncanhsu: updated

zhirongwang (approved):

PR #1682, opened 2026-06-17
feat(gallery): add gallery view with full-width page thumbnail grid

aivkov123:
**Summary**

Adds a gallery view mode to document preview — a full-width, responsive grid of page thumbnails. Users toggle a GridView toolbar icon to enter an overlay showing all pages, single-click to navigate, and return to the document viewer.

The gallery renders thumbnails progressively outward from the current page, with lazy scroll-triggered loading for pages beyond the initial buffer. A persistent thumbnail cache means re-opening gallery is instant. The toolbar collapses to just the gallery toggle + fullscreen when gallery is open (will also later have pinch to zoom), and sidebar state is preserved across open/close.

**Feature flag**
  
Gallery toggle visibility is gated on features.galleryView.enabled (consumed via isFeatureEnabled). BCP never contacts Split — EUA evaluates preview_gallery_view and passes the boolean. Additionally requires page count between 2–200.

**Other changes**
  
  - New GalleryToggle toolbar button with bp-is-active highlight state and aria-pressed
  - New GalleryGrid React component mounted directly by DocBaseViewer via createRoot
  - Dedicated Thumbnail instance for gallery (not shared with sidebar or ACI - in a future version will be shared with sidebar thumbnails) — destroyed on rotation
  - Keyboard: Tab/Shift-Tab cycles tiles with focus wrap, Escape exits, Enter/Space navigates
  - Focused page tracked via onFocusChange callback — exiting via toggle navigates to the focused tile
  - Sidebar scroll synced after gallery exit via delayed setCurrentPage
  - New i18n string: gallery_view (more later in a11y pr)

**Screenshots**
Screenshot 2026-06-17 at 10 22 38 AM

Screenshot 2026-06-17 at 10 23 00 AM

Screenshot 2026-06-17 at 10 24 38 AM

**Test plan**

  - [x] 27 unit tests (8 GalleryToggle, 19 GalleryGrid) — all passing
  - [x] Lint and type check clean
  - [x] Manual: flag ON + multi-page PDF → grid icon appears, click opens grid, thumbnails load outward from current page
  - [x] Manual: single-click tile → navigates, gallery closes, sidebar restores
  - [x] Manual: Tab between tiles → highlight moves, exit navigates to focused page
  - [x] Manual: Escape exits without navigation
  - [x] Manual: rotate PDF → re-open gallery → fresh thumbnails at new rotation
  - [x] Manual: close + re-open → instant from cache
  - [x] Manual: flag OFF / single-page / 200+ pages → no grid icon

**Known bugs**

- Markup, comment on region, and highlight and comment have some issues after clicking gallery view - will be fixed in next PR
- Scrolling hard up at top or bottom will show document under gallery view for a little bit

### DIFF
--- a/src/lib/viewers/controls/gallery/GalleryToggle.tsx
+++ b/src/lib/viewers/controls/gallery/GalleryToggle.tsx
@@ -1,0 +1,27 @@
+import React from 'react';
+import IconGridView24 from '../icons/IconGridView24';
+import './GalleryToggle.scss';
+
+export type Props = {
+    isGalleryOpen?: boolean;
+    onGalleryToggle?: () => void;
+};
+
+export default function GalleryToggle({ onGalleryToggle, isGalleryOpen }: Props): JSX.Element | null {
+    if (!onGalleryToggle) {
+        return null;
+    }
+
+    return (
+        <button
+            aria-pressed={isGalleryOpen}
+            className={`bp-GalleryToggle${isGalleryOpen ? ' bp-is-active' : ''}`}
+            data-resin-target="galleryView"
+            onClick={onGalleryToggle}
+            title={__('gallery_view')}
+            type="button"
+        >
+            <IconGridView24 />
+        </button>
+    );
+}
--- a/src/lib/viewers/doc/DocBaseViewer.js
+++ b/src/lib/viewers/doc/DocBaseViewer.js
@@ -1,10 +1,12 @@
 import React from 'react';
+import { createRoot } from 'react-dom/client';
 import throttle from 'lodash/throttle';
 import BaseViewer from '../BaseViewer';
 import Browser from '../../Browser';
 import ControlsRoot from '../controls/controls-root';
 import DocControls from './DocControls';
 import DocFindBar from './DocFindBar';
+import GalleryGrid from '../gallery/GalleryGrid';
 import PageTracker from '../../PageTracker';
 import Popup from '../../Popup';
 import PreviewError from '../../PreviewError';
@@ -121,6 +123,18 @@ class DocBaseViewer extends BaseViewer {
     /** @property {Thumbnail} - Thumbnail reference */
     advancedInsightsThumbs;
 
+    /** @property {number} - Page focused in gallery via Tab */
+    galleryFocusedPage = null;
+
+    /** @property {Thumbnail} - Dedicated gallery thumbnail instance (persists between opens) */
+    galleryThumbnail;
+
+    /** @property {boolean} - Whether gallery mode is active */
+    isGalleryOpen = false;
+
+    /** @property {boolean} - Whether sidebar was open before entering gallery */
+    sidebarWasOpen = false;
+
     /** @property {PageTracker} - PageTracker instance */
     pageTracker;
 
@@ -153,6 +167,7 @@ class DocBaseViewer extends BaseViewer {
         this.handleAnnotationCreateEvent = this.handleAnnotationCreateEvent.bind(this);
         this.handleAnnotationCreatorChangeEvent = this.handleAnnotationCreatorChangeEvent.bind(this);
         this.handleDocElKeydown = this.handleDocElKeydown.bind(this);
+        this.handleGalleryNavigate = this.handleGalleryNavigate.bind(this);
         this.handlePageSubmit = this.handlePageSubmit.bind(this);
         this.onThumbnailSelectHandler = this.onThumbnailSelectHandler.bind(this);
         this.pagechangingHandler = this.pagechangingHandler.bind(this);
@@ -167,6 +182,7 @@ class DocBaseViewer extends BaseViewer {
         this.setPage = this.setPage.bind(this);
         this.throttledScrollHandler = this.getScrollHandler().bind(this);
         this.toggleFindBar = this.toggleFindBar.bind(this);
+        this.toggleGallery = this.toggleGallery.bind(this);
         this.toggleThumbnails = this.toggleThumbnails.bind(this);
         this.updateExperiences = this.updateExperiences.bind(this);
         this.updateDiscoverabilityResinTag = this.updateDiscoverabilityResinTag.bind(this);
@@ -273,6 +289,21 @@ class DocBaseViewer extends BaseViewer {
 
         if (this.printPopup) {
             this.printPopup.destroy();
+        }
+
+        if (this.galleryRoot) {
+            this.galleryRoot.unmount();
+            this.galleryRoot = null;
+        }
+
+        if (this.galleryEl) {
+            this.galleryEl.remove();
+            this.galleryEl = null;
+        }
+
+        if (this.galleryThumbnail) {
+            this.galleryThumbnail.destroy();
+            this.galleryThumbnail = null;
         }
 
         if (this.thumbnailsSidebar) {
@@ -881,6 +912,11 @@ class DocBaseViewer extends BaseViewer {
             this.thumbnailsSidebar.refresh();
         }
 
+        if (this.galleryThumbnail) {
+            this.galleryThumbnail.destroy();
+            this.galleryThumbnail = null;
+        }
+
         this.renderUI();
 
         // Emit scale with rotation angle so annotations transform to match rotated content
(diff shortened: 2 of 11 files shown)


src/lib/viewers/controls/gallery/GalleryToggle.tsx
```diff
@@ -0,0 +1,27 @@
+import React from 'react';
+import IconGridView24 from '../icons/IconGridView24';
+import './GalleryToggle.scss';
+
+export type Props = {
+    isGalleryOpen?: boolean;
+    onGalleryToggle?: () => void;
```
kduncanhsu, line 7: do these props need to be optional?

src/lib/viewers/doc/DocBaseViewer.js
```diff
@@ -1449,6 +1485,11 @@ class DocBaseViewer extends BaseViewer {
         const canDownload = checkPermission(this.options.file, PERMISSION_DOWNLOAD);
         const isAnnotationsMode = this.currentAnnotatorViewMode === ANNOTATOR_VIEW_MODES.ANNOTATIONS;
         const canRotate = this.featureEnabled('rotate.enabled');
+        const canGallery =
+            isFeatureEnabled(this.options.features, 'galleryView.enabled') &&
+            enableThumbnailsSidebar &&
```
kduncanhsu: why do we need to check `enableThumbnailsSidebar` here?

src/lib/viewers/doc/DocControls.tsx
```diff
@@ -45,6 +49,19 @@ export default function DocControls({
     pageNumber,
     scale,
 }: Props): JSX.Element {
+    if (isGalleryOpen) {
+        return (
+            <ExperiencesProvider experiences={experiences}>
+                <ControlsBar>
```
kduncanhsu: can we DRY out these two lines (these are repeated in the default return statement)

src/lib/viewers/gallery/GalleryGrid.tsx
```diff
@@ -0,0 +1,255 @@
...
+                ) : (
+                    <span className="bp-gallery-tile-placeholder" />
+                )}
+            </button>,
+        );
+    }
+
+    return (
+        // eslint-disable-next-line jsx-a11y/no-static-element-interactions
+        <div
```
kduncanhsu, line 264: consider adding a `role` and/or `aria-label` to this `div` container. screen readers might not know the context of what they opened

src/lib/viewers/gallery/GalleryGrid.tsx
```diff
@@ -0,0 +1,255 @@
...
+const SCROLL_THROTTLE_MS = 200;
+
+export type Props = {
+    pageCount: number;
+    currentPage: number;
+    onFocusChange?: (pageNum: number) => void;
+    onPageNavigate: (n: number) => void;
+    onClose: () => void;
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    thumbnail: any;
```
kduncanhsu: is it worth replacing the the `any` here with a minimal interface?

src/lib/viewers/doc/DocBaseViewer.js
```diff
@@ -2009,6 +2057,88 @@ class DocBaseViewer extends BaseViewer {
...
+                    this.galleryFocusedPage = pageNum;
+                }}
+                onPageNavigate={this.handleGalleryNavigate}
+                pageCount={this.pdfViewer.pagesCount}
+                thumbnail={this.galleryThumbnail}
+            />,
+        );
+    }
+
+    toggleGallery() {
```
kduncanhsu, line 2103: what is the focus behavior on gallery close? does the focus return the the toggle button?

src/lib/viewers/doc/DocBaseViewer.js
```diff
@@ -2009,6 +2057,88 @@ class DocBaseViewer extends BaseViewer {
...
+    }
+
+    handleGalleryNavigate(pageNum) {
+        this.toggleGallery();
+        this.setPage(pageNum);
+
+        if (this.sidebarWasOpen && this.thumbnailsSidebar) {
+            setTimeout(() => {
+                this.thumbnailsSidebar.setCurrentPage(pageNum);
+            }, THUMBNAILS_SIDEBAR_TRANSITION_TIME);
```
kduncanhsu: `handleGalleryNavigate` calls` this.toggleGallery() which already restores the sidebar and calls `setCurrentPage(navigateToPage)` when `sidebarWasOpen` is true. do we still need this identical code block which calls `setCurrentPage` here?

zhirongwang:
Looks great. One thing to optimize. When the gallery is open, scrolling past the top reveals the document pages underneath.  
Screenshot 2026-06-18 at 2 34 25 PM

Might be worth locking the scroll so it doesn't go past the top, or hiding the page behind, or maybe adding a fixed div behind the gallery layer — or other better approach.

src/lib/viewers/gallery/GalleryGrid.tsx
```diff
@@ -0,0 +1,255 @@
...
+}: Props): JSX.Element {
+    const [loadedImages, setLoadedImages] = useState<Record<number, string>>({});
+    const [focusedPage, setFocusedPage] = useState(currentPage);
+    const gridRef = useRef<HTMLDivElement>(null);
+    const queueRef = useRef<number[]>([]);
+    const isProcessingRef = useRef(false);
+    const initialLoadDoneRef = useRef(false);
+    const initialLoadCountRef = useRef(0);
+
+    function processQueue() {
```
zhirongwang, line 45: It looks like when the gallery page is open, load all thumbnails when the gallery even it they are not in the view.  Is this intentional ?   Curious if there's a plan to optimize this for larger documents. Ideally, thumbnails should be load rendered when it's close to the view.

src/lib/viewers/controls/gallery/GalleryToggle.tsx
aivkov123: They follow the existing toggle pattern (ThumbnailsToggle, RotateControl for example), onGalleryToggle is optional so `renderUI()` in DocBaseViewer passes undefined to hide the button in GalleryToggle when the flag is off or the document is not between 2- 200 pages. I have isGalleryOpen as optional following isThumbnailsOpen? but I think neither technically have to be optional.

src/lib/viewers/doc/DocBaseViewer.js
aivkov123: Yea we definitely don't need to, will remove in next commit.

src/lib/viewers/doc/DocControls.tsx
aivkov123: Yep, I just didn't think the DRY approach was cleaner/easier to understand from a human perspective, but in terms of principles it's probably better.

src/lib/viewers/gallery/GalleryGrid.tsx
aivkov123: I agree, a11y improvements are planned for a future pr that will address a11y, polish, watermarking, and thumbnail load failures.

src/lib/viewers/gallery/GalleryGrid.tsx
aivkov123: Was debating this and the interface is definitely the right approach.

src/lib/viewers/doc/DocBaseViewer.js
aivkov123: Currently the focus behaviour on gallery close is just the document body which is an a11y concern, but similar to the gallery grid a11y improvement focus return to the toggle button is planned to be in that future pr.

src/lib/viewers/doc/DocBaseViewer.js
aivkov123: Miss on my part, we don't need this duplicate block, we can decrease `handleGalleryNavigate` to be `this.galleryFocusedPage = pageNum` and`this.toggleGallery()`.

src/lib/viewers/gallery/GalleryGrid.tsx
aivkov123: So it actually doesn't load all the thumbnails, it loads up to 40 (a couple rows more than the visible amount on an average screen) right away and then after that it only loads thumbnails when you scroll close to them.

aivkov123:
> Looks great. One thing to optimize. When the gallery is open, scrolling past the top reveals the document pages underneath. Screenshot 2026-06-18 at 2 34 25 PM
> 
> Might be worth locking the scroll so it doesn't go past the top, or hiding the page behind, or maybe adding a fixed div behind the gallery layer — or other better approach.

Yep I saw this issue before but forgot to put it in known bugs, thank you for the reminder

src/lib/viewers/doc/DocBaseViewer.js
```diff
@@ -126,6 +127,12 @@ class DocBaseViewer extends BaseViewer {
     /** @property {number} - Page focused in gallery via Tab */
     galleryFocusedPage = null;
 
+    /** @property {number} - Pending setTimeout ID for delayed gallery mount */
```
aivkov123, line 130: In future I plan to have a GalleryController class considering how many properties and methods have been added to DocBaseViewer for gallery view

kduncanhsu (approved):
lgtm!

zhirongwang (approved):

aivkov123:
@mergifyio queue

PR #1683, opened 2026-06-18
fix(video): restore media container width for v1 player

kduncanhsu:
## Summary
- Restore `mediaContainerEl.style.width` assignment in `setVideoDimensions()`, gated to the V1 player only
- This line was inadvertently removed in #1673 without being guarded behind the `videoPlayerV2.enabled `split, which broke video sizing for the V1 player

## Test plan
- [ ] Verify V1 player (feature flag off) correctly sizes the media container to match the video width
- [ ] Verify V2 player (`videoPlayerV2.enabled`) is unaffected (no container width override)
- [ ] Test with various video aspect ratios and viewport sizes in V1 mode

### DIFF
--- a/src/lib/viewers/media/VideoBaseViewer.js
+++ b/src/lib/viewers/media/VideoBaseViewer.js
@@ -633,6 +633,10 @@ class VideoBaseViewer extends MediaBaseViewer {
             this.preloader.wrapperEl.style.top = '50%';
             this.preloader.wrapperEl.style.transform = 'translate(-50%, -50%)';
         }
+
+        if (!this.featureEnabled('videoPlayerV2.enabled') && this.mediaContainerEl && this.mediaEl.style.width) {
+            this.mediaContainerEl.style.width = this.mediaEl.style.width;
+        }
     }
 
     /**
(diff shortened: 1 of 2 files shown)


src/lib/viewers/media/VideoBaseViewer.js
```diff
@@ -633,6 +633,10 @@ class VideoBaseViewer extends MediaBaseViewer {
             this.preloader.wrapperEl.style.top = '50%';
             this.preloader.wrapperEl.style.transform = 'translate(-50%, -50%)';
         }
+
+        if (!this.featureEnabled('videoPlayerV2.enabled') && this.mediaContainerEl && this.mediaEl.style.width) {
```
jackiejou: should we add tests? then we can catch these earlier

kduncanhsu: done

kduncanhsu: i also refactored it slightly so that it's overriden when v2 flag is on, so we arent doing the weird check `!this.featureEnabled('videoPlayerV2.enabled')`

kduncanhsu: this way it's also a direct revert of the lines we deleted: https://github.com/box/box-content-preview/pull/1673/changes#diff-273d01742d3cacd017b25a48d01a4290275bb15620e09e3d2b1b40e9517d12f8L615-L618

jackiejou (approved):

PR #1684, opened 2026-06-22
feat(video-player-v2): add fullscreen button and fix controls auto-hide

kduncanhsu:
## Summary
- Add a fullscreen toggle button to the V2 video player, positioned 24px from the top-right corner of the video element. The button is a 48px translucent dark circle with backdrop blur and drop shadow, containing a white maximize/minimize icon.
- Fix controls not auto-hiding after interacting with the control bar (e.g. clicking play/pause) — focused elements inside the controls layer are now blurred on mouse leave so the 2-second hide timer can proceed.
- Fix controls not auto-hiding after entering fullscreen — the controls layer state is reset on fullscreen enter to clear stale cursor/focus refs.
- Separate fix: wrapped V2 video controls with `ExperienceProvider` for parity with v1 video controls so that ftux tooltips don't incorrectly appear

**Note:** The new fullscreen button only appears when the v2 video player split is on, but the controls auto-hide behavior change in this PR (blur on mouse leave, reset on fullscreen enter) are not gated behind the split for simplicity. These are minor UX improvements that apply to both V1 and V2 and should be safe without gating.

##  Changes
- VideoFullscreenButton.tsx/scss — New component that dynamically positions itself relative to the video element using a ResizeObserver.
- VideoControlsV2.tsx/scss — Renders the fullscreen button, expands the controls root to cover the full video area with `pointer-events: none` passthrough.
- DashViewer.js — Passes onFullscreenToggle via shared props.
- ControlsLayer.tsx — Blurs the active element inside the layer on mouse leave to allow auto-hide.
- VideoBaseViewer.js — Overrides handleFullscreenEnter to reset controls layer state and trigger a show→hide cycle.

## Test plan
  
  - [ ] Hover over video → controls and fullscreen button appear; stop moving → they disappear after 2s
  - [ ] Click fullscreen button → enters fullscreen; controls auto-hide after 2s
  - [ ] In fullscreen, click minimize button → exits fullscreen
  - [ ] Click play/pause in control bar, then move mouse away → controls auto-hide after 2s
  - [ ] Keyboard-only: Tab to controls → they stay visible while focused
  - [ ] Verify button stays positioned at top-right of video on window resize
  - [ ] Verify V1 player (non-V2) is unaffected


src/lib/viewers/controls/media/VideoFullscreenButton.tsx
```diff
@@ -0,0 +1,63 @@
...
+import IconArrowsMinimizeMedium24 from '../icons/IconArrowsMinimizeMedium24';
+import useFullscreen from '../hooks/useFullscreen';
+import './VideoFullscreenButton.scss';
+
+export type Props = {
+    mediaEl?: HTMLVideoElement | null;
+    onFullscreenToggle: (isFullscreen: boolean, toggleFullscreenEl: EventTarget | null) => void;
+};
+
+export default function VideoFullscreenButton({ mediaEl, onFullscreenToggle }: Props): JSX.Element {
```
jackiejou, line 12: can we add unit test for this?

kduncanhsu: added

src/lib/viewers/media/VideoControlsV2.tsx
```diff
@@ -94,105 +97,108 @@ export default function VideoControlsV2({
...
-                            />
-                            <DrawingControls
-                                annotationColor={annotationColor}
-                                annotationMode={annotationMode}
-                                onAnnotationColorChange={onAnnotationColorChange}
-                            />
-                        </>
-                    )}
+        <>
+            <VideoFullscreenButton mediaEl={mediaEl} onFullscreenToggle={onFullscreenToggle} />
```
jackiejou: is there a reason we cannot put this inside `<div className="bp-VideoControlsV2" data-testid="media-controls-wrapper-v2">`? is this only available for V2?

We should avoid return fragments for components so that we can assume each react component is 1 contained element vs here would be 2 inside VideoControlsV2.

ahorowitz123: +1

kduncanhsu: I see, thanks for the feedback. there's no reason not to put it inside the `<div>` wrapper, have updated

ahorowitz123 (approved):

PR #1687, opened 2026-06-24
feat(video-player-v2): add comment markers to scrubber bar

kduncanhsu:
## Summary
- Adds comment markers to the V2 video scrubber bar that display timestamped comments and annotations as interactive markers
- Each marker shows the author's avatar (with profile picture or color-coded initial fallback) positioned above a white tick on the scrubber bar
  - Clicking a marker seeks the video to that timestamp, activates the annotation overlay (for annotation-type markers), and emits a `commentmarkerselect` event so the sidebar can scroll
  to the corresponding item
- The scrubber track uses a CSS mask to punch transparent gaps around each marker for a clean visual separation
- Introduces a `commentmarkers` inbound event on the viewer that the host app (BUIE) emits whenever feed items change

## New files
- `MarkerAvatar.tsx` / `MarkerAvatar.scss` — Avatar component for markers with image, color-coded initial, or anonymous icon fallback
- `TimeControlsV2-test.tsx` — Unit tests for marker rendering, positioning, and click behavior
- `MarkerAvatar-test.tsx` — Unit tests for avatar states and color logic

## Modified files
- `TimeControlsV2.tsx` — Added `CommentMarker` type, marker rendering, track mask computation
- `TimeControlsV2.scss` — Marker styles, track mask application, hover effects
- `VideoControlsV2.tsx` — Threads `commentMarkers` and `onCommentMarkerClick` props
- `VideoBaseViewer.js` — `commentmarkers` event listener, `handleCommentMarkerClick` with annotator activation
- `DashViewer.js` — Passes marker props to V2 controls
- `VideoBaseViewer-test.js` — Tests for new handler methods

  ## Test plan
  - [ ] Load a video with timestamped comments and annotations in the activity sidebar
  - [ ] Verify markers appear at correct positions on the scrubber bar
  - [ ] Verify each marker shows the correct avatar (image or initial with matching color)
  - [ ] Click a comment marker — video seeks to timestamp, sidebar scrolls to comment
  - [ ] Click an annotation marker — video seeks, annotation overlay appears, sidebar scrolls
  - [ ] Create a new timestamped comment — verify a new marker appears
  - [ ] Delete a comment/annotation — verify the marker disappears
  - [ ] Hover over markers — verify tick scales and avatar expands with white border
  - [ ] Verify markers don't appear for non-video files

### DIFF
--- a/src/lib/viewers/controls/media/MarkerAvatar.tsx
+++ b/src/lib/viewers/controls/media/MarkerAvatar.tsx
@@ -1,20 +1,18 @@
 import React from 'react';
 import './MarkerAvatar.scss';
 
-const AVATAR_COLORS = [
-    '#7fb0ea',
-    '#003c84',
-    '#ffeb7f',
-    '#92e0c0',
-    '#fad98d',
-    '#91c2fd',
-    '#f69bab',
-    '#cf9ff6',
-    '#f8c08c',
-    '#a392e0',
+const AVATAR_PALETTE: ReadonlyArray<{ bg: string; fg: string }> = [
+    { bg: '#7fb0ea', fg: '#222' },
+    { bg: '#003c84', fg: '#fff' },
+    { bg: '#ffeb7f', fg: '#222' },
+    { bg: '#92e0c0', fg: '#222' },
+    { bg: '#fad98d', fg: '#222' },
+    { bg: '#91c2fd', fg: '#222' },
+    { bg: '#f69bab', fg: '#222' },
+    { bg: '#cf9ff6', fg: '#222' },
+    { bg: '#f8c08c', fg: '#222' },
+    { bg: '#a392e0', fg: '#222' },
 ];
-
-const DARK_COLOR_INDICES = new Set([1]);
 
 export type Props = {
     avatarUrl?: string;
@@ -34,9 +32,7 @@ function AnonymousAvatarIcon(): JSX.Element {
 }
 
 export default function MarkerAvatar({ avatarUrl, colorIndex = 0, initial }: Props): JSX.Element {
-    const index = colorIndex % AVATAR_COLORS.length;
-    const bgColor = AVATAR_COLORS[index];
-    const textColor = DARK_COLOR_INDICES.has(index) ? '#fff' : '#222';
+    const { bg: bgColor, fg: textColor } = AVATAR_PALETTE[colorIndex % AVATAR_PALETTE.length];
 
     const [imgFailed, setImgFailed] = React.useState(false);
     const showImage = Boolean(avatarUrl) && !imgFailed;
--- a/src/lib/viewers/controls/media/MarkerAvatar.tsx
+++ b/src/lib/viewers/controls/media/MarkerAvatar.tsx
@@ -32,7 +32,8 @@ function AnonymousAvatarIcon(): JSX.Element {
 }
 
 export default function MarkerAvatar({ avatarUrl, colorIndex = 0, initial }: Props): JSX.Element {
-    const { bg: bgColor, fg: textColor } = AVATAR_PALETTE[colorIndex % AVATAR_PALETTE.length];
+    const safeIndex = Number.isFinite(colorIndex) ? Math.abs(colorIndex) % AVATAR_PALETTE.length : 0;
+    const { bg: bgColor, fg: textColor } = AVATAR_PALETTE[safeIndex];
 
     const [imgFailed, setImgFailed] = React.useState(false);
     const showImage = Boolean(avatarUrl) && !imgFailed;
--- a/src/lib/viewers/controls/media/MarkerAvatar.tsx
+++ b/src/lib/viewers/controls/media/MarkerAvatar.tsx
@@ -1,0 +1,60 @@
+import React from 'react';
+import './MarkerAvatar.scss';
+
+const AVATAR_COLORS = [
+    '#7fb0ea',
+    '#003c84',
+    '#ffeb7f',
+    '#92e0c0',
+    '#fad98d',
+    '#91c2fd',
+    '#f69bab',
+    '#cf9ff6',
+    '#f8c08c',
+    '#a392e0',
+];
+
+const DARK_COLOR_INDICES = new Set([1]);
+
+export type Props = {
+    avatarUrl?: string;
+    colorIndex?: number;
+    initial?: string;
+};
+
+function AnonymousAvatarIcon(): JSX.Element {
+    return (
+        <svg aria-hidden="true" className="bp-MarkerAvatar-anonymousIcon" focusable="false" viewBox="0 0 16 16">
+            <path
+                d="M8 8a3 3 0 1 0 0-6 3 3 0 0 0 0 6Zm0 1.5c-2.5 0-5 1.25-5 3.75V14h10v-.75c0-2.5-2.5-3.75-5-3.75Z"
+                fill="currentColor"
+            />
+        </svg>
+    );
+}
+
+export default function MarkerAvatar({ avatarUrl, colorIndex = 0, initial }: Props): JSX.Element {
+    const index = colorIndex % AVATAR_COLORS.length;
+    const bgColor = AVATAR_COLORS[index];
+    const textColor = DARK_COLOR_INDICES.has(index) ? '#fff' : '#222';
+
+    const [imgFailed, setImgFailed] = React.useState(false);
+    const showImage = Boolean(avatarUrl) && !imgFailed;
+
+    let avatar = <AnonymousAvatarIcon />;
+    if (showImage) {
+        avatar = <img alt="" onError={() => setImgFailed(true)} src={avatarUrl} />;
+    } else if (initial) {
+        avatar = (
+            <span className="bp-MarkerAvatar-initial" style={{ color: textColor }}>
+                {initial}
+            </span>
+        );
+    }
+
+    return (
+        <span className="bp-MarkerAvatar" style={!showImage ? { backgroundColor: bgColor } : undefined}>
+            {avatar}
+        </span>
+    );
+}
--- a/src/lib/viewers/controls/media/TimeControlsV2.scss
+++ b/src/lib/viewers/controls/media/TimeControlsV2.scss
@@ -48,12 +48,18 @@
     cursor: pointer;
     transition: transform 150ms;
 
-    &:hover {
+    &:hover,
+    &:focus-visible {
         transform: translate(-50%, -50%) scale(1.7);
 
         .bp-MarkerAvatar {
             box-shadow: 0 0 0 2px $white;
             transform: translate(-50%, calc(5px / 1.7));
         }
+    }
+
+    &:focus-visible {
+        outline: 2px solid $white;
+        outline-offset: 2px;
     }
 }
--- a/src/lib/viewers/controls/media/TimeControlsV2.scss
+++ b/src/lib/viewers/controls/media/TimeControlsV2.scss
@@ -6,6 +6,10 @@
     margin-left: 10px;
 }
 
+.bp-TimeControlsV2-scrubber {
+    position: relative;
+}
+
 .bp-TimeControlsV2-slider {
     height: 20px;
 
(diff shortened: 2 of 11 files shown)


jmcbgaston:
🟠 The `commentmarkers` listener is attached in `VideoBaseViewer.loadUIReact()` (the shared base for `DashViewer` + `MP4Viewer` + `MP3Viewer` + `Video360Viewer`), so all four viewers accept and store markers — but only `DashViewer.renderUI()` threads them into `VideoControlsV2`. `MP4Viewer.renderUI()` renders the V1 `VideoControls` unconditionally and has no V2 path, so `this.commentMarkers` is set but never rendered.

Could you verify on a `.mp4` file with timestamped comments whether markers actually appear? If BUIE emits `commentmarkers` for MP4 the way it does for Dash, this is a silent drop on MP4. Markers should work on all video viewers.

----------------
^ resolved

src/lib/viewers/media/VideoBaseViewer.js
```diff
@@ -811,6 +817,18 @@ class VideoBaseViewer extends MediaBaseViewer {
         }
     }
 
+    handleCommentMarkerClick = marker => {
```
jmcbgaston, line 820: 🟡 The new event names — `commentmarkers` (inbound) and `commentmarkerselect` (outbound) — read as one token. Existing public viewer events use snake_case (`annotator_create`, `annotations_active_set`, `scrolltoannotation` being the exception, not the rule). Worth aligning on `comment_markers` + `comment_marker_select` so they grep/sort with the rest of the contract surface.

src/lib/viewers/media/VideoBaseViewer.js
```diff
@@ -811,6 +817,18 @@ class VideoBaseViewer extends MediaBaseViewer {
...
+        if (this.mediaEl) {
+            this.mediaEl.pause();
+            this.mediaEl.currentTime = marker.time;
+        }
+        if (marker.type === 'annotation' && this.annotator) {
+            this.annotator.emit('annotations_active_set', marker.id);
+        }
+        this.emit('commentmarkerselect', { id: marker.id, time: marker.time });
```
jmcbgaston: 🟡 `handleCommentMarkerClick` emits `annotations_active_set` *before* `commentmarkerselect`. The BUIE-side `commentmarkerselect` handler triggers a `scrollIntoView` on the sidebar card; the `annotations_active_set` path activates the in-video overlay (orange glow, `PopupV2` for threaded). Order isn't fatal but the user perceives "overlay flashes → sidebar scrolls" rather than "sidebar lands → overlay highlights." Worth flipping for the more natural reveal:

```js
this.emit('commentmarkerselect', { id: marker.id, time: marker.time });
if (marker.type === 'annotation' && this.annotator) {
    this.annotator.emit('annotations_active_set', marker.id);
}
```

Minor — won't block. Flag for next iteration if you'd rather not churn this PR.

src/lib/viewers/controls/media/TimeControlsV2.tsx
```diff
@@ -43,6 +57,25 @@ export default function TimeControlsV2({
...
+        sorted.forEach(pos => {
+            stops.push(`black calc(${pos}% - ${HALF_GAP}px)`);
+            stops.push(`transparent calc(${pos}% - ${HALF_GAP}px)`);
+            stops.push(`transparent calc(${pos}% + ${HALF_GAP}px)`);
+            stops.push(`black calc(${pos}% + ${HALF_GAP}px)`);
+        });
+        stops.push('black 100%');
+
+        return `linear-gradient(to right, ${stops.join(', ')})`;
+    }, [commentMarkers, durationValue]);
```
jmcbgaston: 🟡 `trackMask` rebuilds the gradient string on every change to `commentMarkers` (array identity) or `durationValue`. The BUIE side emits a fresh `markers` array on every `filteredItems` change, which fires every render of `ActivityFeedV2`. Combined with the `commentMarkers.map(...).sort(...)` inside `useMemo`, you're rebuilding an N*4-stop CSS string on every sidebar render even when the timestamps haven't changed.

Cheap fix: memoize on a stable signature (`commentMarkers.map(m => m.time).join('|')`) rather than array identity, or move the sort+stop generation behind a key derived from sorted times.

Not load-bearing — gradient construction is microseconds — but the smell is real because the BUIE-side `useEffect` deps include `filteredItems` directly, so this fires more than it needs to.

src/lib/viewers/controls/media/MarkerAvatar.tsx
```diff
@@ -0,0 +1,60 @@
...
+            <path
+                d="M8 8a3 3 0 1 0 0-6 3 3 0 0 0 0 6Zm0 1.5c-2.5 0-5 1.25-5 3.75V14h10v-.75c0-2.5-2.5-3.75-5-3.75Z"
+                fill="currentColor"
+            />
+        </svg>
+    );
+}
+
+export default function MarkerAvatar({ avatarUrl, colorIndex = 0, initial }: Props): JSX.Element {
+    const index = colorIndex % AVATAR_COLORS.length;
```
jmcbgaston: 🟡 `colorIndex` is typed as `number` here, but on the BUIE side (`ActivityFeedV2.tsx`) it's populated from `author?.id ?? 0` — and `author.id` on the wire is a **string** (e.g. `"7340978551"`). The `%` operator coerces the string to a number, which gives `NaN` for any non-numeric id and `NaN % 10` returns `NaN`, so `AVATAR_COLORS[NaN]` is `undefined` → the avatar renders with no background color.

For real user ids it happens to work (numeric strings coerce cleanly). For test fixtures, anonymous-id placeholders, or any future id format change (UUIDs, opaque tokens), this silently breaks the color palette.

Worth either typing `colorIndex` strictly and casting on the BUIE side, or hashing the id (`hashString(id) % AVATAR_COLORS.length`) so the input contract isn't "must be numeric."

mrscobbler (dismissed):

src/lib/viewers/controls/media/MarkerAvatar.tsx
```diff
@@ -0,0 +1,60 @@
...
+    '#003c84',
+    '#ffeb7f',
+    '#92e0c0',
+    '#fad98d',
+    '#91c2fd',
+    '#f69bab',
+    '#cf9ff6',
+    '#f8c08c',
+    '#a392e0',
+];
```
jmcbgaston, line 15: 🟡 BUIE already exposes an avatar palette at `box-ui-elements/src/styles/variables` (`avatarColors`, 6 colors) used by `AvatarInitials.tsx`. This PR's `TimeControlsV2.tsx` already imports from that path (`bdlGray65, white`), so the import is free.

Worth reusing it rather than redefining a 10-color palette inline — keeps the scrubber markers aligned with sidebar avatar coloring and avoids two sources of truth.

Side note: `DARK_COLOR_INDICES = new Set([1])` is implicit — if the palette gets reordered, the contrast logic silently breaks for the wrong color. If you stick with a local palette, a `{ bg, fg }` tuple per entry would be self-checking.

(Blueprint isn't the answer — BCP doesn't import `@box/blueprint-web` at runtime today and adding it pulls in CSS-loader complications.)

mrscobbler:
> 🟠 The `commentmarkers` listener is attached in `VideoBaseViewer.loadUIReact()` (the shared base for `DashViewer` + `MP4Viewer` + `MP3Viewer` + `Video360Viewer`), so all four viewers accept and store markers — but only `DashViewer.renderUI()` threads them into `VideoControlsV2`. `MP4Viewer.renderUI()` renders the V1 `VideoControls` unconditionally and has no V2 path, so `this.commentMarkers` is set but never rendered.
> 
> Could you verify on a `.mp4` file with timestamped comments whether markers actually appear? If BUIE emits `commentmarkers` for MP4 the way it does for Dash, this is a silent drop on MP4. Markers should work on all video viewers.

@jmcbgaston -- if we aren't rendering the v2 controls for MP4 videos then we shouldn't be rendering the markers? V1 controls with markers is not in scope for this project.

src/lib/viewers/media/VideoBaseViewer.js
```diff
@@ -776,6 +777,11 @@ class VideoBaseViewer extends MediaBaseViewer {
         }
     }
 
+    handleCommentMarkersUpdated = (markers = []) => {
+        this.commentMarkers = markers;
+        this.renderUI();
+    };
```
jmcbgaston, line 783: 🟢 `handleCommentMarkersUpdated = (markers = []) => { this.commentMarkers = markers; this.renderUI(); }` is fine, but `renderUI()` is also called from at least two other paths (`annotator_create` listener, plus the original `loadUIReact` end-of-setup). On a `commentmarkers` event during initial load these can fire in rapid succession — three `renderUI()` calls in a tick.

`ControlsRoot.render` is React 18's `createRoot` which batches synchronously inside one event tick, so this is harmless. Just noting in case anyone later wonders why `renderUI` is called every which way.

No change requested.

src/lib/viewers/controls/media/TimeControlsV2.scss
```diff
@@ -28,3 +33,27 @@
...
+    transition: transform 150ms;
+
+    &:hover {
+        transform: translate(-50%, -50%) scale(1.7);
+
+        .bp-MarkerAvatar {
+            box-shadow: 0 0 0 2px $white;
+            transform: translate(-50%, calc(5px / 1.7));
+        }
+    }
```
jmcbgaston: 🟡 `.bp-TimeControlsV2-marker` has `border: 0` and no `:focus` / `:focus-visible` style — only `:hover`. UA default outline is typically suppressed when border is reset, so the marker has no visible focus indicator when tabbed to, failing WCAG 2.4.7 Focus Visible (AA).

Mirror the `:hover` transform on `:focus-visible`, or add a ring (`box-shadow: 0 0 0 2px ${white}` style) that's distinct from hover.

Lower-priority side note: `aria-label="Comment marker"` is identical across N markers. WCAG 2.4.6 (Headings and Labels, AA) suggests labels should describe purpose — including author/time (e.g., `"Comment by Jose Gaston at 0:20"`) would also help screen-reader users distinguish markers.

src/lib/viewers/controls/media/MarkerAvatar.tsx
kduncanhsu: The BUIE avatarColors is ['#0061d5', '#003c84', '#6f6f6f', '#222', '#4826c2', '#9f3fed'] — 6 dark colors with white text. This is the legacy BUIE avatar palette used by `AvatarInitials.tsx`. However, the activity feed sidebar uses the Blueprint Avatar (10 pastel colors from `@box/blueprint-web-assets/tokens`). These are two different systems.

kduncanhsu: though i have updated DARK_COLOR_INDICES to use a tuple per entry approach instead

kduncanhsu: this should all be updated with blueprint lands in box-content-preview

src/lib/viewers/media/VideoBaseViewer.js
kduncanhsu: updated

src/lib/viewers/media/VideoBaseViewer.js
kduncanhsu: consider for next iteration - i just tested it out and there does not seem to be a visible difference

src/lib/viewers/controls/media/TimeControlsV2.scss
kduncanhsu: updated

src/lib/viewers/controls/media/TimeControlsV2.tsx
kduncanhsu: updated

src/lib/viewers/controls/media/MarkerAvatar.tsx
kduncanhsu: updated

mrscobbler (approved):

PR #1688, opened 2026-06-26
fix(file): cache-bust authenticated content URL per request

jmcbgaston:
### Description

Browser disk cache can replay a cached 302 redirect long after the signed URL inside the `Location` header expires. When a viewer re-fetches the authenticated content URL, the cached redirect points at a dead signed URL and the preview fails to load.

The fix gives content fetches a unique disk-cache key via a `_cache_buster` query parameter, resolved **at request time** rather than stamped once:

- `addOriginalRepresentation` (`file.js`) tags the synthesized `authenticated_download_url` with a `_cache_buster` marker. It's stamped unconditionally because the file object is cached before the viewer (and its feature flags) is known.
- `createContentUrl` (`util.js`) re-resolves `_cache_buster` to a fresh value per request, but only when its caller asks for it.
- `BaseViewer.createContentUrl` asks for re-resolution only when `migrateAccessTokenToHeader` is enabled.

The gate matters because the rotating in-URL `access_token` already varies the disk-cache key on its own. When `migrateAccessTokenToHeader` is off, the token stays in the URL and does that job, so the marker is left frozen. When the flag is on, the token moves to the `Authorization` header and the URL loses its only varying component — that's the configuration where the stale-302 replay was actually observed, and where `_cache_buster` takes over.

Resolving at request time (rather than freezing one value into the cached file object) is what makes this hold across in-SPA navigation: a value stamped once gets replayed unchanged on file -> folder -> same-file reopen, so the key never changes and the stale 302 comes back. Re-resolving in `createContentUrl` keeps every actual fetch a unique key regardless of how the file object is cached or reused.

The fix lives at the URL-construction chokepoint (`createContentUrl`), so it covers all viewers that build content URLs through it (PDF, SVG, GIF, MP3, SWF, HTML, XML), including native-tag viewers where request-header workarounds aren't an option.

### Verification

Captured customer HAR files for this incident show every failing `/content?preview=true` request carried `access_token=false` in the URL — i.e. `migrateAccessTokenToHeader` was on in 100% of observed failures, which is what scopes the fix to that flag.

Verified on devpod against the previously-failing in-SPA navigation flow with the flag on:

- Each `/content?preview=true` request carries a distinct `_cache_buster` value across reopens of the same file.
- After idling past the signed-URL TTL, reopening a file (gif, mp3, pdf) issues a fresh `302` and renders correctly.

### Test plan

- [ ] With `migrateAccessTokenToHeader` enabled, open a file in a viewer that uses the original representation (e.g., PDF, SVG)
- [ ] Confirm the `/content` request URL includes a `_cache_buster=` query parameter and no `access_token`
- [ ] Navigate away within the app and reopen the same file; confirm `_cache_buster` changes between opens
- [ ] Leave the tab idle past the signed-URL TTL, reopen the file, confirm it renders
- [ ] With the flag off, confirm the `access_token` stays in the URL and `_cache_buster` does not rotate

### Changes

code 22% · tests 78%

### DIFF
--- a/src/lib/__tests__/util-test.js
+++ b/src/lib/__tests__/util-test.js
@@ -242,15 +242,23 @@ describe('lib/util', () => {
             expect(util.createContentUrl('https://dl6.boxcloud.com', 'bar')).toBe('https://dl.boxcloud.com');
         });
 
-        test('should resolve the _cache_buster marker to a fresh value at request time', () => {
+        test('should resolve the _cache_buster marker to a fresh value at request time when enabled', () => {
             jest.spyOn(Date, 'now').mockReturnValue(0);
             expect(
-                util.createContentUrl('https://dl.boxcloud.com/content?preview=true&_cache_buster=0&version=9'),
+                util.createContentUrl(
+                    'https://dl.boxcloud.com/content?preview=true&_cache_buster=0&version=9',
+                    null,
+                    true,
+                ),
             ).toBe('https://dl.boxcloud.com/content?preview=true&_cache_buster=0&version=9');
 
             jest.spyOn(Date, 'now').mockReturnValue(parseInt('zzz', 36));
             expect(
-                util.createContentUrl('https://dl.boxcloud.com/content?preview=true&_cache_buster=0&version=9'),
+                util.createContentUrl(
+                    'https://dl.boxcloud.com/content?preview=true&_cache_buster=0&version=9',
+                    null,
+                    true,
+                ),
             ).toBe('https://dl.boxcloud.com/content?preview=true&_cache_buster=zzz&version=9');
         });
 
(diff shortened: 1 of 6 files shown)


src/lib/file.js
```diff
@@ -149,6 +149,10 @@ function addOriginalRepresentation(file) {
 
     const queryParams = {
         preview: 'true',
+        // Cache-buster so a cached 302 redirect isn't replayed after its signed URL
+        // expires. Re-resolved to a fresh value per request in createContentUrl(); the
+        // value here is a fallback for any path that skips createContentUrl().
+        _bcs: Date.now().toString(36),
```
mickr: Does using Date.now() still allow caching to be used since this would bust the cache on every minute? Should we perform a retry or check the expiry of the token instead?

src/lib/__tests__/util-test.js
```diff
@@ -241,6 +241,42 @@ describe('lib/util', () => {
             DownloadReachability.isDownloadHostBlocked.mockReturnValue(true);
             expect(util.createContentUrl('https://dl6.boxcloud.com', 'bar')).toBe('https://dl.boxcloud.com');
         });
+
+        test('should resolve the _bcs cache-buster marker to a fresh value at request time', () => {
```
mickr: Should be have a test to ensure that caching still works?

src/lib/file.js
jmcbgaston: We've got this cache busting per call - not in intervals so no need for a retry.

Seems like token expiry wasn't the issue - it's the signed URL


from cursor:
Date.now() busts per request by design, but only on the synthesized ORIGINAL-rep URL — converted reps have no _bcs and cache normally, and the signed-URL bytes were never cacheable (cache-control: private), so general caching is intact (verified prod vs devpod: only the 302 redirect is disk-cached). On token/retry: the bug isn't token expiry — the browser replays the cached 302 from disk without hitting the server and follows the expired signed URL inside it (which uses its own signature, not our bearer token), so a fresh token doesn't help. And a plain retry replays the same cached redirect → same 404; the only way to make a retry work is to bust the cache key, which is what _bcs does proactively (and it's the only approach that reaches native-tag viewers). The real fix is server-side: shorten the redirect's Cache-Control to ≤ the signed-URL TTL — tracked as a follow-up so _bcs can be removed.

src/lib/__tests__/util-test.js
jmcbgaston: this doesn't affect the other cacheing pieces outside of the original representation call - so we have a test to make sure of that here https://github.com/box/box-content-preview/pull/1688/changes#diff-c3f35c7de888fa53f97fc51ab417e6170266ecff6880738678910ddd2830e75eR269

src/lib/file.js
mickr: Ah got it, thanks for walking through it. So the cached thing is really the 302 redirect (max-age 28800), not the bytes, and the bug is that 8h cache outliving the ~1h signed URL inside it. That clicks now.

One thing I'm curious about, not blocking: since Date.now() busts on every request, we never reuse that redirect even inside the TTL. Was eager the choice mainly because the native-tag viewers can't intercept a failed fetch to retry-and-bust? If so that makes total sense, just wanted it captured here so the tradeoff is on record for the next person.

zhirongwang (dismissed):

JChan106 (dismissed):

mickr (approved):

zhirongwang (approved):

PR #1691, opened 2026-07-02
feat(gallery): a11y polish, entry-state reset, focus return

aivkov123:
## Summary

  Polish + a11y baseline pass on the gallery view.
  
  The grid now implements the full ARIA listbox pattern: `role="listbox"` on the container with `aria-label` sourced from a new `page_gallery` i18n key, `role="option"` + `aria-selected` on each tile, roving tabindex so exactly one tile is in the tab order at a time, and keyboard handlers for Arrow keys / Home / End / Enter / Space / Escape. Tiles are extracted into a memoized `<GalleryTile>` so grids with hundreds of pages don't re-render on every focus change.

  Entering the gallery now resets viewer state that would otherwise leak: the find bar closes, the annotation controls FSM is reset, andthe annotator's mode is switched to NONE. Without this, users could enter the gallery with the drawing tool still active and mark up thumbnails.

  Two rendering bugs are fixed: annotations no longer bleed through the overlay (gallery `z-index` bumped above the annotation layer, toolbar bumped above the gallery), and hard-scrolling the grid on macOS/iOS no longer exposes the doc behind it (`overscroll-behavior: none`).

  Focus returns to the Gallery toggle when the gallery closes. This required two things: `DocControls.tsx` was restructured from an `isGalleryOpen ? A : B` ternary into one tree with conditional groups so React preserves the toggle's DOM node across state flips, and `GalleryController` gained a `focusToggle` option that the viewer wires to a live DOM query for `.bp-GalleryToggle`. The controller calls it after `requestUiUpdate()` when transitioning to closed, so both toolbar-click and Escape-key close paths restore focus.

  ## Modifies

  - `src/lib/viewers/gallery/GalleryGrid.tsx` — listbox semantics, roving tabindex, arrow-key navigation, memoized `<GalleryTile>`
  - `src/lib/viewers/gallery/GalleryGrid.scss` — `z-index: 4`, `overscroll-behavior: none`
  - `src/lib/viewers/gallery/GalleryController.tsx` — `focusToggle` option, invoked after UI update on close
  - `src/lib/viewers/doc/DocBaseViewer.js` — `handleGalleryToggle` resets find bar + annotation state on entry; `focusToggle` callback queries the live toggle DOM node
  - `src/lib/viewers/doc/DocControls.tsx` — unified JSX tree so `<GalleryToggle>` stays mounted across `isGalleryOpen` flips
  - `src/lib/viewers/controls/controls-root/ControlsRoot.scss` — toolbar `z-index: 5` (above gallery)
  - `src/i18n/en-US.properties` — `page_gallery` key
  - `src/lib/viewers/gallery/__tests__/GalleryGrid-test.tsx` — listbox role assertions, arrow-key coverage, fallback tile test
  - `src/lib/viewers/gallery/__tests__/GalleryController-test.tsx` — `focusToggle` called on close, not on open
  - `src/lib/viewers/doc/__tests__/DocBaseViewer-test.js` — four tests covering `handleGalleryToggle` reset behavior

  ## Test plan
  
  - [ ] Open a multi-page doc; toggle gallery — grid opens with the current page focused and centered
  - [ ] Tab into the grid — focus lands on the selected tile only (roving tabindex)
  - [ ] Arrow / Home / End move focus tile-to-tile without leaving the grid; Enter or Space navigates to that page and closes gallery
  - [ ] Escape closes gallery; focus returns to the Gallery toggle button
  - [ ] Click the toggle to close gallery; focus returns to the toggle
  - [ ] Enable a drawing tool, then open the gallery — tool deselects, no draw overlay on thumbnails
  - [ ] Open the find bar, then open the gallery — find bar closes
  - [ ] Scroll the grid past its edges on macOS/iOS — no rubber-band exposes the doc underneath
  - [ ] With annotations visible, open the gallery — no annotations bleed through the overlay
  - [ ] Watermarked doc: gallery thumbnails render with the same watermark stamp as the sidebar
  - [ ] Axe DevTools on the open gallery — 0 critical violations

### DIFF
--- a/src/lib/viewers/doc/DocBaseViewer.js
+++ b/src/lib/viewers/doc/DocBaseViewer.js
@@ -157,6 +157,7 @@ class DocBaseViewer extends BaseViewer {
         this.handleAnnotationCreateEvent = this.handleAnnotationCreateEvent.bind(this);
         this.handleAnnotationCreatorChangeEvent = this.handleAnnotationCreatorChangeEvent.bind(this);
         this.handleDocElKeydown = this.handleDocElKeydown.bind(this);
+        this.handleGalleryToggle = this.handleGalleryToggle.bind(this);
         this.handlePageSubmit = this.handlePageSubmit.bind(this);
         this.onThumbnailSelectHandler = this.onThumbnailSelectHandler.bind(this);
         this.pagechangingHandler = this.pagechangingHandler.bind(this);
@@ -244,6 +245,12 @@ class DocBaseViewer extends BaseViewer {
             setPage: n => this.setPage(n),
             toggleThumbnails: () => this.toggleThumbnails(),
             requestUiUpdate: () => this.renderUI(),
+            focusToggle: () => {
+                const toggle = this.containerEl.querySelector('.bp-GalleryToggle');
+                if (toggle && toggle.focus) {
+                    toggle.focus();
+                }
+            },
         });
     }
 
@@ -1492,7 +1499,7 @@ class DocBaseViewer extends BaseViewer {
                 onAnnotationModeEscape={this.handleAnnotationControlsEscape}
                 onFindBarToggle={!this.isFindDisabled() ? this.toggleFindBar : undefined}
                 onFullscreenToggle={this.toggleFullscreen}
-                onGalleryToggle={canGallery ? this.galleryController.toggle : undefined}
+                onGalleryToggle={canGallery ? this.handleGalleryToggle : undefined}
                 onPageChange={this.setPage}
                 onPageSubmit={this.handlePageSubmit}
                 onRotateLeft={canRotate ? this.rotateLeft : undefined}
@@ -1992,6 +1999,26 @@ class DocBaseViewer extends BaseViewer {
     toggleFindBar(findBarToggleEl) {
         this.findBarToggleEl = findBarToggleEl;
         this.findBar.toggle();
+    }
+
+    /**
+     * Callback when the gallery toggle button is clicked. Closes the find bar and resets
+     * annotation mode before entering gallery, then delegates to the gallery controller.
+     *
+     * @protected
+     * @return {void}
+     */
+    handleGalleryToggle() {
+        if (!this.galleryController.isOpen) {
+            if (this.findBar) {
+                this.findBar.close();
+            }
+            this.processAnnotationModeChange(this.annotationControlsFSM.transition(AnnotationInput.RESET));
+            if (this.annotator) {
+                this.annotator.toggleAnnotationMode(AnnotationMode.NONE);
+            }
+        }
+        this.galleryController.toggle();
     }
 
     /**
--- a/src/lib/viewers/doc/DocBaseViewer.js
+++ b/src/lib/viewers/doc/DocBaseViewer.js
@@ -157,7 +157,6 @@ class DocBaseViewer extends BaseViewer {
         this.handleAnnotationCreateEvent = this.handleAnnotationCreateEvent.bind(this);
         this.handleAnnotationCreatorChangeEvent = this.handleAnnotationCreatorChangeEvent.bind(this);
         this.handleDocElKeydown = this.handleDocElKeydown.bind(this);
-        this.handleGalleryToggle = this.handleGalleryToggle.bind(this);
         this.handlePageSubmit = this.handlePageSubmit.bind(this);
         this.onThumbnailSelectHandler = this.onThumbnailSelectHandler.bind(this);
         this.pagechangingHandler = this.pagechangingHandler.bind(this);
@@ -251,6 +250,8 @@ class DocBaseViewer extends BaseViewer {
                     toggle.focus();
                 }
             },
+            onBeforeOpen: () => this.handleGalleryEnter(),
+            onAfterClose: () => this.handleGalleryExit(),
         });
     }
 
@@ -1499,7 +1500,7 @@ class DocBaseViewer extends BaseViewer {
                 onAnnotationModeEscape={this.handleAnnotationControlsEscape}
                 onFindBarToggle={!this.isFindDisabled() ? this.toggleFindBar : undefined}
                 onFullscreenToggle={this.toggleFullscreen}
-                onGalleryToggle={canGallery ? this.handleGalleryToggle : undefined}
+                onGalleryToggle={canGallery ? this.galleryController.toggle : undefined}
                 onPageChange={this.setPage}
                 onPageSubmit={this.handlePageSubmit}
                 onRotateLeft={canRotate ? this.rotateLeft : undefined}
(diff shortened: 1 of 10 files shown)


src/lib/viewers/doc/DocControls.tsx
```diff
@@ -51,16 +51,9 @@ export default function DocControls({
...
-                    <ControlsBarGroup>
-                        <GalleryToggle isGalleryOpen={isGalleryOpen} onGalleryToggle={onGalleryToggle} />
-                        <FullscreenToggle onFullscreenToggle={onFullscreenToggle} />
-                    </ControlsBarGroup>
-                </ControlsBar>
-            ) : (
-                <>
-                    <ControlsBar>
+            <ControlsBar>
+                {!isGalleryOpen && (
```
jackiejou, line 55: now it looks like if isGalleryOpen is true, it will just be an empty ControlsBar? is that intended?

src/lib/viewers/gallery/GalleryGrid.tsx
```diff
@@ -25,6 +26,39 @@ export type Props = {
...
 
+interface TileProps {
+    pageNum: number;
+    isFocused: boolean;
+    imageSrc?: string;
+    onClick: (pageNum: number) => void;
+    onFocus: (pageNum: number) => void;
+}
+
+const GalleryTile = React.memo(function GalleryTile({
```
jackiejou, line 37: can we just declare a normal functional component? do we need the memo?

src/lib/viewers/doc/DocControls.tsx
aivkov123: if isGalleryOpen is true, then lines 89 and 90 render the gallery toggle and the fullscreen toggle

src/lib/viewers/doc/DocBaseViewer.js
```diff
@@ -1994,6 +2001,26 @@ class DocBaseViewer extends BaseViewer {
...
+     * @return {void}
+     */
+    handleGalleryToggle() {
+        if (!this.galleryController.isOpen) {
+            if (this.findBar) {
+                this.findBar.close();
+            }
+            this.processAnnotationModeChange(this.annotationControlsFSM.transition(AnnotationInput.RESET));
+            if (this.annotator) {
+                this.annotator.toggleAnnotationMode(AnnotationMode.NONE);
```
jmcbgaston: 🟠 Resetting to `NONE` on entry is right, but nothing restores `REGION` when the gallery closes with `enableAnnotationsDiscoverability` on. `handleFullscreenExit` does the equivalent restore after the same `NONE` reset on enter:

```js
if (this.annotator && this.areNewAnnotationsEnabled() && this.options.enableAnnotationsDiscoverability) {
}
```

Without it, the click-drag region-comment gesture is dead after closing the gallery until some other interaction resets the mode. Worth mirroring here on the close path (and covering in the `handleGalleryToggle` tests).

jmcbgaston (commented):
Solid a11y pass — the listbox/roving-tabindex implementation is correct and the DocControls restructure is the right fix for focus return. One behavioral gap in the annotation-mode reset worth addressing before merge.

**Approve with minor changes.**

src/lib/viewers/gallery/GalleryGrid.tsx
```diff
@@ -198,75 +241,88 @@ export default function GalleryGrid({
...
+                    onClose();
+                    return;
+                case 'ArrowUp':
+                case 'ArrowLeft':
+                    if (focusedPage > 1) {
+                        event.preventDefault();
+                        focusTile(focusedPage - 1);
+                    }
+                    return;
+                case 'ArrowDown':
```
jmcbgaston, line 278: 🟡 `ArrowDown`/`ArrowRight` both move +1, so in the wrapped grid layout pressing Down visually moves focus right, not down a row. That's consistent with the listbox's 1-D semantics for screen readers, but surprising for sighted keyboard users. Consider row-aware navigation (jump by the current column count, derivable from tile `offsetTop`/`offsetWidth`), or a comment documenting the 1-D choice so it reads as intentional.

src/lib/viewers/gallery/GalleryGrid.tsx
```diff
@@ -25,6 +26,39 @@ export type Props = {
...
+    pageNum,
+    isFocused,
+    imageSrc,
+    onClick,
+    onFocus,
+}: TileProps): JSX.Element {
+    return (
+        // eslint-disable-next-line jsx-a11y/click-events-have-key-events
+        <div
+            aria-label={`Page ${pageNum}`}
```
jmcbgaston: 🟡 The container label goes through i18n (`__('page_gallery')`), but each tile's label is still hardcoded English. Since this pass already touches these labels (dropping the ", current page" suffix), it's the natural moment to add a properties key with a placeholder for the page number — the tests asserting the literal `Page 1` strings would need updating alongside.

src/lib/viewers/gallery/GalleryGrid.tsx
aivkov123: We could, but then every time a new tile would load in, setLoadedImages would rerender GalleryGrid which rerenders every tile's components which is a lot of unnecessary work. With the memo, we only render each tile's components once (when it loads in)

src/lib/viewers/gallery/GalleryGrid.tsx
aivkov123: Yea ArrowDown and ArrowRight as well as ArrowUp and ArrowLeft currently only go side to side, in v2 when grid is used instead of listbox, ArrowUp and ArrowDown will function as a user expect (going up and down), I'll add a comment documenting this

src/lib/viewers/doc/DocBaseViewer.js
aivkov123: Good catch, addressing in commit

src/lib/viewers/gallery/GalleryGrid.tsx
aivkov123: Good catch, addressed in latest commit

src/lib/viewers/doc/DocBaseViewer.js
jmcbgaston: 🟠 Re-reviewed `72d2223` — moving the entry/exit logic onto `onBeforeOpen`/`onAfterClose` is a real improvement (the resets now fire on every close path, not just toggle click), but the restore condition dropped the `enableAnnotationsDiscoverability` check that `handleFullscreenExit` has:

```js
if (this.annotator && this.areNewAnnotationsEnabled() && this.options.enableAnnotationsDiscoverability) {
}
```

Without it, closing the gallery on an instance with new annotations enabled but discoverability **off** force-activates REGION mode — the click-drag region tool turns on even though the user never selected it and `getInitialAnnotationMode()` would return `NONE` for that config. Needs `&& this.options.enableAnnotationsDiscoverability` in `handleGalleryExit`, plus a test covering the flag-off case (the new tests only vary `areNewAnnotationsEnabled`).

jackiejou (dismissed):

kduncanhsu (approved):

PR #1698, opened 2026-07-08
fix(annotations): default fallback annotator to threaded modernized UI

JChan106:
## Problem
When a host does not pass a `boxAnnotations` instance into Preview options, `BaseViewer.createAnnotator()` builds its own via:

```js
const boxAnnotations = this.options.boxAnnotations || new global.BoxAnnotations(viewerOptions);
```

This fallback passes only `viewerOptions` (the first constructor arg) and no `features` (the second arg). Every feature then resolves through `getProp(features, name, false)` → `false`, so `isThreadedAnnotation` is `false` and the annotator renders the **legacy `PopupReply`** instead of the GA'd threaded `PopupV2`.

This surfaces during a race where a host constructs its annotator instance asynchronously (e.g. inside `requestIdleCallback`) but Preview loads before the instance is ready — Preview receives `boxAnnotations: undefined`, hits this fallback, and shows the legacy popup even though the global `BoxAnnotations` bundle supports the threaded UI.

## Fix
Default the fallback annotator's features to the two GA'd flags:

```js
const boxAnnotations =
    this.options.boxAnnotations ||
    new global.BoxAnnotations(viewerOptions, {
        features: { isThreadedAnnotation: true, modernization: true },
    });
```

- `isThreadedAnnotation` and `modernization` are GA and set unconditionally by hosts, so a shared default of `true` is always correct.
- Flag-gated features (`drawing`, `discoverability`) are intentionally **left off** in the fallback — force-enabling them for all consumers would be a behavior change. A host that wants them passes its own instance (the normal path, which this fallback yields to).

## Test
Added a `createAnnotator()` test asserting the fallback constructor is called with `{ features: { isThreadedAnnotation: true, modernization: true } }`. All 11 `createAnnotator` tests pass; lint clean.

### DIFF
--- a/src/lib/viewers/BaseViewer.js
+++ b/src/lib/viewers/BaseViewer.js
@@ -1079,7 +1079,11 @@ class BaseViewer extends EventEmitter {
             return;
         }
 
-        const boxAnnotations = this.options.boxAnnotations || new global.BoxAnnotations(viewerOptions);
+        const boxAnnotations =
+            this.options.boxAnnotations ||
+            new global.BoxAnnotations(viewerOptions, {
+                features: { isThreadedAnnotation: true, modernization: true },
+            });
         this.annotatorConf = boxAnnotations.determineAnnotator(this.options, this.viewerConfig);
 
         if (!this.annotatorConf) {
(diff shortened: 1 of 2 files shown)


JChan106:
Closing — root-cause analysis showed the fallback path loads box-content-preview's own bundled annotations.js (pinned box-annotations beta.0, no PopupV2) and overwrites global.BoxAnnotations, so this fallback default is a no-op without also bumping box-annotations in BCP (a heavy dependency upgrade: react-redux + several @box/* peers). Fixing this correctly belongs host-side: gate the preview render on the annotator instance being ready (EUA-parity), so the host's own beta.18 instance is always passed and the fallback is never hit. Pursuing that in preview-client instead.

PR #1700, opened 2026-07-09
fix(gallery): trap keys while open

aivkov123:
Contain gallery view's keyboard behavior without walling it off: keystrokes can't leak into the underlying document or the host page's file browser, but Tab now flows out of the gallery to the sidebar and header, and Escape closes only the gallery from anywhere unless an open menu consumes it first.

## What changed
- Add `bp-is-gallery-open` on the container while the gallery is open, used to hide Preview's Prev/Next File nav (`_navigation.scss`), hide box-ui-elements' file-nav arrows that sit outside `.bp` (`GalleryGrid.scss`), and lock the underlying `.bp-doc` overflow so arrow keys can't scroll the doc or flash its scrollbar.
- Mark `.bp-doc` as `inert` while open, removing PDF.js text/annotation layers from the tab order and accessibility tree. With the doc inert, natural tab order runs tile → Gallery toggle → Fullscreen toggle → sidebar → header — no custom Tab cycling needed (the grid mounts before the controls root to keep that order).
- Centralize the gallery key policy in `DocBaseViewer.onKeydown`, which hosts invoke for any keydown inside the preview root (including the sidebar): Escape closes only the gallery and is consumed before it reaches the host's window-level Escape hotkey; arrow and bracket (`[` `]`) keys are swallowed so they can't flip the doc page or trigger collection navigation. A `defaultPrevented` guard defers to menus/dropdowns that already handled the key. `DocumentViewer` and `PresentationViewer` defer their own paging/zoom shortcuts to this policy while the gallery is open.
- Emit `galleryOpen`/`galleryClose` viewer events and expose a guarded `closeGallery()` method so hosts (preview-client/EUA header, follow-up PR) can dismiss the gallery when Escape is pressed with focus in the header.
- Teach `FocusTrap` to include roving-tabindex listbox options (`[role="option"][tabindex="0"]`, i.e. gallery tiles) and skip inert subtrees, so the fullscreen tab cycle reaches the focused tile. Scoped so no other viewer's fullscreen behavior changes.
- Apply the open state in `toggle()` (not `mountGrid()`) so it also covers the deferred grid mount behind the thumbnails-sidebar transition.

## Test plan
- [ ] Tab flows tile → Gallery toggle → Fullscreen toggle → sidebar → header and wraps via the host's focus trap; Shift+Tab reverses it
- [ ] In fullscreen, Tab cycles tile → Gallery toggle → Fullscreen toggle
- [ ] Escape closes only the gallery from a tile, either toggle, or with focus/click in the sidebar — the preview stays open
- [ ] Escape inside an open menu/dropdown closes only that menu; the gallery stays open
- [ ] With the gallery open, Arrow keys and `[` `]` don't flip pages underneath, no scroll into the host file browser
- [ ] Prev/Next File nav (both `.bp-navigate-*` and the host's arrows) is not tab-reachable while open
- [ ] Underlying document does not show a scrollbar on arrow-key nav
- [ ] Behavior is correct whether the thumbnails sidebar was open or closed when the gallery was toggled
- [ ] Regular (non-gallery) fullscreen tab behavior is unchanged for all viewer types

## Out of scope
- Escape with focus in the header still closes the preview — fixed by the follow-up preview-client and EUA PRs consuming `closeGallery()`

### DIFF
--- a/src/lib/viewers/doc/DocBaseViewer.js
+++ b/src/lib/viewers/doc/DocBaseViewer.js
@@ -930,6 +930,23 @@ class DocBaseViewer extends BaseViewer {
      * @return {boolean} consumed or not
      */
     onKeydown(key, event) {
+        if (this.galleryController && this.galleryController.isOpen) {
+            if (event && event.defaultPrevented) {
+                return false;
+            }
+
+            if (key === 'Escape') {
+                this.galleryController.handleEscape();
+                return true;
+            }
+
+            // Swallow page-nav keys so they can't flip the doc page underneath the gallery
+            // or trigger the host's collection navigation.
+            if (['ArrowUp', 'ArrowDown', 'ArrowLeft', 'ArrowRight', '[', ']'].includes(key)) {
+                return true;
+            }
+        }
+
         switch (key) {
             case 'ArrowLeft':
                 this.previousPage();
@@ -1739,10 +1756,6 @@ class DocBaseViewer extends BaseViewer {
     handleDocElKeydown(event) {
         const key = decodeKeydown(event);
 
-        if (key === 'Escape' && this.galleryController.handleEscape()) {
-            return;
-        }
-
         if (event.altKey && key.includes('Arrow')) {
             event.stopPropagation(); // Prevent collection/page navigation for caret navigation users
         }
@@ -2009,6 +2022,7 @@ class DocBaseViewer extends BaseViewer {
      * Called before the gallery opens. Closes the find bar and resets annotation mode.
      *
      * @protected
+     * @emits galleryOpen
      * @return {void}
      */
     handleGalleryEnter() {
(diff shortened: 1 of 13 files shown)


src/lib/viewers/doc/DocBaseViewer.js
```diff
@@ -930,6 +930,23 @@ class DocBaseViewer extends BaseViewer {
      * @return {boolean} consumed or not
      */
     onKeydown(key, event) {
+        if (this.galleryController && this.galleryController.isOpen) {
+            if (event && event.defaultPrevented) {
+                return false;
```
zhirongwang, line 935: It needs to return a boolean ?

src/lib/viewers/gallery/GalleryController.tsx
```diff
@@ -148,6 +149,8 @@ export default class GalleryController {
                 this.galleryRoot = null;
             }
 
+            this.clearGalleryOpenState();
```
zhirongwang, line 152: Is this redundant ? It's already called in `destroy()`

src/lib/viewers/doc/DocBaseViewer.js
aivkov123: Yes, it returns a boolean (and false in this case) because `event.defaultPrevented` returning true means something else already handled the key, so we don't want to do anything additional (like we had some menu open and we pressed escape, we would only want to escape that menu instead of escaping gallery view as well)

src/lib/viewers/gallery/GalleryController.tsx
aivkov123: It's not redundant because this `this.clearGalleryOpenState();` is called when we toggle (clicking the gallery toggle, escape while in gallery view, clicking a tile in gallery view to go to a page), and the `destroy()` you're talking about with `this.clearGalleryOpenState();` is called when we actually leave the file while in gallery view (using the arrow in the top left).

zhirongwang (approved):

PR #1708, opened 2026-07-14
fix(gallery): gallery row spacing, scrollbar fixes, and loading updates

aivkov123:
Fix gallery view's grid layout, reflow, and loading behavior: tiles keep their shape and position while thumbnails load, visible tiles always load first, and keyboard navigation works when focus is parked on a toolbar toggle or after exiting fullscreen.
## What changed
**Layout & reflow (`fix(gallery): fix reflow gaps & nav`)**
- Size grid rows with `grid-auto-rows: min-content` so landscape thumbnails no longer leave large vertical gaps between rows.
- Size tile placeholders from real PDF page dimensions — per page via PDF.js page metadata (`getPageView`), falling back to the first-page ratio from `Thumbnail.init()` — so tiles are born at their final shape and the grid no longer shifts or creeps the scrollbar as images arrive (~500px of `scrollHeight` drift per load reduced to single-digit px).
- Preserve the viewed area when the grid reflows: an unthrottled scroll handler tracks the topmost visible tile as an anchor, and a `ResizeObserver` scrolls that tile back to the top after fullscreen enter/exit or window resize (previously the raw scroll offset was kept while tile positions shifted, drifting the view).
**Keyboard navigation**
- Redirect arrow keys pressed outside the grid back into it: `GalleryController.handleArrowKey()` refocuses the selected tile and replays the key so the first press navigates (like the thumbnails sidebar), called from `DocBaseViewer`'s gallery key policy (#1700) where those keys were already being swallowed. A `contains()` check ignores the replayed event to prevent re-entry, and brackets (`[` `]`) stay swallowed without redirecting.
- Restore focus to the fullscreen toggle on fullscreen exit (`DocBaseViewer.handleFullscreenExit`, mirroring what enter has always done) — the browser otherwise drops focus to the page body, leaving arrow keys dead until the user clicks back into the preview. Doc-family viewers only.
**Viewport-driven lazy loading**
- Replace the fixed 40-tile initial load with a viewport-driven queue: after every 4-concurrent batch, the queue is re-derived from actual grid geometry (unloaded tiles in the viewport first, then a 3-screen buffer, both radiating outward from the viewed area) and the loader idles once that set is empty. Fixes screens that were left partially gray (large monitors + landscape tiles exceeded the old fixed budget) and the slow serial fill around the entry point.
- Scroll and resize catch-ups run through the same path, so entry fill, scroll catch-up, and fullscreen reveal all load at the same rate with visible tiles always winning over off-screen prefetch. Pages already being rendered are tracked in-flight and never re-queued.
- Loading is strictly lazy now: only viewport ± buffer ever loads per trigger (previously the first scroll kicked off background-loading the entire document).
## Test plan
- [ ] Grid shows consistent 16px gaps between rows for landscape/mixed-orientation documents
- [ ] Placeholders match page shape immediately; no tile resizing or scrollbar creep while thumbnails load (mixed-size docs included)
- [ ] Open gallery anywhere in a long document — the visible screen fills first, radiating outward, then loading stops
- [ ] Fling-scroll into an unloaded area — visible tiles fill before off-screen ones
- [ ] Scroll deep into a long document, enter/exit fullscreen — the same tiles remain in view and newly revealed tiles load without scrolling
- [ ] Resize the window while scrolled — the viewed area is preserved
- [ ] With focus on the Gallery or Fullscreen toggle, the first arrow press moves tile selection; subsequent presses keep navigating
- [ ] Click the fullscreen toggle to exit fullscreen — arrow keys navigate tiles immediately, no click needed
- [ ] `[` and `]` still do nothing with the gallery open
- [ ] Keyboard behavior from #1700 is unchanged (Escape, Tab flow, no page flips underneath)

### DIFF
--- a/src/lib/viewers/doc/DocBaseViewer.js
+++ b/src/lib/viewers/doc/DocBaseViewer.js
@@ -941,8 +941,10 @@ class DocBaseViewer extends BaseViewer {
             }
 
             // Swallow page-nav keys so they can't flip the doc page underneath the gallery
-            // or trigger the host's collection navigation.
+            // or trigger the host's collection navigation. Arrows pressed outside the grid
+            // are redirected into it so the first press navigates the tiles.
             if (['ArrowUp', 'ArrowDown', 'ArrowLeft', 'ArrowRight', '[', ']'].includes(key)) {
+                this.galleryController.handleArrowKey(key, event && event.target);
                 return true;
             }
         }
--- a/src/lib/viewers/doc/DocBaseViewer.js
+++ b/src/lib/viewers/doc/DocBaseViewer.js
@@ -944,7 +944,7 @@ class DocBaseViewer extends BaseViewer {
             // or trigger the host's collection navigation. Arrows pressed outside the grid
             // are redirected into it so the first press navigates the tiles.
             if (['ArrowUp', 'ArrowDown', 'ArrowLeft', 'ArrowRight', '[', ']'].includes(key)) {
-                this.galleryController.handleArrowKey(key, event && event.target);
+                this.galleryController.handleArrowKey(key);
                 return true;
             }
         }
@@ -1780,12 +1780,9 @@ class DocBaseViewer extends BaseViewer {
             this.annotator.toggleAnnotationMode(AnnotationMode.REGION);
         }
 
-        // Restore focus to the toggle, mirroring BaseViewer.handleFullscreenEnter — the browser
-        // otherwise drops focus to the body on exit, and keyboard input (e.g. gallery arrow
-        // keys) no longer reaches the viewer.
-        if (this.fullscreenToggleEl && this.fullscreenToggleEl.focus) {
-            this.fullscreenToggleEl.focus();
-        }
+        // Restore focus to the toggle, the browser otherwise drops focus to the body on exit,
+        // and keyboard input (e.g. gallery arrow keys) no longer reach the viewer
+        this.fullscreenToggleEl?.focus?.();
     }
 
     /**
--- a/src/lib/viewers/doc/DocBaseViewer.js
+++ b/src/lib/viewers/doc/DocBaseViewer.js
@@ -1779,6 +1779,13 @@ class DocBaseViewer extends BaseViewer {
         if (this.annotator && this.areNewAnnotationsEnabled() && this.options.enableAnnotationsDiscoverability) {
             this.annotator.toggleAnnotationMode(AnnotationMode.REGION);
         }
+
+        // Restore focus to the toggle, mirroring BaseViewer.handleFullscreenEnter — the browser
+        // otherwise drops focus to the body on exit, and keyboard input (e.g. gallery arrow
+        // keys) no longer reaches the viewer.
+        if (this.fullscreenToggleEl && this.fullscreenToggleEl.focus) {
+            this.fullscreenToggleEl.focus();
+        }
     }
 
     /**
--- a/src/lib/viewers/gallery/GalleryController.tsx
+++ b/src/lib/viewers/gallery/GalleryController.tsx
@@ -195,14 +195,14 @@ export default class GalleryController {
     /**
      * Redirects an arrow key pressed outside the grid (e.g. focus parked on a toggle after
      * toggling fullscreen) back into it: refocuses the selected tile and replays the key so
-     * the first press navigates, like the thumbnail sidebar. The contains() check makes the
-     * replayed event (which lands inside the grid) a no-op, preventing re-entry.
+     * the first press navigates, like the thumbnail sidebar. Keys pressed inside the grid
+     * never arrive here — GalleryGrid stops propagation on every arrow it handles.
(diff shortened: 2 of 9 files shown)


src/lib/viewers/doc/DocBaseViewer.js
```diff
@@ -1777,6 +1779,13 @@ class DocBaseViewer extends BaseViewer {
         if (this.annotator && this.areNewAnnotationsEnabled() && this.options.enableAnnotationsDiscoverability) {
             this.annotator.toggleAnnotationMode(AnnotationMode.REGION);
         }
+
+        // Restore focus to the toggle, mirroring BaseViewer.handleFullscreenEnter — the browser
+        // otherwise drops focus to the body on exit, and keyboard input (e.g. gallery arrow
+        // keys) no longer reaches the viewer.
+        if (this.fullscreenToggleEl && this.fullscreenToggleEl.focus) {
+            this.fullscreenToggleEl.focus();
+        }
```
kduncanhsu: can we simplify to `this.fullscreenToggleEl?.focus?.();`

kduncanhsu:
can we update the PR title to use the `fix(...)` pattern

src/lib/viewers/doc/DocBaseViewer.js
aivkov123: Yep, copied this pattern from `handleFullscreenEnter()` in `BaseViewer.js` which is multiple years old at this point

src/lib/viewers/gallery/GalleryController.tsx
```diff
@@ -189,6 +192,25 @@ export default class GalleryController {
...
+     * Redirects an arrow key pressed outside the grid (e.g. focus parked on a toggle after
+     * toggling fullscreen) back into it: refocuses the selected tile and replays the key so
+     * the first press navigates, like the thumbnail sidebar. The contains() check makes the
+     * replayed event (which lands inside the grid) a no-op, preventing re-entry.
+     */
+    handleArrowKey(key: string, target: EventTarget | null): void {
+        if (!this.isGalleryOpen || !key.startsWith('Arrow')) return;
+
+        const grid = this.galleryEl;
+        if (!grid || grid.contains(target as Node)) return;
```
kduncanhsu: i know it's mentioned in the comment, but what does `grid.contains(target as Node))` do? what is a "replayed event"

src/lib/viewers/gallery/GalleryController.tsx
```diff
@@ -189,6 +192,25 @@ export default class GalleryController {
...
+
+        const tile = grid.querySelector<HTMLElement>('[role="option"][tabindex="0"]');
```
kduncanhsu, line 207: Which tile will this `querySelector` select?

src/lib/viewers/gallery/GalleryController.tsx
aivkov123: This is kind of an artifact left over from before I made a change to how arrow keys get handled since I wrote this before I changed that, but I got the arrow key change merged first. Before, there was no `stopPropagation()` in `GalleryGrid` so an arrow key that was pressed inside the grid would make it to `handleArrowKey`, but after the `stopPropagation()` addition, this can no longer happen so we don't actually need the `grid.contains(target as Node)` check anymore, good catch. I'll explain how it worked before though because I think it's kind of interesting. Basically when a user pressed an arrow key with focus outside the grid, we still wanted that arrow to cycle through the grid. `if (!grid || grid.contains(target as Node))` checked if the arrow key press happened with no grid (don't want to cycle if the grid hasn't initialized yet or something) or it happened inside the grid (focus in the grid). If focus was inside the grid when the arrow key press happened, `handleGridKeyDown` in `GalleryGrid` had already handled it as the event bubbled up, so this line returned to avoid handling the same press a second time (without it, there would basically be a two tile move even though the arrow key was only pressed once). If focus wasn't inside the grid though, this line doesn't return and we continue through this method, get the tile that is currently selected (the one with a blue outline), and then we "replay" the event on line 210 where we use the arrow key that was pressed to artificially press it with focus now actually on the tile that needs to have it (lines 207 through 209).

aivkov123: tldr, leftover from before I made another change, will remove

src/lib/viewers/gallery/GalleryController.tsx
aivkov123: This will select the tile that is currently selected (the one with the blue outline) since in the grid the selected tile has tabindex 0 and all other tiles have tabindex -1 so tile will always be that tile.

src/lib/viewers/gallery/GalleryController.tsx
aivkov123: and replayed event just means the artificial key press on line 210 since when we make our arrow key press focus isn't in the grid and our arrow key press won't actually move tiles

kduncanhsu (approved):

PR #1710, opened 2026-07-16
fix(doc): disable wheel zoom handler on mobile behind query param gate

kduncanhsu:
## Summary
  - After enabling pinch-to-zoom features (`pinchToZoom.enabled`), mobile users (i.e. using the web app on a mobile browser, not to be confused with mobile Box app) experienced page refreshes when attempting to zoom in/out
- The wheel event handler (intended for trackpad pinch-to-zoom on desktop) may conflict with mobile browser gesture handling, causing erratic behavior and page refreshes
- Adds a `disableMobileWheelZoom` query param gate to skip registering the wheel zoom handler on mobile for testing before a full rollout (not sure how else to test this works on mobile)
    
## Test plan
- [ ] Open a document preview on mobile browser without query param — behavior unchanged
- [ ] Open a document preview on mobile browser with `?disableMobileWheelZoom` — wheel zoom handler is not registered, no page refresh on pinch
- [ ] Open a document preview on desktop — trackpad pinch-to-zoom still works regardless of query param

### DIFF
--- a/src/lib/viewers/doc/DocBaseViewer.js
+++ b/src/lib/viewers/doc/DocBaseViewer.js
@@ -1554,7 +1554,8 @@ class DocBaseViewer extends BaseViewer {
             this.docEl.addEventListener('touchend', this.pinchToZoomEndHandler);
         }
 
-        if (this.featureEnabled('pinchToZoom.enabled')) {
+        const disableMobileWheelZoom = new URLSearchParams(window.location.search).has('disableMobileWheelZoom');
+        if (this.featureEnabled('pinchToZoom.enabled') && !(this.isMobile && disableMobileWheelZoom)) {
             this.docEl.addEventListener('wheel', this.trackpadPinchToZoomHandler, { passive: false });
         }
     }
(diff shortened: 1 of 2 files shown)


src/lib/viewers/doc/DocBaseViewer.js
aivkov123, line 1: Just curious if we could enable/disable based on how wide the screen is instead of with a query param gate?

kduncanhsu: -  the URL query param gate is more similar to a feature flag - the reason I'm not just using a regular feature flag though is because im not sure how I'd be able to toggle the feature flag on a mobile device (i.e. no Devtools/switchbox on mobile chrome browser afaik). however with this approach changing the URL on the mobile browser can toggle the change on/off, so we can see if it fixes the issue

- I'm also adding the `this.isMobile` check as this is an issue which is happening only on mobile/touch devices, doesn't depend on screen width

aivkov123 (approved):

PR #1716, opened 2026-07-21
fix(doc): disable native pinch-to-zoom on touch devices via passive false

kduncanhsu:
## Summary

- Users have reported a bug where pinching to zoom on documents in mobile browsers can trigger an unintended page refresh.
- I previously attempted to fix this in #1710, but that approach did not resolve the issue because mobile browsers do not emit `wheel` events for pinch gestures.
- This PR reverts the changes from #1710 and introduces three potential fixes, each gated behind a URL query parameter so they can be evaluated independently before a full rollout:
    - `disableNativePinchToZoom` — Sets `touch-action: pan-x pan-y` on the document element, disabling native pinch-to-zoom while preserving scrolling.
     - `disableAllTouchActions` — Sets `touch-action: none` on the document element, disabling all native touch gestures (including pinch-to-zoom and scrolling).
    - `forceNonPassiveTouchListeners` — Registers `touchstart` and `touchmove` listeners with `{ passive: false }`, allowing `preventDefault()` to suppress native gesture handling.

Note: These changes are gated behind URL query parameters because I currently can't reproduce or validate the issue in my local development environment. My personal mobile device has restrictions that prevent testing against local builds, and I have been unable to reproduce the issue using mobile browser emulators. Once this code reaches production, I'll be able to test each approach independently on my device and remove the experimental code after identifying the correct fix. I'm using URL query parameters instead of a traditional feature flag because I can't configure the required feature flag cookies on my personal mobile browser.

## Suspected root cause
- Modern mobile browsers ignore `user-scalable=no` in the viewport meta tag for accessibility reasons, so native pinch-to-zoom is no longer reliably disabled.
- As a result, native browser pinch-to-zoom can conflict with our custom pinch-to-zoom implementation, potentially causing the browser to refresh the page.

 ## Test plan
  - [ ] Open a document preview on mobile browser without query params — behavior unchanged
  - [ ] Open with `?disableNativePinchToZoom` — browser no longer handles pinch-zoom natively, custom zoom works without page refresh, scrolling still works
  - [ ] Open with `?disableAllTouchActions` — all native gestures disabled, JS handles everything
  - [ ] Open with `?forceNonPassiveTouchListeners` — `preventDefault()` blocks native gesture handling
  - [ ] Open with params combined to compare approaches
  - [ ] Open on desktop — behavior unchanged regardless of query params (no touch events)
  - [ ] Verify wheel-based trackpad pinch-to-zoom still works on desktop

### DIFF
--- a/src/lib/viewers/doc/DocBaseViewer.js
+++ b/src/lib/viewers/doc/DocBaseViewer.js
@@ -1551,13 +1551,16 @@ class DocBaseViewer extends BaseViewer {
         this.docEl.addEventListener('scroll', this.throttledScrollHandler);
 
         if (this.hasTouch) {
-            this.docEl.addEventListener('touchstart', this.pinchToZoomStartHandler);
-            this.docEl.addEventListener('touchmove', this.pinchToZoomChangeHandler);
+            const disableNativePinchToZoom = new URLSearchParams(window.location.search).has(
+                'disableNativePinchToZoom',
+            );
+            const passiveOption = disableNativePinchToZoom ? { passive: false } : undefined;
+            this.docEl.addEventListener('touchstart', this.pinchToZoomStartHandler, passiveOption);
+            this.docEl.addEventListener('touchmove', this.pinchToZoomChangeHandler, passiveOption);
             this.docEl.addEventListener('touchend', this.pinchToZoomEndHandler);
         }
 
-        const disableMobileWheelZoom = new URLSearchParams(window.location.search).has('disableMobileWheelZoom');
-        if (this.featureEnabled('pinchToZoom.enabled') && !(this.isMobile && disableMobileWheelZoom)) {
+        if (this.featureEnabled('pinchToZoom.enabled')) {
             this.docEl.addEventListener('wheel', this.trackpadPinchToZoomHandler, { passive: false });
         }
     }
@@ -1579,8 +1582,7 @@ class DocBaseViewer extends BaseViewer {
                 this.docEl.removeEventListener('touchend', this.pinchToZoomEndHandler);
             }
 
-            const disableMobileWheelZoom = new URLSearchParams(window.location.search).has('disableMobileWheelZoom');
-            if (this.featureEnabled('pinchToZoom.enabled') && !(this.isMobile && disableMobileWheelZoom)) {
+            if (this.featureEnabled('pinchToZoom.enabled')) {
                 this.docEl.removeEventListener('wheel', this.trackpadPinchToZoomHandler);
             }
         }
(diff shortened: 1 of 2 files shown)


src/lib/viewers/doc/DocBaseViewer.js
```diff
@@ -1551,13 +1551,16 @@ class DocBaseViewer extends BaseViewer {
         this.docEl.addEventListener('scroll', this.throttledScrollHandler);
 
         if (this.hasTouch) {
-            this.docEl.addEventListener('touchstart', this.pinchToZoomStartHandler);
-            this.docEl.addEventListener('touchmove', this.pinchToZoomChangeHandler);
+            const disableNativePinchToZoom = new URLSearchParams(window.location.search).has(
+                'disableNativePinchToZoom',
+            );
+            const passiveOption = disableNativePinchToZoom ? { passive: false } : undefined;
+            this.docEl.addEventListener('touchstart', this.pinchToZoomStartHandler, passiveOption);
```
aivkov123, line 1566: I think that by default docEl has passive : false (since it's a div and not something like window or document which by default would have passive : true) and so passing in passive : false doesn't change anything?

aivkov123: https://stackoverflow.com/questions/61114830/how-to-prevent-native-browser-default-pinch-to-zoom-behavior could be a fix

kduncanhsu: thanks for the comment, good catch!

kduncanhsu: It seems like you're correct, `passive` defaults to `false` for a regular div so it probably doesn't make a different here

kduncanhsu: I've updated the PR to include the alternative fix suggested in the link you sent (setting `touch-action` styles to `none` or `pan x pan y`). 

However, I have left the `{ passive: false }` gate there for now behind a separate query gate - while it likely it won't fix the issue, I wasn't able to find a good way to definitely verify this _doesn't_, so I still want to test it on prod anyways. Admittedly this is not best practice, but given the urgency of the CRI  I want to test potential fixes on prod asap. I think it is okay here given there will definitely be a cleanup PR and it definitely won't introduce a regression either as it behind the url param gate. Lmk if you feel strongly against this.

aivkov123: Cool, that sounds good to me.

aivkov123 (approved):

PR #1718, opened 2026-07-24
fix(preview): release PDF memory

aivkov123:
Fix document Preview memory retention that is amplified by Gallery View: closing, reloading, or navigating away now releases PDF.js page resources and thumbnail caches instead of retaining them after Preview exits.

## What changed
**PDF.js teardown**
- Call `pdfViewer.setDocument(null)` during `DocBaseViewer.destroy()` so PDF.js releases the previous document, page-view array, rendered canvases, text layers, and associated page resources.
- Call `pdfLinkService.setDocument(null)` so the link service no longer retains the previous PDF document.
- Destroy Gallery and sidebar resources before detaching PDF.js, ensuring owned thumbnail resources are released first.
- Keep teardown safe when called repeatedly.

**Thumbnail cleanup**
- Destroy the `Thumbnail` instance owned by `ThumbnailsSidebar` instead of only destroying its virtual scroller.
- Release the sidebar’s bounded thumbnail cache and clear references to its PDF viewer and preloader.
- Ignore thumbnail initialization, rendering, and cache updates that complete after the thumbnail has been destroyed.
- Ignore delayed sidebar rendering and refresh callbacks after teardown.

**Async document loading**
- Check whether the viewer was destroyed before counting spreadsheet PDF operations or attaching a loaded document.
- Prevent delayed PDF loading from calling `setDocument()` on a destroyed viewer.
- Ignore expected PDF loading failures after teardown instead of logging or retrying them.
- Preserve normal loading behavior for active document and presentation viewers.

**Impact and compatibility**
- The underlying retention affected document Preview independently of Gallery View; Gallery amplified it by rendering thumbnails across many pages.
- During investigation with a complex 198-page document, repeated Gallery usage peaked around 4.5 GB and retained approximately 2.6 GB after leaving Preview. Explicit teardown reduced the footprint to approximately 895 MB.
- Gallery remains controlled by its existing feature flag. Users without Gallery access receive the same Preview behavior with improved document cleanup.
- Sidebar cleanup is conditional, so documents without the thumbnails sidebar are unaffected.

## Test plan
- [ ] Open a large document, render Gallery thumbnails, close Preview, and verify memory drops after garbage collection
- [ ] Reload Preview in place and verify the previous PDF.js document and page views are released
- [ ] Navigate to the next file and verify the previous viewer releases its resources
- [ ] Close Preview without opening Gallery and verify normal document teardown
- [ ] Verify Gallery remains hidden when its feature flag is omitted or disabled
- [ ] Verify documents without the thumbnails sidebar continue to open and close normally
- [ ] Destroy Preview while PDF loading is pending and verify the document is never attached
- [ ] Destroy Preview while spreadsheet operation counting is pending and verify loading does not resume
- [ ] Destroy thumbnails while initialization or rendering is pending and verify no cache or DOM work resumes
- [ ] Rotate a document and verify sidebar and Gallery thumbnails continue to refresh normally
- [ ] Verify repeated destruction does not throw
- [ ] Verify document and presentation viewers both detach their PDF.js documents
- [x] 499 focused Jest tests pass
- [x] 12 Gallery Cypress tests pass
- [x] ESLint passes
- [x] `git diff --check` passes

### DIFF
--- a/src/lib/ThumbnailsSidebar.js
+++ b/src/lib/ThumbnailsSidebar.js
@@ -137,6 +137,11 @@ class ThumbnailsSidebar {
             this.virtualScroller = null;
         }
 
+        if (this.thumbnail) {
+            this.thumbnail.destroy();
+            this.thumbnail = null;
+        }
+
         this.pdfViewer = null;
         this.currentThumbnails = [];
         this.currentPage = null;
@@ -211,6 +216,10 @@ class ThumbnailsSidebar {
      * @return {void}
      */
     renderNextThumbnailImage() {
+        if (!this.thumbnail || !this.virtualScroller) {
+            return;
+        }
+
         const scrollerNotInited =
             this.virtualScroller &&
             this.virtualScroller instanceof VirtualScroller &&
@@ -293,8 +302,13 @@ class ThumbnailsSidebar {
      * @return {void}
      */
     requestThumbnailImage(itemIndex, thumbnailEl) {
+        if (!this.thumbnail) {
+            return;
+        }
+
         const requestId = requestAnimationFrame(() => {
             this.animationFrameRequestIds = this.animationFrameRequestIds.filter(id => id !== requestId);
+
             this.thumbnail
                 .createThumbnailImage(itemIndex)
                 .then(imageEl => {
@@ -407,7 +421,7 @@ class ThumbnailsSidebar {
 
     refresh() {
         if (!this.virtualScroller) {
-            return;
+            return Promise.resolve();
         }
 
         this.virtualScroller.destroy();
@@ -421,7 +435,11 @@ class ThumbnailsSidebar {
         const savedPreloader = this.thumbnail.preloader;
         this.thumbnail.preloader = null;
 
-        this.thumbnail.init().then(thumbnailHeight => {
+        return this.thumbnail.init().then(thumbnailHeight => {
+            if (!this.thumbnail || !this.virtualScroller) {
+                return;
+            }
+
             this.thumbnail.preloader = savedPreloader;
 
             if (thumbnailHeight) {
--- a/src/lib/ThumbnailsSidebar.js
+++ b/src/lib/ThumbnailsSidebar.js
@@ -147,6 +147,7 @@ class ThumbnailsSidebar {
         this.currentPage = null;
         this.preloader = null;
         this.anchorEl.removeEventListener('click', this.thumbnailClickHandler);
+        this.anchorEl.removeEventListener('keydown', this.onKeydown);
     }
 
     /**
--- a/src/lib/viewers/doc/DocBaseViewer.js
+++ b/src/lib/viewers/doc/DocBaseViewer.js
@@ -289,6 +289,11 @@ class DocBaseViewer extends BaseViewer {
             this.thumbnailsSidebar.destroy();
         }
 
+        if (this.advancedInsightsThumbs) {
+            this.advancedInsightsThumbs.destroy();
+            this.advancedInsightsThumbs = null;
+        }
+
         // Clean up PDF network requests
         if (this.pdfLoadingTask) {
             try {
@@ -308,6 +313,8 @@ class DocBaseViewer extends BaseViewer {
         if (this.pdfLinkService) {
             this.pdfLinkService.setDocument(null);
         }
+
+        this.doc = null;
 
         if (this.printPopup) {
             this.printPopup.destroy();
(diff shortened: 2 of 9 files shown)


src/lib/viewers/doc/DocBaseViewer.js
```diff
@@ -285,6 +285,10 @@ class DocBaseViewer extends BaseViewer {
             this.galleryController.destroy();
         }
 
+        if (this.thumbnailsSidebar) {
+            this.thumbnailsSidebar.destroy();
```
jmcbgaston, line 289: **Also tear down `advancedInsightsThumbs` here.** `destroy()` releases `galleryController` and `thumbnailsSidebar`, but not `advancedInsightsThumbs` — a lazily-created `Thumbnail` (ACI render path ~L2337) that holds a `pdfViewer` ref and its own bounded image cache. `pdfViewer.setDocument(null)` detaches the document, but the cached image elements still leak after teardown. Suggest adding alongside the sidebar block:

```js
if (this.advancedInsightsThumbs) {
    this.advancedInsightsThumbs.destroy();
    this.advancedInsightsThumbs = null;
}
```

src/lib/ThumbnailsSidebar.js
```diff
@@ -137,6 +137,11 @@ class ThumbnailsSidebar {
             this.virtualScroller = null;
         }
 
+        if (this.thumbnail) {
+            this.thumbnail.destroy();
+            this.thumbnail = null;
```
jmcbgaston, line 142: **The `keydown` listener is never removed.** The constructor registers both `click` and `keydown` on `anchorEl`, but `destroy()` only removes `click` (a few lines below). `onKeydown` closes over the sidebar, so `anchorEl → listener → sidebar` stays reachable after teardown. Add next to the existing `click` removal:

```js
this.anchorEl.removeEventListener('keydown', this.onKeydown);
```

(plus a matching unit test for listener removal).

src/lib/viewers/doc/DocBaseViewer.js
aivkov123: I agree that `advancedInsightsThumbs` should also be destroyed on exit, I decided not to because I couldn't test it and I wanted to limit this pr to the main offenders in case this would cause some unforseen issue.

src/lib/ThumbnailsSidebar.js
aivkov123: I was able to test this and it takes almost 0 memory, I can remove it though since it didn't cause issues through my testing. Just wanted to keep the pr focused.

src/lib/viewers/doc/DocBaseViewer.js
```diff
@@ -294,17 +298,19 @@ class DocBaseViewer extends BaseViewer {
             }
         }
 
-        // Clean up viewer
+        // Clean up and detach the PDF.js document so retained viewer instances
+        // do not keep page views and document resources alive.
         if (this.pdfViewer) {
             this.pdfViewer.cleanup();
+            this.pdfViewer.setDocument(null);
```
jmcbgaston, line 310: 🟠 **`this.doc` still retains the loaded PDF proxy after teardown.** `initViewer()` assigns `this.doc = doc`, but `destroy()` only detaches the PDF viewer and link service. Since retained viewer instances are the scenario this PR addresses, that direct reference remains reachable after Preview closes. Clear `this.doc` during teardown and assert it in the release tests.

src/lib/viewers/doc/DocBaseViewer.js
jmcbgaston: Keeping unrelated cleanup out of this PR makes sense, but this is another `Thumbnail` owner in the same viewer teardown path. Its cache can retain up to 500 image elements, and because `destroy()` is never called, pending thumbnail work can still populate it after Preview closes. I still consider this part of the PR’s thumbnail-cache teardown scope.

mickr (dismissed):

src/lib/viewers/doc/DocBaseViewer.js
aivkov123: Fair enough, I think Advanced Content Insights is for enterprises to track things about how users view documents so I wasn't sure if destroying the thumbnails would mess with something there. I'll add the destruction for it in next commit.

src/lib/viewers/doc/DocBaseViewer.js
aivkov123: Will clear this as well, didn't seem to really be a main culprit in testing though

jmcbgaston (approved):

PR #1726, opened 2026-07-30
Gallery resin tags

aivkov123:
Add Resin tagging and usage metrics to Gallery View so tile interactions are attributable and the open-to-navigate rate is measurable.

### Resin tagging
- Tag gallery tiles with `data-resin-target="galleryTile"`, alongside the `galleryView` target the toggle already carried.
- Tag the gallery root with `data-resin-component="gallery"`. The grid mounts as a sibling of `.bp-ControlsRoot`, so tile clicks had no component ancestor and Resin could not attribute them.

### Usage metrics
- Add `user_document_gallery_open`, `_navigate`, and `_dismiss`, which Gallery had none of previously. Open and dismiss carry the document's page count; navigate carries the page landed on.
- Report outcomes from `GalleryController` through a single `onClose(landedPage)` callback, where a page number means the document ended up there and `null` means it never moved. One nullable argument makes "exactly one outcome per close" a property of the type rather than a convention.
- Count an outright pick and a close that commits the browsed page both as navigations, since closing commits the focused page either way. Picking the page already being viewed counts too.
- Report a dismissal when Preview is torn down with Gallery still open, so the open resolves instead of looking abandoned. A closed tab or crash cannot be observed.
- Leave these events out of `METRICS_WHITELIST`, unlike the thumbnail events, so every open is counted against its outcome rather than flattened to one of each.

## Test plan
- [ ] Open Gallery and click a tile, and verify `_open` then `_navigate` with the clicked page
- [ ] Open Gallery and press Escape, and verify `_open` then `_dismiss` with the page count
- [ ] Arrow to another page then press Escape, and verify `_navigate` with the browsed page rather than a dismissal
- [ ] Pick the page already being viewed and verify `_navigate`, not `_dismiss`
- [ ] Open and close Gallery twice and verify two `_open` events, confirming they are not de-duplicated
- [ ] Switch files with Gallery open and verify a dismissal is reported
- [ ] Click a tile and verify Resin no longer reports a missing component
- [ ] Verify tile clicks report component `gallery` and toggle clicks report component `toolbar`
- [x] 483 focused Jest tests pass
- [x] ESLint passes
- [x] `tsc --noEmit` passes

### DIFF
--- a/src/lib/events.js
+++ b/src/lib/events.js
@@ -112,6 +112,13 @@ export const USER_DOCUMENT_THUMBNAIL_EVENTS = {
     OPEN: 'user_document_thumbnails_open',
 };
 
+// Events fired when using gallery view
+export const USER_DOCUMENT_GALLERY_EVENTS = {
+    DISMISS: 'user_document_gallery_dismiss', // The gallery closes without the user settling on a page
+    NAVIGATE: 'user_document_gallery_navigate', // The user picks a page, or closes on one they browsed to
+    OPEN: 'user_document_gallery_open', // The user opens the gallery
+};
+
 export const MISSING_EXTERNAL_REFS = 'missing_x_refs';
 
 export const MEDIA_METRIC = {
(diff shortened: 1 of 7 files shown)


JChan106 (approved):
lgtm

zhirongwang (approved):

PR #1729, opened 2026-08-05
feat(gallery): add pinch to zoom

aivkov123:
## Summary

- Add zoom controls to the document gallery.
- Support toolbar, keyboard, trackpad, and touch pinch zooming.
- Preserve the viewed tile while the gallery grid reflows.
- Keep gallery zoom independent from document zoom.
- Record gallery zoom Resin events.
- Keep thumbnails at their existing resolution; adaptive quality will follow in a separate stacked PR.

## Test plan

- [x] 542 affected unit tests pass.
- [x] TypeScript typecheck passes.
- [x] ESLint passes.
- [ ] Verify toolbar and keyboard zoom controls.
- [ ] Verify trackpad and touch pinch zoom.
- [ ] Verify scroll position remains anchored during zoom.
- [ ] Verify gallery scale persists after reopening and resets after teardown.
- [ ] Verify zoom controls remain hidden when Gallery V2 is disabled.

### DIFF
--- a/src/lib/viewers/gallery/GalleryController.tsx
+++ b/src/lib/viewers/gallery/GalleryController.tsx
@@ -384,8 +384,8 @@ export default class GalleryController {
             <GalleryGrid
                 currentPage={pdfViewer.currentPageNumber}
                 getPageRatio={this.getPageRatio}
-                hasTouch={this.isZoomEnabled && this.hasTouch}
                 isPinchZoomEnabled={this.isZoomEnabled && isFeatureEnabled(this.features, 'pinchToZoom.enabled')}
+                isTouchZoomEnabled={this.isZoomEnabled && this.hasTouch}
                 onClose={this.toggle}
                 onFocusChange={this.handleFocusChange}
                 onPageNavigate={this.handleGalleryNavigate}
--- a/src/lib/viewers/gallery/GalleryController.tsx
+++ b/src/lib/viewers/gallery/GalleryController.tsx
@@ -226,9 +226,13 @@ export default class GalleryController {
         return true;
     }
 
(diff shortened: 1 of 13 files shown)


JChan106 (commented):
Pinching past min/max still stashes a focal even though scale doesn’t change, so the next real zoom can scroll-anchor off stale coords. Should we clear the focal on the no-op path?

src/lib/viewers/gallery/GalleryController.tsx
```diff
@@ -317,21 +361,34 @@ export default class GalleryController {
...
+        if (!this.galleryRoot || !this.galleryThumbnail) {
+            return;
+        }
+
+        const pdfViewer = this.getPdfViewer();
         this.galleryRoot.render(
             <GalleryGrid
                 currentPage={pdfViewer.currentPageNumber}
                 getPageRatio={this.getPageRatio}
+                hasTouch={this.isZoomEnabled && this.hasTouch}
```
kduncanhsu: nit: consider moving the `this.isZoomEnabled` check elsewhere or renaming the prop. having a prop named `hasTouch` that is only true when zoom is enabled might be a bit confusing - I initially interpreted it as indicating whether the device supports touch input, rather than whether touch interactions for zoom are enabled.

aivkov123: Will rename `hasTouch` instances to `isTouchZoomEnabled` for more clarity

aivkov123:
> Pinching past min/max still stashes a focal even though scale doesn’t change, so the next real zoom can scroll-anchor off stale coords. Should we clear the focal on the no-op path?

Yep, good catch. Fixed by having position get discarded if zoom is already at the min or max limit and the user tries to zoom more.

JChan106 (approved):

PR #1730, opened 2026-08-05
feat(gallery): sharpen thumbs

aivkov123:
## Summary

Depends on [#1729](https://github.com/box/box-content-preview/pull/1729).

- Render sharper gallery thumbnails at 880px and 1320px as tiles grow.
- Keep persistent 440px thumbnails as a fallback.
- Limit high-resolution thumbnails to visible and nearby pages.
- Enforce a 64MiB decoded-pixel budget, 16-page limit, and two concurrent renders.
- Cancel stale PDF renders and release canvas backing stores.
- Evict offscreen high-resolution images back to base quality.
- Prevent scrollbar interaction from jumping to the highlighted tile.

## Test plan

- [x] Affected unit tests pass.
- [x] TypeScript typecheck passes.
- [x] ESLint passes.
- [ ] Verify visible thumbnails sharpen when zooming.
- [ ] Verify nearby thumbnails sharpen before entering the viewport.
- [ ] Verify offscreen pages return to base quality.
- [ ] Aggressively zoom and scroll through a large document and confirm memory plateaus.
- [ ] Verify closing and reopening the gallery releases high-resolution resources.
- [ ] Verify dragging the gallery scrollbar does not jump to the highlighted tile.

PR #1733, opened 2026-08-07
feat(gallery): add ARIA grid for 2D navigation

aivkov123:
## Summary

Upgrades Gallery View from 1D listbox semantics to a WAI-ARIA grid with 2D keyboard navigation, gated behind `galleryViewV2` (requires `galleryView`).

## ARIA grid (v2 flag only)

- `role="grid"` + `aria-rowcount`/`aria-colcount`; tiles become `gridcell`s with `aria-colindex`, chunked into `role="row"` wrappers with `aria-rowindex`
- Row wrappers use `display: contents` — tiles stay direct CSS-grid items, so layout and SCSS are unchanged
- Column count is measured from the rendered layout (not duplicated CSS math); re-measured on resize/fullscreen/zoom, guarded against zero-layout containers
- Keyboard: Left/Right stay ±1 wrapping across rows; Up/Down move by one row, with Down clamping on a short final row; Home/End/Enter/Space/Escape unchanged
- Focus survives re-chunks (column changes recreate tile nodes), including first open; restores with `preventScroll`
- Flag off → V1 listbox markup and behavior unchanged (pinned by tests)

## Scope note: memory fix removed

An earlier revision of this PR also included an unflagged `Thumbnail.js` change intended to release PDF.js per-page decoded resources after thumbnail renders. It was removed (`revert(doc): drop page memory release`) because the `isPageCached` guard it relied on does not exist in the bundled PDF.js (Box `2.107.0`, built from `pdfjs-dist` 4.3.136), so the cleanup never executed — and A/B verification on devpod with a 180-page image-heavy stress document showed no memory difference vs master (the bundled PDF.js already releases decoded bitmaps for thumbnail-only pages in Chrome). 

## Also

- Gallery constants extracted to `gallery/constants.ts` (follow-up from #1730 review)
- 2D nav math isolated in `galleryGridNavigation.ts`; FocusTrap recognizes gridcells

## Testing

- Unit tests, tsc, eslint, stylelint all green
- New unit coverage: grid semantics, keyboard boundaries (full + partial final rows), resize/zoom re-measure, focus retention across node recreation, no focus-stealing, zero-layout guard
- Cypress: v2 grid + 2D nav at wide/narrow viewports, v1 flag-off regression; existing specs made role-agnostic
- Manual on devpod EUA with both flag cookies: entry focus, arrows from first keypress, zoom + scroll

### DIFF
--- a/src/i18n/en-US.properties
+++ b/src/i18n/en-US.properties
@@ -52,7 +52,7 @@ toggle_findbar=Toggle findbar
 toggle_thumbnails=Toggle thumbnails
 # Button tooltip for Gallery view
 gallery_view=Gallery view
-# Accessible name for the gallery page-thumbnail grid (listbox pre-V2)
+# Accessible name for the gallery of page thumbnails
 page_gallery=Page gallery
 # Accessible name for a single gallery tile; {1} is the page number
 page_gallery_tile=Page {1}
--- a/src/i18n/en-US.properties
+++ b/src/i18n/en-US.properties
@@ -52,7 +52,7 @@ toggle_findbar=Toggle findbar
 toggle_thumbnails=Toggle thumbnails
 # Button tooltip for Gallery view
 gallery_view=Gallery view
-# Accessible name for the gallery page-thumbnail listbox
+# Accessible name for the gallery page-thumbnail grid (listbox pre-V2)
 page_gallery=Page gallery
 # Accessible name for a single gallery tile; {1} is the page number
 page_gallery_tile=Page {1}
--- a/src/lib/viewers/gallery/GalleryController.tsx
+++ b/src/lib/viewers/gallery/GalleryController.tsx
@@ -130,14 +130,10 @@ export default class GalleryController {
         return this.galleryScale;
     }
 
-    /** V2 requires the V1 flag: parent apps target the splits independently. */
-    get isV2Enabled(): boolean {
+    /** Both gallery flags must be on: parent apps target the splits independently. */
+    get isEnhancedGalleryEnabled(): boolean {
         const { features } = this;
         return isFeatureEnabled(features, 'galleryView.enabled') && isFeatureEnabled(features, 'galleryViewV2.enabled');
-    }
-
-    get isZoomEnabled(): boolean {
-        return this.isV2Enabled;
     }
 
     canRender(pageCount: number): boolean {
@@ -328,7 +324,7 @@ export default class GalleryController {
     };
 
     private commitScale = (scale: number): boolean => {
-        if (!this.isZoomEnabled || !Number.isFinite(scale)) {
+        if (!this.isEnhancedGalleryEnabled || !Number.isFinite(scale)) {
             return false;
         }
 
@@ -390,9 +386,11 @@ export default class GalleryController {
             <GalleryGrid
                 currentPage={pdfViewer.currentPageNumber}
                 getPageRatio={this.getPageRatio}
-                isAriaGridEnabled={this.isV2Enabled}
-                isPinchZoomEnabled={this.isZoomEnabled && isFeatureEnabled(this.features, 'pinchToZoom.enabled')}
-                isTouchZoomEnabled={this.isZoomEnabled && this.hasTouch}
+                isAriaGridEnabled={this.isEnhancedGalleryEnabled}
+                isPinchZoomEnabled={
+                    this.isEnhancedGalleryEnabled && isFeatureEnabled(this.features, 'pinchToZoom.enabled')
+                }
+                isTouchZoomEnabled={this.isEnhancedGalleryEnabled && this.hasTouch}
                 onClose={this.toggle}
                 onFocusChange={this.handleFocusChange}
                 onPageNavigate={this.handleGalleryNavigate}
--- a/src/lib/viewers/gallery/GalleryController.tsx
+++ b/src/lib/viewers/gallery/GalleryController.tsx
@@ -6,19 +6,19 @@ import React from 'react';
 import { createRoot, Root } from 'react-dom/client';
 import Thumbnail from '../../Thumbnail';
 import { FeatureConfig, isFeatureEnabled } from '../../featureChecking';
+import {
+    GALLERY_MAX_PAGES,
+    GALLERY_MAX_SCALE,
+    GALLERY_MIN_SCALE,
+    GALLERY_SCALE_STEP,
+    THUMBNAILS_SIDEBAR_TRANSITION_TIME,
+} from './constants';
 import GalleryGrid, { GalleryThumbnail } from './GalleryGrid';
 import { PinchDirection } from './useGalleryPinch';
 
 // Controller-owned shape: GalleryGrid only sees the read surface (GalleryThumbnail),
 // but the controller also needs destroy() to invalidate the cache on rotate/teardown.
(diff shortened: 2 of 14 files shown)


src/lib/Thumbnail.js
```diff
@@ -281,13 +283,40 @@ class Thumbnail {
...
+     * @param {Object|null} pdfPage - The PDF.js page proxy, or null if the render never started
+     * @return {void}
+     */
+    releasePageResources(pdfPage) {
+        if (!pdfPage) {
+            return;
+        }
+
+        const pageView = this.pdfViewer?.getPageView?.(pdfPage.pageNumber - 1);
+        const bufferedViews = pageView && this.pdfViewer?.getCachedPageViews?.();
```
jmcbgaston: **Bugbot (high):** `getCachedPageViews()` isn’t part of the PDF.js `PDFViewer` API (only `isPageCached` exists). Optional chaining makes this guard a no-op, so gallery/sidebar thumbnail renders can `cleanup()` shared page proxies still held in the document viewer’s buffer — visible pages may flicker or blank while the gallery is open.

Prefer `this.pdfViewer?.isPageCached?.(pdfPage.pageNumber)` (or equivalent) before calling `pdfPage.cleanup()`.

src/lib/Thumbnail.js
```diff
@@ -281,13 +283,40 @@ class Thumbnail {
             .finally(() => {
                 canvas.width = 0;
                 canvas.height = 0;
+                this.releasePageResources(pdfPage);
```
jmcbgaston, line 286: **Bugbot (medium):** In this same `renderPageImage` path, if high-res sharpening runs before `Thumbnail.init()` finishes, `this.pageRatio` is still `undefined` (see the portrait branch around the canvas sizing above). The portrait check is skipped, so taller pages can render at the wrong aspect ratio, and `HighResThumbnailStore` may cache that bad data URL until a zoom change evicts it.

Guard this path until `pageRatio` is initialized (or derive ratio from the current page only before init completes).

src/lib/Thumbnail.js
aivkov123: Addressed in new commit

src/lib/Thumbnail.js
aivkov123: Addressed in new commit

zhirongwang:
Heads-up on the `Thumbnail.releasePageResources` guard: `PDFViewer.isPageCached` does not exist in the PDF.js this repo ships (`third-party/doc/2.107.0`, built from `pdfjs-dist` 4.3.136) — it was removed upstream and only survives in the stale vendored `2.84.0`–`2.100.0` bundles. So `typeof isPageCached !== 'function'` is always true at runtime, `releasePageResources` always returns early, and `pdfPage.cleanup()` never runs — the memory fix is a no-op in production. The unit tests pass because they mock `isPageCached` into existence.

The original `getCachedPageViews()` approach from ac6f865 was actually correct for the shipped bundle (the earlier automated review comment had it backwards — `getCachedPageViews()` exists in 4.3.136 and `isPageCached` doesn't).

The supported replacement is dispatching `thumbnailrendered` on the viewer's event bus — `PDFViewer` registers a handler that does the buffer check against its private page cache and calls `pdfPage.cleanup()` itself:

```js
this.pdfViewer?.eventBus?.dispatch('thumbnailrendered', {
    source: this,
    pageNumber: pdfPage.pageNumber,
    pdfPage,
});
```

That's what #1745 (the memory fix split out of this PR, per PREVIEW-1548) now does — happy to sync on rebasing this PR on top of it so the `Thumbnail.js` changes drop out here.

zhirongwang:
Pushed `9bcc587c` (with maintainer-edit access) removing the memory-release change and its tests from this PR, and updated the description accordingly — following up on my earlier comment about the `isPageCached` dead guard. Beyond the dead guard, A/B verification on devpod (180-page stress doc, unique ~7.3 MB-decoded image per page) showed no memory difference vs master even with a working release path: the bundled PDF.js already frees decoded bitmaps for thumbnail-only pages in Chrome (`queryObjects(ImageBitmap)` showed 4 live bitmaps on unfixed master after a full gallery scroll). The page-release work continues in #1745 (PREVIEW-1548) via the supported `thumbnailrendered` event, reframed as convention alignment; full investigation notes are on the ticket. This keeps this PR scoped to the PREVIEW-1246 ARIA grid work. @aivkov123 please yell if you'd rather I hadn't touched the branch — happy to walk anything back.

src/lib/Thumbnail.js
zhirongwang: removed this memory fix to another ticket. [PREVIEW-1548](https://jira.inside-box.net/browse/PREVIEW-1548)

src/i18n/en-US.properties
```diff
@@ -52,7 +52,7 @@ toggle_findbar=Toggle findbar
 toggle_thumbnails=Toggle thumbnails
 # Button tooltip for Gallery view
 gallery_view=Gallery view
-# Accessible name for the gallery page-thumbnail listbox
+# Accessible name for the gallery page-thumbnail grid (listbox pre-V2)
```
jackiejou: can we avoid mentioning v2? or versions?

src/lib/viewers/gallery/__tests__/GalleryController-test.tsx
```diff
@@ -572,6 +574,18 @@ describe('GalleryController', () => {
         });
     });
 
+    describe('isV2Enabled', () => {
+        test.each`
+            v1       | v2       | expected
+            ${true}  | ${true}  | ${true}
+            ${true}  | ${false} | ${false}
+            ${false} | ${true}  | ${false}
+        `('should return $expected when v1=$v1 and v2=$v2', ({ v1, v2, expected }) => {
```
jackiejou: can we just update v1 to flagOn and v2 to ZoomFlagOn?

src/lib/viewers/gallery/__tests__/GalleryGrid-test.tsx
```diff
@@ -778,4 +819,282 @@ describe('GalleryGrid', () => {
...
+            // arrow keys go dead until the user clicks.
+            test('should keep focus on the current page tile after the mount re-measure re-chunks the grid', () => {
+                setupGrid(3);
+                // Guard against the mount re-chunk never happening, which would make this vacuous
+                expect(screen.getByRole('grid')).toHaveAttribute('aria-colcount', '3');
+                expect(screen.getByLabelText('Page 3')).toHaveFocus();
+            });
+        });
+
+        describe('semantics', () => {
```
jackiejou: can we update this describe to be clearer?

src/lib/viewers/gallery/GalleryController.tsx
```diff
@@ -130,11 +130,16 @@ export default class GalleryController {
         return this.galleryScale;
     }
 
-    get isZoomEnabled(): boolean {
+    /** V2 requires the V1 flag: parent apps target the splits independently. */
+    get isV2Enabled(): boolean {
```
jackiejou: do we need this relabel? can we just set isAriaGridEnabled={this.isZoomEnabled} below?

zhirongwang: relabeled to isEnhancedGalleryEnabled and use it for grid and zoom prop

jackiejou (approved):

PR #1734, opened 2026-08-11
fix(npm): expose the PDF.js worker asset

### LINKED ISSUE #1722: expose the export to the pdf.worker.min.mjs file
hi! i'm switching from CDN-hosted box-content-preview to the npm package and having troubles with passing the workerSrc option

according to https://github.com/box/box-content-preview/pull/1704

> The pdfjs-dist version installed by the consumer must match the version pinned by this package (currently 4.3.136).

i want to avoid having a separate pdfjs-dist package and i found the pdf.worker.min.mjs inside box-content-preview's `dist/lib`. Unfortunately, it can't be accessed because the package.json only allows `.`, `./styles.css`, and `./package.json` exports.

do you think it would be an improvement to add the `./pdf.worker.min.mjs` export to the box-content-preview package.json's `exports` field so that it can be accessed as

```js
import pdfjsWorkerUrl from 'box-content-preview/pdf.worker.min.mjs?url';
```

danieloleary:
A quick hello from the Box alumni crew 👋, and thanks for keeping this project moving.

I took a pass at #1722. The npm build already emits the pinned PDF.js worker, but only under a generated filename that consumers cannot access through the package exports.

This gives the existing worker a stable `pdf.worker.min.mjs` filename and exports it through the package.

For Vite consumers:

```js
import workerUrl from 'box-content-preview/pdf.worker.min.mjs?url';
```

The build verifier now checks that the export resolves, the emitted worker exactly matches the pinned `pdfjs-dist` worker, and no duplicate worker is shipped.

Tested with:

- the library build and verifier
- 171 Jest suites and 4,205 tests
- ESLint and `git diff --check`
- the real npm tarball installed in a clean Vite app using the import above

Fixes #1722.

Meow. 🐈

### DIFF
--- a/build/verifyLibBundle.js
+++ b/build/verifyLibBundle.js
@@ -8,6 +8,9 @@ const path = require('path');
 // files. output.chunkLoading: false and optimization.splitChunks: false in
 // webpack.config.lib.js prevent it; this check keeps that from silently regressing.
 const bundlePath = path.resolve(__dirname, '../dist/lib/index.js');
+const packagePath = path.resolve(__dirname, '../package.json');
+const pdfjsWorkerPath = path.resolve(__dirname, '../node_modules/pdfjs-dist/build/pdf.worker.min.mjs');
+const emittedWorkerPath = path.resolve(__dirname, '../dist/lib/pdf.worker.min.mjs');
 
 if (!fs.existsSync(bundlePath)) {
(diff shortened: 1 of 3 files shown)


jackiejou (approved):

jackiejou:
@danieloleary , thanks for resolving this! Please help click on the CLA link to sign the agreement and then we can merge it

PR #1739, opened 2026-08-14
fix(video transcription): re-fetch transcription rep on cache hits

kronosmichall:
## Summary
- When `skipServerUpdate` is set and a cached video file lacks an `extracted_text` rep, refresh file metadata from the server (same pattern as missing playable video reps).
- Defer loading the Dash viewer until that server refresh completes when playable video reps are already cached but `extracted_text` is not — avoids a race where `loadeddata` fires before transcription metadata arrives.
- After the server response includes `extracted_text`, update the active Dash viewer and call `loadTranscription()` instead of treating the cache hit as complete.

Fixes: on dev, subtitles only appeared on the first video preview because reopening reused cached metadata without `extracted_text`.

## Test plan
- [x] Unit tests for `loadFromCache` server refresh when transcription rep is missing
- [x] Unit tests for deferred viewer load when playable reps exist but transcription is missing
- [x] Unit tests for `handleFileInfoResponse` calling `loadTranscription` when cache lacked `extracted_text`
- [ ] Preview a video with AI transcription enabled, close preview, reopen same video — subtitles should still appear
- [ ] Preview a video, navigate away, return — subtitles should still appear

### DIFF
--- a/src/lib/Preview.js
+++ b/src/lib/Preview.js
@@ -35,6 +35,7 @@ import {
     uncacheFile,
     isWatermarked,
     getCachedFile,
+    getRepresentation,
     normalizeFileVersion,
     canDownload,
     shouldDownloadWM,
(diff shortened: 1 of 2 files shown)


kronosmichall:
Superseded by #1743 — branch renamed to `fix/video-transcription-cache-hits` (no internal ticket refs) and rebased onto latest master.

PR #1740, opened 2026-08-14
chore(deps): Replace mojito-rb-gen and bump Babel preset-env

jackiejou:
## Summary

Replaces `mojito-rb-gen` with a local properties-to-JSON converter, bumps `@babel/preset-env` to the same `^7.24.7` range box-ui-elements uses, and removes unused `eslint-import-resolver-webpack`.

`yarn build:i18n` still reads `src/i18n/*.properties` and writes `src/i18n/json/{locale}.json`. Webpack, `__()`, CDN, and the npm bundle are unchanged.

## Why

The i18n build depended on an unmaintained 0.0.1 CLI. Babel `preset-env` was still on the 7.18 range. The webpack import resolver is unused because `import/no-unresolved` is already off.

## Test plan

- [ ] `yarn build:i18n` writes JSON for all 26 locales
- [ ] Locale JSON keys and values match the previous converter
- [ ] `yarn lint` and unit tests pass
- [ ] CDN and npm webpack configs are unchanged

### DIFF
--- a/build/buildI18n.js
+++ b/build/buildI18n.js
@@ -1,0 +1,42 @@
+const fs = require('fs');
+const path = require('path');
+
+const srcDir = path.resolve(__dirname, '../src/i18n');
+const outDir = path.join(srcDir, 'json');
+const baselinePath = path.join(srcDir, 'en-US.properties');
+
+if (!fs.existsSync(baselinePath)) {
+    throw new Error(`buildI18n: missing baseline ${baselinePath}`);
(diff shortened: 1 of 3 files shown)


ehoogerbeets (approved):
Since this is temporary code, it is fine like it is. Later, when you move to react-intl, we should follow the standards of migrated localized projects.

jmcbgaston (approved):

PR #1741, opened 2026-08-14
fix(video): prevent V2 filmstrip thumbnail drift on non-standard aspect ratios

kduncanhsu:
## Summary
- `FilmstripV2.tsx` drifted out of sync because it stretched the whole sheet 1.5× (`background-size`) and panned in those scaled coordinates, so frame width and position could disagree.
- Crop one native cell first (90px tall, pan with `-frameIndex * sourceFrameWidth` / `-row * 90`, `overflow: hidden`), then `transform: scale(1.5)` that crop to 135px. Scale never participates in which cell is shown, so later timestamps cannot leak the next row (the black sliver under the thumbnail).
- Adds unit tests for the ticket’s 12800px sheet, odd source widths at `time=53`, and row 2 at `time=110`.

BEFORE

Screenshot 2026-08-18 at 2 56 24 PM

AFTER

Screenshot 2026-08-18 at 2 55 42 PM

## Test plan
- [ ] Hover the V2 player scrubber on a 16:9 video and confirm filmstrip thumbnails still look correct
- [ ] Reproduce with the PREVIEW-1354 test clip (or another non-standard ratio such as 3060×2160): scrub past ~0:50 and confirm each thumbnail is a single frame, not split
- [ ] Scrub toward the end of a longer video and confirm there is no black sliver under the thumbnail
- [ ] Confirm the filmstrip still appears while the JPEG is loading (crawler, then thumbnail) without a split frame
- [ ] `NODE_ENV=test yarn jest src/lib/viewers/controls/media/__tests__/FilmstripV2-test.tsx --watchman=false`

### DIFF
--- a/src/lib/viewers/controls/media/__tests__/FilmstripV2-test.tsx
+++ b/src/lib/viewers/controls/media/__tests__/FilmstripV2-test.tsx
@@ -110,7 +110,7 @@ describe('FilmstripV2', () => {
     });
 
     describe('frame width', () => {
-        const mockFilmstripLoad = (naturalWidth: number): (() => void) => {
+        const mockFilmstripLoad = (naturalWidth: number, naturalHeight = 90): (() => void) => {
             let capturedImg: HTMLImageElement | null = null;
             const originalCreateElement = document.createElement.bind(document);
             jest.spyOn(document, 'createElement').mockImplementation((tag: string) => {
@@ -123,6 +123,7 @@ describe('FilmstripV2', () => {
                 act(() => {
                     if (capturedImg?.onload) {
                         Object.defineProperty(capturedImg, 'naturalWidth', { value: naturalWidth });
+                        Object.defineProperty(capturedImg, 'naturalHeight', { value: naturalHeight });
                         (capturedImg.onload as (this: GlobalEventHandlers, ev: Event) => void).call(
                             capturedImg,
                             new Event('load'),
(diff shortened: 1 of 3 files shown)


mrscobbler (approved):

src/lib/viewers/controls/media/__tests__/FilmstripV2-test.tsx
```diff
@@ -50,8 +49,8 @@ describe('FilmstripV2', () => {
 
         test('should calculate background position based on time and interval', () => {
             render(<FilmstripV2 imageUrl="https://example.com/filmstrip.jpg" interval={1} time={10} />);
-            const frame = screen.getByTestId('bp-FilmstripV2-frame');
-            expect(frame.style.backgroundPositionX).toBeDefined();
+            const frameImage = screen.getByTestId('bp-FilmstripV2-frameImage');
+            expect(frameImage.style.backgroundPositionX).toBeDefined();
```
mrscobbler, line 53: Non-blocking: This assertion cannot fail.

**Concern**: `style.backgroundPositionX` is always a string, so `toBeDefined()` will not catch incorrect positioning.

**Recommendation**:
```tsx
expect(frameImage.style.backgroundPositionX).toBe("-1600px");
```
Alternatively, remove this test because the new exact-position tests provide stronger coverage.

PR #1743, opened 2026-08-18
fix(video transcription): re-fetch transcription rep on cache hits

kronosmichall:
## Summary
- When a cached video is missing an `extracted_text` transcription rep, refresh file metadata from the server (`needsTranscriptionRep`), including when `skipServerUpdate` is set.
- Load the Dash viewer immediately from cache. Captions are optional; do not defer playback for a missing transcription rep.

Fixes: on cache hits (e.g. reopen on dev), subtitles were skipped because cached file info lacked `extracted_text`.

## Test plan
- [x] Unit tests for `loadFromCache` server refresh when transcription rep is missing
- [x] Unit tests for `hasTranscriptionRep`
- [ ] Preview a video with AI transcription enabled, close preview, reopen same video — subtitles should still appear
- [ ] Preview a video, navigate away, return — subtitles should still appear

### DIFF
--- a/src/lib/Preview.js
+++ b/src/lib/Preview.js
@@ -35,6 +35,7 @@ import {
     uncacheFile,
     isWatermarked,
     getCachedFile,
+    getRepresentation,
     normalizeFileVersion,
     canDownload,
     shouldDownloadWM,
@@ -1207,11 +1208,23 @@ class Preview extends EventEmitter {
         // Log cache hit
         this.logger.setCached();
 
-        // Finally load the viewer
-        this.loadViewer();
-
         const needsVideoReps = this.isVideoFileByExtension() && !this.hasPlayableVideoReps(this.file);
-        if (!this.options.skipServerUpdate || needsVideoReps) {
+        const needsTranscriptionRep =
+            isFeatureEnabled(this.options.features, AI_TRANSCRIPTION_FOR_VIDEO_SUBTITLES) &&
+            this.isVideoFileByExtension() &&
+            !this.hasTranscriptionRep(this.file);
+        const needsServerRefresh = !this.options.skipServerUpdate || needsVideoReps || needsTranscriptionRep;
+
+        // When playable video reps are already cached but extracted_text is not, defer loading the
+        // viewer until handleFileInfoResponse can supply transcription metadata. Loading dash first
+        // races loadeddata (no captions) against the server refresh.
+        const deferViewerForTranscription = needsTranscriptionRep && !needsVideoReps && needsServerRefresh;
+
+        if (!deferViewerForTranscription) {
+            this.loadViewer();
+        }
+
+        if (needsServerRefresh) {
             this.loadFromServer();
         }
     }
@@ -1336,6 +1349,12 @@ class Preview extends EventEmitter {
                 throw new PreviewError(ERROR_CODE.ACCOUNT, __('error_account'));
             }
 
+            const needsTranscriptionReload =
+                isFeatureEnabled(this.options.features, AI_TRANSCRIPTION_FOR_VIDEO_SUBTITLES) &&
+                this.isVideoFileByExtension() &&
+                !this.hasTranscriptionRep(cachedFile) &&
+                this.hasTranscriptionRep(file);
+
             // Should load viewer for first time if:
             //   - File isn't cached OR
             //   - Cached file doesn't have a valid structure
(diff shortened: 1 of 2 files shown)


src/lib/Preview.js
```diff
@@ -1740,6 +1806,12 @@ class Preview extends EventEmitter {
 
         // Check if we hit the retry limit for fetching file info
         if (this.retryCount > RETRY_COUNT) {
+            // Viewer load was deferred for transcription refresh; fall back to cached playback.
+            if (this.viewerLoadDeferredForTranscription && !this.viewer) {
+                this.loadViewer();
```
jackiejou: If file-info keeps failing, we uncache and retry via load(this.file.id). After that, does loadViewer() still have a file with dash/mp4 to play, or is this.file only { id }, so loadViewer() throws and the spinner never clears?  we shouldnt block the viewer if only the file-info calls fail

kronosmichall: This path no longer defers the viewer. Cache hits call `loadViewer()` immediately; file-info is a background refresh.

`needsTranscriptionRep` is no longer treated like missing dash/mp4 under `skipServerUpdate` — captions are optional, so skip-server mode does not force a file-info call just to fetch `extracted_text`. Default reopen already refreshes via `loadFromServer()` because `skipServerUpdate` is false.

A failed captions-only refresh will not uncache or destroy a playing viewer: if a viewer is already showing dash/mp4, `loadFromServer()` swallows the file-info error instead of going through `handleFetchError` → `uncacheFile` → `load(this.file.id)` → `triggerError`.

src/lib/Preview.js
```diff
@@ -1220,11 +1208,22 @@ class Preview extends EventEmitter {
...
-        // Finally load the viewer
+        const needsVideoReps = this.isVideoFileByExtension() && !this.hasPlayableVideoReps(this.file);
+        const needsTranscriptionRep =
+            isFeatureEnabled(this.options.features, AI_TRANSCRIPTION_FOR_VIDEO_SUBTITLES) &&
+            this.isVideoFileByExtension() &&
+            !this.hasTranscriptionRep(this.file);
+        // Captions are optional — do not treat a missing extracted_text rep like missing
+        // dash/mp4. skipServerUpdate should still skip the refresh; default reopen already
+        // calls loadFromServer() because skipServerUpdate is false.
+        const needsServerRefresh = !this.options.skipServerUpdate || needsVideoReps || needsTranscriptionRep;
```
jackiejou, line 1219: would this be true for videos without captions? should we skip refreshing them?

kronosmichall: we should, but `needsVideoReps` already forces refresh for every dash preview, so it's not that big of a deal? 
I could not get it to work without forcing server refresh, but i don't know preview that well

jackiejou: do you have a video with no caption that can be tested?

would this work?
```suggestion
        const needsServerRefresh = !this.options.skipServerUpdate || needsVideoReps;
```

kronosmichall: First open (/file/21920937907):

  • File came from preview-client / EUA getItems, whose hints are pdf/jpg plus [extracted_text] — no [dash,mp4].
  • Logged reps: jpg, extracted_text, ORIGINAL.
  • needsVideoReps was true, so box-content-preview still called file-info with its own hints ([dash,mp4][filmstrip][extracted_text]).
  • ~500ms later DashViewer had dash and extracted_text. 

  Refresh (?sb=/activity):

  • File came from BUE ContentPreview.fetchFile, whose X_REP_HINTS are pdf/jpg plus [dash,mp4][filmstrip] — no [extracted_text].
  • Logged reps: jpg, dash, filmstrip, ORIGINAL.
  • Dash shows up because this GET asked for it. The API already had it (your first box-content-preview file-info already returned dash: success).
  • extracted_text is omitted because this GET did not ask. Representations are hint-filtered; a missing entry does not mean the rep is gone.

i worked around this with `needsTranscriptionRep` but i think i could change BUE

kronosmichall: i think the alternative is https://github.com/box/box-ui-elements/pull/4854

but this pr works on videos which do not have extracted_text (because the audio path is missing all together) but it forces the server reload every time if its missing

jackiejou (dismissed):

kronosmichall:
mergifyio queue

kronosmichall:
@Mergifyio queue

kronosmichall:
https://github.com/Mergifyio queue

PR #1746, opened 2026-08-18
chore(i18n): Replace mojito-rb-gen with props2es wrapper

jackiejou:
## Summary

- `yarn build:i18n` runs `@box/frontend` `props2es` in a temp `i18n` directory, then writes `src/i18n/json/{locale}.json` with an English merge for missing keys.
- Removes unmaintained `mojito-rb-gen`. Properties stay in `src/i18n`. Webpack, Jest, and `dist/{version}/{locale}/preview.js` are unchanged.

Related: #1740 is a local properties-to-JSON script without `props2es`.

## Why

`mojito-rb-gen` is unmaintained (npm 0.0.1). `props2es` is the shared converter, but it only reads repo-root `i18n/` and does not merge English unless extract JSON exists. The wrapper keeps our `src/i18n` layout and the `-b en-US.properties` fallback (needed for incomplete locales such as `en-x-pseudo`).

## Test plan

- [ ] `yarn build:i18n` writes 26 files under `src/i18n/json`; `en-US` and `en-x-pseudo` both have 228 keys
- [ ] `yarn test` (Jest `global.__` still reads `src/i18n/json/en-US.json`)
- [ ] `yarn start` / `yarn build:prod` still emit per-locale `preview.js`
- [ ] Travis `clean build:i18n build:ci test:ci`

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,13 +21,8 @@
         "./pdf.worker.min.mjs": "./dist/lib/pdf.worker.min.mjs",
         "./package.json": "./package.json"
     },
-    "files": [
-        "dist/lib"
-    ],
-    "sideEffects": [
-        "*.css",
-        "*.scss"
-    ],
+    "files": ["dist/lib"],
+    "sideEffects": ["*.css", "*.scss"],
     "publishConfig": {
         "access": "public"
     },
@@ -114,7 +109,6 @@
         "lodash": "^4.18.1",
         "mini-css-extract-plugin": "^2.10.2",
         "mock-local-storage": "^1.1.15",
-        "mojito-rb-gen": "^0.0.1",
         "npm-run-all": "^4.1.5",
         "postcss": "^8.5.18",
         "postcss-loader": "^8.1.1",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -149,7 +143,7 @@
         "build": "yarn setup && yarn clean && yarn build:i18n && yarn build:dev",
         "build:ci": "yarn setup && yarn webpack --progress --config build/webpack.config.js",
         "build:dev": "BABEL_ENV=dev NODE_ENV=dev yarn webpack --progress --config build/webpack.config.js",
-        "build:i18n": "mojito-rb-gen -s src/i18n -o src/i18n/json -b en-US.properties",
+        "build:i18n": "node build/buildI18n.js",
```
ehoogerbeets, line 146:
```suggestion
        "build:i18n": "cd src && node ../node_modules/@box/frontend/i18n/props2es.js",
```

I think the above should work. It's a simpler solution and then you don't need the buildI18n.js wrapper.

PR #1751, opened 2026-08-21
feat(audio): add wavesurfer waveform view

abhishek1128:
## Summary

- Adds a peaks-only `WaveformView` that draws unit peaks with wavesurfer (`7.12.11`). It does not fetch audio or attach a media element.
- Playhead, hover time, buffer/hover fills, and a placeholder→decoded peak morph (with reduced-motion skip) live in this module. Duration can change without remounting the wavesurfer instance.
- No MP3 player chrome, client decode, or zoom in this PR. Those follow in stacked PRs.

### DIFF
--- a/src/lib/viewers/media/waveform/WaveformView.tsx
+++ b/src/lib/viewers/media/waveform/WaveformView.tsx
@@ -1,0 +1,320 @@
+import React, { useCallback, useEffect, useLayoutEffect, useRef, useState } from 'react';
+import WaveSurfer from 'wavesurfer.js';
+import {
+    getBufferedProgress,
+    getWaveformFills,
+    toCanvasFill,
+    WAVEFORM_COLOR_PLAYED,
+    WAVEFORM_COLOR_UNPLAYED,
+} from './colors';
+import { formatTime, morphPeaks, toChannels, WAVEFORM_PEAK_TRANSITION_MS } from './peaks';
+import './WaveformView.scss';
+
+export type WaveformViewProps = {
+    bufferedRange?: TimeRanges;
+    currentTime?: number;
+    durationSec: number;
+    height?: number;
+    interactive?: boolean;
+    mediaEl?: HTMLMediaElement | null;
+    onSeek?: (timeSec: number) => void;
+    peaks: ArrayLike<number>;
+};
+
+export const WAVEFORM_BAR_GAP = 2;
+export const WAVEFORM_BAR_WIDTH = 2;
+export const WAVEFORM_BAR_RADIUS = WAVEFORM_BAR_WIDTH / 2;
+/** Total bar height so the top and bottom radii meet as a circle on the mirror. */
+export const WAVEFORM_BAR_MIN_HEIGHT = WAVEFORM_BAR_WIDTH;
+export const WAVEFORM_HEIGHT = 140;
+
+function leftPercent(timeSec: number, durationSec: number): string {
+    const progress = durationSec > 0 ? Math.min(1, Math.max(0, timeSec / durationSec)) : 0;
+    return `${progress * 100}%`;
+}
+
+function fillWidth(container: HTMLElement | null): number {
+    const width = container ? container.clientWidth : 0;
+    const pixelRatio = typeof window !== 'undefined' ? window.devicePixelRatio || 1 : 1;
+    return width * pixelRatio;
+}
+
+type WaveSurferRenderer = {
+    render?: (data: unknown) => void;
+};
+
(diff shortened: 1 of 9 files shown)


src/lib/viewers/media/waveform/WaveformView.tsx
```diff
@@ -0,0 +1,320 @@
...
+    if (wavesurfer.setOptions) {
+        wavesurfer.setOptions({
+            duration: durationSec,
+            peaks: channelData,
+        });
+    }
+
+    const decoded = wavesurfer.getDecodedData ? wavesurfer.getDecodedData() : null;
+    // Private wavesurfer renderer (7.12.11): force a redraw after setOptions(peaks).
+    const { renderer } = (wavesurfer as unknown) as { renderer?: WaveSurferRenderer };
```
mrscobbler: Private wavesurfer renderer used to redraw peaks

**Concern:** This is the path that actually changes the bars (placeholder → real peaks, duration updates, later zoom). After passing in the new peaks, it reaches into wavesurfer’s internals and redraws from the *old* audio data. That renderer is undocumented, so a wavesurfer upgrade can change or remove it without warning and break the waveform. Follow-up PRs will all go through this.

**Recommendation:** Use the documented public API:

```ts
wavesurfer.load('', toChannels(peaks), durationSec);
```

src/lib/viewers/media/waveform/WaveformView.tsx
```diff
@@ -0,0 +1,320 @@
...
+            } else {
+                displayedPeaksRef.current = peaks;
+            }
+        };
+        peakTransitionRafRef.current = window.requestAnimationFrame(tick);
+
+        return () => window.cancelAnimationFrame(peakTransitionRafRef.current);
+    }, [durationSec, peaks]);
+
+    useEffect(() => {
```
mrscobbler, line 253: Hover/buffer fills are sized once and never updated on resize

**Concern:** `toCanvasFill` builds a gradient from the current canvas width. This effect only reruns on hover/buffer changes, so a layout or zoom resize leaves the played/buffer/hover bands misaligned. This module owns those fills; parent PRs should not have to patch width.

**Recommendation:** Match `TimeControlsV2` (~54–67): observe the canvas with `ResizeObserver`, keep width in state, and recompute fills when it changes.

src/lib/viewers/media/waveform/__tests__/WaveformView-test.tsx
```diff
@@ -0,0 +1,229 @@
...
+
+describe('WaveformView', () => {
+    beforeEach(() => {
+        jest.clearAllMocks();
+    });
+
+    test('should mount a wavesurfer instance from peaks without a media url', () => {
+        render(<WaveformView durationSec={8} peaks={[0.2, 0.8, 0.1]} />);
+
+        expect(screen.getByTestId('bp-waveform-view')).toBeInTheDocument();
```
mrscobbler, line 69: Click/seek contract is untested

**Concern:** `onSeek` is the API later PRs will wire. Hover and morph are covered; a click at 25% of an 8s file is not.

**Recommendation:**

```ts
test('should seek from a wavesurfer click', () => {
    const onSeek = jest.fn();
    let clickHandler: (relativeX: number) => void = () => {};
    mockOn.mockImplementation((event, handler) => {
        if (event === 'click') {
            clickHandler = handler;
        }
        return jest.fn();
    });

    render(<WaveformView durationSec={8} onSeek={onSeek} peaks={[0.2, 0.8]} />);
    clickHandler(0.25);

    expect(onSeek).toHaveBeenCalledWith(2);
});
```

mrscobbler (commented):
Foundation looks right: peaks-only wavesurfer, no media fetch, duration can change without remounting. Three things to lock in here so later chrome / decode / zoom PRs do not inherit a fragile redraw path or broken fills.

src/lib/viewers/media/waveform/WaveformView.tsx
abhishek1128: Switched peak updates to the public API

src/lib/viewers/media/waveform/WaveformView.tsx
abhishek1128: Added a ResizeObserver on the canvas container

src/lib/viewers/media/waveform/__tests__/WaveformView-test.tsx
abhishek1128: Added a click test

mrscobbler (approved):
approved!

PR #1752, opened 2026-08-21
Feat/audio player v2

abhishek1128:
## Summary

- Adds opt-in MP3 v2 chrome (`MP3ControlsV2`) on top of the waveform view from the stacked waveform PR. Placeholder peaks only; no client decode and no zoom.
- Gate is `audioPlayerV2.enabled` and React controls. Default off. `MP3ControlsV2` is a lazy chunk; a failed load drops v2 classes and keeps the classic player.
- Play overlay unlocks the waveform after metadata; skip buttons stay on v1 and are hidden on v2.

Depends on: `feat/audio-waveform-view` (waveform renderer + wavesurfer).

### DIFF
--- a/src/lib/viewers/controls/media/PlayPauseToggle.tsx
+++ b/src/lib/viewers/controls/media/PlayPauseToggle.tsx
@@ -12,9 +12,15 @@ export type Props = {
     isPlaying?: boolean;
     onPlayPause: (isPlaying: boolean) => void;
     movePlayback?: (forward: boolean, duration: number) => void;
+    showSkipButtons?: boolean;
 };
 
-export default function PlayPauseToggle({ isPlaying, onPlayPause = noop, movePlayback = noop }: Props): JSX.Element {
+export default function PlayPauseToggle({
+    isPlaying,
+    onPlayPause = noop,
+    movePlayback = noop,
+    showSkipButtons = true,
+}: Props): JSX.Element {
     const PlayPauseIcon = isPlaying ? IconPause24 : IconPlay24;
     const title = isPlaying ? __('media_pause') : __('media_play');
     const skipForwardTitle = __('media_skip_forward');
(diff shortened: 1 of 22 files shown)


src/lib/viewers/media/__tests__/MP3ControlsV2-test.jsx
```diff
@@ -0,0 +1,142 @@
+import React from 'react';
+import { render, screen, within } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
+import MP3ControlsV2 from '../MP3ControlsV2';
+
+jest.mock('../waveform/WaveformView', () => {
+    const React = require('react');
```
jackiejou: do we need to require again? we have import react at the top?

src/lib/viewers/media/__tests__/MP3ControlsV2-test.tsx
```diff
@@ -0,0 +1,142 @@
+import React from 'react';
```
jackiejou, line 1: can we create a tsx test instead?

src/lib/viewers/controls/media/PlayPauseToggle.tsx
```diff
@@ -12,9 +12,15 @@ export type Props = {
     isPlaying?: boolean;
     onPlayPause: (isPlaying: boolean) => void;
     movePlayback?: (forward: boolean, duration: number) => void;
+    showSkipButtons?: boolean;
```
jackiejou: should we use a var name with is- , has- , can-? since it is a boolean?  showSkipButtons sound like a function

src/lib/viewers/media/__tests__/MP3ControlsV2-test.jsx
abhishek1128: addressed. not required

src/lib/viewers/controls/media/PlayPauseToggle.tsx
abhishek1128: changed to hasSkipButtons

src/lib/viewers/media/__tests__/MP3ControlsV2-test.tsx
abhishek1128: addressed

jackiejou (approved):

PR #1756, opened 2026-08-24
feat(audio): decode client waveform peaks under size and duration caps

abhishek1128:
## Summary
- Add a client decode path that turns compressed audio into waveform peaks when the file is under **6 MiB and 5 minutes**. Skip decode (fail-closed) if size or duration is missing or over cap.
- Caps, skip reasons, and peak extract live in `waveform/decode.ts`. No viewer wiring in this PR — the chunk is not loaded yet.
- In-memory peaks skip the JSON byte-length check. Shared `isAbortError` / `errorFromUnknown` so decode and the later viewer use the same mapping.

## Out of scope
MP3Viewer, `media_metric_waveform_decode`, and waveform CSS. Those land in the follow-up stacked on this branch.

### DIFF
--- a/src/lib/viewers/media/waveform/decode.ts
+++ b/src/lib/viewers/media/waveform/decode.ts
@@ -7,44 +7,18 @@ import {
     WAVEFORM_PAYLOAD_VERSION,
 } from './constants';
 import { createCappedWaveformState, errorFromUnknown, isAbortError, WaveformLoadError } from './createWaveformLoader';
-import { WaveformCaps, WaveformError, WaveformLoadState } from './types';
+import {
+    ClientDecodeOutput,
+    ClientDecodeRequest,
+    ClientDecodeResult,
+    DecodeDecision,
+    DecodeSkipReason,
+    DecodeTimings,
+    DecodeToPeaksFn,
+    MediaInfo,
+    WaveformCaps,
+} from './types';
 import { isRetryableWaveformError, validateWaveformPayload } from './validateWaveformPayload';
-
-export type DecodeSkipReason = 'missing_metadata' | 'compressed_size' | 'duration';
-
-export type DecodeDecision = { isAllowed: true } | { isAllowed: false; reason: DecodeSkipReason };
-
-export type MediaInfo = {
-    compressedBytes?: number;
-    durationSec?: number;
-};
-
-export type ClientDecodeOutput = {
-    durationSec: number;
-    peaks: number[];
-    extractMs: number;
-};
-
-export type DecodeToPeaksFn = (signal: AbortSignal) => Promise<ClientDecodeOutput>;
-
-export type DecodeTimings = {
-    attemptMs: number | null;
-    extractMs: number | null;
-};
-
-export type ClientDecodeResult = WaveformLoadState & {
-    isDecodeSkipped: boolean;
-    timings: DecodeTimings;
-    reason?: DecodeSkipReason;
-    error?: WaveformError;
-};
-
-export type ClientDecodeRequest = {
-    compressedBytes?: number;
-    durationSec?: number;
-    fetchArrayBuffer: (signal: AbortSignal) => Promise<ArrayBuffer>;
-    signal?: AbortSignal;
-};
 
 type DecodeAudioContext = {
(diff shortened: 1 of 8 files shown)


src/lib/viewers/media/waveform/decode.ts
```diff
@@ -0,0 +1,380 @@
...
+
+    const AudioContextConstructor = getAudioContextConstructor();
+    if (!AudioContextConstructor) {
+        throw new WaveformLoadError('DECODE_FAILED', 'AudioContext is not available');
+    }
+
+    const context = new AudioContextConstructor();
+
+    try {
+        const audioBuffer = await decodeWithContext(context, arrayBuffer.slice(0), signal);
```
kduncanhsu, line 230: since the only caller to this function is `loadPeaks`, do we need to call `array.slice(0)` here?

abhishek1128: The copy from `arrayBuffer.slice(0)` is not because there is only one caller. `decodeWithContext` / `decodeAudioData` detaches the buffer it receives, and `decodeToPeaks` should not do that to its argument, so it passes a copy. That keeps the helper robust so you can test on its own. The tradeoff is a short-lived copy, at most 6 MiB.

src/lib/viewers/media/waveform/decode.ts
```diff
@@ -0,0 +1,380 @@
...
+};
+
+export type ClientDecodeResult = WaveformLoadState & {
+    isDecodeSkipped: boolean;
+    timings: DecodeTimings;
+    reason?: DecodeSkipReason;
+    error?: WaveformError;
+};
+
+export type ClientDecodeRequest = {
```
jackiejou: can we put these export types in the types file?

src/lib/viewers/media/waveform/decode.ts
```diff
@@ -0,0 +1,380 @@
...
+    decodeAudioData: (
+        buffer: ArrayBuffer,
+        successCallback?: (decoded: AudioBuffer) => void,
+        errorCallback?: (error?: unknown) => void,
+    ) => Promise<AudioBuffer> | void;
+};
+
+const EMPTY_TIMINGS: DecodeTimings = { attemptMs: null, extractMs: null };
+
+function now(): number {
```
jackiejou: can we update the function name? it is not clear what now() does from a caller's perspective

src/lib/viewers/media/waveform/types.ts
```diff
@@ -85,6 +85,8 @@ export type WaveformValidationOptions = {
     /** Override default caps (used in tests and perf harness). */
     maxPeakCount?: number;
     maxPayloadBytes?: number;
+    /** Skip JSON byte-length check (in-memory client peaks, not wire JSON). */
+    isPayloadByteCheckSkipped?: boolean;
```
jackiejou: can we alphabetize?

src/lib/viewers/media/waveform/decode.ts
```diff
@@ -0,0 +1,354 @@
...
+    decodeAudioData: (
+        buffer: ArrayBuffer,
+        successCallback?: (decoded: AudioBuffer) => void,
+        errorCallback?: (error?: unknown) => void,
+    ) => Promise<AudioBuffer> | void;
+};
+
+const EMPTY_TIMINGS: DecodeTimings = { attemptMs: null, extractMs: null };
+
+function timestampMs(): number {
```
jackiejou: timestampMs is still not very clear of what it does. could we do something like getNowTimestampMs or similar?

src/lib/viewers/media/waveform/decode.ts
```diff
@@ -0,0 +1,354 @@
...
+        const succeed = (decoded: AudioBuffer) => {
+            if (isSettled) {
+                return;
+            }
+            isSettled = true;
+            signal?.removeEventListener('abort', onAbort);
+            resolve(decoded);
+        };
+
+        const fail = (error?: unknown) => {
```
jackiejou: can we rename these function callbacks to onSuccess and onFailure?

jackiejou (approved):

PR #1757, opened 2026-08-24
feat(audio): load client waveform peaks into the MP3 viewer

abhishek1128:
## Summary
- Wire the merged client decode path into `MP3Viewer` when audio player v2 is on. Lazy-load `loadPeaks` so playback never waits on the decode chunk.
- Prefer the blob URL playback already fetched; only fall back to an authenticated GET if that blob is missing. Overlay play retries decode once when the error is retryable (Safari user-gesture).
- Round the playhead and increase waveform / control-bar padding to match v2.

## Test plan
- [ ] MP3 under 6 MiB and 5 min: peaks replace the placeholder after metadata; play is not blocked
- [ ] File over either cap, or missing size/duration: placeholder stays
- [ ] Safari: first overlay play still starts audio; retryable decode failure retries once on the next overlay play
- [ ] Destroy / navigate away aborts an in-flight decode

### DIFF
--- a/src/lib/events.js
+++ b/src/lib/events.js
@@ -133,6 +133,7 @@ export const MEDIA_METRIC = {
 export const MEDIA_METRIC_EVENTS = {
     bufferFill: 'media_metric_buffer_fill',
     endPlayback: 'media_metric_end_playback',
+    waveformDecode: 'media_metric_waveform_decode',
 };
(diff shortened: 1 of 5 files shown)


kduncanhsu (approved):
lgtm

PR #1759, opened 2026-08-26
feat(audio): keep the zoomed playhead in view

abhishek1128:
## Summary

When the waveform is zoomed and audio is playing, a playhead camera keeps the current time in view. Fit-to-width (1×) does not move the camera.

**Walk, then pin.** The playhead crosses the visible window with the waveform still. When it reaches ~200px from the right edge (capped at one third of a narrow player), the line pins there and the waveform scrolls under it.

**Jump on play.** If playback starts with the playhead off-screen (left or right), the view jumps so the line sits at that inset, then follows. Reduced motion skips the 400ms ease and applies the scroll immediately.

**Keep following off the right.** If the line walks off the right while already playing, the camera keeps following. It does not jump unless this is a play-start.

**End of file.** Follow unpins when there is no more waveform to scroll. The playhead can leave the inset so a pan is not stolen.

**User pan.** A pan (including while pinned) lets the camera go for the duration of the gesture. After scroll settles (~150ms):
- If the playhead is at or past the inset (including off the right), do **not** jump back. Follow stays off until the line is left of the inset again (user panned it there, or time walked in from the left). Then it walks and pins as usual.
- Pause during that settle window does not pin or scroll a stopped waveform. Play / `playing` clears the hold and may jump.

**Hook.** Pin vs pan lives in `usePlayheadCamera` (flags, settle timer, jump, `apply`). `WaveformView` still owns WaveSurfer, zoom/pinch, peak morph, and the media rAF loop. Camera math (`getPlayheadCameraAction`) stays in `viewport.ts`.

## Test plan

- [ ] Zoom in, press play near the left: playhead travels across, then the waveform scrolls and the line stays ~200px from the right
- [ ] Press play with the playhead off-screen (left or right): it jumps into view, then follows
- [ ] Let it play to the end: following unpins; the playhead can leave the inset
- [ ] While following, pan: camera lets go for the gesture; after you stop, it does not jump if the line is off the right
- [ ] After a pan, put the playhead left of the inset (or let time walk there): it walks, then pins at the inset
- [ ] Pan, then pause before the camera would settle: the paused view does not pin or jump
- [ ] Fit-to-width (no zoom): no camera movement


src/lib/viewers/media/waveform/types.ts
```diff
@@ -160,6 +160,11 @@ export type WaveformViewport = {
 
 export type WaveformViewportInput = Omit<WaveformViewport, 'endSec' | 'pixelsPerSecond' | 'startSec'>;
 
+export type PlayheadCameraAction =
+    | { kind: 'none' }
```
jmcbgaston: why `kind` over `type`?

src/lib/viewers/media/waveform/usePlayheadCamera.ts
```diff
@@ -0,0 +1,266 @@
...
+        },
+        [],
+    );
+
+    const apply = useCallback(
+        (timeSec: number, playJustStarted = false): void => {
+            mediaTimeRef.current = timeSec;
+            const wavesurfer = wavesurferRef.current;
+            const viewport = viewportRef.current;
+            if (!wavesurfer || viewport.zoomLevel <= WAVEFORM_ZOOM_MIN) {
```
jmcbgaston: [risk] This early return skips the `kind: 'none'` cleanup below, which is the only path that clears `isFollowPinnedRef` when zoom is 1×. `getPlayheadCameraAction` already returns `none` at fit-to-width.

If the user zooms out to 1× while following, the pin stays set. `updatePlayheadPosition` then skips `style.left`, so the line sits at the inset percent instead of tracking time. Fit-to-width must not move the camera — it still has to move the line.

Drop the `zoomLevel` check (keep `!wavesurfer`) so the `none` branch unpins, or call `clearFollowPin()` on this return.

src/lib/viewers/media/waveform/WaveformView.tsx
```diff
@@ -412,21 +424,29 @@ export default function WaveformView({
...
         const stopLoop = (): void => {
             window.cancelAnimationFrame(playheadAnimationRef.current);
+            cancelJump();
+            releaseUserPanHold();
+            clearFollowPin();
+            syncViewport();
             updatePlayheadPosition(media.currentTime);
         };
 
         const handleSeeked = (): void => {
```
jmcbgaston, line 476: [risk] `seeked` clears the pin but never calls `cancelJump()`. While the 400ms jump is running, `apply(..., false)` is a no-op, and the jump tick still retargets with `playJustStarted: true`. If the new time is in view (`kind: 'none'`), `to` falls back to the original jump target, so the waveform keeps sliding toward the pre-seek time.

Call `cancelJump()` here (pause already does). If a live jump action is `none`, stop at the current scroll.

src/lib/viewers/media/waveform/usePlayheadCamera.ts
```diff
@@ -0,0 +1,266 @@
...
+                scrollSettleTimerRef.current = 0;
+                if (!mediaElRef.current || mediaElRef.current.paused) {
+                    return;
+                }
+                applyRef.current?.(mediaTimeRef.current, false);
+            }, WAVEFORM_FOLLOW_SCROLL_SETTLE_MS);
+            cancelJump();
+            isFollowPinnedRef.current = false;
+            programmaticScrollRef.current = false;
+            onUserPan(mediaTimeRef.current);
```
jmcbgaston: [risk] `mediaTimeRef` starts at 0 and is only written in `apply()`. Pan-before-play, pan-after-pause, and seek-then-pan-while-paused all feed this into `onUserPan` and the settle timeout, so the playhead can snap to 0 or a stale time.

Read `mediaElRef.current?.currentTime` here and in the settle callback, and write `mediaTimeRef` from `handleSeeked`.

src/lib/viewers/media/waveform/usePlayheadCamera.ts
```diff
@@ -0,0 +1,266 @@
...
+    }, []);
+
+    const applyScrollLeft = useCallback(
+        (scrollLeftPx: number, shouldCommitState: boolean): void => {
+            const wavesurfer = wavesurferRef.current;
+            if (!wavesurfer || !wavesurfer.setScroll) {
+                return;
+            }
+            programmaticScrollRef.current = true;
+            lastCameraScrollRef.current = scrollLeftPx;
```
jmcbgaston, line 122: [risk] `lastCameraScrollRef` is set to the requested value before `setScroll`. If WaveSurfer emits `scroll` synchronously, `handleScroll` compares against that request, not the applied/clamped value. Follow is only treated as camera-owned when `|scroll - lastCamera| <= 1`, so a >1px clamp or round trips a false user-pan (unpin + `holdFollowUntilInset`).

Set `lastCamera` from `getScrollLeft()` after apply (or pass the applied value into the sync handler). Treat clamp-to-the-same-max-scroll as camera-owned.

src/lib/viewers/media/waveform/WaveformView.tsx
```diff
@@ -186,8 +185,30 @@ export default function WaveformView({
...
+        releaseUserPanHold,
+        shouldFreezeViewport,
+    } = usePlayheadCamera({
+        mediaElRef,
+        onViewportCommit,
+        playheadRef,
+        viewportRef,
+        wavesurferRef,
+    });
+    viewportRef.current = shouldFreezeViewport()
```
jmcbgaston: [risk] Freezing `viewportRef` to the pre-zoom scroll during render, while the zoom effect still calls `applyScrollLeft` (pointer origin or time-center) and the rAF camera is pinned, fights for one frame. Pinch/zoom-while-following loses the zoom origin as follow-scroll jumps back to the inset.

On zoom change, release the pin (same as a pan) or skip camera `apply` until the zoom `setScroll` commits. Also avoid assigning `viewportRef.current` during render — Strict Mode double-render can clobber live follow scroll.

src/lib/viewers/media/waveform/__tests__/viewport-test.ts
```diff
@@ -56,23 +182,9 @@ describe('viewport', () => {
         expect(getWaveformZoomMax({ durationSec: 3600, peakCount: 0, viewWidthPx: 800 })).toBe(1);
 
         expect(getWaveformZoomMax({ durationSec: 2, peakCount: 16384, viewWidthPx: 800 })).toBe(1);
```
jmcbgaston, line 191: [consistency] The last commit dropped these as overlapping camera cases, but they are not camera tests. The only real overlap was the off-right follow (now inlined above).

Restore the `getWaveformZoomMax` cases for `durationSec === WAVEFORM_MIN_VIEW_WINDOW_SEC`, 60s/400px, 60s/800px, and 300s. Also restore `timeLeftPercent` (the helper moved into `viewport.ts`; its tests did not) and `getPinnedPlayheadLeft(200)` for the one-third cap.

src/lib/viewers/media/waveform/__tests__/WaveformView-test.tsx
```diff
@@ -407,4 +424,269 @@ describe('WaveformView', () => {
...
+        });
+
+        expect(playhead.style.left).not.toBe(pinnedLeft);
+
+        mockGetScroll.mockReturnValue(0);
+        act(() => {
+            scrollHandler?.();
+        });
+
+        expect(playhead.style.left).toBe('87.5%');
```
jmcbgaston: [consistency] `87.5%` is `currentTime={3.5}` at scroll 0, but the last rAF tick set `media.currentTime` to 3.8 (95% at this viewport). This either fails once pan uses `mediaTimeRef`, or only passes because the paused layout effect rewrote the line from the prop.

Assert against `media.currentTime` after the pan. A seek-then-pan-while-paused case would lock the stale-time bug above.

jmcbgaston (commented):
Walk/pin/jump math in `getPlayheadCameraAction` matches the writeup. Pin teardown and the time source do not: fit-to-width while following leaves the line stuck, pan/seek paths can read a stale `mediaTimeRef`, and seek during the jump keeps interpolating to the old target.

**Request changes.**

src/lib/viewers/media/waveform/usePlayheadCamera.ts
```diff
@@ -0,0 +1,266 @@
...
+    wavesurferRef,
+}: UsePlayheadCameraOptions): {
+    apply: (timeSec: number, playJustStarted?: boolean) => void;
+    applyScrollLeft: (scrollLeftPx: number, shouldCommitState: boolean) => void;
+    cancelJump: () => void;
+    clearFollowPin: () => void;
+    handleScroll: (onUserPan: (mediaTimeSec: number) => void) => void;
+    isFollowPinned: () => boolean;
+    releaseUserPanHold: () => void;
+    shouldFreezeViewport: () => boolean;
```
jmcbgaston: [consistency] Eight return methods means `WaveformView` has to pick the right combo on play / pause / seek / unmount — that is how `seeked` missed `cancelJump()`.

Collapse to lifecycle and drop the `playJustStarted` boolean arg:
- `onPlay(time)` — release hold + today's `apply(time, true)`
- `onFrame(time)` — today's `apply(time, false)`
- `onPause()` — cancel jump, release hold, clear pin
- `onSeek(time)` — cancel jump, clear pin
- `onScroll()` — keep; write playhead CSS in here (`playheadRef` is already in the hook)

`applyScrollLeft` can stay exported until zoom has another way to set the programmatic flag.

src/lib/viewers/media/waveform/usePlayheadCamera.ts
```diff
@@ -0,0 +1,266 @@
...
+    const onViewportCommitRef = useRef(onViewportCommit);
+    onViewportCommitRef.current = onViewportCommit;
+
+    const programmaticScrollRef = useRef(false);
+    const jumpAnimationRef = useRef(0);
+    const lastCameraScrollRef = useRef<number | null>(null);
+    const mediaTimeRef = useRef(0);
+    const isFollowPinnedRef = useRef(false);
+    const userIsScrollingRef = useRef(false);
+    const holdFollowUntilInsetRef = useRef(false);
```
jmcbgaston, line 64: [consistency] `programmaticScrollRef`, `isFollowPinnedRef`, `userIsScrollingRef`, and `holdFollowUntilInsetRef` are one story split across four booleans. The reader has to know programmatic stays true while pinned, and that a user pan is "pinned + scroll drifted > 1px."

One mode: `'idle' | 'jumping' | 'pinned' | 'panning' | 'holdUntilInset'`. Keep `jumpAnimationRef`, `lastCameraScrollRef`, and the settle timer as implementation, not as extra sources of truth.

src/lib/viewers/media/waveform/usePlayheadCamera.ts
```diff
@@ -0,0 +1,266 @@
...
+            }
+            if (userIsScrollingRef.current && !playJustStarted) {
+                return;
+            }
+            if (jumpAnimationRef.current && !playJustStarted) {
+                return;
+            }
+
+            const liveViewport = getViewportAtScroll(viewport, getScrollLeft(wavesurfer, viewport.scrollLeftPx));
+            if (holdFollowUntilInsetRef.current && !playJustStarted) {
```
jmcbgaston: [consistency] This restates the inset test already in `getPlayheadCameraAction` (`followX` / `viewX >= followX`).

Ask the action instead: if `kind !== 'none'`, stay held; if `'none'`, clear the hold and fall through.

src/lib/viewers/media/waveform/usePlayheadCamera.ts
```diff
@@ -0,0 +1,266 @@
...
+                return;
+            }
+            const scrollLeftPx = getScrollLeft(wavesurfer);
+            const isCameraOwnedScroll = programmaticScrollRef.current || jumpAnimationRef.current !== 0;
+            const lastCameraScroll = lastCameraScrollRef.current;
+            const isFollowCameraScroll =
+                isFollowPinnedRef.current && lastCameraScroll != null && Math.abs(scrollLeftPx - lastCameraScroll) <= 1;
+            // Jump, matching follow scroll, or unpinned programmatic → ours.
+            // Pinned + scroll ≠ lastCamera → user pan (programmatic stays true while pinned).
+            if (
```
jmcbgaston: [consistency] Invert this to "is this a user pan?" and put it in a named `isUserPan(scrollLeftPx)`. Jump / matching follow scroll / unpinned programmatic are camera-owned; pinned + `|delta| > 1` is a pan. A one-line truth table next to the helper would save the next reader from reconstructing it.

src/lib/viewers/media/waveform/usePlayheadCamera.ts
```diff
@@ -0,0 +1,266 @@
...
+                programmaticScrollRef.current = false;
+            });
+            if (shouldCommitState) {
+                onViewportCommitRef.current(appliedScrollLeft, viewportRef.current);
+            }
+        },
+        [viewportRef, wavesurferRef],
+    );
+
+    const pinFollowPlayhead = useCallback(
```
jmcbgaston, line 141: [consistency] Three writers of `playhead.style.left`: `pinFollowPlayhead`, `updatePlayheadPosition`, and the scroll `onUserPan` callback. Unpin here only flips the ref — CSS is someone else's job.

One `setPlayheadLeft(timeSec | 'pinned')` in this hook. `WaveformView` should stop touching the node.

src/lib/viewers/media/waveform/viewport.ts
```diff
@@ -121,3 +123,93 @@ export function zoomFromSliderValue(value: number, maxZoom: number = WAVEFORM_ZO
...
+    return `${progress * 100}%`;
+}
+
+/**
+ * Walk across the view until the playhead nears the right edge, then follow.
+ * Off-screen at play start jumps in; off-screen right while playing keeps following.
+ */
+export function getPlayheadCameraAction({
+    insetPx = WAVEFORM_FOLLOW_INSET_PX,
+    isPlaying,
```
jmcbgaston, line 172: [consistency] `isPlaying` is never false — `apply()` always passes `true`, and pause never calls this. `insetPx` is only so tests can pass `40`; production always uses `WAVEFORM_FOLLOW_INSET_PX` (same on `getFollowInsetPx` / `getPinnedPlayheadLeft`).

Drop both. Tests should use the real 200px inset (or the `/3` cap) and assert those numbers.

src/lib/viewers/media/waveform/viewport.ts
```diff
@@ -121,3 +123,93 @@ export function zoomFromSliderValue(value: number, maxZoom: number = WAVEFORM_ZO
     const t = Math.min(WAVEFORM_ZOOM_SLIDER_MAX, Math.max(0, value)) / WAVEFORM_ZOOM_SLIDER_MAX;
     return clampWaveformZoom(WAVEFORM_ZOOM_MIN + t * (max - WAVEFORM_ZOOM_MIN), max);
 }
+
+/** Scroll offset that still shows a full window (no overscroll). */
+function clampScrollLeft(scrollLeftPx: number, viewport: WaveformViewport): number {
+    const maxScroll = Math.max(0, viewport.durationSec * viewport.pixelsPerSecond - viewport.widthPx);
```
jmcbgaston: [consistency] `durationSec * pixelsPerSecond - widthPx` is computed here, again in `getPlayheadCameraAction` for `isPlayheadPinned`, and a third time in the zoom effect. Third occurrence: `maxScrollLeft(viewport)` next to `clampScrollLeft`.

src/lib/viewers/media/waveform/constants.ts
```diff
@@ -25,10 +25,14 @@ export const CLIENT_DECODE_PEAK_COUNT = 16384;
 
 export const WAVEFORM_ZOOM_MIN = 1;
 export const WAVEFORM_ZOOM_MAX = 24;
-/** Visible window never shorter than this. */
+/** Visible window never shorter than this, so follow-playhead stays smooth. */
```
jmcbgaston: [nit] This constant predates the camera. Restore "Visible window never shorter than this." — the follow-playhead clause attributes a zoom-density limit to a feature that did not create it.

jmcbgaston (commented):
Second pass: YAGNI / KISS / DRY / legibility. Discriminated `PlayheadCameraAction` is the right shape. The hook leaks too many controls, and a few parameters exist only for tests.

Independent of the first-pass bugs.

src/lib/viewers/media/waveform/__tests__/WaveformView-test.tsx
```diff
@@ -407,4 +424,269 @@ describe('WaveformView', () => {
         expect(onZoomChange).not.toHaveBeenCalled();
         expect(screen.getByTestId('bp-waveform-view')).not.toHaveClass('bp-WaveformView--zoomed');
     });
+
+    test('should jump the playhead into view on resume even when WaveSurfer emits extra scroll events', () => {
```
jmcbgaston, line 483: [consistency] The four new camera tests copy the same media / rAF / `setScroll` / zoom-2 fixture. Extract `renderZoomedWaveform({ currentTime })` in this file and vary only the behavior under test.

src/lib/viewers/media/waveform/types.ts
abhishek1128: Renamed to type

src/lib/viewers/media/waveform/usePlayheadCamera.ts
abhishek1128: clearFollowPin() on the 1× return so the line tracks time. Left the zoom check so fit-to-width does not run camera apply / scroll.

src/lib/viewers/media/waveform/WaveformView.tsx
abhishek1128: onSeek calls cancelJump(), same as pause. Not aborting a running play-start jump when live action is none, that is how an off-left resume reaches the inset.

src/lib/viewers/media/waveform/usePlayheadCamera.ts
abhishek1128: Pan and the settle callback read media.currentTime. onSeek writes the ref. Paused layout uses the element too, so a seek-then-pan-while-paused cannot snap to 0

src/lib/viewers/media/waveform/usePlayheadCamera.ts
abhishek1128: Still write lastCamera before setScroll (so a synchronous scroll matches the request) and again from getScrollLeft() after apply. The After-only is what makes the sync handler look like a pan. Clamp-to-the-same-max is now treated as camera-owned.

src/lib/viewers/media/waveform/WaveformView.tsx
abhishek1128: Zoom calls onZoom (unpin + hold follow) so follow does not steal the new scroll. Slider still time-centers; pinch/wheel keep the pointer origin. A short pointer-zoom sticky window covers pinch-end / maxZoom clamp so those do not time-center. viewportRef is updated in a layout effect from live scroll + latest zoom/width. Not skipping camera apply for all playing zooms; that breaks slider centering.

src/lib/viewers/media/waveform/__tests__/viewport-test.ts
abhishek1128: Restored the getWaveformZoomMax cases, timeLeftPercent, and getPinnedPlayheadLeft(200) for the one-third cap.

src/lib/viewers/media/waveform/__tests__/WaveformView-test.tsx
abhishek1128: Pan asserts 95% (media.currentTime 3.8 at this viewport). Added a seek-then-pan-while-paused case so the stale-time bug cannot regress.

src/lib/viewers/media/waveform/usePlayheadCamera.ts
abhishek1128: Added onSeek and onZoom. Left apply / playJustStarted and applyScrollLeft so play/pause/zoom wiring stays the same.

src/lib/viewers/media/waveform/usePlayheadCamera.ts
abhishek1128: Leaving the flags. Pin vs pan detection stays the same; a mode enum is a follow-up.

src/lib/viewers/media/waveform/usePlayheadCamera.ts
abhishek1128: Hold asks the action: stay held if kind !== 'none', otherwise clear and fall through.

src/lib/viewers/media/waveform/usePlayheadCamera.ts
abhishek1128: Named isUserPan with that camera-owned vs pan table. Clamp-to-max is included as camera-owned.

src/lib/viewers/media/waveform/usePlayheadCamera.ts
abhishek1128: User-pan CSS is written in the hook (playheadRef). WaveformView still writes on rAF / layout via updatePlayheadPosition. Not collapsing to one writer in this PR.

src/lib/viewers/media/waveform/viewport.ts
abhishek1128: Not in this PR. Leaving both so camera tests keep the 40px inset. Production still defaults to WAVEFORM_FOLLOW_INSET_PX.

src/lib/viewers/media/waveform/viewport.ts
abhishek1128: maxScrollLeft(viewport) next to clampScrollLeft; the zoom effect uses it too.

src/lib/viewers/media/waveform/constants.ts
abhishek1128: Restored “Visible window never shorter than this.”

src/lib/viewers/media/waveform/__tests__/WaveformView-test.tsx
abhishek1128: Extracted renderZoomedWaveform({ currentTime }) and varied only the behavior under test.

jmcbgaston (approved):

PR #1761, opened 2026-09-01
feat(audio): waveform comment markers wiring

abhishek1128:
## Summary

Wires the waveform comment-marker overlay to the audio player host contract, and jumps the zoomed camera when a selected time is off-screen.

- Listen for `comment_markers` in `loadUIReact` (same attach point as video). Store the list, paint ticks, and **seek + pause** when a **new** `selected` id arrives (feed / deeplink). Same id does not seek again.
- Overlay click: pause, seek, emit `comment_marker_select`. The ring is click-optimistic until the host re-emits `selected`.
- Empty `comment_markers` (`[]`) still clears ticks.
- Zoomed seek camera: jump only when zoomed **and** the time is off-screen; no-op at 1× or when the time is already in view.
- `MarkerAvatar` keeps the initial + palette until the photo `onLoad` (and for a cached `complete` image). Shared with video ticks.

## Intentional

**Empty list clears the overlay.** Honor `[]`. A host that emits `[]` on every filter change will flash the overlay; that is a host bug. Ignoring empty payloads here would leave stale ticks after a real clear.

**`comment_markers` is attached in `loadUIReact`.** Same as video. If the sidebar emits before the listener exists, the first payload is dropped until a later emit. This PR does not buffer a pre-listen list.

**Host `selected` always pauses.** Audio-only. Video does not seek on `comment_markers`. Pause is the product contract (listen to the comment), not a toggle.

**Skip `currentTime` when it already equals the marker time.** Avoids a second `seeked` when the feed already sought. Equality is strict (`!==`); a float remainder after a feed seek could still assign. Only widen that if it shows up in practice.

**Parent `setViewport` still runs every follow frame.** Overlay placement is `%` of the visible window, so `scrollLeftPx` must update or badges freeze while the camera slides. `React.memo(WaveformView)` + `commitReactState: false` keep WaveSurfer off that React path. Do not drop the parent viewport update to “save renders.”

**Seek camera does not follow every playhead tick.** `getSeekCameraAction` jumps only when zoomed and the target is off-screen (comment click, host select, media `seeked`). Playback follow is unchanged.

**Overlay hit target stays `pointer-events: auto`.** The 24px strip has to receive badge clicks. Do not set the overlay to `none`.

**Play overlay `preventDefault` on `mousedown` + focus `.bp-media-container`.** Stops the overlay from stealing focus so Space still plays after the button unmounts.

**Zoomed edge badges can be half-clipped.** Inherited from the overlay PR: zoomed waveform is `padding: 0` / `overflow: clip`. 0:00 sitting on the left edge is cut on purpose (no 60px zoomed gutter).

**`MarkerAvatar` hold-until-load is shared with video.** First `comment_markers` often has `initial` and no `avatarUrl`; swapping to the photo later must not flash. Do not show the `` until `onLoad` / cached `complete`.

## Test plan

- [ ] Host emits `comment_markers` → badges on the waveform; empty list removes them
- [ ] Click a badge: pause, seek, ring, `comment_marker_select`; host ack of the same id does not seek again
- [ ] Feed / deeplink `selected`: seek + pause; already-current time does not assign `currentTime` again
- [ ] Selected marker while zoomed and off-screen: camera jumps to center it; 1× and already-on-screen do not jump
- [ ] Follow-playhead: badges keep sliding with the window
- [ ] Play overlay click: play starts, Space still works
- [ ] Trackpad overscroll on the zoomed scroller does not rubber-band

### DIFF
--- a/src/lib/viewers/media/MP3ControlsV2.tsx
+++ b/src/lib/viewers/media/MP3ControlsV2.tsx
@@ -1,4 +1,4 @@
-import React, { useCallback, useEffect, useRef, useState } from 'react';
+import React, { useCallback, useEffect, useMemo, useRef, useState } from 'react';
 import isFinite from 'lodash/isFinite';
 import { Props as DurationLabelsProps } from '../controls/media/DurationLabels';
 import MediaSettings, { Props as MediaSettingsProps } from '../controls/media/MediaSettings';
@@ -11,7 +11,7 @@ import { ICON_PLAY_LARGE } from '../../icons';
 import { WAVEFORM_ZOOM_DISMISS_MS, WAVEFORM_ZOOM_MIN } from './waveform/constants';
 import { PLACEHOLDER_DURATION_SEC, placeholderPeaks } from './waveform/peaks';
 import { WaveformViewport } from './waveform/types';
-import { clampWaveformZoom } from './waveform/viewport';
+import { clampWaveformZoom, viewportEquals } from './waveform/viewport';
 import WaveformCommentMarkers from './waveform/WaveformCommentMarkers';
 import WaveformView from './waveform/WaveformView';
 import WaveformZoomControl from './waveform/WaveformZoomControl';
@@ -63,21 +63,7 @@ export default function MP3ControlsV2({
     const [viewport, setViewport] = useState<WaveformViewport | null>(null);
     const handleViewportChange = useCallback((next: WaveformViewport) => {
         setMaxZoom(prev => (prev === next.maxZoom ? prev : next.maxZoom));
-        setViewport(prev => {
-            if (
-                prev &&
-                prev.durationSec === next.durationSec &&
-                prev.endSec === next.endSec &&
-                prev.maxZoom === next.maxZoom &&
-                prev.scrollLeftPx === next.scrollLeftPx &&
-                prev.startSec === next.startSec &&
-                prev.widthPx === next.widthPx &&
-                prev.zoomLevel === next.zoomLevel
-            ) {
-                return prev;
-            }
-            return next;
-        });
+        setViewport(prev => (viewportEquals(prev, next) ? prev : next));
     }, []);
 
     const revealZoomControl = useCallback(() => {
@@ -116,8 +102,10 @@ export default function MP3ControlsV2({
         mediaEl?.closest<HTMLElement>('.bp-media-container')?.focus();
     }, [mediaEl, onPlayPause]);
 
-    const waveformMarkers = commentMarkers || [];
-    const selectedMarkerId = waveformMarkers.find(marker => marker.isSelected)?.id ?? null;
+    const waveformMarkers = useMemo(() => commentMarkers || [], [commentMarkers]);
+    const selectedMarkerId = useMemo(() => waveformMarkers.find(marker => marker.isSelected)?.id ?? null, [
+        waveformMarkers,
+    ]);
     const handleCommentMarkerClick = useCallback(
         (marker: CommentMarker) => {
             setPlayRequested(true);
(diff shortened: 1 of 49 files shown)


src/lib/viewers/media/MP3ControlsV2.tsx
```diff
@@ -90,7 +112,24 @@ export default function MP3ControlsV2({
     const handlePlayOverlayClick = useCallback(() => {
         setPlayRequested(true);
         onPlayPause(true);
-    }, [onPlayPause]);
+        // Overlay unmounts on click; without this, focus lands on body and Space is lost.
+        mediaEl?.closest<HTMLElement>('.bp-media-container')?.focus();
+    }, [mediaEl, onPlayPause]);
+
+    const waveformMarkers = commentMarkers || [];
+    const selectedMarkerId = waveformMarkers.find(marker => marker.isSelected)?.id ?? null;
```
kduncanhsu: can we use `useMemo` here?

src/lib/viewers/media/waveform/usePlayheadCamera.ts
```diff
@@ -253,6 +255,59 @@ export default function usePlayheadCamera({
...
+            if (action.type !== 'jump') {
+                return;
+            }
+
+            const from = liveViewport.scrollLeftPx;
+            const to = action.scrollLeftPx;
+            if (prefersReducedMotion() || typeof window.requestAnimationFrame !== 'function' || from === to) {
+                applyScrollLeft(to, true);
+                return;
+            }
```
kduncanhsu, line 284: should we add a test for this branch?

src/lib/viewers/media/MP3ControlsV2.tsx
```diff
@@ -54,8 +60,24 @@ export default function MP3ControlsV2({
...
+        setViewport(prev => {
+            if (
+                prev &&
+                prev.durationSec === next.durationSec &&
+                prev.endSec === next.endSec &&
+                prev.maxZoom === next.maxZoom &&
+                prev.scrollLeftPx === next.scrollLeftPx &&
+                prev.startSec === next.startSec &&
+                prev.widthPx === next.widthPx &&
+                prev.zoomLevel === next.zoomLevel
```
kduncanhsu: nit: can we extract this to a e.g. `viewportEquals` helper?

src/lib/viewers/media/waveform/WaveformView.tsx
```diff
@@ -64,6 +64,15 @@ function applyPeaks(wavesurfer: WaveSurfer, peaks: ArrayLike<number>, durationSe
...
 
+/** WaveSurfer pans `.scroll` inside a shadow root; Chrome rubber-bands that node, not our canvas. */
+function disableScrollOverscroll(container: HTMLElement): void {
+    const host = container.firstElementChild;
+    const scroll = host instanceof HTMLElement ? host.shadowRoot?.querySelector('.scroll') : null;
+    if (scroll instanceof HTMLElement) {
+        scroll.style.overscrollBehaviorX = 'none';
+    }
+}
+
```
kduncanhsu, line 75: could you clarify what this function does... comment is a bit confusing 

how does this differ from the corresponding CSS change (i.e. `::part(scroll)` in WaveformView.scss), do we need both?

src/lib/viewers/media/MP3ControlsV2.tsx
abhishek1128: Extracted to viewport.ts

src/lib/viewers/media/waveform/WaveformView.tsx
abhishek1128: We need both. WaveSurfer’s scroller lives in a shadow root, so the SCSS ::part(scroll) rule applies while zoomed. The JS sets 'overscroll-behavior-x: none' as soon as WaveSurfer is created, before that zoomed class exists.

Made the comment more descriptive as well.

src/lib/viewers/media/waveform/usePlayheadCamera.ts
abhishek1128: Yes. added

kduncanhsu (approved):

PR #1764, opened 2026-09-03
feat(audio): split waveform zoom into in and out icons

abhishek1128:
## Summary
Replace the single waveform zoom glyph with dedicated zoom-in and zoom-out icons.

**Collapsed:** 48px circle with zoom-in. The button stays focusable at max zoom (`aria-disabled`).

**Expanded** (hover, focus, or trackpad zoom): zoom-out, slider, zoom-in. The icon buttons step the slider by 10; the slider still sets zoom precisely.

Reuses existing `zoom_in` / `zoom_out` strings.

## Test plan
- [ ] After play, collapsed control is zoom-in in the top-right of the waveform
- [ ] Hover/focus expands to zoom-out | slider | zoom-in
- [ ] Zoom-in / zoom-out step zoom; slider still scrubs precisely
- [ ] At min zoom, zoom-out is `aria-disabled` and does not change zoom
- [ ] At max zoom, zoom-in is `aria-disabled` and stays focusable
- [ ] Keyboard: tab to zoom-in opens the flyout; zoom-out and slider join the tab order
- [ ] Trackpad/ctrl-wheel zoom still reveals the control


src/lib/viewers/media/waveform/WaveformZoomControl.tsx
```diff
@@ -82,12 +71,27 @@ export default function WaveformZoomControl({
                     setHovered(false);
                 }, WAVEFORM_ZOOM_DISMISS_MS);
             }}
+            role="group"
         >
             <div
```
jmcbgaston, line 90: [risk] The flyout (zoom-out + slider) comes before zoom-in in the DOM. Tab lands on zoom-in and opens the flyout, but the next Tab leaves the control — zoom-out and the slider are only reachable with Shift+Tab.

Put zoom-in first in the DOM and keep the flyout visually on the left with CSS `order` (or `row-reverse`) so forward Tab reaches zoom-out, then the slider.

src/lib/viewers/media/waveform/__tests__/WaveformZoomControl-test.tsx
```diff
@@ -18,18 +18,24 @@ describe('WaveformZoomControl', () => {
...
         expect(flyout).toHaveClass('bp-is-open');
+        expect(screen.getByRole('button', { name: __('zoom_out') })).toHaveAttribute(
+            'data-resin-target',
+            'waveformZoomOut',
+        );
 
         const slider = screen.getByRole('slider', { name: __('media_zoom_slider') });
         expect(slider).toHaveAttribute('data-resin-target', 'waveformZoomSlider');
-        await user.tab();
+        act(() => {
```
jmcbgaston: [consistency] This focuses the slider with `.focus()` instead of Tab, so it cannot catch the keyboard-path break. After Tab to zoom-in, the next Tab should hit zoom-out, then the slider.

src/lib/viewers/media/waveform/__tests__/WaveformZoomControl-test.tsx
```diff
@@ -86,19 +92,44 @@ describe('WaveformZoomControl', () => {
...
+        );
+
+        await user.click(screen.getByTestId('bp-waveform-zoom-out'));
+        expect(onZoomChange).not.toHaveBeenCalled();
+        expect(screen.getByTestId('bp-waveform-zoom-out')).toHaveAttribute('aria-disabled', 'true');
+
+        rerender(<WaveformZoomControl maxZoom={4} onZoomChange={onZoomChange} zoomLevel={4} />);
+        await user.click(screen.getByTestId('bp-waveform-zoom-in'));
+        expect(onZoomChange).not.toHaveBeenCalled();
+        expect(screen.getByTestId('bp-waveform-zoom-in')).toHaveAttribute('aria-disabled', 'true');
```
jmcbgaston: [consistency] This asserts `aria-disabled` and that the click is ignored, not that zoom-in stays in the tab order. Tab to it (or assert it is `document.activeElement`) so a future `disabled` / `tabIndex={-1}` fails this test.

jmcbgaston (changes requested):
Split zoom glyph into in/out buttons, `aria-disabled` at the ends, and the group label all look right. Forward Tab never reaches the new flyout controls because of DOM order.

**Request changes.**

src/lib/viewers/media/waveform/__tests__/WaveformZoomControl-test.tsx
abhishek1128: Replaced .focus() with Tab. After zoom-in, the next Tab hits zoom-out, then the slider.

src/lib/viewers/media/waveform/__tests__/WaveformZoomControl-test.tsx
abhishek1128: Added a Tab (and document.activeElement check) so this fails if zoom-in ever gets disabled or tabIndex={-1}. Still asserting the click is ignored and aria-disabled is set.

src/lib/viewers/media/waveform/WaveformZoomControl.tsx
abhishek1128: Moved zoom-in first in the DOM and kept the flyout on the left with order: -1. Forward Tab now goes zoom-in → zoom-out → slider.

jmcbgaston (approved):

PR #1765, opened 2026-09-04
feat(media): add comment-marker hover, focus, and selected badge states

abhishek1128:
## Summary
- Adds hover, keyboard focus, and selected treatments for comment-marker avatars on the audio waveform and V2 video scrubber.
- Default badge is 20×20 with a 1px `#101010` ring (box-shadow, so Preview’s `border-box` rule does not shrink the photo). Hover and `:focus-visible` grow to 24×24 with a 1px white ring. Selected is 24×24 with a 2px white ring plus 1px `#101010`.
- Clicking a marker selects it immediately. Pointer-down outside the selected badge or its group dismisses locally. The host is not notified of dismiss; a later host ack of the same id does not restore the ring.
- Avatar stacks fan out on hover and keyboard `:focus-visible` only (not `:focus-within`). The waveform dim overlay lifts the same way.
- Same-timestamp video groups use a tick button plus avatar stack (no nested buttons). Tick clicks select the first member unless one is already selected.

## Test plan
- [ ] Audio with waveform comment markers: default size/ring, hover grow + white ring, Tab `:focus-visible` matches hover, click shows selected ring, click elsewhere dismisses
- [ ] After local dismiss, host still holding the same selected id does not bring the ring back; selecting a different marker does
- [ ] Stacked avatars fan out on hover and keyboard focus, not on mouse click into the stack
- [ ] Waveform dim lifts on waveform/marker hover and keyboard focus, then returns
- [ ] Video V2 same-timestamp group: tick + avatars, no nested buttons; tick selects first member, or keeps the already-selected one
- [ ] Overflow `+N` chip uses Gray/90 fill; clicking it selects a marker in that group


src/lib/viewers/controls/media/markers/MarkerCluster.tsx
```diff
@@ -17,18 +19,15 @@ export default function MarkerCluster({ cluster, onMarkerClick }: Props): JSX.El
     };
 
     return (
-        <div className="bp-MarkerCluster" data-testid="bp-marker-cluster" style={style}>
+        <div className="bp-MarkerCluster" data-bp-marker-group="" data-testid="bp-marker-cluster" style={style}>
```
jackiejou, line 22: is `data-bp-marker-group=""` intentional?

src/lib/viewers/controls/media/markers/MarkerTick.tsx
```diff
@@ -1,39 +1,59 @@
...
     const isGroup = markers.length > 1;
+    const isSelected = !isGroup && markers[0].id === selectedId;
+    const className = `bp-TimeControlsV2-marker${isGroup ? ' bp-TimeControlsV2-marker--group' : ''}${
+        isSelected ? ' bp-TimeControlsV2-marker--selected' : ''
+    }`;
+    const style = { left: `${position}%` };
+
+    if (isGroup) {
+        return (
+            <div className={className} data-bp-marker-group="" data-testid="bp-time-controls-marker" style={style}>
```
jackiejou, line 24: same question

src/lib/viewers/controls/media/markers/helpers/onMarkerGroupTickClick.ts
```diff
@@ -0,0 +1,16 @@
+import { MouseEvent } from 'react';
+import { CommentMarker } from './types';
+
+/** Click the range/group tick: first member, unless one is already selected. */
+export default function onMarkerGroupTickClick(
```
jackiejou, line 5: should we put this into a helper folder or utils?

src/lib/viewers/controls/media/markers/MarkerCluster.tsx
abhishek1128: Yes. empty string is a presence flag.
The const group = el.closest('[data-bp-marker-group]') matches this, regardless of value.

I could also use data-bp-marker-group={true}, but that's not correctly indicating the use. So empty string for presence/rendering of the attirbute

src/lib/viewers/controls/media/markers/MarkerTick.tsx
abhishek1128: same

src/lib/viewers/controls/media/markers/helpers/onMarkerGroupTickClick.ts
abhishek1128: moved

jackiejou (approved):

PR #1768, opened 2026-09-09
feat(audio): load waveform peaks from conversion JSON

abhishek1128:
## Summary
Audio player v2 can paint waveform peaks from conversion JSON (`waveform` representation) or from client decode. Playback does not wait on either. The first source to apply peaks wins; later results are ignored.

### Conversion load
- `load()` paints the v2 shell and starts conversion in parallel with the audio blob.
- Look up the `waveform` representation. Missing template, missing status, or `error` falls through to client decode.
- `none` / `pending` / `success` poll `RepStatus`, then GET the JSON with header auth and an `AbortSignal`.
- Conversion `durationSec` can lay out peaks and markers before the blob has metadata. The transport bar, seek, and zoom stay locked until `mediaEl.duration` is known.

### Client decode race
- After `loadeddata`, client decode starts if v2 is on, no source has applied peaks yet, and the file is under the compressed-size and duration caps (6MB / 5 minutes).
- It races even while conversion is still polling or the JSON GET is in flight.
- Retryable client-decode failures can retry once on overlay play, only while peaks are still unset.

### First-wins and mismatch
- Applying conversion peaks aborts in-flight client decode. Applying client peaks aborts the JSON fetch only (not `RepStatus`, which does not reject `getPromise()`).
- If conversion JSON arrives before metadata, duration matching is skipped. Once media duration is known, conversion peaks that disagree by more than 1s are dropped and client decode may start.
- Destroy and v1 fallback abort both loads.

## Test plan
- [ ] Conversion JSON applies peaks; client decode does not overwrite them.
- [ ] Client decode wins while conversion is still pending (no JSON GET).
- [ ] Client decode wins while conversion JSON is in flight; late JSON is ignored.
- [ ] Missing / error / invalid conversion falls back to client decode when under cap.
- [ ] Over size or duration cap: no client decode; placeholder remains if conversion never applies.
- [ ] Conversion peaks before metadata: waveform lays out, bar stays locked; bar unlocks after metadata.
- [ ] Conversion duration disagrees with media duration: peaks dropped, client decode starts.
- [ ] Marker click on the loading shell seeks once duration is known.
- [ ] Destroy mid-fetch does not apply peaks.

### DIFF
--- a/src/lib/viewers/media/MP3Viewer.js
+++ b/src/lib/viewers/media/MP3Viewer.js
@@ -11,6 +11,7 @@ import {
     DURATION_MISMATCH_TOLERANCE_SEC,
 } from './waveform/constants';
 import { createWaveformLoader } from './waveform/createWaveformLoader';
+import { isPositiveFinite } from './waveform/validateWaveformPayload';
 import './MP3.scss';
 
 const CSS_CLASS_MP3 = 'bp-media-mp3';
@@ -19,10 +20,6 @@ function createLoadFailedError(message) {
     const error = new Error(message);
     error.name = 'LOAD_FAILED';
     return error;
-}
-
-function isPositiveFinite(value) {
-    return typeof value === 'number' && Number.isFinite(value) && value > 0;
 }
 
 class MP3Viewer extends MediaBaseViewer {
--- a/src/lib/viewers/media/MP3Viewer.js
+++ b/src/lib/viewers/media/MP3Viewer.js
@@ -1,8 +1,16 @@
 import React from 'react';
+import { AUDIO_PLAYER_V2, STATUS_ERROR, WAVEFORM_REP_NAME } from '../../constants';
 import { VIEWER_EVENT } from '../../events';
+import { getRepresentation } from '../../file';
 import MediaBaseViewer from './MediaBaseViewer';
 import MP3Controls from './MP3Controls';
 import MP3ControlsRoot from './MP3ControlsRoot';
+import {
+    CLIENT_DECODE_MAX_COMPRESSED_BYTES,
+    CLIENT_DECODE_MAX_DURATION_SEC,
+    DURATION_MISMATCH_TOLERANCE_SEC,
+} from './waveform/constants';
+import { createWaveformLoader } from './waveform/createWaveformLoader';
 import './MP3.scss';
 
 const CSS_CLASS_MP3 = 'bp-media-mp3';
@@ -11,6 +19,10 @@ function createLoadFailedError(message) {
     const error = new Error(message);
     error.name = 'LOAD_FAILED';
     return error;
+}
+
+function isPositiveFinite(value) {
+    return typeof value === 'number' && Number.isFinite(value) && value > 0;
 }
 
 class MP3Viewer extends MediaBaseViewer {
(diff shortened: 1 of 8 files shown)


src/lib/viewers/media/__tests__/MP3ControlsV2-test.tsx
```diff
@@ -120,20 +124,31 @@ describe('MP3ControlsV2', () => {
...
+    };
+
+    function mediaElWithDuration(durationSec: number): HTMLAudioElement {
+        const mediaEl = document.createElement('audio');
+        Object.defineProperty(mediaEl, 'duration', { configurable: true, value: durationSec });
+        return mediaEl;
+    }
+
+    const getWrapper = (props: Partial<Props> = {}) => {
+        const resolved = { ...props };
```
jackiejou: what does resolved mean?

src/lib/viewers/media/__tests__/MP3ControlsV2-test.tsx
```diff
@@ -181,6 +196,31 @@ describe('MP3ControlsV2', () => {
...
+            expect(await screen.findByTestId('bp-waveform-view')).toHaveAttribute('data-duration-sec', '180');
+            expect(screen.getByTestId('bp-waveform-view')).toHaveAttribute('data-interactive', 'false');
+            expect(screen.queryByTestId('bp-MP3ControlsV2-bar')).not.toBeInTheDocument();
+            expect(screen.getByTestId('bp-MP3ControlsV2-play-overlay')).toBeInTheDocument();
+        });
+
+        test('should keep the waveform inert after overlay click until media metadata', async () => {
+            const mediaEl = document.createElement('audio');
+            Object.defineProperty(mediaEl, 'duration', { configurable: true, value: NaN });
+            const onPlayPause = jest.fn();
```
jackiejou: can we use the default jest.fn rather than a new one?

src/lib/viewers/media/__tests__/MP3Viewer-test.js
```diff
@@ -318,30 +369,68 @@ describe('lib/viewers/media/MP3Viewer', () => {
...
+        });
+
+        test('should keep conversion peaks when media duration is within tolerance', () => {
+            Object.defineProperty(MediaBaseViewer.prototype, 'loadeddataHandler', { value: jest.fn() });
+            mp3.isAudioPlayerV2 = true;
+            mp3.waveformPeaks = [0.2, 0.8];
+            mp3.waveformPeaksSource = 'conversion';
+            mp3.waveformDurationSec = 30 + DURATION_MISMATCH_TOLERANCE_SEC;
+            mp3.options.file = { id: 1, size: 1024 };
+            mp3.mediaEl = { duration: 30 };
```
jackiejou: should we set these props as defaults? since we are resetting them in each test

src/lib/viewers/media/MP3Viewer.js
```diff
@@ -13,6 +21,10 @@ function createLoadFailedError(message) {
     return error;
 }
 
+function isPositiveFinite(value) {
```
jackiejou: we have this same function in decode, can we make a util library to reuse?

abhishek1128: moved to `src/lib/viewers/media/waveform/validateWaveformPayload.ts`

src/lib/viewers/media/__tests__/MP3Viewer-test.js
abhishek1128: moved to beforeEach

src/lib/viewers/media/__tests__/MP3ControlsV2-test.tsx
abhishek1128: Renamed to nextProps. It’s just the props we pass into render

src/lib/viewers/media/MP3Viewer.js
```diff
@@ -182,7 +210,193 @@ class MP3Viewer extends MediaBaseViewer {
...
+     */
+    applyWaveformPeaks(source, payload) {
+        if (this.waveformPeaksSource) {
+            return false;
+        }
+
+        this.waveformPeaks = payload.peaks;
+        this.waveformPeaksSource = source;
+        this.isWaveformDecodeRetryPending = false;
+        if (source === 'conversion') {
```
zhirongwang, line 362: nit: should we enumerate 'conversion' and 'client' ?

zhirongwang (approved):

src/lib/viewers/media/waveform/validateWaveformPayload.ts
```diff
@@ -30,6 +30,10 @@ function fail(code: WaveformError['code'], message: string, retryable = false):
     return { ok: false, error: { code, message }, retryable };
 }
 
+export function isPositiveFinite(value: unknown): value is number {
```
jackiejou, line 33: should decode.ts use this too?

abhishek1128: It already does

src/lib/viewers/media/MP3Viewer.js
abhishek1128: I left them as string literals as they are only used here, and the JSDoc already unions 'conversion' | 'client'. WaveformPayloadSource in types.ts isn’t a runtime value, and it also includes 'fixture', which this path never uses.

PR #1770, opened 2026-09-10
feat(audio): add draft range handles on the waveform

kduncanhsu:
## Summary
- Add checkbox-driven draft range handles on the audio waveform, including collapsed drafts at the playhead, handle drag/snap, and two-tone in-range fills.
- Handles stay inert until a later PR passes a `range` prop from `comment_range_draft` sidebar events.
- No MP3Viewer wiring or range-scoped playback in this PR.

## Test plan
- [ ] `yarn jest src/lib/viewers/media/waveform/__tests__/range-test.ts src/lib/viewers/media/waveform/__tests__/colors-test.ts src/lib/viewers/media/waveform/__tests__/WaveformRangeSelection-test.tsx src/lib/viewers/media/waveform/__tests__/WaveformView-test.tsx --watchman=false`
- [ ] Confirm existing MP3 playback and waveform zoom/pan still behave as before (no range is passed yet, so handles should not appear in product)
- [ ] In WaveformView tests, collapsed drafts sit beside the playhead, open ranges stretch between start/end, drag does not seek, and zoom/pan keep the range glued

Made with [Cursor](https://cursor.com)


src/lib/viewers/media/waveform/WaveformRangeSelection.tsx
```diff
@@ -0,0 +1,375 @@
...
+        const durationSecRef = useRef(durationSec);
+        const onDragChangeRef = useRef(onDragChange);
+        const onPreviewChangeRef = useRef(onPreviewChange);
+        const onRangeChangeRef = useRef(onRangeChange);
+        const getPlayheadSecRef = useRef(getPlayheadSec);
+        const [dragRange, setDragRange] = useState<ResolvedRange | null>(null);
+        const [activeHandle, setActiveHandle] = useState<RangeHandle | null>(null);
+
+        rangeRef.current = range;
+        viewportRef.current = viewport;
```
jmcbgaston, line 132: 🔴 bug: `viewportRef.current = viewport` on every render undoes `applyViewport`.

Parent `onViewportCommit(..., commitReactState=false)` keeps the React `viewport` prop stale during follow-zoom. A drag `setDragRange` re-render copies that stale window onto `viewportRef`, then the layout effect calls `syncPositions(displayed, viewport, …)` from the prop. Handles jump and `moveDrag` math uses the wrong window. Tooltip positioning has the same issue.

Stop copying the prop over the live viewport. Have the layout effect read `viewportRef`.

src/lib/viewers/media/waveform/WaveformRangeSelection.tsx
```diff
@@ -0,0 +1,375 @@
...
+            [startHandleRef.current, endHandleRef.current].forEach(handleEl => {
+                if (handleEl && handleEl.hasPointerCapture(event.pointerId)) {
+                    handleEl.releasePointerCapture(event.pointerId);
+                }
+            });
+            dragRef.current = null;
+            setActiveHandle(null);
+            setDragRange(null);
+            onDragChangeRef.current?.(false);
+            const committed = commitRangeChange(drag.range);
```
jmcbgaston, line 202: 🟡 risk: pointerup always `commitRangeChange` for an open range, even with no move. Click-without-drag emits `onRangeChange`. Commit only when `drag.range` differs from origin.

src/lib/viewers/media/waveform/WaveformRangeSelection.tsx
```diff
@@ -0,0 +1,375 @@
...
+                if (!dragRef.current) {
+                    return;
+                }
+                endDrag(event);
+            };
+            window.addEventListener('pointermove', onMove);
+            window.addEventListener('pointerup', onUp);
+            window.addEventListener('pointercancel', onUp);
+            return () => {
+                window.removeEventListener('pointermove', onMove);
```
jmcbgaston, line 312: 🟡 risk: cleanup removes window listeners only. `range` going null mid-drag unmounts this layer and never calls `onDragChange(false)`. Parent `isRangeDraggingRef` stays true and seeks stay blocked. End the drag in cleanup.

src/lib/viewers/media/waveform/WaveformRangeSelection.tsx
```diff
@@ -0,0 +1,375 @@
...
+                data-highlighted={highlight ? 'true' : 'false'}
+                data-testid="bp-waveform-range"
+            >
+                <div ref={regionRef} className="bp-WaveformRange-region" data-testid="bp-waveform-range-region" />
+                <div
+                    ref={startHandleRef}
+                    className={`bp-WaveformRange-handle bp-WaveformRange-handle--start${
+                        activeHandle === 'start' ? ' bp-WaveformRange-handle--dragging' : ''
+                    }`}
+                    data-testid="bp-waveform-range-handle-start"
```
jmcbgaston, line 344: 🟡 risk: handles have no `tabIndex` / `role` / `aria-*`. Grip is `opacity: 0` until hover, so keyboard users cannot move the range and `focus-visible` CSS never runs.

src/lib/viewers/media/waveform/WaveformView.tsx
```diff
@@ -187,7 +199,14 @@ function WaveformView({
     const isZoomed = zoomLevel > WAVEFORM_ZOOM_MIN;
     const bufferProgress = getBufferedProgress(bufferedRange, durationSec);
     bufferProgressRef.current = bufferProgress;
-    hoverProgressRef.current = hoverProgress;
+    hoverProgressRef.current = isRangeDragging ? null : hoverProgress;
+    const activeRange = isRangeDragging ? previewRange ?? range : range;
+    const activeRangeRef = useRef(activeRange);
+    activeRangeRef.current = activeRange;
+    const activeRangeProgress = activeRange
```
jmcbgaston, line 207: 🟡 risk: `activeRangeProgress` is a new object every render. The fills effect then `setOptions` + `setTime` on every `currentTime` tick while a range is shown. Depend on primitive start/end, or `useMemo`.

src/lib/viewers/media/waveform/WaveformView.tsx
```diff
@@ -709,10 +743,35 @@ function WaveformView({
...
+    }, []);
+
+    const handleRangeDragChange = useCallback(
+        (isDragging: boolean) => {
+            isRangeDraggingRef.current = isDragging;
+            setIsRangeDragging(isDragging);
+            if (isDragging) {
+                setHoverProgress(null);
+            } else {
+                skipNextSeekRef.current = true;
```
jmcbgaston, line 762: 🟡 risk: `skipNextSeekRef` is cleared in `setTimeout(0)`. A WaveSurfer `click` after that timeout still seeks; a click that never arrives wastes the flag. Suppress until the next `pointerup`/`click` pair, not a timer.

src/lib/viewers/media/waveform/range.ts
```diff
@@ -0,0 +1,202 @@
...
+        }
+    } else if (pointer < startMs) {
+        endMs = startMs;
+        startMs = pointer;
+        nextHandle = 'start';
+    } else {
+        endMs = pointer;
+    }
+
+    const opened = startMs !== endMs;
```
jmcbgaston, line 167: 🟡 risk: `opened` is `startMs !== endMs`, so dragging or snapping one edge onto the other collapses an open range. `commitRangeChange` then returns null and the UI snaps back to the prop. Once open, clamp to min span instead of collapsing.

src/lib/viewers/media/waveform/__tests__/WaveformView-test.tsx
```diff
@@ -995,4 +1033,94 @@ describe('WaveformView', () => {
...
+        expect(screen.getByTestId('bp-waveform-range')).toHaveAttribute('data-highlighted', 'true');
+        expect(screen.getByTestId('bp-waveform-range-handle-start')).toHaveStyle({
+            left: `calc(25% - ${WAVEFORM_RANGE_COLLAPSED_OFFSET_PX}px)`,
+        });
+        expect(screen.getByTestId('bp-waveform-range-handle-end')).toHaveStyle({
+            left: `calc(25% + ${WAVEFORM_RANGE_COLLAPSED_OFFSET_PX}px)`,
+        });
+    });
+
+    test('should keep a draft range glued through zoom and pan', () => {
```
jmcbgaston, line 1120: 🟡 risk: this glue test only drives React `viewport` / `scrollHandler` → `setScrollLeft`. It misses pinned follow (`commitReactState: false`), where `applyViewport` is the only glue — that is the L132 bug. Cover the live-viewport path.

src/lib/viewers/media/waveform/__tests__/WaveformView-test.tsx
```diff
@@ -995,4 +1033,94 @@ describe('WaveformView', () => {
...
+
+        mockGetScroll.mockReturnValue(200);
+        act(() => {
+            scrollHandler?.();
+        });
+        expect(screen.getByTestId('bp-waveform-range-handle-start')).toHaveStyle({ left: '-50%' });
+        expect(screen.getByTestId('bp-waveform-range-handle-end')).toHaveStyle({ left: '0%' });
+    });
+
+    test('should not seek while a range handle is dragging', () => {
```
jmcbgaston, line 1152: 🟡 risk: post-`pointerup` `clickHandler` still expects no seek. That passes if drag state never cleared (`isRangeDraggingRef` stuck). Assert a later click seeks, or that `onRangeDragChange(false)` fired.

src/lib/viewers/controls/media/types.ts
```diff
@@ -0,0 +1,30 @@
...
+};
+
+/** Committed range after a handle drag. Always a real span (`endMs > startMs`). */
+export type CommentRangeChange = {
+    endMs: number;
+    startMs: number;
+};
+
+/** Highlight a posted comment's range. Same payload shape as a committed drag. */
+export type CommentRangeSelect = CommentRangeChange;
```
jmcbgaston: ❓ q: `CommentRangeSelect` aliases a committed span (`endMs` required). Current markers are a single `time` in seconds. Timestamp highlight cannot use this type — intentional?

src/lib/viewers/media/waveform/WaveformView.scss
```diff
@@ -86,4 +86,12 @@
     .bp-WaveformView-track {
         cursor: default;
     }
+
+    .bp-WaveformRange-handle {
+        pointer-events: auto;
```
jmcbgaston, line 91: ❓ q: `--inert` re-enables handle `pointer-events`. Draft drags still fire `onRangeChange` while the waveform is non-interactive (pre-play). Intentional?

jmcbgaston (commented):
Follow-zoom drag can desync handles: render copies the stale React `viewport` over `viewportRef`, then the layout effect repositions from that prop. Other risks around fill updates, seek suppression, unmount-during-drag, and tests that miss the pinned-follow path.

The L132 viewport clobber is the one to fix before merge.

src/lib/viewers/controls/media/types.ts
kduncanhsu: yes, this was just a placeholder type for when we implement the scenario "user clicks timespan comment --> range gets highlighted" in a future PR. however we don't really need it for this PR, so I've removed it to avoid confusion

src/lib/viewers/media/waveform/WaveformRangeSelection.tsx
```diff
@@ -0,0 +1,375 @@
...
+                    viewportRef.current = nextViewport;
+                    syncPositions(displayedRef.current, nextViewport, durationSecRef.current);
+                },
+            }),
+            [syncPositions],
+        );
+
+        useLayoutEffect(() => {
+            syncPositions(displayed, viewport, durationSec);
+        }, [displayed, durationSec, syncPositions, viewport]);
```
abhishek1128, line 186: The scroll while zoomed and range visible is jittery. This layout effect is what I was seeing on zoomed pan: handles stutter even after live `applyViewport`.

Scroll already positions via `applyViewport`. Then React re-renders, `displayed` is a new object, and this effect runs `syncPositions` from the prop viewport. If that prop is a frame behind, the handles jump back.

Keep pan/follow on viewportRef only — don’t copy the prop onto the ref during render. Use the React `viewport` for width/zoom/duration, and keep `scrollLeftPx` from the live ref. Depend on `displayed.startMs` / `displayed.endMs`, not the whole object.

Same thing on the tooltip `left` a bit below — if that stays on the React viewport it’ll jitter while you drag during a pan. Easier to set it in `syncPositions`.

The glue test that rerenders a panned `viewport` will need `ref.applyViewport(panned)` for the scroll step. Add a case that pans live, then rerenders the old prop, and the handles stay put.


```suggestion
        useLayoutEffect(() => {
            syncPositions(displayedRef.current, viewportRef.current, durationSec);
        }, [displayed.endMs, displayed.startMs, durationSec, syncPositions]);
        useLayoutEffect(() => {
            const next = createWaveformViewport({
                durationSec: viewport.durationSec,
                heightPx: viewport.heightPx,
                maxZoom: viewport.maxZoom,
                scrollLeftPx: viewportRef.current.scrollLeftPx,
                widthPx: viewport.widthPx,
                zoomLevel: viewport.zoomLevel,
            });
            viewportRef.current = next;
            syncPositions(displayedRef.current, next, durationSecRef.current);
        }, [
            syncPositions,
            viewport.durationSec,
            viewport.heightPx,
            viewport.maxZoom,
            viewport.widthPx,
            viewport.zoomLevel,
        ]);
```

src/lib/viewers/media/waveform/WaveformView.tsx
```diff
@@ -187,7 +200,14 @@ function WaveformView({
     const isZoomed = zoomLevel > WAVEFORM_ZOOM_MIN;
     const bufferProgress = getBufferedProgress(bufferedRange, durationSec);
     bufferProgressRef.current = bufferProgress;
-    hoverProgressRef.current = hoverProgress;
+    hoverProgressRef.current = isRangeDragging ? null : hoverProgress;
+    const activeRange = isRangeDragging ? previewRange ?? range : range;
+    const activeRangeRef = useRef(activeRange);
+    activeRangeRef.current = activeRange;
+    const activeRangeProgress = activeRange
+        ? rangeProgress(resolveRange(activeRange, durationMsFromSec(durationSec)), durationMsFromSec(durationSec))
```
abhishek1128, line 208: Jose's right that `activeRangeProgress` is a new object every render. On zoomed pan that’s stutter: `setScrollLeft` re-renders every frame, the fills effect re-tints every tile, and the playhead/badges jump with the range even though they aren’t the overlay.

Memoize on `startMs` / `endMs` / `duration` (primitives — the parent range object may be new every time). Pan shouldn’t enter that effect. Only re-tint when the range times, hover, or buffer actually change.


```suggestion
    const durationMs = durationMsFromSec(durationSec);
    const rangeStartMs = activeRange?.startMs;
    const rangeEndMs = activeRange?.endMs;
    const activeRangeProgress = useMemo(() => {
        if (rangeStartMs == null || !Number.isFinite(rangeStartMs)) {
            return null;
        }
        return rangeProgress(
            resolveRange({ endMs: rangeEndMs ?? null, startMs: rangeStartMs }, durationMs),
            durationMs,
        );
    }, [durationMs, rangeEndMs, rangeStartMs]);
```

src/lib/viewers/media/waveform/WaveformView.tsx
abhishek1128, line 312: This runs from a layout effect that depends on `viewport`, so every pan frame calls `setTime` with the same media time and WaveSurfer redraws for no reason.

Skip `setTime` when `timeSec` hasn’t changed. Pan should just move the CSS overlays.

Add `const lastSetTimeSecRef = useRef<number | null>(null);` next to the other refs.

```suggestion
            const wavesurfer = wavesurferRef.current;
            if (wavesurfer && wavesurfer.setTime && lastSetTimeSecRef.current !== timeSec) {
                lastSetTimeSecRef.current = timeSec;
                wavesurfer.setTime(timeSec);
            }
```

abhishek1128 (changes requested):
Zoomed pan with a range makes the overlay, playhead, and comment badges stutter together. The canvas scroll is live; the overlays get pulled back a frame.

Two things stack on every pan frame:

1. Live vs React viewport — `applyViewport` writes the WaveSurfer scroll, then React commits an older `viewport` and the range layer redraws from that prop. Same root as Jose's L132 (`viewportRef.current = viewport` on render). It shows up on a normal zoomed drag-scroll, not only while following.
2. Fills work on pan — `activeRangeProgress` is a new object every render, so the fills `useEffect` re-tints every zoomed tile (and `setTime`) even when start/end have not changed. That’s why the playhead and badges look jittery too, even though they aren’t the range layer.

Pan should keep overlay position on the live viewport (`applyViewport` / `viewportRef`). React `viewport` should only drive structural changes (width / zoom / duration), not `scrollLeftPx`. Memoize range fills on `startMs` / `endMs` primitives so pan does not re-tint.

Worth a test: `applyViewport(panned)` then rerender the stale overview/zoomed prop and assert handles do not jump back. The current glue test only feeds React `viewport` / `setScrollLeft`, so it misses this.

Inlines have the direction. Would like this in before merge.

src/lib/viewers/media/waveform/WaveformRangeSelection.tsx
kduncanhsu: Done — stopped copying the React `viewport` onto `viewportRef` during render. Pan/follow stays on the live ref; the React prop only rebuilds for width / zoom / duration. Tooltip left is set in `syncPositions`. The glue test now pans via `applyViewport`, and there’s a case that rerenders the stale zoomed prop and checks the handles don’t jump back.

src/lib/viewers/media/waveform/WaveformView.tsx
kduncanhsu: Done — memoized on primitive startMs / endMs / durationMs so a new parent range object (or a pan rerender) doesn’t re-enter the fills effect.

src/lib/viewers/media/waveform/WaveformView.tsx
kduncanhsu: Done — added lastSetTimeSecRef and skip setTime when timeSec hasn’t changed. Pan now only moves the CSS overlays.

kduncanhsu:
@abhishek1128 Addressed in 416c9977. Overlay position now stays on the live viewport (`applyViewport` / `viewportRef`) — React viewport only drives width / zoom / duration, not `scrollLeftPx`. Range fills are memoized on `startMs` / `endMs` primitives so pan no longer re-tints tiles or calls `setTime`. Added a test that `applyViewport(panned)` then rerenders the stale zoomed prop and asserts the handles stay put.

src/lib/viewers/media/waveform/WaveformRangeSelection.tsx
kduncanhsu: Addressed in 416c9977. We no longer copy the React viewport onto `viewportRef` during render. Pan/follow stay on the live ref (`applyViewport`); the React prop only rebuilds for width / zoom / duration and keeps `scrollLeftPx` from the live window. Tooltip `left` is set in `syncPositions` so it uses the same live viewport.

src/lib/viewers/media/waveform/WaveformView.tsx
kduncanhsu: Addressed in 416c9977. `activeRangeProgress` is now `useMemo`’d on primitive `startMs` / `endMs` / `durationMs`, so a pan rerender doesn’t re-enter the fills effect.

src/lib/viewers/media/waveform/__tests__/WaveformView-test.tsx
kduncanhsu: Covered in the same commit, at the overlay where the bug lived: `applyViewport(panned)` then `rerender` the stale zoomed prop, and assert the handles stay put. The WaveformView glue test also checks that a live pan does not call `setTime` / `setOptions`. Didn’t add a separate follow / commitReactState: false fixture — happy to if you still want that path called out explicitly.

src/lib/viewers/media/waveform/WaveformRangeSelection.tsx
kduncanhsu: will address in follow up ticket

src/lib/viewers/media/waveform/WaveformView.scss
kduncanhsu: good catch, the user shouldn't be able to create a range when waveform is non-interactive (pre-play). have updated accordingly.

src/lib/viewers/media/waveform/WaveformView.tsx
kduncanhsu: I looked at the event order here. WaveSurfer’s seek-click is the browser’s normal `click`, which fires in the same turn as `pointerup`, before `setTimeout(0)`. So a leftover click from that gesture cannot arrive after we clear the flag.

After a handle drag it often never fires anyway — the handles sit on top of the waveform, and we stop the `pointerdown` from reaching WaveSurfer.

If no leftover click comes, the timer just clears the flag; the next real click still seeks. Left the timer as-is.

Strengthened the test: drag end fires `onRangeDragChange(false)`, the leftover click does not seek, and a later click does.

src/lib/viewers/media/waveform/WaveformRangeSelection.scss
```diff
@@ -0,0 +1,99 @@
...
+    inset: 0;
+    z-index: 2;
+    pointer-events: none;
+}
+
+.bp-WaveformRange-region {
+    position: absolute;
+    top: 50%;
+    box-sizing: border-box;
+    height: 150px;
```
abhishek1128, line 14: Tape stage is taller than desktop (max(140px, 25%) of the player, canvas height: 100%). The range band and grips stay 150px, so they sit short on the bars.

Don’t size this in JS from viewport.heightPx. Follow the playhead: in WaveformView.scss under &.bp-WaveformView--tape, next to the existing playhead override:


```suggestion
.bp-WaveformView-playhead {
    height: calc(100% + 10px);
}

.bp-WaveformRange-region,
.bp-WaveformRange-handle {
    height: calc(100% + 10px);
}
```

That keeps desktop at 150px and makes the range fill the tape canvas the same way the pin does.

src/lib/viewers/media/waveform/WaveformRangeSelection.scss
```diff
@@ -0,0 +1,99 @@
...
+    border-width: 2px;
+    border-color: $white;
+}
+
+.bp-WaveformRange-handle {
+    position: absolute;
+    top: 50%;
+    z-index: 1;
+    width: 24px;
+    height: 150px;
```
abhishek1128, line 36: same for this

src/lib/viewers/media/waveform/WaveformRangeSelection.scss
kduncanhsu: Tape already grows the playhead with `calc(100% + 10px)` under .`bp-WaveformView--tape`. I added the region and handles to that same override so desktop stays 150px and tape tracks the max(140px, 25%) stage. Left the 150px defaults in `WaveformRangeSelection.scss`.

src/lib/viewers/media/waveform/WaveformRangeSelection.scss
kduncanhsu: same as above

abhishek1128 (approved):

PR #1774, opened 2026-09-15
feat(audio): add v2 keyboard shortcuts for shuttle, volume, and zoom

abhishek1128:
## Summary

Audio player v2 gets its own keymap. Flag-off audio and video still use `MediaBaseViewer.handleKeydownReact` (Space/K play-pause, J/L ±10s, arrows ±5s, ↑/↓ volume).

The override lives on `MP3Viewer.onKeydown` when `audioPlayerV2` is on, then falls through to the shared React map for mute, Home/0, and F.

### Mapping (current → v2)

| Key | Flag-off / video | Audio v2 |
| --- | --- | --- |
| Space | Play / pause | Play / pause; exits shuttle. Reverse shuttle + Space → pause |
| K | Play / pause | Pause only; exits shuttle |
| J / L | Skip −10s / +10s | Reverse / forward shuttle. Same key steps **2× → 4× → 8× → 16×**. Opposite key resets to **2×** |
| ← / → | Skip ±5s | Skip **1s** |
| Shift+← / → | (none) | Skip **5s** |
| `,` / `.` | Frame step if `frameStep.enabled` | Frame step **always** (~1/30s) |
| ↑ / ↓ | Volume | Previous / next comment marker. No markers → volume |
| Alt+↑ / ↓ | (Alt not decoded; unused) | Volume ±5%; opens the volume flyout |
| M / Shift+M | Mute | Mute (unchanged) |
| 0 / Home | Start | Start (exits shuttle first) |
| End | (none) | Jump to duration (no-op until duration is known) |
| + / = / − | (none) | Waveform zoom, same 10-unit slider step as the zoom buttons |
| F | Fullscreen | Fullscreen (unchanged) |
| I / O / `/` / Shift+L | (none / L would shuttle) | **Consumed no-ops** until range UI exists |

### Shuttle

- Forward uses `HTMLMediaElement.playbackRate`. Reverse ticks `currentTime` every 50ms (HTML audio cannot play backwards).
- Settings speed (0.5 / 1 / 1.25 / 1.5 / 2) is restored when shuttle ends. Shuttle is a momentary scan, not the listen-speed menu, so the ladders stay separate.
- K, Space, End, Home, marker jump/click, host-selected seek, and hitting t=0 all exit shuttle. `destroy()` clears the reverse timer without painting.

### Comment markers

↑ / ↓ walk every timed marker, including same-time stacks (one keypress per badge). If nothing is selected, Down takes the marker at or after the playhead (Up: at or before). No markers → volume, same as today.

### Volume flyout

Alt+↑/↓ steps volume 5% (same increment as video) and opens the flyout for 1s. Tab order matches zoom: mute, then the slider while the flyout is open. Tabbing out closes it. The full-height track is the one `role="slider"`; the fill is visual only.

### Zoom

`+` / `−` go through React as a signed `keyboardZoomStep` (zoom state lives in the controls, not the viewer). No-op until audio is loaded and zoom is enabled.

## Left out / not in this PR

- **I / O** in/out points and **`/` / Shift+L** loop — keys are swallowed so they cannot become shuttle or Preview navigation; no range UI on this branch.
- **M as drop-marker** — stays mute, same as video.
- **Arrows as frame step (~40ms)** — audio has no frames; `,` / `.` are the fine nudge (~33ms), unshifted arrows are 1s, Shift+arrows recover today’s 5s.
- **Shuttle matching the speed menu (0.5–2)** — shuttle is 2/4/8/16, like NLE/YouTube-style scan.
- **Opposite J/L stepping down** — opposite direction starts over at 2×, not 16→8→4.
- **Video / flag-off keymap** — unchanged.

Known rough edges, not blocking: 16× reverse is noisy (50ms seeks); focused volume slider is 1% per arrow vs Alt’s 5% (slider `step={1}`, shared with video).

## Test plan

- [ ] Flag off: Space/K, J/L ±10s, arrows ±5s, ↑/↓ volume — unchanged
- [ ] V2: Space play/pause; K pause; J/L shuttle 2×→4×→8×→16×; opposite J/L resets to 2×
- [ ] Reverse shuttle stops at 0, on K/Space, End/Home, and on comment click
- [ ] ←/→ = 1s, Shift+←/→ = 5s, `,`/`.` = ~1 frame
- [ ] ↑/↓ walk markers (including a same-time stack); no markers → volume
- [ ] Alt+↑/↓ volume + flyout; Tab mute → slider; Tab out closes flyout
- [ ] +/− zoom after audio is loaded; zoom control reveals
- [ ] M mute, 0/Home start, End end, F fullscreen
- [ ] I / O / `/` / Shift+L do nothing (do not shuttle)


src/lib/viewers/controls/hooks/__tests__/useAttention-test.tsx
```diff
@@ -29,4 +29,23 @@ describe('useAttention', () => {
         fireEvent.mouseOut(element);
         expect(element).not.toHaveClass('active'); // Default
     });
+
+    test('should stay active when focus moves to a child', async () => {
+        function NestedComponent(): React.JSX.Element {
```
jackiejou: can we just have one test mock  with  the above TestComponent as well?

also can we use userEvent?

also can we check if we can test this hook with `import { renderHook } from '@testing-library/react';`?

src/lib/viewers/controls/hooks/useAttention.ts
```diff
@@ -1,21 +1,35 @@
...
+    onBlur: (event: React.FocusEvent<HTMLElement>) => void;
     onFocus: () => void;
-    onMouseOut: () => void;
+    onMouseOut: (event: React.MouseEvent<HTMLElement>) => void;
     onMouseOver: () => void;
 };
 
 export type isActive = boolean;
 
+function isInside(event: { currentTarget: EventTarget; relatedTarget: EventTarget | null }): boolean {
```
jackiejou: what does isInside mean? can we elaborate or update the func name?

src/lib/viewers/controls/media/VolumeControls.tsx
```diff
@@ -9,9 +9,13 @@ import './VolumeControls.scss';
 import IconVolumeMax24 from '../icons/IconVolumeMax24';
 import VolumeSliderControl from '../slider/VolumeSliderControl';
 
+/** Keyboard volume keeps the flyout open longer than the 200ms max-width transition. */
+export const VOLUME_FLYOUT_DISMISS_MS = 1000;
+
 export type Props = {
     onMuteChange: (isMuted: boolean) => void;
     onVolumeChange: (volume: number) => void;
+    revealStep?: number;
```
jackiejou: is this a count or index? or something else? the !revealStep check below made it seem like a boolean

src/lib/viewers/media/MP3Viewer.js
```diff
@@ -15,6 +15,41 @@ import { isPositiveFinite } from './waveform/validateWaveformPayload';
...
+        return AUDIO_V2_SHUTTLE_RATES[0];
+    }
+    const index = AUDIO_V2_SHUTTLE_RATES.indexOf(currentRate);
+    if (index < 0) {
+        return AUDIO_V2_SHUTTLE_RATES[0];
+    }
+    return AUDIO_V2_SHUTTLE_RATES[Math.min(index + 1, AUDIO_V2_SHUTTLE_RATES.length - 1)];
+}
+
+function nextCommentMarker(markers, { direction, selectedId, time }) {
```
jackiejou, line 51: can we add return type docstrings

src/lib/viewers/media/MP3Viewer.js
```diff
@@ -75,6 +116,224 @@ class MP3Viewer extends MediaBaseViewer {
...
+                })
+                .catch(() => {});
+        }
+    }
+
+    startReverseShuttle(rate) {
+        this.stopReverseShuttle();
+        this.shuttleDirection = 'reverse';
+        this.shuttleRate = rate;
+        this.pause(undefined, true);
```
JChan106, line 279: **Must fix:** reverse shuttle never looks like playback to the React controls, and the play/pause UI does not call `exitShuttle()`.

`J` pauses and starts this timer, but `isPlaying` stays false and `playRequested` is never set. On a fresh load `MP3ControlsV2` still renders the play overlay (`showPlayOverlay = !playRequested && !isPlaying`) while reverse shuttle runs underneath.

Clicking that overlay (or the bar play/pause) goes through `handlePlayPause` → `handlePlayRequest` → `togglePlay` and never exits shuttle:

- Reverse + play: audio starts forward while this 50ms timer keeps seeking backward.
- Forward shuttle + pause: `shuttleDirection` stays `'forward'`, so the next `play()` → `handleRate()` keeps the 2–16× rate.

Space/K/Home/End/comment click already call `exitShuttle()`. Please do the same at the start of `handlePlayPause` / `handlePlayRequest` (or a v2 `togglePlay` override), and treat shuttle as “playback started” so the overlay goes away.

Waveform seek (`handleTimeupdateFromMediaControls`) has the same gap — a tape click during reverse shuttle will jump, then this timer walks off that time.

A test for “play during reverse shuttle” (and overlay hidden once `J` has been pressed) would lock this in.

src/lib/viewers/media/MP3Viewer.js
```diff
@@ -75,6 +116,224 @@ class MP3Viewer extends MediaBaseViewer {
...
+    /**
+     * V2 audio keymap. Flag off and video keep MediaBaseViewer.handleKeydownReact.
+     * Alt is not in Preview's key decoder — read it from the native event.
+     *
+     * @param {string} key
+     * @param {KeyboardEvent} [event]
+     * @return {boolean} consumed
+     */
+    onKeydown(key, event) {
+        if (this.getIsAudioPlayerV2() && this.handleKeydownAudioV2(key, event)) {
```
JChan106: **Must fix:** this should gate on `this.isAudioPlayerV2`, not `getIsAudioPlayerV2()`.

`fallbackToV1Controls()` sets `this.isAudioPlayerV2 = false` and renders classic `MP3Controls`, but `getIsAudioPlayerV2()` still returns true whenever the feature flag and React controls are on. After a v2 chunk load failure, `J`/`L` still shuttle and arrows still skip 1s against the v1 UI.

The existing “fall back to v1 when the v2 chunk fails” test already checks the UI; please also assert `onKeydown('j')` is the shared ±10s map after fallback.

JChan106 (changes requested):
## Requesting changes

Two shuttle/keymap bugs that will show up in the v2 player:

1. Play/pause UI (overlay and bar) does not exit shuttle, and reverse shuttle never hides the play overlay.
2. After a v1 controls fallback, `onKeydown` still uses the v2 keymap.

Inline comments have the details. Happy to re-review once those paths exit shuttle / honor `this.isAudioPlayerV2`.

src/lib/viewers/media/MP3Viewer.js
JChan106: AI mentioned 2 must fixes, can you check if this is valid?

src/lib/viewers/media/MP3Viewer.js
JChan106: AI mentioned 2 must fixes, can you check if this is valid?

src/lib/viewers/controls/hooks/__tests__/useAttention-test.tsx
abhishek1128: one TestComponent now
used userEvent.
Used renderHook

src/lib/viewers/controls/hooks/useAttention.ts
abhishek1128: tried to generalize the naming like 'isActive' since it's a reusable hook. Renamed to isRelatedTargetInside

src/lib/viewers/controls/media/VolumeControls.tsx
abhishek1128: It's a counter. renamed to keyboardVolumeStep

src/lib/viewers/media/MP3Viewer.js
abhishek1128: Added for nextShuttleRate and nextCommentMarker

src/lib/viewers/media/MP3Viewer.js
abhishek1128: Yup. It's valid. calling exitShuttle() on handlePlayPause / handlePlayRequest

src/lib/viewers/media/MP3Viewer.js
abhishek1128: Also valid. Gated on this.isAudioPlayerV2 now. Wasn't really an issue or blocker, but good to have this edge case present until we retire v1

jackiejou (approved):

JChan106 (approved):

PR #1779, opened 2026-09-16
feat(analytics): programmatic Resin for scrub gestures

JChan106:
## Summary
- Log one programmatic Resin event per scrub gesture (mousedown/touchstart/arrow key, not pointermove), matching pinch zoom. Click autolog never sees `role=slider`.
- `SliderControl` fires from the existing `data-resin-target`, so time scrubbers and waveform zoom share one path. Volume is a separate control and logs from its matching mouse/touch/key handlers.
- Archive file names are not clickable, so they are not resin-tagged. Folder buttons keep `data-resin-target="folder"` and still autolog.

## Test plan
- [ ] Scrub a video/audio timeline: one `timeScrubber` programmatic event per mouse/touch gesture and per arrow key in Resin debug; seeking still works.
- [ ] Adjust volume slider: one `volumeSlider` event per mouse/touch gesture and per arrow key; mute toggle still click-autologs.
- [ ] MP3 waveform zoom slider (if flag on): one `waveformZoomSlider` event per mouse/touch gesture and per arrow key; +/- buttons unchanged.
- [ ] Zip preview: folder names still navigate and keep `data-resin-target="folder"`. File names have no resin target and no pointer cursor.
- [ ] Unit tests: `yarn jest src/lib/viewers/controls/slider/__tests__/SliderControl-test.tsx src/lib/viewers/controls/slider/__tests__/VolumeSliderControl-test.tsx`


abhishek1128 (commented):
The data-resin-target tags are already on the sliders — `timeScrubber`, `waveformZoomSlider`, and `volumeSlider`. Click autolog never sees a drag, so I get why we're doing a programmatic `recordAction` on gesture start, same as pinch.

Logging `timeScrubber` from TimeControls `onPointerDown` only covers that one wrapper. Waveform zoom is the same `SliderControl` with a different target, and it never gets this handler, so `waveformZoomSlider` will still be empty. Volume is a separate control, and pointerdown also misses the arrow-key path that already lives in `handleKeydown`.

Could we fire this once inside `SliderControl` (and the matching mouse/touch/key handlers on VolumeSlider) using the `data-resin-target` that's already on the node? TimeControls could stay tags-only, and zoom would come along for free.

abhishek1128 (approved):

PR #1781, opened 2026-09-17
feat(media): add play-next setting on audio and video v2

abhishek1128:
## Summary
- Adds a Play next file setting (Enabled/Disabled) on audio and video v2, with its own cache key and default off.
- Autoplay is unchanged: it still means play this file on open.
- When Play next is on and playback ends, Preview emits `mediaendplaynext` and does not navigate itself. Hosts that own URL/collection (for example Content Preview) should listen and advance.

## Test plan
- [ ] Audio v2 settings show Play next file; audio v1 does not.
- [ ] Video v2 settings show Play next file; video v1 does not.
- [ ] Enabling Play next does not turn Autoplay on, and the preference survives reload.
- [ ] With Play next on, finishing a file in a collection emits `mediaendplaynext` (host should move to the next item).
- [ ] Last file in the collection does not navigate.
- [ ] Play next label stays on one line in the settings flyout.


src/lib/viewers/media/DashViewer.js
```diff
@@ -1388,6 +1388,8 @@ class DashViewer extends VideoBaseViewer {
                     commentMarkers={this.commentMarkers}
                     onCommentMarkerClick={this.handleCommentMarkerClick}
                     {...sharedProps}
+                    onPlayNextChange={this.setPlayNext}
+                    playNext={this.isPlayNextEnabled()}
```
jackiejou, line 1392: should we put these above the spread operator?

src/lib/viewers/media/__tests__/MP4Viewer-test.js
```diff
@@ -171,6 +171,8 @@ describe('lib/viewers/media/MP4Viewer', () => {
     });
 
     describe('renderUI()', () => {
+        const getProps = instance => instance.controls.render.mock.calls[0][0].props;
```
jackiejou, line 174: instead of checking props, can we check something that is or isnt rendered? since this is in the render block

src/lib/viewers/controls/media/__tests__/MediaSettingsMenuPlayNext-test.tsx
```diff
@@ -0,0 +1,60 @@
...
+
+    describe('render', () => {
+        test('should return a valid wrapper', async () => {
+            getWrapper();
+            const menuBack = await getMenuBack();
+            const enabledRadioItem = await getEnabledRadioItem();
+            const disabledRadioItem = await getDisabledRadioItem();
+
+            expect(menuBack).toBeInTheDocument();
+            expect(enabledRadioItem).toBeInTheDocument();
```
jackiejou: can we use toBeVisible?

src/lib/viewers/media/DashViewer.js
abhishek1128: Umm no. I’d leave them after the spread so they override sharedProps (future-proofing). Same order as MP4. Ideally the spread should happen in the beginning to keep sharedProps as default and the rest of the props as overrides

src/lib/viewers/media/__tests__/MP4Viewer-test.js
abhishek1128: renderUI doesn’t actually mount anything, it just calls a mocked controls.render. The settings row is already covered in MediaSettings. This test is only checking that we pass the handler on v2 and skip it on v1.

src/lib/viewers/controls/media/__tests__/MediaSettingsMenuPlayNext-test.tsx
abhishek1128: done

jackiejou (approved):

PR #1785, opened 2026-09-17
feat(controls): use Blueprint RotateLeft for the file rotate icon

EdDaWord:
Pendo Data shows a 7% frustration click on the rotate button. Our hypothesis is that the Icon currently looks more like a refresh button than a rotate. So we're changing the icon

|before|after|
|----|---|
|image|image|


EdDaWord:
Closing due to duplicate work

PR #1787, opened 2026-09-18
feat(audio): loop playback within an open draft comment range

kduncanhsu:
## Summary
- While an open draft range is selected (`endMs > startMs`), Play stays inside that span and wraps to start instead of pausing at end. Collapsed drafts (`endMs: null`) play the whole file as usual.
- We schedule the wrap for the remaining time to the exact end. `timeupdate` only fires about four times a second, so waiting for it lets the playhead (and audio) slip past the right handle before we loop. `MediaBaseViewer.play(start, end)` is not used because it pauses at end.
- Clicking the waveform outside the range still dismisses it (and unchecks the host toggle), then seeks to the click. Keyboard, control, and host seeks clamp into the span until the draft is cleared.

## Test plan
- [ ] With an open range, Play from outside the span jumps to start and loops
- [ ] Play from inside the span continues from the playhead and wraps at the right handle (no visible overshoot, no pause)
- [ ] Collapsed draft (`endMs: null`) does not loop or confine playback
- [ ] Clearing the draft (checkbox, post, or click-outside) restores whole-file playback from the current playhead
- [ ] Click-outside dismisses the range and seeks to the click
- [ ] Keyboard skip / End / host comment seek stay inside the open span
- [ ] Resizing the range or receiving a new draft while playing updates the loop window immediately
- [ ] `yarn jest src/lib/viewers/media/__tests__/MP3Viewer-test.js src/lib/viewers/media/waveform/__tests__/WaveformView-test.tsx --watchman=false`

Made with [Cursor](https://cursor.com)


src/lib/viewers/media/MP3Viewer.js
```diff
@@ -928,22 +937,178 @@ class MP3Viewer extends MediaBaseViewer {
         this.pendingHostSelectedSeek = null;
         this.exitShuttle();
         this.mediaEl.pause();
-        if (this.mediaEl.currentTime !== marker.time) {
-            this.mediaEl.currentTime = marker.time;
+        const time = this.clampTimeToOpenCommentRange(marker.time);
```
abhishek1128: Do we want to stay in the range and keep the range open when a timestamped comment is clicked in the sidebar? Or should this be more like if marker.time is within the range, apply the seek, and when it's outside the range, dismiss the range and seek? Let's clarify this?

src/lib/viewers/media/MP3Viewer.js
```diff
@@ -928,22 +937,178 @@ class MP3Viewer extends MediaBaseViewer {
...
+     * @return {void}
+     */
+    enforceCommentRangePlayback = () => {
+        const range = this.getOpenCommentRangeSeconds();
+        if (!range || !this.mediaEl) {
+            return;
+        }
+        const { endSec, startSec } = range;
+        const current = this.mediaEl.currentTime;
+        let next = current;
```
abhishek1128: Can we better name next and current?

src/lib/viewers/media/__tests__/MP3Viewer-test.js
```diff
@@ -834,6 +834,190 @@ describe('lib/viewers/media/MP3Viewer', () => {
...
+            mp3.mediaEl.currentTime = 10;
+            setPaused(true);
+
+            mp3.handleCommentRangeDraft({ endMs: 4000, startMs: 2000 });
+
+            expect(mp3.mediaEl.currentTime).toBe(10);
+        });
+
+        test('should clamp a host comment seek into the open span', () => {
+            mp3.handleCommentRangeDraft({ endMs: 4000, startMs: 2000 });
```
abhishek1128, line 985: same product question as on the MP3Viewer

src/lib/viewers/media/MP3Viewer.js
kduncanhsu: yep, renamed to `currPlayhead` and `nextPlayhead`

src/lib/viewers/media/MP3Viewer.js
kduncanhsu: good point. any sidebar/marker comment seek should dismiss the selected. have updated

src/lib/viewers/media/__tests__/MP3Viewer-test.js
kduncanhsu: same as bove

abhishek1128 (approved):

